PR #228, opened 2011-02-10
Adding classList support to the various class manipulation methods

dominicbarnes:
(this is my first pull request with jQuery, so bear with me)

The following functions in attributes were modified: addClass, removeClass, toggleClass, hasClass

Added support for the HTML5 classList feature. As far as I can tell, at least Firefox and Chrome both support this.

I've added this in such a way that if it does not exist in the DOM for a particular element, it continues with the original method.

I'm more than happy to refactor my code if necessary, but I made sure that I passed the JSLint bulid test as well as all the tests in the jQuery Test Suite.

### DIFF
--- a/src/attributes.js
+++ b/src/attributes.js
@@ -51,19 +51,26 @@ jQuery.fn.extend({
 				var elem = this[i];
 
 				if ( elem.nodeType === 1 ) {
-					if ( !elem.className ) {
-						elem.className = value;
-
-					} else {
-						var className = " " + elem.className + " ",
-							setClass = elem.className;
-
-						for ( var c = 0, cl = classNames.length; c < cl; c++ ) {
-							if ( className.indexOf( " " + classNames[c] + " " ) < 0 ) {
-								setClass += " " + classNames[c];
-							}
+					var c, cl;
+					if ( elem.classList ) {
+						for ( c = 0, cl = classNames.length; c < cl; c++ ) {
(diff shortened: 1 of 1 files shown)


HolgerJeromin:
The layouttests from the last comment is now available [here](https://github.com/WebKit/webkit/blob/master/LayoutTests/fast/dom/HTMLElement/script-tests/class-list.js).

Has someone newer performance number that those 6 years old ones? Again waiting for jsperf.com up again.

mgol:
@HolgerJeromin You can fire up test cases in the repository of my [jquery-classlist plugin](https://www.npmjs.com/package/jquery-classlist#rationale). We've since switched from using the `className` property to using the `class` attribute to support SVG class manipulation.

We don't plan to switch to `classList` in any predictable future as even IE 11 doesn't support `classList` in SVG so we'd have to maintain duplicate code paths; performance gains would have to be huge in real-world scenarios (not just jsperf micro-benchmarks) to make the additional code path worth it.

HolgerJeromin:
OK, fine with me. 
I had performance issues with this, but just changed to native DOM api in the critical and easy cases (one element, one className, HTML only, hundreds of calls)

mgol:
@HolgerJeromin If you have performance issues with the current implementation on a real site, we would definitely want to know! Usually it turns out other things are responsible.

It's easiest to check that by including my linked plugin and seeing if it improves performance in a visible way.

PR #764, opened 2012-05-04
Fixes #6724: Wrong $(window).height() in Mobile Safari

farmdawgnation:
Greetings all,

This is my first jQuery pull request, so please do bear with me. I know I probably missed some detail in terms of style, testing, or something. :)

This is a fix for issue #6724, where the value of `$(window).height()` is always the value of the window height minus the address bar and debug console. The problem is that the value returned for `$(window).height` is accurate until you scroll. At which point the value of `window.innerHeight` is adjusted to reflect the correct height of the drawable area the Browser is using, but the `document`'s height properties are not.

This patch fixes the issue by returning `window.innerHeight` for `$(window).height()` if it exists. If it doesn't exist, it will attempt to use height parameters from document and its children. Tests for the dimensions module are passing in the latest Chrome, Firefox, and Mobile Safari.

Size changes:

```
jQuery Size - compared to last make
  252813    (+26) jquery.js
   94786    (+15) jquery.min.js
   33650    (+15) jquery.min.js.gz
```

Let me know what I did wrong! :)

### DIFF
--- a/src/dimensions.js
+++ b/src/dimensions.js
@@ -31,7 +31,10 @@ jQuery.each( { Height: "height", Width: "width" }, function( name, type ) {
 			var doc, orig, ret;
 
 			if ( jQuery.isWindow( elem ) ) {
-				return window[ "inner" + name ] || elem.document.documentElement[ clientProp ];
+				// As of 5/8/2012 this will yield incorrect results for Mobile Safari, but there
+				// isn't a whole lot we can do. See pull request at this URL for discussion:
+				// https://github.com/jquery/jquery/pull/764
+				return elem.document.documentElement[ clientProp ];
 			}
 
 			// Get document width or height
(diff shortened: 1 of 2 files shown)


hashchange:
Given that this behaviour, or bug, hasn't been fixed in mobile Safari for the last five years, and hasn't been dealt with in jQuery either, let me point you to a working solution: [`$.windowHeight()`](https://github.com/hashchange/jquery.documentsize#window-size) in [jQuery.documentSize](https://github.com/hashchange/jquery.documentsize). (I have written that thing.)

I don't want to revive this thread from the late middle ages, but it does come up when you research the topic, so I figured that a pointer to the solution would be helpful here. 

In case you do want to tackle the issue in jQuery, feel free to pillage my code, or the unit tests.

PR #1051, opened 2012-12-02
Fix #12838: hook point for non-jQuery.ajax synchronous script fetch/execute in domManip

gibson042:
I'm not thrilled about the name, but at least it's unambiguous.

Sizes - compared to master @ c2d6847de09a52496f78baebc04f317e11ece6d2

```
    264083       (+44)  dist/jquery.js                                         
     92059       (+60)  dist/jquery.min.js                                     
     32626       (+11)  dist/jquery.min.js.gz
```


CFThomas:
Boy that rwaldron is a charmer!

PR #1191, opened 2013-03-03
Fix for #13514.Set selectedIdx to 0 if a 'wrong' value is set as dropdown-list value

ruado1987:
This commit fixes the issue described in the bug ticket when a dropdown-list is set to a value which doesnt match with any of its options' value.

### DIFF
--- a/src/attributes.js
+++ b/src/attributes.js
@@ -282,10 +282,10 @@ jQuery.extend({
 						optionSet = true;
 					}
 				}
-												
+				
+				// force browsers to behave consistently when non-matching value is set
 				if ( !optionSet ) {
-					// force IE9 to behave like other browsers when a non-matching value is set on a select-one element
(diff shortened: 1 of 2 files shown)


billynoah:
Applications everywhere used `$('select').val('');` to *unset* the selected option and reset a form.  This now results in a blank option appearing and broke a whole ton of code across dozens of projects when I upgraded to jQuery 1.12.  It's not that this new behavior is wrong, the problem is that many, many applications are now broken because of this change.

mgol:
@billynoah Unfortunately, it's too late now to change it; we're no longer maintaining jQuery 1.x/2.x and even if we were, changing that back would be a breaking change now and could break people using jQuery 1.12 and expecting the new behavior.

billynoah:
@mgol - Yes I get that, just taking time out to gripe :-)  I don't think this change was a good one, but I agree it's too late.

PR #1427, opened 2013-11-11
Fix #14503: Cast to string before setting XHR header

hongymagic:
When setting request header to a `null` value, **Microsoft.XMLHTTP** throws a `Type Mismatch` exception. This means that in IE6+ passing a `null` value in the `headers` property causes `$.ajax` to fail. (IE7 and above require users to uncheck _"Enable native XMLHTTP support"_ in settings).

After a discussion the proposed solution was to skip `undefined` and cast to string in other cases.

Relevant tests have been extended and existing tests modified to fit new changes. [See the discussion on jQuery Bug Tracker](http://bugs.jquery.com/ticket/14503).

Changes to existing tests:
- `headers.php` now echo headers that have non-empty values

### DIFF
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -78,7 +78,15 @@ if ( xhrSupported ) {
 
 					// Set headers
 					for ( i in headers ) {
-						xhr.setRequestHeader( i, headers[ i ] );
+						// Support: IE<9
+						// IE's ActiveXObject throws a 'Type Mismatch' exception when setting
+						// request header to a null-value.
+						//
+						// To keep consistent with other XHR implementations, cast the value
+						// to string and ignore `undefined`.
+						if ( headers[ i ] !== undefined ) {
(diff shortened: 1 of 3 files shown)


src/ajax/xhr.js
```diff
@@ -78,7 +78,15 @@ if ( xhrSupported ) {
...
 					// Set headers
 					for ( i in headers ) {
-						xhr.setRequestHeader( i, headers[ i ] );
+						// Support: IE<9
+						// IE's ActiveXObject throws a 'Type Mismatch' exception when setting
+						// request header to a null-value.
+						//
+						// To keep consistent with other XHR implementations, cast the value
+						// to string and ignore `undefined`.
+						if ( headers[ i ] !== undefined ) {
```
dbrans: Should this condition be cherrypicked to the jquery 2.x branch? I'm seeing undefined Accept headers.

PR #1428, opened 2013-11-11
Data: move element cache to element[expando]

### LINKED ISSUE #1734: Move element cache to the element[expando] to avoid cleanup and reduce code.
Originally reported by john.david.dalton@… at: http://bugs.jquery.com/ticket/11570
<p>
Soo way back in 2005 Dean Edwards proposed an event system that slapped the listeners on the actual element. This allowed cache to be destroyed once the element was GC'ed.
<a class="ext-link" href="http://dean.edwards.name/weblog/2005/10/add-event2/" rel="nofollow"><span class="icon">&nbsp;</span>http://dean.edwards.name/weblog/2005/10/add-event2/</a>
</p>
<p>
Over the weekend Peter Michaux proposed a similar solution:
<a class="ext-link" href="https://twitter.com/#!/petermichaux/status/189012245483757568" rel="nofollow"><span class="icon">&nbsp;</span>https://twitter.com/#!/petermichaux/status/189012245483757568</a>
<a class="ext-link" href="https://github.com/petermichaux/evento/blob/bd065a9b254a8d14b956ec438dc06c29622c3ef6/src/eventTarget.js#L165" rel="nofollow"><span class="icon">&nbsp;</span>https://github.com/petermichaux/evento/blob/bd065a9b254a8d14b956ec438dc06c29622c3ef6/src/eventTarget.js#L165</a>
</p>
<p>
I noticed that jQuery supports events on vanilla objects too so this change would align elements and objects.
</p>
<p>
The gist is that since jQuery is already adding an expando for a UID they can just use that property as the actual storage for the element.
</p>
<p>
Then all the code to recursively clean .removed'ed elements event/custom data could be removed as when the element is GC'ed that data is destroyed.
</p><p>Issue reported for jQuery git</p>

rwaldron:
- avoid explicit data.discard() cleanup calls
- explicitly remove the data.events property
- reduces code footprint

Fixes #1734

### DIFF
--- a/src/data/Data.js
+++ b/src/data/Data.js
@@ -5,15 +5,6 @@ define([
 ], function( jQuery, rnotwhite ) {
 
 function Data() {
-	// Support: Android<4,
-	// Old WebKit does not have Object.preventExtensions/freeze method,
-	// return new empty object instead with no [[set]] accessor
-	Object.defineProperty( this.cache = {}, 0, {
-		get: function() {
-			return {};
-		}
-	});
-
 	this.expando = jQuery.expando + Data.uid++;
 }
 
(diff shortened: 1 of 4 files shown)


timmywil:
@rwaldron Could you resolve the conflicts. I think this is read after that.

rwaldron:
- Rebased
- Resigned the CLA
- Updated to include changes based on @dcherman's feedback

rwaldron:
weird

jbedard:
Some of the cleanup from 95fb798980d7e404c413e29e20016db9052e2bf2 got undone here.

`defineProperties` was replaced with `defineProperty`, then `descriptor` can be removed (less code/vars and it is minorly faster). The Android try/catch was also removed.  Both of these have been added back in https://github.com/jquery/jquery/commit/d702b7637a61e1973e08c27b8d8de2ed24a543e2

timmywil:
You're right. @rwaldron was this intentional or is it alright to change back?

PR #1449, opened 2013-11-29
Switch to script injection instead of indirect eval call

markelog:
Right now, 2.x uses two techincs for code execution in global scope: script injection if "use strict" statement is found and indirect eval call if not. 1.x only uses eval path.

With indirect eval call, we have only one problem - if evaluated code uses strict pragma, it will not be executed in global scope, which subverts function purpose. Although current code tries to circumvent it, but fails on edge <a href="http://bugs.jquery.com/ticket/14189">cases</a>.

With script injection we have four problems:
1) It requires browser environments, i.e. DOM
2) It might be executed <a href="http://bugs.jquery.com/ticket/7862">asynchronously</a>
3) Errors do not propagate, but instead would be thrown in different flow, meaning you can't catch them without `window.onerror`
4) Potential performance penalty

One common and obvious issue is difference of `globalEval` functionality between branches. Because now we have all those problems combine, so i propose to start using only one technic, this PR will try to do it with script injection.

Now about issues with this way:

<blockquote>It requires browser env, i.e. DOM</blockquote>

Well, jQuery is DOM-centric library, in other environments you should use other means

<blockquote>It might be executed asynchronously</blockquote>

It's <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Gecko-specific_notes">no longer true</a>, since Firefox 4.0, but just to be sure, i added a test for it.

<blockquote>Errors do not propagate</blockquote>

Yes, that would be a incompatible difference

<blockquote>Potential perfomance penalty</blockquote>

It's actually the <a href="http://jsperf.com/tag-vs-eval-fn-construction">opposite</a> (it's not my test if anyone asks :-)

I think the important thing is that jQuery should use only one technic and it should be identical between branches, i, personally, could live with any of them.

/cc @dmethvin, @rwaldron, @jdalton, @jaubourg


mzgol:
This PR seems to make Android fail AJAX tests even on `compat` (previously it failed only on `master`). That's one of the reasons I created #2483; supporting Android 2.3 fully would sometimes mean reverting things like this PR and I'd like to avoid catering to this browser so much.

markelog:
Okay, if we leave 2.3, i will investigate, if not, then lets leave it as is.

PR #1505, opened 2014-01-31
jQuery.parseHTML: Mitigate XSS vulnerability

fhemberger:
Scripts passed in event attributes are executed in `parseHTML` immediately,
without any possibility for the user to intervene:

`jQuery.parseHTML('');`

To mitigate this vulnerability `document.implementation.createHTMLDocument()`
should be used as standard context instead of `document`. Now the user has
to set a different context deliberately for this issue to occur.

See also jQuery ticket #13921.

**EDIT:** Agreed to the CLA.

### DIFF
--- a/src/core/parseHTML.js
+++ b/src/core/parseHTML.js
@@ -2,7 +2,7 @@ define([
 	"../core",
 	"./var/rsingleTag",
 	"../manipulation" // buildFragment
-], function( jQuery, rsingleTag ) {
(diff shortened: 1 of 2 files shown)


mgol:
This was originally backported to 1.12/2.2 but due to #2941 we're backing out the patch on those branches (as they wen't meant to be non-breaking releases and there may always be more issues around that); we'll try to fix the issue for 3.0.

PR #1506, opened 2014-01-31
jQuery.parseHTML: Mitigate XSS vulnerability

fhemberger:
Scripts passed in event attributes are executed in `parseHTML` immediately,
without any possibility for the user to intervene:

`jQuery.parseHTML('');`

To mitigate this vulnerability `document.implementation.createHTMLDocument()`
should be used as standard context instead of `document`. Now the user has
to set a different context deliberately for this issue to occur.

See also GitHub issue #1505.

**EDIT:** Agreed to the CLA.


mgol:
This PR was reverted on `1.12-stable` so I'm removing the milestone. We'll still try to fix the issues that popped up in 3.0.0 and not revert this patch there so I'm keeping the milestone on #1505.

(the issue to resolve is #2941).

PR #1525, opened 2014-02-22
Ajax: Integrated native datatype, for XHR2 rich responseType support

tarikbenmerar:
XHR2 provides a response attribute, that contains a converted response to a native object, depending on the responseType ("arraybuffer", "blob", "document", "json"). See:
https://dvcs.w3.org/hg/xhr/raw-file/tip/Overview.html#the-response-attribute
http://www.html5rocks.com/en/tutorials/file/xhr2/?redirect_from_locale=fr

I propose in this pull request, to create a native datatype and response, to manage this kind of response. Some of previous closed tickets proposed to include a binary dataType, with adhoc changes on the transport function :
http://bugs.jquery.com/ticket/11461
http://bugs.jquery.com/ticket/9992
http://bugs.jquery.com/ticket/14068

I think it is not appropriate as the response attribute deals with more than binary data, like arraybuffer, but extends to blob, document and json datatype. In my pull request, I create a native datatype to support all the four response types that are abstracted.

I think also that such improvements should be included as this is a new improvement in XHR, that is not already included in jQuery. Also, trying to use a converter in the case of ArrayBuffer for example, leads to a heavy loop for the conversion (See: http://updates.html5rocks.com/2012/06/How-to-convert-ArrayBuffer-to-and-from-String). Simple casting doesn't function.

In my case, I have used Brain dMRI NifTI data formats that are 20MB in size, and all of them need parsing after downloading, which is also a heavy computation (a hundred of milliseconds).

Also, treating a binary data as text, it is just not natural at all, as this is already the case.

In any way, I think we should find a solution, and this is my proposition. Please, take a look at the tests, for the usage. And let's start a discussion around this.

### DIFF
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -2000,7 +2000,7 @@ module( "ajax", {
 	});
 
 	ajaxTest( "jQuery.ajax() - Native datatype", 8, [{
-		url: url("data/native.bin"),
+		url: url( "data/native.bin" ),
 		xhrFields: {
 			responseType: "arraybuffer"
 		},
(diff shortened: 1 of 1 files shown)


tarikbenmerar:
@dmethvin had some free time and worked on a plugin https://github.com/acigna/jquery-ajax-native. The source code is here : https://github.com/acigna/jquery-ajax-native/blob/master/src/jquery-ajax-native.js . I had to copy past a large part of the xhr ajaxTransport as there were no alternative. I hope there is a better way to add this support in the future.

tarikbenmerar:
@ProgerXP here is for the plugin : https://github.com/acigna/jquery-ajax-native, if you want to use the feature now.

PR #1573, opened 2014-05-05
Ajax: Fix the invalid script test in Android 2.3

mgol:
Android 2.3 doesn't support async script tags so if the injected script tag
refers to a throwing script, the error is thrown synchronously making the
window.onerror handler to not fire. Catch the error instead.


mzgol:
A new issue: #2457 and PR: #2458.

PR #1633, opened 2014-07-30
Update ready.js

LowG33kDev:
Add IE support in jQuery.ready.promise. 

When you use jQuery in a big html file, the loading is very slow, and the function jQuery.ready.promise is used. This function call addEventListener, and it don't work on IE . I added the equivalent function attachEvent.

### DIFF
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -59,12 +59,31 @@ jQuery.extend({
 });
 
 /**
+ * Clean-up method for dom ready events
+ */
+function detach() {
+	if ( document.addEventListener ) {
+		document.removeEventListener( "DOMContentLoaded", completed, false );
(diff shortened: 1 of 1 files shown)


rwil02:
Issue exists in ie8 using 1.11.3

mzgol:
What issue? This issue claims addEventListener is used in IE8 which is not true in jQuery 1.x.

rwil02:
Actually, it "kind of" is used.  But NOT jQuery's fault.
I eventually tracked it down to the ScriptBundle minification  stripping out the "if(u.addEventListener)" from the jQuery-1.11.3.min version.
Which it doesn't do to the jQuery-1.11.3 version when I hard coded that instead

PR #1668, opened 2014-09-27
Data: Avoid ES5 defineProperty for speed

jbedard:
As discussed in http://bugs.jquery.com/ticket/14839, removing the use of definePropert(y|ies) brings the jQuery2 data initialization up to speed with jQuery1 for DOM nodes. For non nodes this also switches from defineProperties to defineProperty which is a bit faster, saves an object and saves some bytes.

@gibson042 this has the changed you suggested in https://github.com/jbedard/jquery/commit/d44885b3603dd41ebc47ac889bc55a9a0da0fd4f

### DIFF
--- a/src/data/Data.js
+++ b/src/data/Data.js
@@ -29,24 +29,20 @@ Data.prototype = {
 			return 0;
 		}
 
-		var descriptor = {},
-			// Check if the owner object already has a cache key
-			unlock = owner[ this.expando ];
+		// Check if the owner object already has a cache key
+		var unlock = owner[ this.expando ];
 
 		// If not, create one
 		if ( !unlock ) {
 			unlock = Data.uid++;
 
-			// Secure it in a non-enumerable, non-writable property
-			try {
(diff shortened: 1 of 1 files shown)


timmywil:
It's been a while. Is everyone ok if we land this?

dmethvin:
Looks like @jbedard needs to re-sign the CLA with the email he used for the commit, but otherwise YES, let's get this puppy landed! :dog:

jbedard:
I've resigned with the email of the commit. Does @jquerybot need something to trigger it?

arthurvr:
@jbedard Looks alright. Thanks! The bot doesn't have the ability to re-check PRs yet. (https://github.com/jquery/jquery-license/issues/2)

rwaldron:
[As promised](https://github.com/jquery/jquery/pull/1668#issuecomment-67508811) [data-as-expando](https://github.com/jquery/jquery/pull/1428) has been rebased

rwaldron:
Hm. Looks like I lost something there. Hold the line please.

PR #1714, opened 2014-10-16
Fix #15239: position when parent scrolled

rmcdaniel:


timmywil:
@mikesherov ping

timmywil:
@jquery/core Does anyone have an issue with landing this?

gibson042:
I'm interested in the size difference between two assignment operations on each `parentOffset` property vs. one (e.g., `parentOffset.top += jQuery.css(…) - offsetParent.scrollTop()`), but the logic and test changes look good to me.

timmywil:
> I'm interested in the size difference between two assignment operations on each `parentOffset` property vs. one (e.g., `parentOffset.top += jQuery.css(…) - offsetParent.scrollTop()`)

It saves 3 bytes.

timmywil:
Landing with the proposed change from Richard.

mgol:
This PR caused a couple of problems in jQuery 2.2.0 & 1.12.0 so we've reverted it in 49833f7795d665ff1d543c4f71f29fca95b567e9; we'll try to tackle the problem later.

PR #1821, opened 2014-10-30
Deferred: produce and consume standard thenables

jaubourg:
Up-to-date version of the patch that passes all the tests.


jaubourg:
```
   raw     gz Compared to master @ d7e5fcee519e5f3e840beef9e67a536e75133df9
 +1769   +478 dist/jquery.js
  +629   +235 dist/jquery.min.js
```

100% Promise A+ compliant, smart interoperability (even `.pipe()` can handle standard promises).

Does not protect against exceptions in non-`then` handlers like #1996, which does so at the cost of nearly doubling memory usage (not worth it imo).

jaubourg:
Would appreciate some code review too.

jaubourg:
... and I f***ed up the whole PR pretty badly so any help in handling this mess by some git guru would be much appreciated :/

timmywil:
This PR was essential in working out which changes we wanted to make and how to make them. So, it's not a complete waste, but since we will be landing #1996, this one can be closed.

Issue #1815, opened 2014-10-28
.css method return wrong value for negative percent margin


deneb7000:
.css method return wrong value for margin on an hidden element if it is specified as negative percent.
the value of width is returned instead.

examples:
$("&lt;div/&gt;").css({'margin': '-2%'}).css("marginTop"); // return "auto" - WRONG: expected "-2%"

$("&lt;div/&gt;").css({'margin': '-2%', 'width': '60%'}).css("marginTop"); // return "60%" - WRONG: expected "-2%"

$("&lt;div/&gt;").css({'margin': '2%'}).css("marginTop"); // return "2%" - CORRECT

$("&lt;div/&gt;").css({'margin': '-2px'}).css("marginTop"); // return "-2px" - CORRECT

mzgol:
The patch restoring the `marginRight` hack on Android 4.0-4.3 has landed on `master` so now this problem will reappear in these browsers there.

So we'll most likely need to mark it as wontfix and add a note to the docs.

mzgol:
#1842 has landed and it includes a support test for the `pixelMarginRight` workaround so it seems all we can do here is to make it clear in the docs that the Android Browser suffers from this issue?

timmywil:
@mzgol agreed.

timmywil:
Moved to docs issue: https://github.com/jquery/api.jquery.com/issues/865

PR #1842, opened 2014-11-05
Css: Restore the hack to get pixels for .css('width') etc.

mzgol:
This hack turns out to be needed by Android 4.0-4.3

Refs gh-1815
Refs gh-1820

### DIFF
--- a/src/css/curCSS.js
+++ b/src/css/curCSS.js
@@ -3,11 +3,13 @@ define([
 	"./var/rnumnonpx",
 	"./var/rmargin",
 	"./var/getStyles",
+	"./support",
 	"../selector" // contains
-], function( jQuery, rnumnonpx, rmargin, getStyles ) {
+], function( jQuery, rnumnonpx, rmargin, getStyles, support ) {
 
 function curCSS( elem, name, computed ) {
-	var ret;
+	var width, minWidth, maxWidth, ret,
+		style = elem.style;
 
 	computed = computed || getStyles( elem );
 
@@ -21,6 +23,29 @@ function curCSS( elem, name, computed ) {
 
 		if ( ret === "" && !jQuery.contains( elem.ownerDocument, elem ) ) {
 			ret = jQuery.style( elem, name );
+		}
+
+		// Support: Android 4.0-4.3
+		// A tribute to the "awesome hack by Dean Edwards"
+		// Android Browser returns percentage for some values,
+		// but width seems to be reliably pixels.
+		// This is against the CSSOM draft spec:
+		// http://dev.w3.org/csswg/cssom/#resolved-values
+		if ( !support.pixelMarginRight() && rnumnonpx.test( ret ) && rmargin.test( name ) ) {
+
+			// Remember the original values
+			width = style.width;
+			minWidth = style.minWidth;
+			maxWidth = style.maxWidth;
+
+			// Put in the new values to get a computed value out
+			style.minWidth = style.maxWidth = style.width = ret;
+			ret = computed.width;
+
+			// Revert the changed values
+			style.width = width;
+			style.minWidth = minWidth;
+			style.maxWidth = maxWidth;
 		}
 	}
 
--- a/src/css/support.js
+++ b/src/css/support.js
@@ -6,7 +6,7 @@ define([
 ], function( jQuery, document, documentElement, support ) {
 
 (function() {
-	var pixelPositionVal, boxSizingReliableVal,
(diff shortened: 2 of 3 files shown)


mzgol:
This is ready for initial review, I still have one problem, though. This patch makes the test fail in Android 2.3 & Chrome (40 & Canary) in the [bodyBackground test](https://github.com/jquery/jquery/blob/a5e1c9b44c971fd7046d9a95bd0810e50840b663/test/unit/support.js#L29). The added support test returns `false` in this case as in both Chrome & Android 2.3 the computed `marginRight` is then equal to `50%`, not `4px`.

It seems there's a race condition here; if some operations are executed in the "right" order the browser forgets how to compute styles. O_o Firefox is OK here.

If anyone has any idea what might be a cause of this, that'd be great. Otherwise I'll have to create a simpler test case and probably submit a Chrome issue. :/

dmethvin:
Maybe @mikesherov has some idea, I'm stumped. Otherwise it looks like this is putting back the big lump of code we took out so if we have to, we have to.

mzgol:
> Maybe @mikesherov has some idea, I'm stumped. Otherwise it looks like this is putting back the big lump of code we took out so if we have to, we have to.

Yeah but I'd like to run a support test and apply the hack only when needed as it's costly. So I'm a little surprised that Chrome is buggy but only in rare circumstances that we stumble upon in tests. :/

mzgol:
Stupid me, it was caused by the `div { display: inline }` CSS rule in the `bodyBackground.html` template. I added an explicit `display:block` to the div CSS and now it works fine. \o/

Ready to merge unless anyone has any remarks. :)

mzgol:
If no one will object, I'll land this probably after next Monday meeting.

mzgol:
Currently it adds 130 bytes; it'd be good to compress it.

mzgol:
+119 bytes.

Why do we pass `null` as the second parameter to `getComputedStyle` anyway? Does anything besides ancient Firefox versions require it? Since even Android 2.3 seems to go fine with it...

markelog:
> Why do we pass null as the second parameter to getComputedStyle anyway? 

See https://developer.mozilla.org/en-US/docs/Web/API/Window.getComputedStyle, pretty safe to remove it now.

But it feels like a different issue, different commit, PR etc. Like with can try to omit third argument of `addEventListener` method for `master` branch and see where it would lead us. 

And maybe for other native methods as well, but i would do in separate PR.

gibson042:
>  it'd be good to compress it

Have you tried combining the `if ( computed )` blocks? `jQuery.swap` also comes to mind, and the whole css support suite looks rife with duplication (though cleaning that is probably best suited to another PR).

mzgol:
@markelog 
>> Why do we pass null as the second parameter to getComputedStyle anyway? 
>
>See https://developer.mozilla.org/en-US/docs/Web/API/Window.getComputedStyle, pretty safe to remove it now.
>
>But it feels like a different issue, different commit, PR etc.

Sure, that was a side-question, I didn't mean to do it here.

@gibson042 
> Have you tried combining the `if ( computed )` blocks?

What `if ( computed )` blocks?

The `boxSizingReliable` & `pixelPosition` tests require a different setup than `pixelMarginRight` (see what `width` is set in both cases on the div). Conflating them would mean stacking 2 layouts which is not good.

mzgol:
@gibson042 I could reimplement some of those tests to use `jQuery.swap` and be shorter but I'd still want to keep the logic of computing both `pixelPosition` & `boxSizingReliable` in one go. As you said - that's rather a question for a separate PR.

markelog:
@mzgol created - https://github.com/jquery/jquery/issues/2047

gibson042:
> What `if ( computed )` blocks?

[src/css/curCSS.js#L18](https://github.com/mzgol/jquery/blob/97330e00f97a34cca6b71e80d497cb4975b0b239/src/css/curCSS.js#L18-22). And the same file for `jQuery.swap`: [#L36-48](https://github.com/mzgol/jquery/blob/97330e00f97a34cca6b71e80d497cb4975b0b239/src/css/curCSS.js#L36-48).

> The `boxSizingReliable` & `pixelPosition` tests require a different setup than `pixelMarginRight` (see what `width` is set in both cases on the div)

That's true, but maybe the former could use height. There are many possibilities, all to be explored elsewhere.

mzgol:
> That's true, but maybe the former could use height. There are many possibilities, all to be explored elsewhere.

Sure. I just want to get Android>2.3 green for now, we can tweak it later, even after the beta. :)

mgol:

PR #1905, opened 2014-12-05
Data: Fix to .data update if key was added as-is

meetselva:
When adding data using object argument, the keys are not camel cased as internally it uses `.extend` to create data cache. This fails when trying to update the value currently as the update is always camel cased. 

Test Scenario: Creating a data using [object argument](http://api.jquery.com/data/#data-obj) with keys in object has dash(es).

````
//creates a data for the element with keys data-test-1 and data-test-2 
$('#data-test').data({'data-test-1': 1, 'data-test-2': 2}); 
//logs 1
console.log($('#data-test').data('data-test-1'));
//creates a new data dataTest1 and updates the value 4
$('#data-test').data('data-test-1', 4);
//still logs 1 as the update didn't actually update the value
console.log($('#data-test').data('data-test-1'));
````

Test Case: http://jsfiddle.net/c0k0owkq/

jq

The fix is to either use a different method instead of `.extend` to add data from an object or check if the key exist as-is when updating. The later was an easy change.

### DIFF
--- a/test/unit/data.js
+++ b/test/unit/data.js
@@ -678,18 +678,18 @@ test( "JSON data- attributes can have newlines", function() {
 	x.remove();
 });
 
-/*testIframeWithCallback( "enumerate data attrs on body (#14894)", "data/dataAttrs.html", function( result ) {
+testIframeWithCallback( "enumerate data attrs on body (#14894)", "data/dataAttrs.html", function( result ) {
 	expect(1);
 
 	equal(result, "ok", "enumeration of data- attrs on body" );
-});*/
+});
 
 test(".data should update as-is if added without camel casing (PR 1.x-master #1905)", function() {
 	expect(5);
 	var div = jQuery("<div></div>").prependTo("body");
 	div.data({
 		"data-test-1": "data1",
-		"data-test-2": "data2",
+		"data-test-2": "data2"
 	});
 	equal( div.data()["data-test-1"], "data1", "Verifing data update succesful with object" );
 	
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/unit/data.js [hunk 1]
```diff
@@ -682,4 +682,28 @@ testIframeWithCallback( "enumerate data attrs on body (#14894)", "data/dataAttrs
...
+	equal( div.data()["dataTest1"], undefined, "Verifying it is not camel cased when adding as object" );
+	
+	//update data-test-1
+	div.data("data-test-1", "data1_updated");
+	equal( div.data()["data-test-1"], "data1_updated", "Verifying .update() updating the right data" );
+	
+	//remove data-test-1 and data-test-2
+	div.removeData(["data-test-1", "data-test-2"]);
+	equal( div.data()["data-test-1"], undefined, "Verifying remove of data-test-1 deleting the right data" );
+	equal( div.data()["data-test-2"], undefined, "Verifying remove of data-test-2 deleting the right data" );
```
mzgol: There should be spaces everywhere after the opening bracket and before the closing one if there are any parameters passed, e.g. here `div.data()[ "data-test-2" ]` etc.

mzgol: I see the rest of this file has incorrect spacing as well, though (our fault) so you don't have to change it if you don't want to now.

dmethvin: Typically I have been using current style for creating brand new tests and sticking with the ambient style when patching old tests. I didn't want to have conflicting styles so close to each other but also didn't want to go through and create a lot of ugly whitespace diffs. If we use that as a rule then this newly created test should be written in the current style. At some point (when we finalize all our style issues) we can do through and do a style-only commit.

timmywil:
BTW, this is also an issue in master. @meetselva Would you be willing to fix the same issue in master?

meetselva:
@timmywil The master code looks different from the 1.x-master (refactored and cleaned up) . I will take a look and will let you know.

dmethvin:
@meetselva You're really close here. Can you put together a PR for master? This is blocking some other work so I really want it to land this week. Really appreciate you staying with this!

meetselva:
Sorry for the delay.

Created [PR 2036](https://github.com/jquery/jquery/pull/2036) with a fix to master. Please review.

timmywil:
After [much deliberation](https://docs.google.com/spreadsheets/d/1gf0vzYGwg7IA7yIkyJs-Jk31SWDogQrBkzPb-xl6u-o/), we decided to go a different route. This happened to be a very hairy issue involving lots of issues. Thank you very much for your work here and I'm sorry we weren't able to merge this PR.

PR #1909, opened 2014-12-07
Build: Specify valid components for commit messages

gibson042:
I didn't see a sane place to reference the bypass prefixes (cf. https://github.com/jzaefferer/commitplease#usage ), but they really should be publicized to contributors to prevent unnecessary frustration.

Ref https://github.com/jzaefferer/commitplease/issues/23

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,5 +55,30 @@
     "build": "npm install && grunt",
     "start": "grunt watch",
     "test": "grunt"
+  },
+  "commitplease": {
+    "components": [
+      "Docs",
+      "Tests",
+      "Build",
+      "Release",
+
+      "Core",
+      "Ajax",
+      "Attributes",
+      "Callbacks",
+      "CSS",
+      "Data",
+      "Deferred",
+      "Dimensions",
+      "Effects",
+      "Event",
+      "Manipulation",
+      "Offset",
+      "Queue",
+      "Selector",
+      "Serialize",
+      "Traversing"
+    ]
   }
 }


dmethvin:
Before we land this, we'll need to ensure the the commits done during the release script don't crap out. I'd be most comfortable if we just disabled it for those commits, nothing would be worse than dying in the middle of a release.  @jzaefferer Do commits prefixed with [STUFF] exempt from the component check? The docs weren't clear and I was too lazy to read through the code.

 /cc @timmywil who has the jquery-release tasks

jzaefferer:
Looking at [build/release.js](https://github.com/jquery/jquery/blob/master/build/release.js), you're not doing any custom commits. I've used jquery-release for QUnit's 1.16 release, which [includes the commitplease component list](https://github.com/jquery/qunit/commit/8bfa5d702a9551d5bfea5c4c62c9b51a19d976c2). I see no reason why you'd have any issues either.

jzaefferer:
I think the only commits are for the tag itself, which commitplease has an exception for, [the other](https://github.com/jquery/qunit/commit/c993979b6ad0777cf80780960dc611eea7831ded) is prefixed with "Build:", which you're listing here as a valid component. So yes, its fine!

package.json
```diff
@@ -55,5 +55,30 @@
     "build": "npm install && grunt",
     "start": "grunt watch",
     "test": "grunt"
+  },
+  "commitplease": {
+    "components": [
+      "Docs",
+      "Tests",
+      "Build",
+      "Release",
```
jzaefferer: What's this one? I think for UI and QUnit we dump release-related commits into "Build".

gibson042: 20ea8071b44c23e7915300eacc9ed958fbe4eaf5
de7702ac59889177b0dd622d3b5503004abf0258
1d931068d30b6a516b883cf7f420606de21eab68

They're distinct enough from Build to make this a worthwhile component, in my opinion.

timmywil: @gibson042 +1

PR #1954, opened 2014-12-19
Tests: fix cross-domain detection test for non-default port

victor-homyakov:
Tests: fix cross-domain detection test for non-default port


dmethvin:
Sorry, I lost track of this. Looking at it with fresh eyes, I now realize what you are saying! The change is only in the unit test so I'm good with landing it as-is.

Issue #1765, opened 2014-10-21
offsetParent fails to recognize 3D transform context


mgol:
Originally reported by jlukic at: http://bugs.jquery.com/ticket/14956

<p>
According to W3C spec (and browser behavior) using a 3D transformation creates a new rendering context similar to <strong>position:relative;</strong> 
</p>

<p>
jQuery however will not return this as the offset parent, it will return incorrectly the closest container with position:relative;
</p>

<p>
See this JSFiddle: <a class="ext-link" href="http://jsfiddle.net/uUJDd/1/" rel="nofollow"><span class="icon">&nbsp;</span>http://jsfiddle.net/uUJDd/1/</a>
</p>

<p>
The absolute positioned element is relative to the 3D transformed context, but reports the wrong offsetParent
</p>

<p>
This is crucial for third party libraries which use jQuery (like my library Semantic UI) which need to determine correct offsetParent when using 3D contexts. 
</p><p>Issue reported for jQuery 2.1.0</p>

markelog:
From the `jQuery#offsetParent` <a href="http://api.jquery.com/offsetParent/">documentation</a> - 

> An element is said to be positioned if it has a CSS position attribute of relative, absolute, or fixed. 

So at least `offsetParent` method behave as it should, which means that this would be a behaviour change.

> The code size/speed issue has me worried as well if we have to make .offsetParent() always climb the tree from the element and ignore the DOM .offsetParent property.

and 

> We'd need to look for the a non-none transform (and -webkit-transform?) property in the method as we climb up the tree and there are probably other special cases not anticipated.

plus 

> This appears to be an open issues with the spec
...
Looks like nobodies decided what to do.

This sounds horrible - possible dirty workaround, perf and byte size hit. And since spec is incomplete we can stuck with some code and logic that we don't want?

Whereas fix for it on the user side, sounds like a pretty easy thing to do - add `position:relative` to the transform node? Although i understand that might be not that easy, at least not always.

> This is crucial for third party libraries which use jQuery (like my library Semantic UI) which need to determine correct offsetParent when using 3D contexts.

Should this library use their own algorithm for "offsetParent" search? Although given proposed logic, it's hard to call this method as `offsetParent`, since it would deviate from spec significantly.

jlukic:
[...]

PR #1968, opened 2014-12-25
Offset: simplify jQuery#offsetParent method

markelog:
* It seems, check for html element (and previously for body element)
  was redundant

* Simplify "return" statement

* Add comment about potential errors that didn't find themselves in real life app

This code would work even in IE6 quirks mode

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -143,12 +143,19 @@ jQuery.fn.extend({
 		};
 	},
 
+	// This method will return documentElement in the following cases:
+	// 1) For the element inside the iframe without offsetParent, this method will return
+	//    documentElement of the parent window
+	// 2) For the hidden or detached element
+	// 3) For body or html element, i.e. in case of the html node - it will return itself
+	//
+	// but those exceptions were never presented as a real life use-cases
+	// and might be considered as more preferable results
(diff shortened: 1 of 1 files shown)


src/offset.js
```diff
@@ -143,12 +143,19 @@ jQuery.fn.extend({
...
 	},
 
+	// This method will return documentElement in the following cases:
+	// 1) For the element inside the iframe, this method will return
+	//    documentElement of the parent window
+	// 2) For the hidden or detached element
+	// 3) For body or html element, i.e. in case of the html node - it will return itself
+	//
+	// but those exceptions were never presented as a real life use-cases
+	// and might be considered as more preferable results
```
markelog: How about i will add a notice about it, like "But this is undefined and undocumented behavior, could be changed at any point" or something like that.

It feels weird to know about these edge cases, but not put some info about them somewhere.

dmethvin: It really does seem like this is documentation. I am okay with putting a comment like "this is not guaranteed" nearby though and we can hope we don't need to change it.

dmethvin:
Seems like this should address gh-1765 as well since that will change the same code.

PR #1975, opened 2014-12-26
Ajax: simplify XHR detection.

### LINKED ISSUE #1967: Simplify XHR detection
This is the `XMLHttpRequest` object detection code:

https://github.com/jquery/jquery/blob/master/src/ajax/xhr.js

``` javascript
jQuery.ajaxSettings.xhr = function() {
    try {
        return new XMLHttpRequest();
    } catch ( e ) {}
};
...
xhrSupported = jQuery.ajaxSettings.xhr();
...
support.cors = !!xhrSupported && ( "withCredentials" in xhrSupported );
support.ajax = xhrSupported = !!xhrSupported;
```

`support.ajax` is used nowhere in the code and is not documented, `XMLHttpRequest` object is supported in all modern browsers (including Android 2.3, iOS 6.1, Opera 12 and IE 9 — [caniuse.com](http://caniuse.com/#feat=cors)).
Why not remove `xhrSupported` variable together with `support.ajax` and `try-catch` block?

FarSeeing:
Closes #1967 

```
raw     gz Compared to master @ ab20d9d24f0a95df382c02e9ef3dcc4adb86e4f1    
-124   -30 dist/jquery.js                                                   
-46    -31 dist/jquery.min.js
```

### DIFF
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -1,13 +1,10 @@
 define([
 	"../core",
-	"../var/support",
-	"../ajax"
+	"../var/support"
 ], function( jQuery, support ) {
 
 jQuery.ajaxSettings.xhr = function() {
-	try {
-		return new XMLHttpRequest();
-	} catch ( e ) {}
+	return new XMLHttpRequest();
 };
 
 var xhrId = 0,
@@ -18,8 +15,7 @@ var xhrId = 0,
 		// Support: IE9
 		// #1450: sometimes IE returns 1223 when it should be 204
 		1223: 204
-	},
-	xhrSupported = jQuery.ajaxSettings.xhr();
+	};
 
 // Support: IE9
 // Open requests must be manually aborted on unload (#5280)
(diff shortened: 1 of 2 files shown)


dmethvin:
Per discussion in the meeting today, we were thinking about using XHR as the default transport if no other transport was found. That would eliminate some of the special cases and avoid the need for `jQuery.support.cors` since it would try XHR anyway. @jaubourg does that sound feasible?

jaubourg:
Everything is feasible but I really don't see the appeal:

1. it implies that `$.ajax` cannot be used without the XHR transport which breaks modularity
2. it creates a late failure (at the moment of performing the first request) whereas it's currently possible to pre-emptively not call `$.ajax` requests by checking the support property.
3. it creates a late failure for CORS requests in environments that do not support it with no way to anticipate it.

As of now, `$.ajax` presents a unified error path for when there is no transport, including situations where there is no XHR implementation or CORS is not supported. It also provides a way to anticipate failures and take pre-emptive actions. It is of great importance now that we're actively working on supporting non-browser environments.

I think we're getting much too eager to reduce surface API and byte count and I'd rather document more than cut stuff around because "we gain 10 bytes" or "it's on `$.support` and we don't document it".

That being said, since the XHR factory is on `$.ajaxSettings` we _could_ nullify it when there is no XHR implementation and we _could_ add a `cors` property there too. That would bring those in the realm of things we document but, to be frank, I'd rather document a subset of `$.support` (the ajax stuff) rather than move them around.

timmywil:
I think I'm with @jaubourg here. I don't necessarily agree with 1 – it's still modular because the xhr transport is only required if another transport isn't used first – but I think 2 and 3 are good points. Regardless, whatever we decide, this PR cannot be landed as-is.

timmywil:
Thank you for taking the time to put together a PR, but I'm closing this because it needs modification regardless. This can be reopened if necessary.

Issue #1784, opened 2014-10-21
Issues using ShadowDom and ShadowRoot elements


mzgol:
Originally reported by rictic at: http://bugs.jquery.com/ticket/15201
<p>
JSBin link to reproduce: <a class="ext-link" href="http://jsbin.com/tamugo/2/edit" rel="nofollow"><span class="icon">&nbsp;</span>http://jsbin.com/tamugo/2/edit</a> (try in Chrome (or Firefox with dom.webcomponents.enabled turned on))
</p>
<p>
An element that is in the shadow dom appears in some ways to not be contained in the document. .offset() assumes that if the element isn't contained in the document then it's disconnected and returns an offset of {top: 0, left: 0}.
</p>
<p>
When .offset() checks to see if the element is disconnected it could try walking up the DOM not only by checking .parentNode as .contains() does but also by looking at .host, which is how one travels up from a shadow root into its containing element.
</p><p>Issue reported for jQuery 2.1.1</p>

NekR:
```javascript
if (elem.createShadowRoot) {
  var rect = elem.getBoundingClientRect();
  // disconnected or hidden
  if (!rect.height && !rect.width) {
    return box;
  }

  box = rect;
} else {
  // Make sure it's not a disconnected DOM node
  if ( !jQuery.contains( docElem, elem ) ) {
    return box;
  }

  box = elem.getBoundingClientRect();
}
```

This is another way to fix this issue. This method does not need traversation over tree to detect if element is disconnected. ```getBoundingClientRect()``` does nothing on disconnected nodes and returns always zeroes. It also returns zeroes on hidden elements. So if element is hidden or disconnected -- we return zeroes. If it visible and in the tree (anywhere in the tree) -- then we return real box.

One problem here is that for hidden elements, jQuery currently returns box ```{ top: 0 + pageYOffset, left: 0 + pagetXOffset }``` and for disconnected elements returns always ```{ top: 0, left: 0 }```.
I do not know if this artifact is useful or if someone uses it. Probably it might break compatibility.

Anyway, I will just leave it here as an alternative solution.

NekR:
Also, as far as I see, there is no tests for such behavior. And if you think what that method will not break compatibility, then we can even drop the condition there and execute always first part.

dmethvin:
It's possible we could use something like the solution @NekR proposes. The [docs for .offset()](http://api.jquery.com/offset/) specifically say it doesn't work for  hidden elements.

[...]

PR #1976, opened 2014-12-26
Offset: `$.fn.offset()` on element inside shadowRoot

nazar-pc:
`$.fn.offset()` didn't worked on `html>...>shadowRoot>...>target_element`, while `target_element.getBoundingClientRect` worked fine with correct result.
The reason was that check for disconnected node made by checking whether `document` contains `target_element` fails, which is correct, element inside shadowDOM, not in main DOM.

That is why I've added recursive traversal from element to its parent until it reaches `document` or until `parent_element.parentNode` is `falsy` (for disconnected nodes it will be `null`, if pass some badly composed object - it may be `undefined`, so, just `falsy`, in this case we just return).
If we encounter `DocumentFragment` (`shadowRoot`), we take `parent_element.host` property instead of `parent_element.parentNode` in order to get parent element of that `shadowRoot`.
Also it was necessary to add explicit check for `target_element.getBoundingClientRect` property.

This is modern, Web Components-aware version of ea507b3e998126ae1f94f4fd1618994d645c9cc8

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -96,10 +96,6 @@ jQuery.fn.extend({
 
 		// We have element in ShadowDOM, need recursive traversal from element to its parent
 		if ( elem.matches && elem.matches( ":host *" ) ) {
-			if ( !elem.getBoundingClientRect ) {
-				return box;
-			}
-
 			// Make sure it's not a disconnected DOM node
 			while ( true ) {
 				elemParent = elemParent.parentNode || elemParent.host;
--- a/src/offset.js
+++ b/src/offset.js
@@ -97,7 +97,7 @@ jQuery.fn.extend({
 		// We have element in ShadowDOM, need recursive traversal from element to its parent
 		// First checking for shadowRoot creation support
 		// If that is present - we have non-prefixed element.matches() support
-		if ( Element.prototype.createShadowRoot && elem.matches( ":host *" ) ) {
+		if ( elem.createShadowRoot && elem.matches( ":host *" ) ) {
 			// Make sure it's not a disconnected DOM node
 			while ( true ) {
 				elemParent = elemParent.parentNode || elemParent.host;
--- a/src/offset.js
+++ b/src/offset.js
@@ -92,21 +92,29 @@ jQuery.fn.extend({
 			return;
 		}
 
-		if ( !elem.getBoundingClientRect ) {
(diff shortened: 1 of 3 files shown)


src/offset.js
```diff
@@ -93,9 +94,26 @@ jQuery.fn.extend({
 
 		docElem = doc.documentElement;
 
-		// Make sure it's not a disconnected DOM node
-		if ( !jQuery.contains( docElem, elem ) ) {
-			return box;
+		// We have element in ShadowDOM, need recursive traversal from element to its parent
+		if ( elem.matches && elem.matches( ":host *" ) ) {
+			if ( !elem.getBoundingClientRect ) {
```
dmethvin: Is there any situation where a browser would support shadowDOM but not `getBoundingClientRect`? seems unlikely.

dmethvin:
This needs unit tests.

src/offset.js
mzgol: We don't check form `getBoundingClientRect` at all, I removed this `if` in my commit dropping older browsers.

dmethvin: Oh right, that was for some old BlackBerry.

nazar-pc: Removed that

nazar-pc:
Just merged upstream and dropped `getBoundingClientRect` check.
@dmethvin, how to run unit test properly here? Because to test this we need shadowDOM support from browser.
I've tried `(document.body.shadowRoot === undefined ? skip : test)(...)`, but `skip` is unavailable.

mzgol:
> skip is unavailable.

Try `QUnit.skip`; we'll need to migrate all those globals to the QUnit namespace some time, perhaps `skip` isn't exposed as a global at all since it's a new API.

nazar-pc:
Thanks, @mzgol, it works.
Added unit test, will be skipped if browser do not have native support for shadowDOM.

deshawnbw:
Anything holding this up? Encountered this issue while using web components as well.

nazar-pc:
Please, read discussion #1784 and pull request #2043.
There is different pull request that addresses this issue, also code might be simplified for 3.0 due to dropping support for some old browsers.

gibson042:
Closing in favor of #2043, which is just about ready to land.

PR #1980, opened 2014-12-30
Build: Rearrange grunt/npm tasks into a build/dist/test pattern

gibson042:
Ref 76df9e4e389d80bff410a9e5f08b848de1d21a2f
Ref https://github.com/jquery/jquery/pull/1949#discussion_r22042517

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -155,26 +155,14 @@ module.exports = function( grunt ) {
 	// Integrate jQuery specific tasks
 	grunt.loadTasks( "build/tasks" );
 
-	grunt.registerTask( "lint", [ "jshint", "jscs" ] );
-
-	grunt.registerTask( "node_smoke_test", function() {
-	    var done = this.async();
-		require( "jsdom" ).env( "", function( errors, window ) {
-			if ( errors ) {
-				console.error( errors );
-				done( false );
-			}
-			require( "./" )( window );
-			done();
-		});
-	});
-
 	// Short list as a high frequency watch task
 	grunt.registerTask( "dev", [ "build:*:*", "lint" ] );
 
+	grunt.registerTask( "lint", [ "jsonlint", "jshint", "jscs" ] );
+
 	grunt.registerTask( "test_fast", [ "node_smoke_test" ] );
 
-	grunt.registerTask( "test", [ "default", "test_fast" ] );
+	grunt.registerTask( "test", [ "test_fast" ] );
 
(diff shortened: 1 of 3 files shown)


markelog:
So at `grunt test` we would not lint anything? But at ci-run will execute those tests twice? 

This:

```js
grunt.registerTask( "test_fast", [ "node_smoke_test" ] );
... 
grunt.registerTask( "test", [ "test_fast" ] );
```

Is look like very weird, `test` is `test_fast` and `test_fast` is `node_smoke_test`, confusing, i assume, this is done for the future but it looks pretty incoherent to me.

Whereas this `node_smoke_test` just require stuff, not running qunit tests, i don't see the point doing this everytime for the `default` task, you will find more issues with your code with `jshint` task, not with this one, this smoke stuff could find the problem but in very rare cases. For me, it would be even okay, if we would run it before the release, certainly not everytime, but that's me.

I feel we redoing a lot and make arguably confusing changes for very small benefit. I'd create an npm command, that runs this jsdom require and execute it with ci-run or even without npm command and only with grunt task just like `grunt testswarm`

gibson042:
> So at `grunt test` we would not lint anything? 

Right; testing and linting are separate processes.

> But ci-run will execute those tests twice?

The fast ones, yes. Or we could introduce a currently-empty `test_slow` (for eventually population by the full test suite) and use that to avoid duplication. The goal here is to set up a basic structure: lint (preceded by build for us because our source is deconstructed), output to `dist` (including minification/tree-shaking/etc.), test the result.

Gruntfile.js
```diff
@@ -155,26 +155,14 @@ module.exports = function( grunt ) {
...
-			}
-			require( "./" )( window );
-			done();
-		});
-	});
-
 	// Short list as a high frequency watch task
 	grunt.registerTask( "dev", [ "build:*:*", "lint" ] );
 
+	grunt.registerTask( "lint", [ "jsonlint", "jshint", "jscs" ] );
```
markelog: Probably better to register it above the "dev" task

markelog:
I think it would fine if have something to structure, now we only have this require and couple tasks that do this require, that do the same thing, "test_fast" name now doesn't make sense, at least now.

gibson042:
> "test_fast" name doesn't make sense, at least now.

As it turns out, #1821 or #1996 or similar will introduce such a need... `node_smoke_test` is fast, but `promises-aplus-tests` is slow.

PR #1982, opened 2015-01-01
Updated license date

benhc123:
Updated license date to 2015 for new year.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright 2005, 2014 jQuery Foundation and other contributors,
+Copyright 2005, 2015 jQuery Foundation and other contributors,
 https://jquery.org/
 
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Thanks, but we're actually going to be removing the dates.

Kangaroopower:
I've always wondered about this. What happens if the date wasn't up to date? Like could someone use it for nefarious purposes if they grabbed the copy at exactly 12:00:01 on Jan 1? And if there are no dates, what are the legal implications? Honestly curious. Thanks

scottgonzalez:
Nobody can do anything nefarious for an out of date copyright notice. If you want a more detailed answer, you should consult a lawyer. If you want a definitive answer, you'd probably need to get a case in front of a judge, since lawyers often disagree on what's required in the copyright notice.

PR #1983, opened 2015-01-01
Build: Remove dates from copyright notice

agcolom:
I'm not too sure about the change in Gruntfile.js, so please let me know if you would like it to be different, e,g, whether the jquery.org/license should be on the same line as the Copyright notice, whether it should be https://jquery.org/ or https://jquery.org/license also, or anything else.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -137,7 +137,7 @@ module.exports = function( grunt ) {
 						"ascii_only": true
 					},
 					banner: "/*! jQuery v<%= pkg.version %> | " +
-						"Copyright jQuery Foundation and other contributors " +
+						"(c) jQuery Foundation, Inc. | " +
 						"jquery.org/license */",
 					compress: {
 						"hoist_funs": false,
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -137,7 +137,7 @@ module.exports = function( grunt ) {
 						"ascii_only": true
 					},
 					banner: "/*! jQuery v<%= pkg.version %> | " +
-						"(c) 2005, <%= grunt.template.today('yyyy') %> jQuery Foundation, Inc. | " +
+						"Copyright jQuery Foundation and other contributors " +
 						"jquery.org/license */",
 					compress: {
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -137,7 +137,7 @@ module.exports = function( grunt ) {
 						"ascii_only": true
 					},
 					banner: "/*! jQuery v<%= pkg.version %> | " +
-						"(c) 2005, <%= grunt.template.today('yyyy') %> jQuery Foundation, Inc. | " +
+						"Copyright jQuery Foundation and other contributors " +
```
mzgol: Why the change from `(c)` to `Copyright`? You didn't do the same in the Sizzle PR: https://github.com/jquery/sizzle/pull/307/files

This is the banner put in the minified files, e.g. https://code.jquery.com/jquery-2.1.3.min.js

mzgol: I'd leave `|` at the end, it separates the copyright info from the URL to the license (see the linked minified file for example).

agcolom: @mzgol Thanks for the comments. I'll make the changes now.

mzgol:
LGTM

Gruntfile.js
```diff
@@ -137,7 +137,7 @@ module.exports = function( grunt ) {
 						"ascii_only": true
 					},
 					banner: "/*! jQuery v<%= pkg.version %> | " +
-						"(c) 2005, <%= grunt.template.today('yyyy') %> jQuery Foundation, Inc. | " +
+						"(c) jQuery Foundation, Inc. | " +
```
staabm: In the intro file sou also dropped the `Inc`

markelog: +1, why this is being dropped btw?

dmethvin: We don't need them, according to lawyers. So removing the dates makes it easier to maintain! :smiley:

dmethvin: I'll remove the `Inc.` here for consistency.

PR #1985, opened 2015-01-01
Core: Standardize indexOf comparisons

gibson042:
not present: `< 0`
present: `> -1`
at index: `=== N`

(cherry picked from commit 509406dd7c2763108819f9f29568309f48726237)

```
   raw     gz Compared to compat @ 2866da9e12b7cf339667ae894ffe072f4eb673de    
    +9     +4 dist/jquery.js                                                   
    +1     -4 dist/jquery.min.js
```

Ref #1984

### DIFF
--- a/src/ajax/jsonp.js
+++ b/src/ajax/jsonp.js
@@ -25,7 +25,8 @@ jQuery.ajaxPrefilter( "json jsonp", function( s, originalSettings, jqXHR ) {
 		jsonProp = s.jsonp !== false && ( rjsonp.test( s.url ) ?
 			"url" :
 			typeof s.data === "string" &&
-				!( s.contentType || "" ).indexOf("application/x-www-form-urlencoded") &&
+				( s.contentType || "" )
+					.indexOf("application/x-www-form-urlencoded") === 0 &&
 				rjsonp.test( s.data ) && "data"
 		);
 
--- a/src/ajax/load.js
+++ b/src/ajax/load.js
@@ -24,7 +24,7 @@ jQuery.fn.load = function( url, params, callback ) {
 		self = this,
 		off = url.indexOf(" ");
 
-	if ( off >= 0 ) {
+	if ( off > -1 ) {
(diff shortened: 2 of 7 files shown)


dmethvin:
I'm good with this, consistency is always nice if there's no compelling reason to be inconsistent.

markelog:
Yeah, but i fear we will forget about this convention pretty quicky, plus it's not standardized across all the projects

gibson042:
It's a little esoteric, but we _could_ put this in the style guide. The corresponding text would be something like "strict ordered inequality" (this PR), "ordered comparison to zero" (`index < 0`; `index >= 0`), "strict equality comparison" (`index !== -1`, `index === -1`), etc. At a glance, [globalize](https://github.com/jquery/globalize) seems to use strict equality comparison and [qunit](https://github.com/jquery/qunit) is inconsistent. I'm not feeling ambitious enough to look at [jquery-mobile](https://github.com/jquery/jquery-mobile) or [jquery-ui](https://github.com/jquery/jquery-ui) right now.

gibson042:
I just checked the git builds for each. The [modes](http://en.wikipedia.org/wiki/Mode_(statistics)) for Mobile are (`index === -1`, `index > -1`) and for UI are (`index === -1`, `index !== -1`), although each has a single "not present" `index < 0` and both exhibit a few "present" variations (Mobile even has three `!!~index` patterns).

cc @arschmitz, @scottgonzalez

scottgonzalez:
I feel like this is getting into the weeds, unless there's a strong reason to be extremely consistent.

With that being said, if a linter is going to find these, then I don't care what we do.

arschmitz:
Im not strongly opinionated here either. If we are going to do something though i would like it to be in the style guide or as @markelog said it will quickly be forgotten and not enforced.

gibson042:
Closed by 6ae222a54f687e7950a3f4c4e04bd42ca0387504

PR #1988, opened 2015-01-04
Core: Add support to tag-hyphenated elements

### LINKED ISSUE #1987: Core: Issue while creating dash-delimited nodes
JQuery fails in some cases to create dash-delimited notes:

```javascript
// Works fine and creates the element:
$('<div-d></div-d>');

// Does not create the element:
$('<tr-d></tr-d>');

// ... and it should, since document.createElement works fine too:
document.createElement('tr-d');
```

Reported first here: https://github.com/angular/angular.js/issues/10617

LeonardoBraga:
Fixes #1987

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -16,8 +16,8 @@ define([
 ], function( jQuery, concat, push, access, rcheckableType, support, dataPriv, dataUser ) {
 
 var
-	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([\w:]+)[^>]*)\/>/gi,
-	rtagName = /<([\w:]+)/,
+	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([\w:-]+)[^>]*)\/>/gi,
+	rtagName = /<([\w:-]+)/,
 	rhtml = /<|&#?\w+;/,
 	rnoInnerhtml = /<(?:script|style|link)/i,
 	// checked="checked" or checked
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -637,6 +637,19 @@ test("jQuery('html')", function() {
 	equal( jQuery( "\\<div\\>" ).length, 0, "Ignore escaped html characters" );
 });
 
+test("jQuery(tag-hyphenated elements) gh-1987", function() {
+	expect( 17 );
+
+	jQuery.each( "thead tbody tfoot colgroup caption tr th td".split(" "), function( i, name ) {
+		var j = jQuery("<" + name + "-d></" + name + "-d>"); 
+		ok( j[0], "Create a tag-hyphenated elements" );
+		ok( jQuery.nodeName(j[0], name.toUpperCase() + "-D"), "Tag-hyphenated element has expected node name" );
+	});
+
+	var j = jQuery("<tr-multiple-hyphens></tr-multiple-hyphens>");
+	ok( jQuery.nodeName(j[0], "TR-MULTIPLE-HYPHENS"), "Element with multiple hyphens in its tag has expected node name" );
+});
+
 test("jQuery('massive html #7990')", function() {
 	expect( 3 );
 
--- a/test/unit/manipulation.js
+++ b/test/unit/manipulation.js
@@ -486,6 +486,23 @@ test( "html(String) with HTML5 (Bug #6485)", function() {
 	equal( jQuery("#qunit-fixture").children().children().children().length, 1, "Make sure nested HTML5 elements can hold children." );
 });
 
+test( "html(String) tag-hyphenated elements (Bug #1987)", function() {
+
+	expect( 27 );
+
+	jQuery.each( "thead tbody tfoot colgroup caption tr th td".split(" "), function( i, name ) {
+		var j = jQuery("<" + name + "-d></" + name + "-d><" + name + "-d></" + name + "-d>"); 
+		ok( j[0], "Create a tag-hyphenated element" );
+		ok( jQuery.nodeName(j[0], name.toUpperCase() + "-D"), "Hyphenated node name" );
+		ok( jQuery.nodeName(j[1], name.toUpperCase() + "-D"), "Hyphenated node name" );
+	});
+
+	var j = jQuery("<tr-multiple-hyphens><td-with-hyphen>text</td-with-hyphen></tr-multiple-hyphens>");
+	ok( jQuery.nodeName(j[0], "TR-MULTIPLE-HYPHENS"), "Tags with multiple hypens" );
+	ok( jQuery.nodeName(j.children()[0], "TD-WITH-HYPHEN"), "Tags with multiple hypens" );
+	equal( j.children().text(), "text", "Tags with multple hypens behave normally" );
+});
+
 test( "IE8 serialization bug", function() {
 
 	expect( 2 );
--- a/src/core/var/rsingleTag.js
+++ b/src/core/var/rsingleTag.js
@@ -1,4 +1,4 @@
 define(function() {
 	// Match a standalone tag
-	return (/^<(\w+)\s*\/?>(?:<\/\1>|)$/);
+	return (/^<([\w-]+)\s*\/?>(?:<\/\1>|)$/);
 });


test/unit/core.js
```diff
@@ -627,6 +627,7 @@ test("jQuery('html')", function() {
 
 	ok( jQuery("<div></div>")[0], "Create a div with closing tag." );
 	ok( jQuery("<table></table>")[0], "Create a table with closing tag." );
+	ok( jQuery("<tr-d></tr-d>")[0], "Create a dash-delimited tag." );
```
rwaldron: I don't mean to be pedantic, but that's not "delimited". Suggest: "hyphenated".

LeonardoBraga: Not a problem! I changed as suggested. Please let me know if anything requires adjustments.

test/unit/manipulation.js
```diff
@@ -1620,6 +1620,8 @@ function testHtml( valueObj ) {
 	tmp = " " + tmp.slice( 1 );
 	equal( div.html(valueObj( tmp )).html().replace( />/g, "&gt;" ), tmp, "Escaped html, leading space" );
 
+	equal( jQuery("<tr-d></tr-d><tr-d></tr-d>").length, 2, "Create tag-hyphenated elements." );
```
markelog: I would want more assertions for this, so it would be more thorough, like check for nodeName and for other elements like <a href="https://github.com/jquery/jquery/blob/740e190223d19a114d5373758127285d14d6b71e/src/manipulation.js#L31-47">this</a>. No need to do this for core module though.

And i would prefer if those assertions would be in different test.

src/manipulation.js
```diff
@@ -17,7 +17,7 @@ define([
 
 var
 	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([\w:]+)[^>]*)\/>/gi,
-	rtagName = /<([\w:]+)/,
+	rtagName = /<([\w:-]+)/,
```
markelog: So we have ":" in manipulation but not in core, also no tests for ":".

@dmethvin how you want to handle this? We either could add tests for ":" names, which probably means we would need to support and add tests for the core module too.

Or just remove this sign.

/cc @timmywil

markelog: @LeonardoBraga this is different issue though, nevermind that.

markelog:
Could break this pull on two commits? One for core, other for manipulation, i think it would be best way to do this. You can squash other commits, don't need them.

Also please verify it this change would work in `compat` branch.

test/unit/manipulation.js
mzgol: And maybe a case with two (or more) hyphens, just in case. It's obvious current implementation will work in the same way but it's better to be future-proof.

markelog: +1 the more tests the better, especially if this involves regexps

LeonardoBraga:
I'm implementing the requested changes and will update this PR soon, however I noticed that `rsingleTag` does not contain `:`, while `rtagName` does. Shouldn't we add `:` to `rsingleTag` in order to keep consistency?

markelog:
@LeonardoBraga i guess you missed my comment - https://github.com/jquery/jquery/pull/1988#discussion_r22439258

src/manipulation.js
gibson042: `rsingleTag` is the test for element creation via [`document.createElement`](https://github.com/jquery/jquery/blob/740e190223d19a114d5373758127285d14d6b71e/src/core/parseHTML.js#L21-L27), and therefore shouldn't include ":".

`rtagName` is for [finding the first open tag in an HTML string](https://github.com/jquery/jquery/blob/740e190223d19a114d5373758127285d14d6b71e/src/manipulation.js#L220), and therefore _should_ include ":" (as should its partner `rxhtmlTag`).

All of the above should probably find its way into source comments.

gibson042: ...and all three regexes should include "-" in this PR.

LeonardoBraga: Thanks @gibson042. As soon as I won the fight with `git rebase`, I'll push the update.

markelog: > rsingleTag is the test for element creation via document.createElement

This is true

> and therefore shouldn't include ":".

This isn't

> rtagName is for finding the first open tag in an HTML string

This is true

> and therefore should include ":"

This isn't

`rsingleTag` is needed to identify single tag, single tag name might include ":" sign, which could be processed by `createElement` method.

`rtagName` doesn't need to identify xml tags, only html tags, since only html tags need additional wrapping, so it could leave without that symbol.

Re-reading my previous comment, i probably wasn't clear - should we drop the ":" sign in `rtagName` regexp or add support for those tags in core module? Either way, i would like to be consistent

LeonardoBraga:
@markelog Applied and tested the changes locally on `compat` and they seem to be fine. Should I provide a PR for that branch too?

test/unit/core.js
```diff
@@ -637,6 +637,17 @@ test("jQuery('html')", function() {
 	equal( jQuery( "\\<div\\>" ).length, 0, "Ignore escaped html characters" );
 });
 
+test("jQuery('tag-hyphenated elements #1987')", function() {
+	expect( 3 );
+
+	var j = jQuery("<tr-d></tr-d>"); 
+	ok( j[0], "Create a tag-hyphenated elements" );
+	equal( j[0].nodeName, "TR-D", "Tag-hyphenated element has expected node name" );
```
markelog: use `jQuery.nodeName` here

test/unit/core.js
```diff
@@ -637,6 +637,17 @@ test("jQuery('html')", function() {
 	equal( jQuery( "\\<div\\>" ).length, 0, "Ignore escaped html characters" );
 });
 
+test("jQuery('tag-hyphenated elements #1987')", function() {
```
markelog: Better text would be "jQuery(tag-hyphenated elements) gh-1987"

markelog:
In https://github.com/jquery/jquery/pull/1988#discussion_r22439239, i meant all of them, like here -  https://github.com/jquery/jquery/blob/66e1b6b8d49812239b5712d65922ff94c60f7b02/test/unit/manipulation.js#L264-268, just to be safe

markelog:
> Should I provide a PR for that branch too?

That would be preferable

src/manipulation.js
markelog: > as should its partner rxhtmlTag

Nice catch this is, but parser should be smart enough already.

gibson042: We want no ":" in `rsingleTag` so elements with the character in their name are created by the namespace-aware `innerHTML` instead of the namespace-ignorant `document.createElement`: https://rawgit.com/gibson042/fdedc0a26c25ffbfc014/raw/9a5487e7aad4282641a498bf95ccd9299a747c16/rdf.xhtml

We want a ":" in `rtagName` and `rxhtmlTag` so the former correctly identifies the first open tag and the latter doesn't mangle input: http://jsfiddle.net/5y6p8777/

markelog: > instead of the namespace-ignorant document.createElement:

It doesn't matter if it's ignorant or not, there is no real difference between tags created with explict namespace or not, for example in `attrs` module we always use `setAttribute` which is not namespace aware, comparing to `setAttributeNS` method, which proves that this is not real use-case.

Bottom line is - `createElement` method could handle colon case and can create required element.

> so the former correctly identifies the first open tag and the latter doesn't mangle input:

I wasn't talking about `rxhtmlTag` only about `rtagName` which doesn't requires colon, because, i as i said before, we only need to indentify html tags, and ":" sign could probably be dropped from `rxhtmlTag` too, since even it wouldn't correctly do the replace operation, browser parser is still smart enough, but this requires further investigation if we would want to go down that road.

That said, i can admit, that if we will leave everything as it is now, it would be less hacky. But what we should definitely do is add tests for these cases.

LeonardoBraga: > But what we should definitely do is add tests for these cases.

I'd appreciate if you could be explicit about which tests cases (nodes) you still want to include. I'm glad to add as many as required to ensure the change is well covered.

gibson042:
@gkalpak makes a valid point... in fact, HTML tag names [must start with an ASCII _letter_](https://html.spec.whatwg.org/multipage/syntax.html#tag-open-state), but can then include [almost anything](https://html.spec.whatwg.org/multipage/syntax.html#tag-name-state). Let's update all three regular expressions accordingly, so that the capture consists of a case-insensitive a–z and any number of non-whitespace non-slash non-greater-than non-NULL characters.

dmethvin:
Wow, so ASCII letters are just lowercased and pretty much any other non-space character is allowed. That's mighty liberal. So it does sound like we need to adjust our regex.

LeonardoBraga:
I'll update this PR to also address the concerns regarding valid tag names.

Regarding the previous state, it's not clear to me how we stand on:
1. Should I add `:` to all regexes (trusting Wikipedia on the plural here) to keep consistency or leave them as they are? This might mean a few extra bytes, but it's a "price" worth paying for consistency, IMO.
2. Should I extend all the tests cases that I added for `-` to also test `:` or this is not necessary?

Thank you - I'm glad this exposed some other cases that needed attention as well. I'll try to wrap up this quickly.

gibson042:
I stand by my [previous position](https://github.com/jquery/jquery/pull/1988#discussion_r22447360), because ":" is special.

> We want no ":" in `rsingleTag` so elements with the character in their name are created by the namespace-aware `innerHTML` instead of the namespace-ignorant `document.createElement`: https://rawgit.com/gibson042/fdedc0a26c25ffbfc014/raw/9a5487e7aad4282641a498bf95ccd9299a747c16/rdf.xhtml

> We want a ":" in `rtagName` and `rxhtmlTag` so the former correctly identifies the first open tag and the latter doesn't mangle input: http://jsfiddle.net/5y6p8777/

Where the statements now refer to acceptance of the character, rather than its literal appearance in the regex.

markelog:
I think we should merge this as is, this PR solves real problem that users faced, "almost anything" is nice to have and could be solved in another PR

markelog:
This, and i can't believe we forgot about it, doesn't work in IE8, remember why we have <a href="https://github.com/jquery/jquery/blob/96447575c23f7b313091750ed84cec05db3a25c6/src/manipulation.js#L17-33">shim</a>?

Right, now i do too. My first thought was to revert, now i think we should blacklist those hyphenated tests for the old one. Since if user uses custom elements he definitely doesn't use it with IE8.

dmethvin:
:smiling_imp: 

I don't think we need to revert, we can just skip it for IE8.

markelog:
> we can just skip it for IE8.

Will do that.

@LeonardoBraga note that you would have to do the same thing for https://github.com/jquery/jquery/issues/2005

LeonardoBraga:
> you would have to do the same thing

Do you mean to blacklist the tests I'll write for IE8? If so, is there a place where this is already done in the code so I can follow the same pattern?

markelog:
See the https://github.com/jquery/jquery/commit/87bb713cc0e5b5aca86b6839894e9f3d8d635b36

mzgol:
> This, and i can't believe we forgot about it, doesn't work in IE8, remember why we have the shim?

Isn't it enough to `document.createElement` it to make IE8 recognize it? The developer might be doing that and using custom tags on IE8 so maybe we should test this scenario.

markelog:
> Isn't it enough to document.createElement it to make IE8 recognize it? 

Yes.

> The developer might be doing that and using custom tags on IE8 so maybe we should test this scenario.

Even though this is unlikely use-case, we still can't do that, oldIE will identify only those elements that created in the scope of the document it first added to, which is in our case, is internal `documentFragment` element, which is not exposed.

dmethvin:
I thought for sure IE8 would die on special chars but at least that part works. I don't think we want to guarantee this madness works consistently in IE8 anyway.
capture

gibson042:
> I don't think we want to guarantee this madness works consistently in IE8 anyway.

Agreed. As long as our code is solid, it's OK for some edge cases to fail on IE8.

markelog:
> I thought for sure IE8 would die on special chars but at least that part works.

If wouldn't do something crazy like find those elems in user string and create them on our `documentFragment` it wouldn't work for the same reason hyphenated elements fall behind the cracks.

mbest:
Why is this only being fixed in version 3.0.0? For anyone using jQuery for template-like support across multiple browsers, this change is essential for full support of custom elements. I think it should be back ported to 1.x and 2.x. This shouldn't be hard since it's a small, non-breaking change.

mgol:

mgol:
I moved the milestone from this PR to the issue #1987.

Issue #1990, opened 2015-01-04
Locking memoryless Callbacks aborts current execution


gibson042:
http://jsfiddle.net/2d9gwd2c/

Note how `.lock()` aborts active execution unless the "memory" flag is set. This is arguably by design, but it seems to me that the purpose of the method is preventing effects of _future_ `fire` calls—execution interruption being more properly provided by the "stopOnFalse" flag.

This has probably gone unnoticed for lack of use (all of our internal callbacks have memory), but nonetheless would be a breaking fix and called for with a major version bump.

In code, this assertion fails:
```js
var completed,
    list = jQuery.Callbacks();
list.add( list.lock );
list.add(function() { completed = true; });
list.fire();
equal( completed, true );
```

Issue #1989, opened 2015-01-04
Single-fire Callbacks report as initially locked


gibson042:
http://jsfiddle.net/2d9gwd2c/

Note how the "once" flag breaks `.locked()`.

In code, this assertion fails:
```js
equal( jQuery.Callbacks("once").locked(), false );
```

PR #1991, opened 2015-01-04
Callbacks: Update locking behavior

gibson042:
Fixes gh-1989
Fixes gh-1990

```
   raw     gz Compared to compat @ 2866da9e12b7cf339667ae894ffe072f4eb673de    
  +398   +108 dist/jquery.js                                                   
   +18    +17 dist/jquery.min.js
```

I isolated size reduction into a [followup branch](https://github.com/gibson042/jquery/tree/2015-01-callbacks-reduce) this time. [Applying it](https://github.com/gibson042/jquery/compare/2015-01-callbacks...2015-01-callbacks-reduce) (which should happen together with landing this PR) changes the above as follows:
```
   raw     gz Compared to compat @ 2866da9e12b7cf339667ae894ffe072f4eb673de    
   -85     -8 dist/jquery.js                                                   
   -83    -52 dist/jquery.min.js
```

### DIFF
--- a/src/callbacks.js
+++ b/src/callbacks.js
@@ -190,7 +190,7 @@ jQuery.Callbacks = function( options ) {
 			lock: function() {
 				stack = undefined;
 				locked = true;
-				if ( !memory ) {
+				if ( !memory && !firing ) {
 					self.disable();
 				}
 				return this;
--- a/src/callbacks.js
+++ b/src/callbacks.js
@@ -51,6 +51,8 @@ jQuery.Callbacks = function( options ) {
 		memory,
 		// Flag to know if list was already fired
 		fired,
+		// Flag to prevent .fire/.fireWith
+		locked,
 		// End of the loop when firing
 		firingLength,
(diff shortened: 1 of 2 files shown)


src/callbacks.js
```diff
@@ -78,13 +81,21 @@ jQuery.Callbacks = function( options ) {
 				}
 			}
 			firing = false;
+
+			// If not disabled,
```
arthurvr: Does that comma have to be there?

gibson042: I suppose not, but it will be gone anyway if we land the followup: https://github.com/gibson042/jquery/compare/2015-01-callbacks...2015-01-callbacks-reduce#diff-67fa1bedebcff27953960d35db47f26bL85

dmethvin:
Okay, I'm down with it. Thanks for the extra comments. :+1:

gibson042:
Any thoughts, @jaubourg?

Issue #1986, opened 2015-01-02
Support passing url/data object to $.post and $.get


togakangaroo:
I've just spent yet another 30 minutes helping someone debug this.

The issue:

    $.post( {url: '/foo', data: bar})

generates a post request not at all to /foo. The reason is of course that this is not the signature of the method. The correct signature is

    $.post('/foo', bar)

The problem is that `$.ajax` *does* accept things in the former format and its oh so easy to forget which is which. (And of course it's jquery standard to allow configuration objects in most places.) Of course when you make the mistake, it is very difficult to figure out what is going on as it will still generate a post request, just to the current page with an [object Object] parameter making you question your sanity.

I must have spent days over the years debugging this and helping others debug it.

So, feature request: Can $.post and $.get please support passing of a configuration object as the first parameter? If not, can passing an object at least generate a `console.warn` to hint at what might be going on?

chris-allnutt:
+1, so frustrating

nichcurtis:
+1

CWSpear:
Ah, the dread post to `[object Object]`. Been there. Pretty to recognize by now, but I'm 100% in agreement for a warning to help people who haven't already made that mistake 100 times.

dmethvin:
I really wish we didn't have the "shortcut" methods at all, they just make things worse. Can you pretend they don't exist? You can also use `$.post = undefined` or other stubs in your own code to ensure they're not called.

We're not going to start checking args at runtime and spewing console warnings, period. Take a look at the [code for the shortcut methods](https://github.com/jquery/jquery/blob/76df9e4e389d80bff410a9e5f08b848de1d21a2f/src/ajax.js#L790-807) and see how much code it takes to make them accept an object. If it's not too much we could consider it.

togakangaroo:
You'll get no argument from me about those existing, usually wrap $.ajax up in my own facade but that's not really to the point. As long as that stuff exists people will keep using them incorrectly and wasting many hours debugging by accident. 

I take it you're saying that if I do a PR for this it will be considered?

boulosdib:
+1

dmethvin:
Yes, be sure to include some unit tests.

Also, stop the annoying +1 guys.

CWSpear:
:+1: stop all the +1's

dmethvin:
[...]

PR #1995, opened 2015-01-06
ajax: post and get accept config object [issue #1986]

togakangaroo:
See [this issue](https://github.com/jquery/jquery/issues/1986) for more discussion.

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -796,13 +796,14 @@ jQuery.each( [ "get", "post" ], function( i, method ) {
 			data = undefined;
 		}
 
-		return jQuery.ajax({
+		//extend to handle situation when url is actually a config object
+		return jQuery.ajax(jQuery.extend({
 			url: url,
 			type: method,
 			dataType: type,
 			data: data,
 			success: callback
-		});
+		}, (jQuery.isPlainObject(url) && url)));
 	};
 });
 
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -796,14 +796,14 @@ jQuery.each( [ "get", "post" ], function( i, method ) {
 			data = undefined;
 		}
 
-		//extend to handle situation when url is actually a config object
-		return jQuery.ajax(jQuery.extend({
+		// Extend to handle situation when url is actually a config object
+		return jQuery.ajax( jQuery.extend({
 			url: url,
 			type: method,
 			dataType: type,
 			data: data,
 			success: callback
-		}, (jQuery.isPlainObject(url) && url)));
+		}, jQuery.isPlainObject( url ) && url ) );
 	};
 });
 
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -2034,6 +2034,25 @@ module( "ajax", {
 		});
 	});
 
+	asyncTest( "jQuery[get|post]( config ) - simple with xml", 2, function() {
+		jQuery.when.apply(jQuery,
+			jQuery.map(["get", "post"], function(method) {
+				return jQuery[method]({
+					url: url("data/name.php"),
+					data: {
+						"xml": "5-2"
+					},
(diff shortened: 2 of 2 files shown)


dmethvin:
I love it when someone delivers. :star2:

src/ajax.js
```diff
@@ -796,13 +796,14 @@ jQuery.each( [ "get", "post" ], function( i, method ) {
...
-		return jQuery.ajax({
+		//extend to handle situation when url is actually a config object
+		return jQuery.ajax(jQuery.extend({
 			url: url,
 			type: method,
 			dataType: type,
 			data: data,
 			success: callback
-		});
+		}, (jQuery.isPlainObject(url) && url)));
```
dmethvin: Should be spaces inside the parens here per our style guide.

src/ajax.js
```diff
@@ -796,13 +796,14 @@ jQuery.each( [ "get", "post" ], function( i, method ) {
 			data = undefined;
 		}
 
-		return jQuery.ajax({
+		//extend to handle situation when url is actually a config object
```
dmethvin: You could leave out this comment, it's pretty clear already. If it stayed you'd need to edit it to add a leading space and capitalize per the style guide.

src/ajax.js
```diff
@@ -796,13 +796,14 @@ jQuery.each( [ "get", "post" ], function( i, method ) {
...
+		return jQuery.ajax(jQuery.extend({
```
dmethvin: Space after `.ajax( `

src/ajax.js
togakangaroo: Trying to find it in the style guide...you mean it should be `( jQuery.isPlainObject(url) && url )))`?

dmethvin: Although this allows a mixed call like `$.ajax({ options }, data)` I think we'd only want to document a single `options` object.

dmethvin: `( jQuery.isPlainObject( url ) && url ) ) )`

dmethvin: The outer parens aren't needed tho, right?

src/ajax.js
togakangaroo: I'd rather leave it since when I started reading jquery source this would have confused me quite a bit. I don't feel strongly about it though if you think I should remove it.

src/ajax.js
togakangaroo: No I don't think they're needed, just make it clearer that its a single expression. I'll remove.

dmethvin:
@togakangaroo Be sure that you've signed our CLA, https://contribute.jquery.org/CLA/

togakangaroo:
@dmethvin I signed it, thanks

togakangaroo:
I'm not super sure where/how to update the docs btw

test/unit/ajax.js
```diff
@@ -2034,6 +2034,25 @@ module( "ajax", {
 		});
 	});
 
+	asyncTest( "jQuery[get|post]( config ) - simple with xml", 2, function() {
+		jQuery.when.apply(jQuery,
```
mzgol: The same remark about spaces; there should be a space after `(`, `[` and before `)`, `]` etc.

test/unit/ajax.js
```diff
@@ -2034,6 +2034,25 @@ module( "ajax", {
...
+			jQuery.map(["get", "post"], function(method) {
```
mzgol: e.g. here:
```js
jQuery.map( [ "get", "post" ], function( method ) {
```

dmethvin:
I created https://github.com/jquery/api.jquery.com/issues/620 , The docs are at http://github.com/jquery/api.jquery.com/#readme, they're processed from XML so you'll want to go through the setup listed there or pull down the preconfigured VM so that you can be sure the markup looks good.

PR #1996, opened 2015-01-06
Deferred: Backwards-compatible standards interoperability

gibson042:
Ref https://github.com/jquery/jquery/pull/1821#issuecomment-67047762

Maintained backwards compatibility (outside of specifications):
* Don't process resolution values: `dfd.resolve( thenable ).done( expectsThenable )`
* Multi-valued resolution: `dfd.resolve( value1, value2, ... ).then( expectsMultipleArguments )`
* Progress: `dfd.notify( progress ).then( null, null, expectsInvocation )`
* Contextual resolution: `dfd.resolveWith( context, values ).then( expectsContext )`

Necessarily-breaking changes
* _Any_ `return` value (i.e., not thrown) from onRejected handlers can fulfill a `.then` Deferred: `dfd.reject().then( null, jQuery.noop ).done( expectsInvocation )`

Potentially-breaking behavior decisions (current state of this PR in **bold**):
* Thenables returned from `.then` onProgress handlers
  * progress: **ignore** vs. forward to outer onProgress vs. forward to outer onFulfilled vs. ???
  * rejection: **throw (asynchronous)** vs. ignore vs. forward to outer onRejected vs. ???
  * fulfillment: **forward to outer onProgress** vs. ignore vs. ???
* `.pipe` behavior: **preserve as-is** vs. support non-promiseable thenables
* `.pipe` support: **deprecate** vs. keep vs. drop
* Testing `.pipe`/`.then` incompatibilities: **prefix tests with `[DIVERGENT]`** vs. ???
* `jQuery.when`: **privilege `.promise`** (e.g., provide context) vs. privilege `.then`

To Do:
* [x] update `jQuery.when`
* [ ] reduce size
* [ ] rebase (avoiding for now to keep a stable size impact reference)
* [ ] prototype `unhandledRejections` plugin

```
   raw     gz Sizes
252424  74450 dist/jquery.js
 85114  29662 dist/jquery.min.js

   raw     gz Compared to master @ 66e1b6b
 +5499  +1231 dist/jquery.js
  +921   +267 dist/jquery.min.js
```


test/unit/deferred.js
```diff
@@ -146,30 +150,70 @@ test( "jQuery.Deferred.then - filtering (fail)", function() {
...
+	assert.expect( 4 );
+
+	var value1, value2, value3,
+		defer = jQuery.Deferred(),
+		piped = defer.pipe( null, function( a, b ) {
+			return a * b;
+		}),
+		done = jQuery.map( new Array( 3 ), function() { return assert.async(); } );
+
+	piped.fail(function( result ) {
```
gibson042: That piped deferreds _inherit_ rejection is the reason why I couldn't justify dropping `.pipe`... the standard and our current behavior are just too divergent here.

test/unit/deferred.js
```diff
@@ -253,7 +306,48 @@ test( "jQuery.Deferred.then - deferred (progress)", function() {
...
+	var value1, value2, value3,
+		defer = jQuery.Deferred(),
+		piped = defer.pipe( null, null, function( a, b ) {
+			return jQuery.Deferred(function( defer ) {
+				defer.resolve( a * b );
+			});
+		}),
+		done = assert.async();
 
 	piped.done(function( result ) {
```
gibson042: This totally blew me away. @jaubourg, is it really intentional that bound deferreds can be resolved by _progress_ handlers? Whose defining characteristic is that they are non-unique, and can be followed by other progress calls or resolve/reject? Which would presumably then have no effect?

I just couldn't see maintaining compatibility with this behavior.

test/unit/deferred.js
```diff
@@ -266,29 +360,36 @@ test( "jQuery.Deferred.then - deferred (progress)", function() {
...
-		context = {};
+		context = {},
+		done = jQuery.map( new Array( 4 ), function() { return assert.async(); } );
 
 	jQuery.Deferred().resolveWith( context, [ 2 ] ).then(function( value ) {
 		return value * 3;
 	}).done(function( value ) {
-		strictEqual( this, context, "custom context correctly propagated" );
-		strictEqual( value, 6, "proper value received" );
+		assert.notStrictEqual( this, context, "custom context not propagated through .then" );
```
gibson042: Required by spec.

test/unit/deferred.js
```diff
@@ -309,9 +412,117 @@ test( "jQuery.Deferred.then - context", function() {
...
+
+	expect( 1 );
+
+	var defer = jQuery.Deferred().done(function() {
+		setTimeout( start );
+		throw new Error();
+	});
+
+	defer.then(function() {
+		ok( true, "errors in .done callbacks don't stop .then handlers" );
```
gibson042: First new test.

test/unit/deferred.js
```diff
@@ -309,9 +412,117 @@ test( "jQuery.Deferred.then - context", function() {
...
+	});
+
+	try {
+		defer.resolve();
+	} catch ( _ ) {}
+});
+
+asyncTest( "jQuery.Deferred - 1.x/2.x compatibility", function() {
```
gibson042: Second new test.

dmethvin:
Wow. I'll be reading through this today ... maybe tomorrow as well. :)

test/unit/deferred.js
jaubourg: (first note that these are not the same `defer`)

Yes, it is perfectly intentional. Let's take an example.

Say you have a Deferred that can progress from 0 to 100 (%) and you have a series of actions happening when the progress reaches a certain point (say 50).

The naive approach is to use progress handlers and repeat something like this throughout your app:

```js
upload.progress( function( advancement ) {
    if ( advancement >=50 ) {
        action1();
    }
} );

// ... elsewhere in the code ...

upload.progress( function( advancement ) {
    if ( advancement >=50 ) {
        action2();
    }
} );

// etc
```

Another, less coupled approach is to create a Promise that is resolved when the first one reaches the threshold and to share this one instead:

```js
var uploadNearCompletion = upload.pipe( null, null, function( advancement ) {
    if ( advancement >= 50 ) {
        return jQuery.Deferred( function( defer ) {
            defer.resolve();
        } );
    }
} );

// ... elsewhere in the code ...

uploadNearCompletion.done( action1 );

// ... elsewhere in the code ...

uploadNearCompletion.done( action2 );

// etc
```

That way, when Bob-the-project-manager comes and tells you "near completion" is now 20%, it's just a matter of checking where the near-completion promise is created.

src/deferred.js
```diff
@@ -46,6 +60,123 @@ jQuery.extend({
...
+												args );
+										} else {
+											deferred.resolve( returned );
+										}
+									} catch ( e ) {
+
+										// Support: Promises/A+ section 2.3.3.3.4.1
+										// https://promisesaplus.com/#point-61
+										// Ignore post-resolution exceptions
+										if ( depth + 1 >= maxDepth ) {
```
dmethvin: This is the part in the story where it swallows errors. Some of the libs do complicated "next-turn" checking but I wonder if we could just do a `console.log` if the error that was thrown is one of the developer-screwup errors like `SyntaxError`. Something as simple as this?
```js
if ( /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error/.test(e) ) {
    console.log("[Caught exception] "+e);
    console.trace();
}
```
It could be in a public method that users could override perhaps?

Also note we only have one test case for `throw`n exceptions and none for browser-initiated exceptions.

test/unit/deferred.js
dmethvin: Oh, because it went through `.then` first? I think that's okay.

dmethvin:
This is really awesome. :clap: I want to understand it a little more, the spec references are helping.

markelog:
:clap: indeed

/cc @domenic

test/unit/deferred.js
gibson042: Interesting, but a little more obtuse than the equivalent:
```js
var uploadNearCompletion = jQuery.Deferred();
upload.progress(function( advancement ) {
    if ( advancement >= 50 ) {
        uploadNearCompletion.resolve();
    }
});
```

My chief concern, though, is the non-unique nature of progress handler execution, which can lead to implicit races. But upon more thought, I don't even think this concept is compatible with standard Promises because they exclusively bind to the first exit value from a handler associated with their creation. I **do**, however, need to add some tests and code supporting thenable return from progress handlers, e.g.:
```js
var trigger = jQuery.Deferred().notify();
trigger.then( null, null, function() {
    var notifier = jQuery.Deferred().notify( "foo" );
    setTimeout(function() {
        notifier.notify( "bar" ).resolve( "swallowed" );
    });
    return notifier;
})
    .then( fail, fail, expectFooThenBarThenFooThenBar );
trigger.notify();
```

Thanks for the insight.

dcherman:
It looks like the `$.when` implementation is still using `.promise` as a test to determine whether or not a given argument is a promise or not.  Can that be changed for compatibility with other Promise libs, or is that out of scope for this PR?

gibson042:
That's definitely in scope here. I'll update it.

gibson042:
After some thought, I ended up [nearly reversing](https://github.com/gibson042/jquery/commit/0da2b1a285dd3baea7a6298bcd8342f33b11f768) the last block [above](https://github.com/jquery/jquery/pull/1996#discussion_r22540118)... thenables returned from `.then` progress handlers are used only for their eventual _fulfillment_ values (which translate back into progress values for the subordinate Deferred), with progress ignored and rejection generating (asynchronous) exceptions so their non-effect can be observed. It's possible to take a different direction, however, and I'd like other opinions—progress events, especially multiple independent chains of progress events originating from various points on a resolution chain, can generate the kinds of headaches usually reserved for time travel fiction.

Still to do:
* [ ] update `jQuery.when`
* [ ] reduce size
* [ ] rebase (avoiding for now to keep a stable size impact reference)
* [ ] prototype `unhandledRejections` plugin

dmethvin:
I would be fine with corralling `.then` to only support standard Promise features, leaving back-compatible behavior to `.pipe` and the rest if that makes life easier.

gibson042:
So that would involve `.then` resolved/rejected handlers always receiving exactly one argument and `undefined` context, and progress handlers never being called at all? We can do so, and it will simplify this effort, but I think it will mean committing to keep `.pipe` around for longer (since our internal uses will lean on it more heavily).

It's also a bigger backcompat break, but I don't know if that's a good thing or a bad thing here.

I'd like to further point out that these cases are just as potentially problematic with `.pipe`, it's just unlikely that anything ever comes close to exercising the full power of what we expose with progress—including our own unit tests. For example, I _just_ found a bug in `.pipe`: #2010.

dcherman:
This may be more of a use case for jQuery.Migrate, but what about adding warnings in the unminified versions to warn the user if `thenCallback.length` or `failCallback.length` is > 1 when `then` is called since the standards compliant one will only accept 1 argument?

I'm just thinking about the nasty case of `$.ajax` where each callback accepts 3 arguments;  the normal way to do that with Promises is to resolve it with a tuple and access the properties by index ( or with a `spread` helper that many libs provide ), however that's obviously a massive backcompat break.

gibson042:
> I'm just thinking about the nasty case of `$.ajax` where each callback accepts 3 arguments

Me too :hear_no_evil:

dmethvin:
Ugh.

Yes, the `$.ajax` case is the worst one. When people are using non-Deferreds they seem to just grab the `jQXHR` that is the first arg and not use the other two, which are derived from it anyway. I think that just *not* truncating the args gets us a long way towards not breaking code in the wild since that's going to be like `$.ajax(...).then(accepted, rejected)` -- although I bet a lot of code doesn't bother handling rejections, which is why we really need to bleat about that case.

Whatever we end up with, reporting the unusual stuff via jquery-migrate seems like a good idea.

gibson042:
> When people are using non-Deferreds they seem to just grab the jQXHR that is the first arg and not use the other two, which are derived from it anyway.

Oh, if only we were that consistent. From http://api.jquery.com/jQuery.ajax/ :
* `jqXHR.done(function( data, textStatus, jqXHR ) {});`
* `jqXHR.fail(function( jqXHR, textStatus, errorThrown ) {});`
* `jqXHR.always(function( data|jqXHR, textStatus, jqXHR|errorThrown ) { });`
* `jqXHR.then(function( data, textStatus, jqXHR ) {}, function( jqXHR, textStatus, errorThrown ) {});`

dmethvin:
Oh geez, I can't even remember it right. It's the *data* that's the first arg for success, which is usually what people want. I agree the args are a mess, especially `.always` since the other arg reversals mean it has no idea what it's being passed.

gibson042:
@jaubourg and @jquery/core:
I know we're all getting kind of antsy about Deferred/Promise interop, and having two competing PRs doesn't help. This one is basically ready to land now (with reduction deferred for later), but as I mentioned in the meeting, I'd like to focus more on the behavior changes and their reflection in unit tests than on the implementation. So please look over the description and [test/unit/deferred.js](https://github.com/jquery/jquery/pull/1996/files#diff-7) and provide any comments you may have on the relevant points. Once we agree on them, the code can follow.

benjamingr:
In order to deal with multiple arguments in `then` you can do what most promise libraries do and expose a `.spread` method _in addition_ to `.then` that converts an array to variable arguments. Note that ES6 already ships with destructuring which accomplishes this.

dcherman:
Just another point of concern with synchronous ajax calls:

I just took a look and unless I missed them, there are no unit tests that assert the ordering of callbacks.  Since all handlers ( including the inline success, error, and complete options ) are attached using Deferreds, this change might have the effect of "breaking" synchronous ajax in that the callbacks will no longer be called synchronously.

benjamingr:
@dcherman using deferreds with synchronous AJAX calls is not and has not been supported as of 1.8. Quoting the manual:

> As of jQuery 1.8, the use of async: false with jqXHR ($.Deferred) is deprecated; you must use the success/error/complete callback options instead of the corresponding methods of the jqXHR object such as jqXHR.done() or the deprecated jqXHR.success().

dcherman:
@benjamingr I did forget about that, but my point still stands actually unless I'm misunderstanding something:

https://github.com/jquery/jquery/blob/master/src/ajax.js#L490-L492
https://github.com/jquery/jquery/blob/master/src/ajax.js#L622-L624

The callback options just add the callbacks to the Deferred which is now resolved asynchronously, no?

gibson042:
@dcherman: this change forces `then` callbacks to follow the on-object `done`/`fail`/`always`, by design. The actual sequence is that the latter fire first, synchronously in the order in which they were added, and then the former fire asynchronously in the order _they_ were added. And the ajax module doesn't use `then`.

In short: only `then` is affected.

Do you foresee that causing problems? If so, I'm not sure anything can be done, since the asynchronicity is explicitly specified.

dcherman:
@gibson042 Ah I see how it works now, I misunderstood how this change worked.  Since as @benjamingr pointed it the Deferred/Promise methods have been deprecated for sync XHR since 1.8, it shouldn't officially cause a problem.  I'm sure there are people out there that were depending on `.then` for sync, but there's not much that you can do about that and it's unsupported.

timmywil:
@gibson042 How bout we land this?

gibson042:
I didn't get a chance to merge this, but did [clean up some text](https://github.com/gibson042/jquery/commit/694374cf01bf1cf4ca99395dd2192590f0647598). At this point, I feel like it's good to go and the remaining work can be dealt with later.

benjamingr:
Awesome! Great news

timmywil:
Hey @gibson042, reminder to merge this when you get a chance. Thanks!

PR #1998, opened 2015-01-07
Core: Return empty array instead of null for parseHTML("")

### LINKED ISSUE #1997: $.parseHTML( "" ) should not return null
This behaviour has been there since the version version (e2497c682f26b7916d76cb2896c6fe621b376d82) but I'd like to propose changing that.

The method is [documented](https://api.jquery.com/jQuery.parseHTML/) as taking a string (of HTML) and returning an Array. Having it return null for non-string input is fair, but for empty strings it should probably return an empty array (or maybe array with empty text node).

Krinkle:
Fixes #1997

### DIFF
--- a/src/core/parseHTML.js
+++ b/src/core/parseHTML.js
@@ -15,8 +15,8 @@ define([
 // defaults to document
 // keepScripts (optional): If true, will include scripts passed in the html string
 jQuery.parseHTML = function( data, context, keepScripts ) {
-	if ( !data || typeof data !== "string" ) {
-		return null;
+	if ( typeof data !== "string" ) {
+		return [];
 	}
 	if ( typeof context === "boolean" ) {
 		keepScripts = context;
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -1301,13 +1301,18 @@ test("jQuery.proxy", function(){
 });
 
 test("jQuery.parseHTML", function() {
-	expect( 18 );
+	expect( 23 );
 
 	var html, nodes;
 
-	equal( jQuery.parseHTML(), null, "Nothing in, null out." );
-	equal( jQuery.parseHTML( null ), null, "Null in, null out." );
-	equal( jQuery.parseHTML( "" ), null, "Empty string in, null out." );
+	deepEqual( jQuery.parseHTML(), [], "Without arguments" );
+	deepEqual( jQuery.parseHTML( undefined ), [], "Undefined" );
+	deepEqual( jQuery.parseHTML( null ), [], "Null" );
+	deepEqual( jQuery.parseHTML( false ), [], "Boolean false" );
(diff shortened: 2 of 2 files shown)


src/core/parseHTML.js
```diff
@@ -15,7 +15,7 @@ define([
 // defaults to document
 // keepScripts (optional): If true, will include scripts passed in the html string
 jQuery.parseHTML = function( data, context, keepScripts ) {
-	if ( !data || typeof data !== "string" ) {
+	if ( typeof data !== "string" ) {
 		return null;
 	}
```
gibson042: I'd even support a followup to drop this `if` entirely. If non-string input really causes a problem, it should be visible as an exception.

timmywil: I wouldn't support that. parseHTML via createHTMLDocument does not throw an error with parseHTML(1) or parseHTML({}).

timmywil: However, maybe we could just return an empty array here.

dmethvin: The docs say the first arg is a String and that it returns Array. I don't think we should ever be returning null, despite the unit test. :smile:  If the first arg is not a string it seems either all bets are off or we should turn it into a string for cases we feel are possible (numbers, `null`, `undefined`)? I would be fine with removing this entirely but have a feeling it would be like our `jQuery.parseJSON(null)` fiasco.

gibson042: > I don't think we should ever be returning null, despite the unit test.

This is the intent of my suggestion. Just dropping the `if` and letting values propagate through gives good (and mostly expected) results: [http://jsfiddle.net/evse952q/](http://jsfiddle.net/evse952q/), so we need not check for any special input or construct any special outptu (and it's worth noting that our internal uses always provide string input as documented).

timmywil: The results for numbers and `true` are inconsistent. I'd rather they return empty array.

dmethvin: I dunno @timmywil , looking through the whole execution path (which I should have done originally) it seems like dropping the check only affects really strange inputs and not in a catastrophic way.

timmywil: Yea, it's not terrible. i just think if we're going to change it, why not match native behavior?

gibson042: Native behavior always casts to string except `null` (which is inconsistent between browsers): http://jsfiddle.net/evse952q/1/

Achieving that in jQuery would mean adopting my above `buildFragment` suggestion and casting (some) nullish values to string in `parseHTML`.

gibson042:
Are you willing to update or re-PR for the more comprehensive changes, @Krinkle?

Krinkle:
@gibson042 Done.

Issue #1908, opened 2014-12-07
Remove context and selector properties


gibson042:
Originally reported for jQuery 2.0.3 by anonymous at http://bugs.jquery.com/ticket/14341

> I was lurking a few months ago and just remembered now, but it looks like both these properties are still present.

> Selector has been deprecated since 1.7, and context since 2.0. Can they be removed now? I'd attempt myself however I am not able to sign the CLA due to weird moonlighting policies of my employer.

> http://bugs.jquery.com/ticket/9469 http://bugs.jquery.com/ticket/13801

Both properties have been deprecated since 1.10/2.0, the former courtesy http://bugs.jquery.com/ticket/13801
* http://api.jquery.com/context/
* http://api.jquery.com/selector/

3.0 strikes me as the right time. Pre-1.10 #1249 probably can't be applied _directly_ anymore, but I'll bet it's pretty close.

dmethvin:
Ref https://github.com/jquery/jquery-migrate/issues/79

thdoan:
How can we get the original selector string that was used to invoke a plugin from within the plugin? Example:

```
(function($) {
  $.fn.plugin = function() {
    // This no longer works in jQuery 3.x
    console.log(this.selector);
    return this;
  };
}(jQuery));
```

dmethvin:
A closed ticket isn't the right place to look for support, but in general there is no string that represents the collection a plugin receives, e.g. `$(this).plugin()` or `$("p").children("span").plugin()`.

PR #2000, opened 2015-01-09
Core: Remove deprecated context and selector properties

dmethvin:
Fixes gh-1908

Ref https://github.com/jquery/jquery-migrate/issues/79

### DIFF
--- a/src/core/init.js
+++ b/src/core/init.js
@@ -73,12 +73,9 @@ var rootjQuery,
 
 					if ( elem ) {
 						// Inject the element directly into the jQuery object
-						this.length = 1;
 						this[0] = elem;
+						this.length = 1;
 					}
-
-					this.context = document;
-					this.selector = selector;
 					return this;
 				}
 
@@ -94,7 +91,7 @@ var rootjQuery,
 
 		// HANDLE: $(DOMElement)
 		} else if ( selector.nodeType ) {
-			this.context = this[0] = selector;
+			this[0] = selector;
 			this.length = 1;
 			return this;
 
@@ -105,11 +102,6 @@ var rootjQuery,
 				rootjQuery.ready( selector ) :
 				// Execute immediately if ready is not present
 				selector( jQuery );
-		}
-
-		if ( selector.selector !== undefined ) {
-			this.selector = selector.selector;
-			this.context = selector.context;
 		}
 
 		return jQuery.makeArray( selector, this );
--- a/test/unit/offset.js
+++ b/test/unit/offset.js
@@ -462,9 +462,9 @@ testIframe("offset/body", "body", function( $ ) {
 test("chaining", function() {
 	expect(3);
 	var coords = { "top":  1, "left":  1 };
-	equal( jQuery("#absolute-1").offset(coords).selector, "#absolute-1", "offset(coords) returns jQuery object" );
-	equal( jQuery("#non-existent").offset(coords).selector, "#non-existent", "offset(coords) with empty jQuery set returns jQuery object" );
-	equal( jQuery("#absolute-1").offset(undefined).selector, "#absolute-1", "offset(undefined) returns jQuery object (#5571)" );
+	equal( jQuery("#absolute-1").offset(coords).jquery, jQuery.fn.jquery, "offset(coords) returns jQuery object" );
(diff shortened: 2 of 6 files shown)


src/core/init.js
```diff
@@ -73,12 +73,9 @@ var rootjQuery,
 
 					if ( elem ) {
 						// Inject the element directly into the jQuery object
-						this.length = 1;
 						this[0] = elem;
+						this.length = 1;
```
markelog: We would save more bytes that way?

markelog: Or just for the consistency - https://github.com/jquery/jquery/pull/2000/files#diff-85f162ae43172ebe26b51a4d1c5d4eddR94

dmethvin: Seemed strange to set the length before the element. I pulled several of the changes from @gibson042 's older PR because they just seemed so right. :smile:

test/unit/offset.js
```diff
@@ -462,9 +462,9 @@ testIframe("offset/body", "body", function( $ ) {
 test("chaining", function() {
 	expect(3);
 	var coords = { "top":  1, "left":  1 };
-	equal( jQuery("#absolute-1").offset(coords).selector, "#absolute-1", "offset(coords) returns jQuery object" );
-	equal( jQuery("#non-existent").offset(coords).selector, "#non-existent", "offset(coords) with empty jQuery set returns jQuery object" );
-	equal( jQuery("#absolute-1").offset(undefined).selector, "#absolute-1", "offset(undefined) returns jQuery object (#5571)" );
+	equal( jQuery("#absolute-1").offset(coords).jquery, jQuery.fn.jquery, "offset(coords) returns jQuery object" );
```
markelog: Always wondered why this property was called "jquery", not "version"

src/core/init.js
markelog: :)

test/unit/offset.js
gibson042: It's killing two birds with one stone: the _presence_ of the property identifies an object as a jQuery collection as opposed to Array/NodeList/Prototype/Zepto/etc., and the value specifies which version.

gibson042:
:+1:

test/unit/offset.js
markelog: > It's killing two birds with one stone

Yeah! Or "a bird in the hand is worth two in the bush"

This property indeed serve for two complitly different things, but i don't think that's good, i think these should be exposed explicitly then otherwise. Besides, i'm not sure that was an original intent.

> the presence of the property identifies an object as a jQuery

This could always be done by other means, like: 

```js
function test() {}
new test().constructor.name // test

// i.e. we could do

function jQuery() {}
jQuery.prototype.constructor = jQuery;
$().constructor.name // "jQuery"
```
Or other countless ways to do it. Not saying we should do something about it, it might be too late, like with `$.each vs $.map` callback arguments case :-(

timmywil: Yea, it's too late.

dmethvin: Until I looked at the old PR, I had just planned to see if the `.selector` property was present but the change by @gibson042 seemed like a slightly better assertion. 

I guess `.version` was avoided since it's a pretty common name. You wouldn't think `someobject.fn.version` would be that ambiguous but it's also available via `someobject.prototype.version` so maybe that caused concern.

markelog: > Yea, it's too late.

Isn't that frustrating? We can't improve something because people are dependent upon it.

> I guess .version was avoided since it's a pretty common name.

If i would do it today, i would called it "version" and provided other ways to identify jquery object

gibson042: > This property indeed serve for two complitly different things, but i don't think that's good, i think these should be exposed explicitly then otherwise.

The purposes aren't completely different; they're analogous to `nodeType` checks in that both properties are considered to have a universal meaning regardless of where they appear. `jquery` specifies the version of jQuery used to construct a collection, which—since it must be `undefined` for objects that are _not_ jQuery collections—can be cast as a boolean for type checking, just like casting `nodeType` to boolean checks for DOM nodes. That said, though, there might well have been a better choice for the name, but it's definitely too late to change.

Issue #1744, opened 2014-10-21
Enhancement: selector for Unwrap


mzgol:
Originally reported by nlte@… at: http://bugs.jquery.com/ticket/13967
<p>
Unwrap simple removes the parent of selected elements, and accepts no arguments.
In some cases it may be desirable to unwrap only if a condition on the parent is met. Enabling to pass a selector to Unwrap would allow this: if the parent matches the selector, then proceed, otherwise do nothing.
</p><p>Issue reported for jQuery 1.10.0</p>

PR #2003, opened 2015-01-12
Wrap: Support .unwrap( selector) for selective unwrapping

dmethvin:
gh-1744

Really should be the `Wrap` component but that's missing from the list. (Note, updated the list in a4133ffafa6ac43b3aa22bc4edb4444b09f5250c)

### DIFF
--- a/src/wrap.js
+++ b/src/wrap.js
@@ -64,14 +64,10 @@ jQuery.fn.extend({
 	},
 
 	unwrap: function( selector ) {
-		return this
-			.parent()
-				.filter( selector || ":not(body)" )
-					.each(function() {
-						jQuery( this ).replaceWith( this.childNodes );
-					})
-				.end()
-			.end();
+		this.parent( selector ).not( "body" ).each(function() {
+			jQuery( this ).replaceWith( this.childNodes );
+		});
+		return this;
 	}
 });
 
--- a/src/wrap.js
+++ b/src/wrap.js
@@ -63,12 +63,15 @@ jQuery.fn.extend({
 		});
 	},
 
-	unwrap: function() {
-		return this.parent().each(function() {
-			if ( !jQuery.nodeName( this, "body" ) ) {
-				jQuery( this ).replaceWith( this.childNodes );
-			}
-		}).end();
+	unwrap: function( selector ) {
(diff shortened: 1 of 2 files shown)


markelog:
Is there a popular plugin for this? How did we come to conclusion that selector filter is most common case for `unwrap`?

dmethvin:
The main case is when you are unwrapping some elements but want to be sure the parent element(s) are the same ones used to wrap them, or there's the chance they could be unwrapped twice based on the code paths. I have run into it in the past (probably 3 or 4 years ago at this point) so I do think its a valid use case.

markelog:
What troubling me about this: as far as i know, we received only one request for it, there was no plugin that could provide representative data of probable popularity of this feature and it might be considered as breakage of argument coherence with opposite method, since they would be more heterogeneity then they are now.

dmethvin:
I don't think it's useful for `.wrap()` and `.unwrap()` to be symmetrical in that way. When you `.wrap(html)` you're providing the content to be wrapped. When you `.unwrap(selector)` you're saying you want to be sure that the elements are wrapped the way you think they are.

So you might `.wrap("<div class='error'></div>")` but the corresponding call would be `.unwrap(".error")` most likely. I definitely don't think you'd want to provide the actual wrapper html to match it during unwrapping.

src/wrap.js
```diff
@@ -63,12 +63,15 @@ jQuery.fn.extend({
...
-	unwrap: function() {
-		return this.parent().each(function() {
-			if ( !jQuery.nodeName( this, "body" ) ) {
-				jQuery( this ).replaceWith( this.childNodes );
-			}
-		}).end();
+	unwrap: function( selector ) {
+		return this
+			.parent()
+				.filter( selector || ":not(body)" )
```
gibson042: We probably need to no-op on body children even when a selector is passed. I recommend `this.parent( selector ).not( "body" )`, and checking to see if storing it in a variable is smaller than this use of `.end()` to enable a one-liner.

dmethvin: Yeah, I agonized on that one and figured a selector that included body would be pretty rare. Your solution is safer, I'll play with this.

PR #2011, opened 2015-01-13
.css(), the values like +=10% aren't ignored anymore (like .animate())

Mr21:
Hi :)

Actually when we make:

`$element.animate({width: "+=10%"}, 500);`
it's work fine, the value is converted into percentage BUT if we write:

`$element.css("width", "+=10%");`
We don't have the same behavior...
The `.css()` function doesn't look the unity, it's `"px"` every time.

The deletion of my two old commit has closed [my last pull request](https://github.com/jquery/jquery/pull/1494)... and even when I make another new commit I can't reopen it, sorry for that doublon.

So, @gibson042 told me to see this part: https://github.com/jquery/jquery/blob/master/src/effects.js#L38-L60
Before, I didn't understand I could make a new CSS change on the element with `$.style` **just** for make the calcul conversion about the different CSS unities.
When I understand this, I understand the code BUT I didn't understand (again) why it's necessary to make the `do{}while()` loop, because every time this loop is passed **only one time** for each call.


gibson042:
Thanks for the followup! I think you will see some problems if you run the effects test suite against this code. Also, I think more thought needs to go into handling extreme values (zero, very small, very large)—which is the reason for do...while in effects.js.

And finally, this new functionality should come with unit tests so we can ensure its correct behavior on all platforms into the future.

Mr21:
Hi :)

I forgot to check the division about the zero, shame on me... I have update my commit.

I didn't understand what you mean by *very small value* it's between 0 and 1 pixel?
Like `0.5pt == 0.666px` ?

It make sens for me only if we could multiply a value like when we multiply a `float` casted in `int`:
`((int)2.5) * 100 == 200` instead of `2.5 * 100 == 250`
I mean the tiny difference less than one pixel would increase with the multiplication. But we cannot multiply anyway.

gibson042:
"Very small" in the sense of "risking problems from rounding errors". The easiest way to get them is mixing "%" and non-"%", e.g. between `width: 0.2%` and `7em`.

And again, all of this should have unit tests.

Mr21:
Here I put several examples with the current jQuery
http://jsfiddle.net/Mr21/o4qbt6sf/

And here it's the same but with my changes (without the loop).
http://jsfiddle.net/Mr21/78yb124y/

The width of the **blue** bars are hard-fixed with the correct value and never move.
The width of the red and the green bars are also initialised with a specific values but I modify it to reach to the blue value:
But the **red** bars is modified with `.css("width", "+=10px");`
But the **green** bars is modified with `.animate({width: "+=10px"}, 500);`

I wrote your example (`0.2%` to `7em`) at the end of list.
But I cannot find any example who show me the need of the `do...while();`

gibson042:
You're still dealing with smallish values, but dimensions can get arbitrarily large (remember, "%" is always _relative_)—the loop is a guarantee that we can establish an accurate scale even for extreme values.

All of which, of course, is completely beside the point while this PR is not adding unit tests for new functionality and is **failing** existing effects tests.

Mr21:
I saw these failing tests, I will extract the loop in the common function this evening.
Is there a particular norme for adding some new unit tests?

gibson042:
Best judgment, and trying to preserve the surrounding style. In this case, I'd like to see a new test immediately after the existing relative-value test in [test/unit/css.js](https://github.com/jquery/jquery/blob/8ebd84418265f7d7db1ec46ee98ebae47a792883/test/unit/css.js#L119) that looks something like:

```js
test( "css() non-px relative values (gh-1711)", <N>, function() {
    …
});
```

and covers various non-px scenarios.

Mr21:
I made a new commit (all tests are passing).
But it was a little complicated to find something in common finally.
Tell me if you think it's okay.

I made the new unit tests just after this.

And another thing: why not make an alias on:
`.css(name, value)` or
`.css({prop: value})` to call directly:
`.animate({prop: value}, 0);`
And on the *animate* side we would find a `if` to check if we launch or not an real animation.

I think it can be a great save a code, no?

Mr21:
@gibson042 I add 10 new tests here; https://github.com/Mr21/jquery/blob/css/test/unit/css.js#L206-L245

src/css.js
```diff
@@ -285,7 +286,7 @@ jQuery.extend({
 
 			// If a number was passed in, add 'px' (except for certain CSS properties)
 			if ( type === "number" && !jQuery.cssNumber[ origName ] ) {
-				value += "px";
+				value += ret && ret[3] ? ret[3] : "px";
```
arthurvr: Nitpick but you should have spaces inside the square brackets.

Mr21:
> `value += ret && ret[3] ? ret[3] : "px";`
> Nitpick but you should have spaces inside the square brackets.
> @arthurvr

Oh, thanks (it's done) :)

src/css.js
Mr21: Oh, thanks (it's done) :)

src/css/cssCastUnit.js
```diff
@@ -0,0 +1,60 @@
+define([
+	"../core",
+	"../var/rfxnum"
+], function( jQuery, rfxnum ) {
+
+function cssCastUnit( elem, prop, parts, tween ) {
```
gibson042: Variables here could have better names, especially now that this code is a fully-fledged function serving two purposes. I recommend something like:
* cssCastUnit → adjustCSS
* parts → value
* getCSS → currentValue
* target → initial
* start → recasted
* end → adjusted

src/css/cssCastUnit.js
```diff
@@ -0,0 +1,60 @@
...
+	var getCSS = tween ?
+			function() { return tween.cur(); } :
+			function() { return jQuery.css( elem, prop, "" ); },
```
gibson042: Very nice.

src/css/cssCastUnit.js
```diff
@@ -0,0 +1,60 @@
...
+		target = getCSS(),
+		unit = parts && parts[ 3 ] || ( jQuery.cssNumber[ prop ] ? "" : "px" ),
+		// Starting value computation is required for potential unit mismatches
+		start = ( jQuery.cssNumber[ prop ] || unit !== "px" && +target ) &&
+			rfxnum.exec( jQuery.css( elem, prop ) ),
+		end,
```
gibson042: Declare this uninitialized variable first: http://contribute.jquery.org/style-guide/js/#assignments

src/css/cssCastUnit.js
```diff
@@ -0,0 +1,60 @@
...
+			// If previous iteration zeroed out, double until we get *something*.
+			// Use string for doubling so we don't accidentally see scale as unchanged below
+			scale = scale || ".5";
+
+			// Adjust and apply
+			start = start / scale;
+			jQuery.style( elem, prop, start + unit );
+
+			// Update scale, tolerating zero or NaN from tween.cur(),
+			// break the loop if scale is unchanged or perfect, or if we've just had enough
```
gibson042: These comments are indented one tab too far. Also, this is the perfect opportunity to fix my bad grammar by removing the trailing comma and capitalizing "break".

src/css/cssCastUnit.js
```diff
@@ -0,0 +1,60 @@
...
+	if ( parts ) {
+		start = +start || +target || 0;
+		end = parts[ 1 ] ?
+			start + ( parts[ 1 ] + 1 ) * parts[ 2 ] :
+			+parts[ 2 ];
+		if ( tween ) {
+			tween.unit = unit;
+			tween.start = start;
+			tween.end = end;
+		}
```
gibson042: Fantastic way to isolate effects-specific needs.

src/css.js
```diff
@@ -273,7 +274,7 @@ jQuery.extend({
 
 			// Convert "+=" or "-=" to relative numbers (#7345)
 			if ( type === "string" && (ret = rrelNum.exec( value )) ) {
```
gibson042: Optimization opportunity (i.e., apply if it yields a smaller jquery.min.js.gz): get rid of `rrelNum` completely by using `rfxnum` in its place here: `if ( type === "string" && (ret = rfxnum.exec( value )) && ret[ 1 ] )`.

src/css.js
```diff
@@ -283,9 +284,9 @@ jQuery.extend({
 				return;
 			}
 
-			// If a number was passed in, add 'px' (except for certain CSS properties)
+			// If a number was passed in, add the unit (except for certain CSS properties)
 			if ( type === "number" && !jQuery.cssNumber[ origName ] ) {
-				value += "px";
+				value += ret && ret[ 3 ] ? ret[ 3 ] : "px";
```
gibson042: Optimization opportunity: `value += ret && ret[ 3 ] || "px"` (or, if `rrelNum` is no more, `ret && ret[ 1 ] && ret[ 3 ] || "px"`).

test/unit/css.js
```diff
@@ -203,6 +203,47 @@ test( "css() explicit and relative values", 29, function() {
...
+	add( 10, "em");
+	add( 10, "pt");
+	add(-2.3, "pt");
+	add( 5, "pc");
+	add(-5, "em");
+	add(+2, "cm");
+	add(-15, "mm");
+	add(21, "px");
+});
+
```
gibson042: Could you also add some tests against `line-height` instead of `width`? That property has caused problems in the past, and I'd like to know it's solid.

gibson042:
:clap: @Mr21; this is looking great!

Mr21:
All is done @gibson042 :)

I found a probleme with `line-height`, this property is set in the `cssNumber` object for *Don't automatically add `"px"` to these possibly-unitless properties*.
And if we not pass any unit to `line-height` so unit per default is `em`.
So I modified some code again here: https://github.com/jquery/jquery/pull/2011/files#diff-414c8f59bff0d1b63680b64763d8c529R288.

src/css.js
Mr21: `rfxnum` is different of the old other regex because these both  case: "10px" and "+=10px" match. So I had this verification here: https://github.com/jquery/jquery/pull/2011/files#diff-414c8f59bff0d1b63680b64763d8c529R276

gibson042:
> I found a probleme with `line-height`, this property is set in the `cssNumber` object

Yep, that's exactly the kind of thing I was hoping to uncover.

> Thanks for all these indications!

Thank you; I will try to look it over again today. Also, squashing everything into a single commit makes it hard to see changes over the course of this pull request... in the future, please add new commits and fast-forward instead of force-pushing replacements.

Mr21:
Understood :+1:

src/css.js
```diff
@@ -17,8 +19,8 @@ define([
 	"./css/swap",
 	"./core/ready",
 	"./selector" // contains
-], function( jQuery, pnum, access, rmargin, rnumnonpx, cssExpand, isHidden,
-	getStyles, curCSS, defaultDisplay, addGetHookIf, support, dataPriv ) {
+], function( jQuery, pnum, rfxnum, access, rmargin, rnumnonpx, cssExpand, isHidden,
+	getStyles, curCSS, adjustCSS, defaultDisplay, addGetHookIf, support, dataPriv ) {
```
gibson042: Let's keep the "r" variables together (e.g., `rfxnum` before or after `rnumnonpx`).

src/css.js
```diff
@@ -283,9 +284,13 @@ jQuery.extend({
...
-			// If a number was passed in, add 'px' (except for certain CSS properties)
-			if ( type === "number" && !jQuery.cssNumber[ origName ] ) {
-				value += "px";
+			// If a number was passed in, add the unit (except for certain CSS properties)
+			if ( type === "number" ) {
+				if ( ret && ret[ 3 ] ) {
+					value += ret[ 3 ];
+				} else if ( !jQuery.cssNumber[ origName ] ) {
+					value += "px";
+				}
```
gibson042: This would probably be smaller as `value += ret && ret[ 3 ] || ( jQuery.cssNumber[ origName ] ? "" : "px" )`.

src/css/adjustCSS.js
```diff
@@ -0,0 +1,60 @@
...
+		scale = 1,
+		maxIterations = 20,
+		currentValue = tween ?
+			function() { return tween.cur(); } :
+			function() { return jQuery.css( elem, prop, "" ); },
+		initial = currentValue(),
+		unit = value && value[ 3 ] || ( jQuery.cssNumber[ prop ] ? "" : "px" ),
+		// Starting value computation is required for potential unit mismatches
+		recasted = ( jQuery.cssNumber[ prop ] || unit !== "px" && +initial ) &&
+			rfxnum.exec( jQuery.css( elem, prop ) );
```
gibson042: Seeing it all together, maybe a couple more renames are in order:
* `value` → `valueParts`
* `recasted` → `initialInUnit`

src/css/adjustCSS.js
```diff
@@ -0,0 +1,60 @@
...
+		// Update scale, tolerating zero or NaN from tween.cur()
+		// Break the loop if scale is unchanged or perfect, or if we've just had enough.
+		} while (
+			scale !== (scale = currentValue() / initial) && scale !== 1 && --maxIterations
+		);
+	}
+
+	if ( value ) {
+		recasted = +recasted || +initial || 0;
+		adjusted = value[ 1 ] ?
```
gibson042: This needs a comment like `Apply relative offset (+=/-=) if specified`.

src/var/rfxnum.js
```diff
@@ -0,0 +1,7 @@
+define([
+	"../var/pnum"
+], function( pnum ) {
+
+return new RegExp( "^(?:([+-])=|)(" + pnum + ")([a-z%]*)$", "i" );
```
gibson042: One more rename, now that it holds: `rfxnum` → `rcssNum`.

test/unit/css.js
```diff
@@ -203,6 +203,59 @@ test( "css() explicit and relative values", 29, function() {
...
+			equal(
+				Math.round( parseFloat( $child.css( prop ) ) ),
+				Math.round( cssCurrent += val * units[ prop ][ unit ] ),
+				"'" + str + "' on " + prop
+			);
+		},
+		getUnits = function( prop ) {
+			units[ prop ] = {
+				"px": 1,
+				"em": parseFloat( $child.css( "fontSize", "16px" ).css( "fontSize" )),
```
gibson042: "em" appears to be buggy with respect to the following lines; why not `parseFloat( $child.css( prop, "100em" ).css( prop ) ) / 100`?

test/unit/css.js
```diff
@@ -203,6 +203,59 @@ test( "css() explicit and relative values", 29, function() {
 	equal( $elem.css("opacity"), "1", "'+=0.5' on opacity (params)" );
 });
 
+test( "css() non-px relative values (gh-1711)", 17, function() {
+	var cssCurrent,
+		units = {},
+		$parent = jQuery("#nothiddendiv"),
+		$child = jQuery("#nothiddendivchild", $parent),
+		add = function( prop, val, unit ) {
+			var str = ( val < 0 ? "-=" : "+=" ) + Math.abs(val) + unit;
```
gibson042: The `jQuery` and `Math.abs` function calls need spaces around their arguments.

test/unit/css.js
```diff
@@ -203,6 +203,59 @@ test( "css() explicit and relative values", 29, function() {
...
+			$child.css( prop, str );
+			equal(
+				Math.round( parseFloat( $child.css( prop ) ) ),
+				Math.round( cssCurrent += val * units[ prop ][ unit ] ),
+				"'" + str + "' on " + prop
```
gibson042: Let's use `prop + ": '" + str + "'"` (which reads as a valid key-value pair in a `$el.css({ … })` call).

test/unit/css.js
```diff
@@ -203,6 +203,59 @@ test( "css() explicit and relative values", 29, function() {
...
+				"mm": parseFloat( $child.css( prop, "100mm" ).css( prop ) ) / 100,
+				"%" : parseFloat( $child.css( prop, "100%"  ).css( prop ) ) / 100
+			};
+		};
+
+	$parent.css({ height: 1, padding: 0, width: 400 });
+	$child.css({ height: 1, padding: 0 });
+
+	getUnits( "width" );
+	cssCurrent = $child.css( "width", "50%" ).width();
```
gibson042: There's no need to pull in code from other modules here; this should use `.css( "width" )` in analogy to the "lineHeight" block below.

test/unit/css.js
```diff
@@ -203,6 +203,59 @@ test( "css() explicit and relative values", 29, function() {
 	equal( $elem.css("opacity"), "1", "'+=0.5' on opacity (params)" );
 });
 
+test( "css() non-px relative values (gh-1711)", 17, function() {
+	var cssCurrent,
+		units = {},
+		$parent = jQuery("#nothiddendiv"),
+		$child = jQuery("#nothiddendivchild", $parent),
```
gibson042: Drop this `$parent`, id is enough to get the right element.

gibson042:
Almost there! Thanks again, @Mr21.

src/css.js
Mr21: You're right, I didn't know what is the best between less charactere or a concat with an empty string `""`.

test/unit/css.js
Mr21: Yeah it's more logical that way, I used the `font-size` to *set* the unit `em` and not only to find it.
But finally it's the same :D

test/unit/css.js
Mr21: Yeah, I prefer that way, but I did that because the previous unit test `"css() explicit and relative values"` write the debug like: `"'-=9px' on width (hash)"`.
Do you want change this part too? It's fast and easy to do that with multi cursor ^^

Mr21:
All is done (in a new commit this time).

gibson042:
Just one more thing before this lands (and sorry I didn't check earlier)... make sure to sign our [CLA](http://contribute.jquery.org/CLA/) with the same email address you use for GitHub commits.

Mr21:
Done :)

timmywil:
Thanks @Mr21! :+1:

Mr21:
Great ^^
So now, do I have to merge my two commits?

gibson042:
Nope; just leave this as-is and I expect to merge within the next seven days (and _definitely_ before the 3.0 beta). Thanks again!

gibson042:
Thanks again, @Mr21!

Mr21:
You're welcome :smile:

PR #2016, opened 2015-01-17
support css value !important

bluelovers:
$('div').css('blue', '!important');

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -271,6 +271,13 @@ jQuery.extend({
 		if ( value !== undefined ) {
 			type = typeof value;
 
+			var _m, priority;
+			if (type === 'string' && (_m = value.match(/(.+)!\s*(important)[\s;]*$/)))
+			{
+				value = _m[1];
+				priority = _m[2];
+			}
+
 			// Convert "+=" or "-=" to relative numbers (#7345)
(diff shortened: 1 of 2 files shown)


dmethvin:
That example doesn't even make sense, there is no property named `blue`. That probably would have been a bit more obvious if the PR had added unit tests. See #1385 for a previous failed attempt, and check out our [Contributing site](http://contribute.jquery.org) for more help.

mzgol:
> That example doesn't even make sense, there is no property named `blue`.

I guess the OP meant:
```js
$('div').css('color', 'blue !important');
```

That said, the PR introduces the `nativeCSSStyleDeclaration` support test but doesn't use it and doesn't test it (see test/unit/support.js; btw, I don't think the support test is needed on `master`) and, as @dmethvin pointed out, there are no unit tests attached.

Also, this is a pretty critical code path and the `!important` use case is not terribly common; I'd like to see perf tests to make sure it doesn't make it worse for the 90% use cases.

bluelovers:
xd sorry type wrong demo code

$('div').css('color', 'blue !important');

timmywil:
Thank you for taking the time to submit a PR, but this will not make it in.

PR #2020, opened 2015-01-21
Ajax: auto stringify data

coderhaoxin:
- auto stringify data when contentType is json and data is not string

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -526,7 +526,11 @@ jQuery.extend({
 
 		// Convert data if not already a string
 		if ( s.data && s.processData && typeof s.data !== "string" ) {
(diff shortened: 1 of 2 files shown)


coderhaoxin:
when contentType is `'application/json'`, or `'application/json-patch+json'`, ...
should stringify the data when it's not string

dmethvin:
This is pretty easy to do with a plugin, I don't see a lot of value in building it in, and I suspect there are other contentType values that some people use to pass JSON. It would require further documentation changes as well, and ajax is already our most complicated API. I do appreciate you adding a unit test though. The best place to start a discussion about features is to open an issue first.

Issue #1835, opened 2014-11-04
Don't auto-insert tbody?


mzgol:
Now that we're talking breaking changes - jQuery inserts a tbody when appending to an empty table because of a "special" behavior of IE6-7 (http://bugs.jquery.com/ticket/14290). The `master` branch does the same for compat with jQuery 1.x. Should we revert this behavior now that we're dropping IE<8?

dmethvin:
@markelog Do you recall the issues you encountered with the `tbody` problem? Maybe I'm remembering it wrong.

markelog:
That was a long time ago, a bit busy today, will refresh my memory tomorrow

markelog:
> Now that we're talking breaking changes

Doesn't look like one

> because of a "special" behavior of IE6-7

Not exactly

> http://bugs.jquery.com/ticket/14290

Don't understand how this related, it just references `tbody` creation

> I recall @markelog trying to remove more of that in the early days of jQuery 2.0

Was not involve in it, you probably thinking of https://github.com/jquery/jquery/commit/19def21df41d6be6419e4d2480a656d9e1a0e6b7.

<a href="<a href="http://htmlhelp.com/reference/html40/tables/tbody.html">According</a> to html4 spec, `tbody` should always be present, which is true for IE6-7 implementation, `tbody` will be auto-inserted if table created through html serialization (declared in html or through `innerHTML` property), regardless if content is present or not. `tbody` should be created if DOM methods are used, otherwise IE6-7 will improperly render it.

According to html5 spec, `tbody` could be <a href="http://www.w3.org/html/wg/drafts/html/master/single-page.html#the-tbody-element">ommited</a> but will be <a href="http://www.w3.org/html/wg/drafts/html/master/syntax.html#parsing-main-intable">auto-inserted</a> if table has `td`, `th`, `tr` tags.

Modern browsers and IE8+ do not create `tbody` through html serialization for empty table and behave according to specification if there is a specific content, with DOM-methods, `tbody` will not be created, regardless of the content, even Opera 9 and FF 3.6 behave this way.

When `tbody` creation is nessary? For IE6-9, when empty table is statically created and `innerHTML` property is used, you basically can't use `innerHTML` on that node in these environments.

We no longer use `innerHTML` directly on tables (even true for `jQuery#html` method), so this code-path could be safely removed.

markelog:
[...]

PR #2021, opened 2015-01-21
Simplification of `manipulationTarget` helper

markelog:


test/unit/manipulation.js
```diff
@@ -2452,6 +2452,81 @@ test( "Validate creation of multiple quantities of certain elements (#13818)", 4
 	});
 });
 
+test( "Make sure tr element will be appended to tbody element of table when it present", function() {
```
dmethvin: You can either say "when it is present" or just "when present" for the descriptions on these.

src/manipulation.js
```diff
@@ -46,14 +46,14 @@ wrapMap.optgroup = wrapMap.option;
 wrapMap.tbody = wrapMap.tfoot = wrapMap.colgroup = wrapMap.caption = wrapMap.thead;
 wrapMap.th = wrapMap.td;
 
-// Manipulating tables requires a tbody
 function manipulationTarget( elem, content ) {
-	return jQuery.nodeName( elem, "table" ) &&
-		jQuery.nodeName( content.nodeType !== 11 ? content : content.firstChild, "tr" ) ?
+	if ( jQuery.nodeName( elem, "table" ) &&
+	    jQuery.nodeName( content.nodeType !== 11 ? content : content.firstChild, "tr" ) ) {
```
dmethvin: Did the spacing get messed up here?

test/unit/manipulation.js
markelog: Fixed

src/manipulation.js
markelog: Fixed

src/manipulation.js
```diff
@@ -46,14 +46,14 @@ wrapMap.optgroup = wrapMap.option;
...
 function manipulationTarget( elem, content ) {
-	return jQuery.nodeName( elem, "table" ) &&
-		jQuery.nodeName( content.nodeType !== 11 ? content : content.firstChild, "tr" ) ?
+	if ( jQuery.nodeName( elem, "table" ) &&
+		jQuery.nodeName( content.nodeType !== 11 ? content : content.firstChild, "tr" ) ) {
 
-		elem.getElementsByTagName("tbody")[0] ||
-			elem.appendChild( elem.ownerDocument.createElement("tbody") ) :
-		elem;
+		return elem.getElementsByTagName( "tbody" )[ 0 ] || elem;
```
dmethvin: In the old code, we guaranteed that a `<tr>` never got appended to a `<table>`. Do all the supported browsers in master auto-insert the `<tbody>` for us now? Do we have a unit test verifying that? I see some weak existing tests [here](https://github.com/markelog/jquery/blob/fdd04a4d47bc599b2615ec2eda01d19d469e24c5/test/unit/manipulation.js#L262-271) and [here](https://github.com/markelog/jquery/blob/fdd04a4d47bc599b2615ec2eda01d19d469e24c5/test/unit/manipulation.js#L1364-1377) but don't think they are making sure of this.

markelog: I guess you missed my <a href="https://github.com/jquery/jquery/issues/1835#issuecomment-70049946">comment</a>

dmethvin: Indeed I forgot you put in that much detail!

> According to html5 spec, tbody could be ommited but will be auto-inserted if table has td, th, tr tags.

If so should we ensure that with a test? Is it true for Android 2.x for example?

markelog: I will complete my previous comment there.

Basically, you can't use `innerHTML` property on the `table` element (only on `tbody`) in IE6-9 (i suppose at one point `innerHTML` might have been used like that), plus, in IE6-7 it would protect you from the case like this - 
```js
$( "<table/>" ).append( "<tr>...</tr>" );
```
No need for that now.

> Is it true for Android 2.x for example?

There could be situation when table exist with rows and columns without `tbody` parent - when you use DOM-manipulation methods. But it doesn't matter, since table should be rendered correctly with or without `tbody` in modern browsers. 

Search for `tbody` though, still necessary, since if user created table through html serialization (like through `innerHTML` property or defining it directly in html) will create `tbody` element even if it was omitted, so in order to correctly and consistently manipulate that table, we still need to search for it. 

And yeah, i checked it in Android 2.3... and IE6-7, Opera 9, Firefox 3.6 (Just to understand genesis of that code-path). They all follow that logic.

dmethvin:
How could I ever doubt @markelog, this LTGM. :smiley_cat:

Issue #1994, opened 2015-01-05
Don't test PATCH requests working in browsers other than IE


mzgol:
BrowserStack has frequent, recurring problems with PATCH requests not working through their proxy. Since we have no special logic for request type for browsers other than IE, we might as well skip the PATCH test for other browsers if we're on TestSwarm.

mzgol:
Related: #1938

mzgol:
Fixed in 2524da0.

mzgol:
I forgot to close the issue, doh.

PR #2026, opened 2015-01-25
Ajax: Run the PATCH test only in IE8 on TestSwarm

mzgol:
Fixes gh-1994

### DIFF
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -1578,16 +1578,19 @@ module( "ajax", {
 		}
 	} );
 
-	ajaxTest( "#13240 - jQuery.ajax() - support non-RFC2616 methods", 1, {
-		url: "data/echoQuery.php",
-		method: "PATCH",
-		success: function() {
-			ok( true, "success" );
-		},
-		error: function() {
-			ok( false, "error" );
-		}
-	});
+	// BrowserStack PATCH support sometimes breaks so on TestSwarm run the test in IE8 only.
+	if ( location.search.indexOf( "swarmURL=" ) === -1 || document.documentMode < 9 ) {
+		ajaxTest( "#13240 - jQuery.ajax() - support non-RFC2616 methods", 1, {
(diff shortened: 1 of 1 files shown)


test/unit/ajax.js
```diff
@@ -1578,16 +1578,19 @@ module( "ajax", {
...
-		url: "data/echoQuery.php",
-		method: "PATCH",
-		success: function() {
-			ok( true, "success" );
-		},
-		error: function() {
-			ok( false, "error" );
-		}
-	});
+	// BrowserStack PATCH support sometimes breaks so on TestSwarm run the test in IE8 only.
```
markelog: Hm, so browserstack can't fix that issue huh? Would need to be careful here when we switch from testswarm

mzgol: Well, I've already exchanged 22 mails with BrowserStack people, the first one at the end of October. This problem disappears and then re-appears again after some time. I'd rather not spend more time on this so it's better to restrict this to IE8.

As for switching from TestSwarm: sure, though if we forget about it then nothing terrible will happen. We'll notice it the first time PATCH fails again. ;) And then we'll fix the exclusion logic.

markelog: > Well, I've already exchanged 22 mails with BrowserStack people, the first one at the end of October. This problem disappears and then re-appears again after some time.

Yeah, i remember, what i fun read that is :-).

> We'll notice it the first time PATCH fails again

Would be kind of weird of the bug, if by the time they fix that issue we leave the testswarm. Anyways, fully agree on the matter.

test/unit/ajax.js
```diff
@@ -1578,16 +1578,19 @@ module( "ajax", {
...
+	if ( location.search.indexOf( "swarmURL=" ) === -1 || document.documentMode < 9 ) {
```
markelog: Hm, we probably need to start use `documentMode` everywhere where IE exclusion is needed.

PR #2027, opened 2015-01-25
Build: Speed up the Travis build

mzgol:
Run Travis tests on a container infrastructure to make it start faster.

See http://docs.travis-ci.com/user/workers/container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
 language: node_js
+sudo: false
 node_js:
 - '0.10'


dmethvin:
Are any other changes we'd need to make just optimizations? I saw the [section on caching](http://docs.travis-ci.com/user/caching/#Arbitrary-directories) but wasn't sure if there was any action required to make sure it all worked.

mzgol:
The only thing we could cache is node_modules but I'd like to not do it as
we don't shrinkwrap and in this way we'll never know if a newer version of
a sub-dependency broke sth terribly for us.

dmethvin:
It definitely makes sense *not* to cache that.

mzgol:
The main change is that builds start almost instantly whereas you have to
wait for regular ones, sometimes a minute, sometimes longer than that.

mzgol:
Maybe we could cache node_modules if we did `npm update` instead of `npm
install`, though?

dmethvin:
If that works reliably I'd be good with it. I'd bet the `npm install` takes a pretty good while.

mzgol:
> If that works reliably I'd be good with it. I'd bet the npm install takes a pretty good while.

OTH if sb submitted a PR making a mess out of `node_modules` we'd get into a bad state for subsequent builds. And seeing how currently not just `npm install` but the whole build [finishes in ~30 seconds](https://travis-ci.org/jquery/jquery/builds/48714215), we don't really need to tweak it further.

Issue #1938, opened 2014-12-17
Disable test(s) relying on Java applets on BrowserStack


mzgol:
Per our IRC conversation - from time to time BrowserStack has an out-of-date Java which causes the test run to hang, causing failures for the whole module that uses it. It seems it's just the `jQuery.acceptData` test:
https://github.com/jquery/jquery/blob/d39fef8472e5ca970e77c211d3b116f98fc10b14/test/unit/data.js#L136-L162

This seems to matter for IE8 only, the `master` branch doesn't have any special logic for `classid` handling. It seems there's a hight chance someone from the team will run the test suite on IE8 manually at least once in a while which would catch potential problems; we should disable this test just on BrowserStack then.

mzgol:
Seems we didn't achieve the intended result... Reopening.
![screen shot 2015-03-25 at 19 33 20](https://cloud.githubusercontent.com/assets/1758366/6832331/fdb4e75e-d325-11e4-92bf-e3269b52c697.png)

mzgol:
The following code triggers the dialog:
```js
jQuery.acceptData( document.createElement( "applet" ) );
```
see https://github.com/jquery/jquery/blob/0065e1f3e3021fc2bcb32e915177498bca034b34/test/unit/data.js#L776

mzgol:
Fixes were landed on `master`, `compat`, `2.1-stable` & `1.11-stable`. This should have been really fixed this time.

PR #2028, opened 2015-01-25
Data: Use a PDF object instead of a Java applet for acceptData testing

mzgol:
This should fix "Java out of date" errors on BrowserStack.

Fixes gh-1938

### DIFF
--- a/test/unit/data.js
+++ b/test/unit/data.js
@@ -767,7 +767,7 @@ test(".data doesn't throw when calling selection is empty. #13551", function() {
 });
 
 test("jQuery.acceptData", 11, function() {
-	var flash, applet;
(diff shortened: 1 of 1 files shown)


dmethvin:
So always BrowserStack pre-installs the PDF viewer on IE? If so then this is a good solution, at least until Adobe desupports *that*. :smile:

rwaldron:
This is BRILLIANT!

mzgol:
> So always BrowserStack pre-installs the PDF viewer on IE?

At least it does so on IE9 & IE8. :)

Credits to @gibson042 for the idea!

gibson042:
LGTM :+1:

PR #2029, opened 2015-01-25
Remove deprecated JSHint options

ntwb:
JSHint no longer supports `onevar`, `smarttabs` or `trailing` options (JSHint source [here](https://github.com/jshint/jshint/blob/3d9c430259afbecd002ca65662ea3103cae512b8/src/options.js#L875))

> JSHint is moving away from purely stylistic warnings. You should use JSCS for that.

Bonus: Includes "final new line" based on `.editorconfig` rule :stuck_out_tongue_winking_eye:

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -6,10 +6,7 @@
 	"expr": true,
 	"immed": true,
 	"noarg": true,
-	"onevar": true,
 	"quotmark": "double",
-	"smarttabs": true,
-	"trailing": true,
 	"undef": true,
 	"unused": true,
 


arthurvr:
These are indeed deprecated, but removing them would be against [our style guide](https://github.com/jquery/contribute.jquery.org/blob/master/pages/style-guide/js.md#linting). We recently had a discussion about it here: https://github.com/jquery/css-chassis/commit/24c71e2d2b35f7216a588bc4bb5895dd74955bdc.

markelog:
Core uses jshint 2.5, those options no longer there and jscs 1.7.3 where jquery preset have included substitutes for them.

@scottgonzalez what is the plan?

scottgonzalez:
This conversation has been had so many times in so many places that I won't even bother trying to track down the actual conversations. Here's the gist of it: We are waiting for @mikesherov to say that JSCS has implemented the right amount of checks before making this a required standard across all projects. Until then, each project is free to do what they want in order to test what works well and what doesn't (so long as their testing doesn't actually regress). Some projects ended up updating JSHint too early, which was bad because they lost the ability to enforce certain style rules. Today, it's safe to use JSCS to catch all the style-related rules that were dropped from JSHint, so it's safe to upgrade both. There is no requirement today other than being able to programmatically enforce the [JSHint options listed in the style guide](http://contribute.jquery.org/style-guide/js/#linting), even if there's a different tool doing that enforcement (such as JSCS + JSHint).

ntwb:
Thanks for the feedback everyone, greatly appreciated.

markelog:
> Some projects ended up updating JSHint too early, which was bad because they lost the ability to enforce certain style rules

Core uses fixed dependency versions but it doesn't help agaisnt auto-update of jshint, since if project uses `grunt-contrib-jshint` it uses 2.x version, cause grunt plugin uses tilde for jshint dependency.
So removal of style related options was inevitable.

> There is no requirement today other than being able to programmatically enforce the JSHint options listed in the style guide

I don't understand what is you trying to say, for example, `onevar` is listed as the requirement, but it is removed, it is however enforced by jscs.

You still want it in the jshint config?

> even if there's a different tool doing that enforcement (such as JSCS + JSHint).

This is not mentioned is the guide, it explicitly said that jshint should do it and such config of jshint should be used.

It seems those options should be removed from jshint config and guide should be updated.

scottgonzalez:
> I don't understand what is you trying to say, for example, onevar is listed as the requirement, but it is removed, it is however enforced by jscs.
> 
> You still want it in the jshint config?

No. If you're using a version of JSHint that doesn't support `onevar`, then it shouldn't be in the config. What matters is that something is enforcing what JSHint's previous `onevar` option means.

markelog:
Okay, so if understood you correctly need to wait on guide update (although it still seems weird to have it there) and you're okay with merging this PR and creating new one for chassis?

scottgonzalez:
>> even if there's a different tool doing that enforcement (such as JSCS + JSHint).

> This is not mentioned in the guide, it explicitly said that jshint should do it and such config of jshint should be used.
>
> It seems those options should be removed from jshint config and guide should be updated.

Because as of today, JSCS is not a tool that's required across projects. What is required and agreed upon is a set of JSHint settings. We're aiming for common sense here rather than being super pedantic about settings and specific versions of tools. Once we're ready to standardize JSCS, it will be listed in the style guide and the JSHint config will be updated, along with the addition of the JSCS config.

scottgonzalez:
> Okay, so if understood you correctly - need to wait on guide update (although it still seems weird to have it there) and you're okay with merging this PR and creating new one for chassis?

I'm not sure what you think is weird. You can merge this today, as I've already said. Please ignore Chassis right now, I don't want to get into a discussion about a dozen individual repositories. We have a standard, some projects have opted to test the future implementation by simultaneously using JSCS. Please just stick with what you're doing and when we're ready to change the standard, it will change.

markelog:
> I'm not sure what you think is weird.

Guide, currently, ask you to use `grunt jshint`, which as of version 0.10.0 (which was released a year ago) uses jshint 2.5.x, which doesn't include those options, but guide ask you to enforced them with `.jshintrc` config.

Guide doesn't say use different tool if those options aren't available in jshint. Currently, guide ask you to follow instructions that are not possible to follow. 

I think this is weird. 

If you like, we could move this discussion to the contribute repo.

markelog:
Had to change two other `.jshintrc` before merging, nevertheless, thank you!

ntwb:
:+1: Thanks @markelog

PR #2031, opened 2015-01-26
Manipulation: simplify html wrappers

markelog:
Pro - code obscuration 
Con - saves some bytes

Don't know, but felt to bring this up.

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -33,9 +33,27 @@ var
 		option: [ 1, "<select multiple='multiple'>", "</select>" ],
 
 		thead: [ 1, "<table>", "</table>" ],
-		col: [ 2, "<table><colgroup>", "</colgroup></table>" ],
-		tr: [ 2, "<table><tbody>", "</tbody></table>" ],
-		td: [ 3, "<table><tbody><tr>", "</tr></tbody></table>" ],
+
+		// Following wrappers are not fully defined because
+		// their parent elements (except for "table" element) could be omitted
+		// since browser parsers are smart enough to auto-insert them
+		col: [
+
+			2,
+			// Support: Android 2.3
+			// Android browser doesn't auto-insert colgroup element
+			"<table><colgroup>", "</colgroup></table>"
+		],
+
+		tr: [
+			2, // - auto-inserts "tbody" element
+			"<table>", "</table>"
+		],
+
+		td: [
+			3, // - auto-inserts "tbody" and "tr" elements
+			"<table>", "</table>"
+		],
 
 		_default: [ 0, "", "" ]
 	};
--- a/test/unit/manipulation.js
+++ b/test/unit/manipulation.js
@@ -2452,6 +2452,18 @@ test( "Validate creation of multiple quantities of certain elements (#13818)", 4
 	});
 });
 
+test( "Make sure col element is appended correctly", function() {
+	expect( 1 );
+
+	var table = jQuery( "<table cellpadding='0'><tr><td>test</td></tr></table>" );
+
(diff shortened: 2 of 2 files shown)


src/manipulation.js
```diff
@@ -33,9 +33,27 @@ var
...
-		td: [ 3, "<table><tbody><tr>", "</tr></tbody></table>" ],
+
+		// Following wrappers are not fully defined because
+		// their parent elements (except for "table" element) could be omitted
+		// since browser parsers are smart enough to auto-insert them
+		col: [
+
+			2,
+			// Support: Android 2.3
+			// Android browser doesn't auto-inserts colgroup element
```
dmethvin: auto-insert

dmethvin:
Sure, why not? The comments make it clear what cases we still have to fix, which is helpful.

src/manipulation.js
markelog: Fixed

src/manipulation.js
```diff
@@ -33,9 +33,27 @@ var
...
+		col: [
+
+			2,
+			// Support: Android 2.3
+			// Android browser doesn't auto-insert colgroup element
+			"<table><colgroup>", "</colgroup></table>"
+		],
+
+		tr: [
+			2, // - auto-inserts "tbody" element
```
gibson042: This comment and the one below should have lines of their own.

markelog: Could provide the full code-example how you would like to look?

gibson042: http://contribute.jquery.org/style-guide/js/#comments

I'm actually not sure, because it isn't clear if these are continuations of the "Support: Android 2.3". But a minimal change from what's here would look like:

```js
tr: [
	// - auto-inserts "tbody" element
	2,
	"<table>", "</table>"
],
```

Though I also think the comments would make more sense a level up, above the `tagName: [`.

markelog: > I'm actually not sure

Yep, me too, hence the question :-). 

> http://contribute.jquery.org/style-guide/js/#comments

Yeah, checked that, you example is also not applicable, since 

> Comments are always preceded by a blank line

So it sounds like, this should look like this - 

```js
// Following wrappers are not fully defined because
// their parent elements (except for "table" element) could be omitted
// since browser parsers are smart enough to auto-insert them

// Support: Android 2.3
// Android browser doesn't auto-insert colgroup element
col: [ 2, "<table><colgroup>", "</colgroup></table>" ],

...

// - auto-inserts "tbody" and "tr" elements
td: [ 3, "<table>", "</table>" ],
```

Which i guess is fine, but less obvious

dmethvin: In these cases we should ask, "What would JSCS do?" :smiley_cat: 

I would remove the leading ` - ` for the `td` case, but otherwise I think the less-spaced-out version is nicer to read.

gibson042: I think you misunderstood. Assuming `auto-inserts "tbody" element` and `auto-inserts "tbody" and "tr" elements` are not also limited to Android 2.3, I would expect something like this:
```js
thead: [ 1, "<table>", "</table>" ],

// Browsers auto-insert tbody
tr: [ 2, "<table>", "</table>" ],

// Browsers auto-insert tbody and tr
td: [ 3, "<table>", "</table>" ],

// Support: Android <=2.3 only
// Android browser doesn't auto-insert colgroup
col: [ 2, "<table><colgroup>", "</colgroup></table>" ],
```

markelog: Not sure what significant diff here is, but sure, let's do it like that.

> are not also limited to Android 2.3

Don't know about previous version of Android and honestly, don't want to know :-). We have established support comments style right?

PR #2032, opened 2015-01-26
Simplification of manipulation wrappers

### LINKED ISSUE #2002: Simplification of manipulation wrappers
Talking about [those](https://github.com/jquery/jquery/blob/8e738f0aa89ca333dac61a1d1d0c4964762464c1/src/manipulation.js#L49-56).

We can remove most of the wrappers for `col` element (effectively removing it entirely) from `master` and `compat` and remove `area` and `legend` wrappers, since they was needed for IE6-7 only.

PR tomorrow.

markelog:
Since support for IE6-7 was dropped these wrappers no longer required
Ref 90d7cc1d8b2ea7ac75f0eacb42439349c9c73278
Fixes gh-2002

Note: if this would be merged, it would not auto-close #2002.
Should we mention "Fixes ..." for compat only commits?

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -47,9 +47,10 @@ var nodeNames = "abbr|article|aside|audio|bdi|canvas|data|datalist|" +
 	// We have to close these tags to support XHTML (#13200)
 	wrapMap = {
 		option: [ 1, "<select multiple='multiple'>", "</select>" ],
-		legend: [ 1, "<fieldset>", "</fieldset>" ],
-		area: [ 1, "<map>", "</map>" ],
+
+		// Support: IE8
 		param: [ 1, "<object>", "</object>" ],
+
 		thead: [ 1, "<table>", "</table>" ],
 		tr: [ 2, "<table><tbody>", "</tbody></table>" ],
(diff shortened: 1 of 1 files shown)


gibson042:
`Fixes gh-2002` will still be valuable for anyone looking at the commit in isolation, even if GitHub doesn't auto-close because it's against a non-master branch. We should keep the reference.

markelog:
> We should keep the reference.

Didn't ask should or shouldn't we need to keep the reference. Did ask about the *way* we need to store the ticket number. `Ref`, `Fixes`? Judging by <a href="http://contribute.jquery.org/commits-and-pull-requests/">guide</a> neither of those are applicable.

I'd say `Fixes` even if github doesn't close anything.

gibson042:
> I'd say Fixes even if github doesn't close anything.

I agree. `Fixes gh-2002`.

dmethvin:
Agreed. The statement is true even if Github doesn't close the ticket. In our case for cherry-picked commits that's fine since the one from `master` will close it. It's just the `compat`-only ones that we have to do the closing.

markelog:
Merged

Issue #1895, opened 2014-12-03
DOM inserted scripts should run asynchronously


markelog:
See discussion in https://github.com/jquery/jquery/issues/1887

markelog:
should this be a blocker?

timmywil:
While the old behavior may be expected by some users, I don't think we've documented that users could count on synchronicity. I wouldn't call it a blocker.

gibson042:
That's answering a slightly different question, but one upon which blocker status is partially dependent: is this a breaking change? It's currently [marked as such](https://github.com/jquery/jquery/labels/Behavior%20Change), but as you point out, we haven't documented synchronous execution. Given the lengths the code goes to with respect to script execution, however, that's arguably a documentation bug. So should giving up script completion certainty (point 2 of https://github.com/jquery/jquery/pull/2126#issuecomment-78273289 ) after a manipulation method returns be considered backwards incompatible? I'm inclined towards "yes", but could be talked out of it if we want to introduce this in [^3.1](https://docs.npmjs.com/misc/semver#caret-ranges-1-2-3-0-2-5-0-0-4).

And of course, if it _is_ breaking then it requires a major version bump—miss 3.0 and it can't be introduced until 4.0.

dmethvin:
Even though we never documented it, I am pretty sure there is code out there depending on it. I think it should go into a major-point release.

timmywil:
It's still scheduled for 3.0, so whether we label it a blocker is somewhat moot.

timmywil:
As for whether it is a breaking change, I'm fine with saying "yes".

sandstrom:
This is great! Will make jQuery & Content Security Policy good friends :boat:

timmywil:
Per meeting, pushing to 4.0.

mgol:
A relevant comment by @gibson042 from https://github.com/jquery/jquery/issues/3969#issuecomment-365063998:
>  To document what was discussed in the meeting today, we SHOULD support restrictive CSP configuration but MUST NOT break currently-working strict ordering of script evaluation in `domManip` (property 1 in https://github.com/jquery/jquery/pull/2126#issuecomment-78273289 ). A proper fix will probably adopt script elements for even same-site content, but also implement manual serialization similar to the [suggestion](https://github.com/jquery/jquery/pull/2126#issuecomment-78762735) (leveraging something like [`rel=preload`](https://www.w3.org/TR/preload/) if performance is an issue).

[...]

PR #2033, opened 2015-01-26
Simplification, clarify and update some of the ajax logic

markelog:


gibson042:
> Remove "type: GET" from "evalUrl" method
since "get" type is by default already

Anything that can be controlled by global ajaxSettings should be explicit here.

dmethvin:
Oh yeah, really good point. I've seen some strange problems caused by that in projects.

markelog:
1. We have <a href="https://github.com/jquery/jquery/blob/4fae91141c04dffb5094b5961eb5d7b2a1f3f423/src/ajax/load.js#L49">this</a> for more then two years now, no complains for far.
2. We already warn people about it in the <a href="http://api.jquery.com/jQuery.ajaxSetup/">docs</a>.
3. if user does set, for example "POST" method, as a default (for some unknown reason), that would  mean he wants to make request with this method.

dmethvin:
Wait, won't the $.ajax call go away (or change to script transport) when gh-1895 lands? Why make that change now?

markelog:
Nice catch! 

Although i did only preliminary work on gh-1895, so i'm not sure how much we can change there. So it might only making `async: true` (removing this property that is) in `evalUrl` and change a bunch of tests, or removing this method entirely and rewriting manipulation module a bit and change a bunch of tests.

We can keep things separate, atomic, improve code at hand and deal with gh-1895 in relevant PR.

markelog:
@gibson042 comment is taken to the account, `load` is now explicitly uses `GET` method too. Test added.

markelog:
Added another commit that replaces `jqXHR.complete` in `jQuery#load` since it was deprecated since 1.8

src/ajax/load.js
```diff
@@ -64,7 +66,7 @@ jQuery.fn.load = function( url, params, callback ) {
 				// Otherwise use the full result
 				responseText );
 
-		}).complete( callback && function( jqXHR, status ) {
+		}).always( callback && function( jqXHR, status ) {
```
gibson042: We could really use some comments here, because the code takes advantage of some tricky behavior. If the request succeeds, this function gets `( data, status, jqXHR )`, but they are ignored because `response` was set above. If it fails, this function gets `( jqXHR, status, error )` (the first two of which coincide with `complete`) and are used in the body. It's probably also worth documenting that the `self.each` takes advantage of `.each`'s undocumented second parameter.

markelog:
Comment for `always` callback added.

`jQuery#each` second argument was always internal and `each` performance matter plus byte size and all that - that is the only place left that uses it.

@timmywil, @dmethvin @gibson042 i think you know where i'm going with this. How about it?

dmethvin:
Yeah, i was wondering if we could just get rid of it when i saw the comment from @gibson042 earlier.

timmywil:
I LIKE it

markelog:
Created - https://github.com/jquery/jquery/issues/2090

PR #2036, opened 2015-01-26
Data: Fix to update if key was added as-is when using dash

meetselva:
Updating fix to master with reference to the original [PR 1905](https://github.com/jquery/jquery/issues/1905)

Unit test passed. Please review.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -151,7 +151,12 @@ jQuery.fn.extend({
 				// First, attempt to store a copy or reference of any
 				// data that might've been store with a camelCased key.
 				var data = dataUser.get( this, camelKey );
-
+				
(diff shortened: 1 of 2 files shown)


meetselva:
@dmethvin It was my bad, I think i abused `==` with that `null` check. Corrected to check for undefined. Also fixed formatting for the comments.

timmywil:
After [much deliberation](https://docs.google.com/spreadsheets/d/1gf0vzYGwg7IA7yIkyJs-Jk31SWDogQrBkzPb-xl6u-o/), we decided to go a different route. This happened to be a very hairy issue involving lots of issues. Thank you very much for your work here and I'm sorry we weren't able to merge this PR.

PR #2038, opened 2015-01-26
Build: Update QUnit to latest (1.17.1)

jzaefferer:
Running `grunt npmcopy` also updated requirejs and sinon. The sinon changes looked like they'd actually break something, so I ignored them both. Can't tell why the dependencies were updated without updating the copies in the repo.

### DIFF
--- a/external/qunit/LICENSE.txt
+++ b/external/qunit/LICENSE.txt
@@ -1,5 +1,4 @@
-Copyright 2006, 2014 jQuery Foundation and other contributors,
-https://jquery.org/
+Copyright jQuery Foundation and other contributors, https://jquery.org/
 
 This software consists of voluntary contributions made by many
 individuals. For exact contribution history, see the revision history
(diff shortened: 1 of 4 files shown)


mzgol:
> Running `grunt bowercopy`

You mean `grunt npmcopy`?

jzaefferer:
Yes yes, that.

jzaefferer:
* [ ] Yes
* [ ] No
* [ ] Maybe

mzgol:
Since nothing changed in Core files, LGTM.

We should get the libs copies synced with what's in package.json btw...

jzaefferer:
Well, I'll leave that exercise to someone who's more familiar with those updated dependencies.

markelog:
If there is no objections - will land this tomorrow

markelog:
Hm, since this update broke couple traversing and manipulation tests, since new qunit added another form to the test-suite and some of jQuery selectors weren't specific.

Will fix that up.

PR #2043, opened 2015-01-29
Fix .offset() to work in ShadowDOM

### LINKED ISSUE #1784: Issues using ShadowDom and ShadowRoot elements

NekR:
This pull request fixes #1784 issue.

Quick summary:
* ```element.contains()``` and ```element.compareDocumentPosition()``` does not work with ShadowDOM and always treated as disconnected even if it in the tree
* ```element.getBoundingClientRect()``` work fine with elements in ShadowDOM, returns correct values when element is in the tree and zeroes when not

Behavior of ```getBoundingClientRect```:
* For hidden elements returns all zeros
* For disconnected elements return all zeros
* In IE8 throws error when uses with disconnected element

What was made:
* Added guard which returns dummy block when rect has only zeroes (disconnected or hidden)
* ```jQuery.contains()``` check was not removed since without it IE8 will throw error and hence we need to return early for IE8 if element is disconnected
* But because ```jQuery.contains()``` has no effect with ShadowDOM (does not work as expected) I added check for existence of ShadowDOM before calling ```jQuery.contains()``` so we do not rely on it when it does not work.

Questions:
* Maybe it's better to move check of ShadowDOM existence to ```jQuery.supports.ShadowDOM``` and then use it?
* @nazar-pc mentioned in #1784 what he already made tests for ```.offset()``` in ShadowDOM. I can made my own test or we can use his since they are already made. If you decide to use his test, then probable it's better what he will commint them by himself.

Also @gibson042 said in #1784
> I don't like this conflation of unrelated behaviors.

About this line of code: ```if ( !elem.createShadowRoot && !jQuery.contains( docElem, elem ) ) {```
I agree with it, but this check is more performant than tree traversal or ```try..catch``` block. Also it's not so much unrelated, it's used to prevent use ```jQuery.contains()``` in situation when it does not work _and_, by the way, prevents IE8 to fail.

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -32,7 +32,7 @@ jQuery.offset = {
 			elem.style.position = "relative";
 		}
 
-		curOffset = curElem.offset();
+		curOffset = curElem.offset() || { top: 0, left: 0 };
 		curCSSTop = jQuery.css( elem, "top" );
 		curCSSLeft = jQuery.css( elem, "left" );
 		calculatePosition = ( position === "absolute" || position === "fixed" ) &&
@@ -57,10 +57,10 @@ jQuery.offset = {
 		}
 
 		if ( options.top != null ) {
-			props.top = ( options.top - ( curOffset ? curOffset.top : 0 ) ) + curTop;
+			props.top = ( options.top - curOffset.top ) + curTop;
 		}
 		if ( options.left != null ) {
-			props.left = ( options.left - ( curOffset ? curOffset.left : 0 ) ) + curLeft;
+			props.left = ( options.left - curOffset.left ) + curLeft;
 		}
 
 		if ( "using" in options ) {
@@ -93,18 +93,15 @@ jQuery.fn.extend({
 		rect = elem.getBoundingClientRect();
 
 		// Make sure element is not hidden (display: none) or disconnected
-		if ( !rect.width && !rect.height && !rect.left &&
-			!rect.top && !rect.right && !rect.bottom ) {
-			return;
+		if ( rect.width || rect.height || elem.getClientRects().length ) {
+			win = getWindow( doc );
+			docElem = doc.documentElement;
+
+			return {
+				top: rect.top + win.pageYOffset - docElem.clientTop,
+				left: rect.left + win.pageXOffset - docElem.clientLeft
+			};
 		}
-
-		win = getWindow( doc );
-		docElem = doc.documentElement;
-
-		return {
-			top: rect.top + win.pageYOffset - docElem.clientTop,
-			left: rect.left + win.pageXOffset - docElem.clientLeft
-		};
 	},
 
 	position: function() {
--- a/src/offset.js
+++ b/src/offset.js
@@ -92,7 +92,7 @@ jQuery.fn.extend({
 
 		rect = elem.getBoundingClientRect();
 
-		// Make sure element is not hidden or disconnected
+		// Make sure element is not hidden (display: none) or disconnected
 		if ( !rect.width && !rect.height && !rect.left &&
 			!rect.top && !rect.right && !rect.bottom ) {
 			return;
--- a/src/offset.js
+++ b/src/offset.js
@@ -82,7 +82,7 @@ jQuery.fn.extend({
 				});
 		}
 
-		var docElem, win,
+		var docElem, win, rect,
 			elem = this[ 0 ],
 			box = { top: 0, left: 0 },
 			doc = elem && elem.ownerDocument;
@@ -91,18 +91,19 @@ jQuery.fn.extend({
 			return;
 		}
 
-		docElem = doc.documentElement;
+		rect = elem.getBoundingClientRect();
 
-		// Make sure it's not a disconnected DOM node
-		if ( !jQuery.contains( docElem, elem ) ) {
+		// Make sure element is not hidden or disconnected
+		if ( !rect.width && !rect.height ) {
 			return box;
 		}
 
-		box = elem.getBoundingClientRect();
+		docElem = doc.documentElement;
 		win = getWindow( doc );
+
 		return {
-			top: box.top + win.pageYOffset - docElem.clientTop,
-			left: box.left + win.pageXOffset - docElem.clientLeft
+			top: rect.top + win.pageYOffset - docElem.clientTop,
+			left: rect.left + win.pageXOffset - docElem.clientLeft
 		};
 	},
 
--- a/src/offset.js
+++ b/src/offset.js
@@ -57,10 +57,10 @@ jQuery.offset = {
 		}
 
 		if ( options.top != null ) {
-			props.top = ( options.top - curOffset.top ) + curTop;
+			props.top = ( options.top - ( curOffset ? curOffset.top : 0 ) ) + curTop;
 		}
 		if ( options.left != null ) {
-			props.left = ( options.left - curOffset.left ) + curLeft;
+			props.left = ( options.left - ( curOffset ? curOffset.left : 0 ) ) + curLeft;
 		}
 
 		if ( "using" in options ) {
@@ -93,8 +93,9 @@ jQuery.fn.extend({
 		rect = elem.getBoundingClientRect();
 
 		// Make sure element is not hidden or disconnected
-		if ( !rect.width && !rect.height ) {
-			return { top: 0, left: 0 };
+		if ( !rect.width && !rect.height && !rect.left &&
+			!rect.top && !rect.right && !rect.bottom ) {
+			return;
 		}
 
 		win = getWindow( doc );
--- a/src/offset.js
+++ b/src/offset.js
@@ -91,6 +91,8 @@ jQuery.fn.extend({
 			return;
 		}
 
+		docElem = doc.documentElement;
+
 		// Make sure it's not a disconnected DOM node,
 		// which is enough only if user-agent does not support ShadowDOM
 		if ( !elem.createShadowRoot && !jQuery.contains( docElem, elem ) ) {
@@ -104,7 +106,6 @@ jQuery.fn.extend({
 			return box;
 		}
 
-		docElem = doc.documentElement;
 		win = getWindow( doc );
 
 		return {
--- a/src/offset.js
+++ b/src/offset.js
@@ -91,6 +91,12 @@ jQuery.fn.extend({
 			return;
 		}
 
+		// Make sure it's not a disconnected DOM node,
+		// which is enough only if user-agent does not support ShadowDOM
+		if ( !elem.createShadowRoot && !jQuery.contains( docElem, elem ) ) {
+			return box;
+		}
+
 		rect = elem.getBoundingClientRect();
 
 		// Make sure element is not hidden or disconnected
--- a/src/offset.js
+++ b/src/offset.js
@@ -84,29 +84,21 @@ jQuery.fn.extend({
 
 		var docElem, win, rect,
 			elem = this[ 0 ],
-			box = { top: 0, left: 0 },
 			doc = elem && elem.ownerDocument;
 
 		if ( !doc ) {
 			return;
 		}
 
-		docElem = doc.documentElement;
-
-		// Make sure it's not a disconnected DOM node,
-		// which is enough only if user-agent does not support ShadowDOM
-		if ( !elem.createShadowRoot && !jQuery.contains( docElem, elem ) ) {
-			return box;
-		}
-
 		rect = elem.getBoundingClientRect();
 
 		// Make sure element is not hidden or disconnected
 		if ( !rect.width && !rect.height ) {
-			return box;
+			return { top: 0, left: 0 };
 		}
 
 		win = getWindow( doc );
+		docElem = doc.documentElement;
 
 		return {
 			top: rect.top + win.pageYOffset - docElem.clientTop,
--- a/test/unit/offset.js
+++ b/test/unit/offset.js
@@ -61,7 +61,7 @@ test("disconnected node", function() {
 
 	var result = jQuery( document.createElement("div") ).offset();
 
-	equal(typeof result, "undefined", "Check result typeof" );
+	equal( typeof result, "undefined", "Check result typeof" );
 });
 
 testIframe("offset/absolute", "absolute", function($, iframe) {
--- a/test/unit/offset.js
+++ b/test/unit/offset.js
@@ -57,12 +57,11 @@ test("object without getBoundingClientRect", function() {
 });
 
 test("disconnected node", function() {
-	expect(2);
+	expect(1);
 
 	var result = jQuery( document.createElement("div") ).offset();
 
-	equal( result.top, 0, "Check top" );
-	equal( result.left, 0, "Check left" );
+	equal(typeof result, "undefined", "Check result typeof" );
 });
 
 testIframe("offset/absolute", "absolute", function($, iframe) {
--- a/test/unit/offset.js
+++ b/test/unit/offset.js
@@ -72,7 +72,7 @@ test("hidden (display: none) element", function() {
 	expect(1);
 
 	var result,
-		node = jQuery("<div style='display: none' />").appendTo(document.body);
+		node = jQuery("<div style='display: none' />").appendTo("#qunit-fixture");
 
 	try {
 		result = node.offset();
--- a/test/unit/offset.js
+++ b/test/unit/offset.js
@@ -56,12 +56,31 @@ test("object without getBoundingClientRect", function() {
 	equal( result.left, 0, "Check left" );
 });
 
-test("disconnected node", function() {
+test("disconnected element", function() {
 	expect(1);
 
-	var result = jQuery( document.createElement("div") ).offset();
+	var result;
 
-	equal( typeof result, "undefined", "Check result typeof" );
+	try {
+		result = jQuery( document.createElement("div") ).offset();
+	} catch ( e ) {}
+
+	ok( !result, "no position for disconnected element" );
+});
+
+test("hidden (display: none) element", function() {
+	expect(1);
+
+	var result,
+		node = jQuery("<div style='display: none' />").appendTo(document.body);
+
+	try {
+		result = node.offset();
+	} catch ( e ) {}
+
+	node.remove();
+
+	ok( !result, "no position for hidden (display: none) element" );
 });
 
 testIframe("offset/absolute", "absolute", function($, iframe) {


NekR:
Oh, one more thing -- this PR slightly changes things of returned data for hidden elements:

**Before**: ```{ top: 0 + pageYOffset - clientTop, left: 0 + pageYOffset - clientLeft }```
**After**: always ```{ top: 0, left: 0 }```

nazar-pc:
Once more, probably I've missed something, why do we need `if ( !rect.width && !rect.height ) {` if documentation says:
> While it is possible to get the coordinates of elements with visibility:hidden set, display:none is excluded from the rendering tree and thus has a position that is undefined.

We can save few bytes by removing it without breaking behavior defined before.

Also
```js
if ( !doc ) {
	return;
}
```

can likely be removed (https://github.com/jquery/jquery/issues/1784#issuecomment-71702258), it makes no sense at all to call `$(document).offset()` because of:
> Get the current coordinates of the first element, or set the coordinates of every element, in the set of matched elements, relative to the document.

It is just ridiculous to do that (also, it does not even returns zeros), thus we can save few more bytes.


**If mentioned changes will be applied I like it otherwise.**

NekR:
> Once more, probably I've missed something, why do we need if ( !rect.width && !rect.height ) { if documentation says:
> We can save few bytes by removing it without breaking behavior defined before.

```getBoundingClientRect()``` always returns ClientRect and for hidden elements (```display: none;``` for example) it returns: ```ClientRect {height: 0, width: 0, left: 0, bottom: 0, right: 0}```

> can likely be removed (#1784 (comment)), it makes no sense at all to call $(document).offset() because of:

I do not decide such things and even more, that lane of code is not related to ShadowDOM issue which I want to fix with this PR. So, probably, if you want to change that behavior of jQuery then I think you can create separate PR which resolves that.

Anyway, thank you, because your comment helped me remember this thing about IE8:
> [1] In IE8 and below, the TextRectangle object returned by getBoundingClientRect() lacks height and width properties. Also, additional properties (including height and width) cannot be added onto these TextRectangle objects.

So I need to change that line of code to support IE8. Will do soon.

**Comment by mzgol** [comment]
> So I need to change that line of code to support IE8. Will do soon.

This PR is reported against the `master` branch where we support IE9+ only. The `compat` branch supports IE8+.

**Comment by NekR** [comment]
> This PR is reported against the master branch where we support IE9+ only. The compat branch supports IE8+.

I understand it. But in this comment https://github.com/jquery/jquery/issues/1784#issuecomment-71842046 @gibson042 said that for this case it would be enough to have one PR for both versions (If I correctly understod him).

So:
* ShadowDOM problem should be fixed if both versions, I think.
* If you want, I may remove all IE8 tricks from this code (and do not add more) and will make separate PR for ```compat``` version.
* If you want same code in both versions, then I will add fix for IE8 here.

**Comment by mzgol** [comment]
> I understand it. But in this comment https://github.com/jquery/jquery/issues/1784#issuecomment-71842046 @gibson042 said that for this case it would be enough to have one PR for both versions (If I correctly understod him).

If the patch needed for both branches is similar, we usually just do the cherry-picking with those minor necessary modifications ourselves. But the PR against the `master` branch should not have fixes for IE8.

You may either remove the IE8 workarounds from this PR and leave it at that or you can report a separate PR to `compat` that would contain the IE8 quirks.

File: src/offset.js [hunk 1]
```diff
@@ -93,16 +93,24 @@ jQuery.fn.extend({
 
 		docElem = doc.documentElement;
 
-		// Make sure it's not a disconnected DOM node
-		if ( !jQuery.contains( docElem, elem ) ) {
+		// Make sure it's not a disconnected DOM node,
+		// which is enough only if user-agent does not support ShadowDOM
+		if ( !elem.createShadowRoot && !jQuery.contains( docElem, elem ) ) {
```
gibson042: The [conflation of unrelated behaviors](https://github.com/jquery/jquery/issues/1784#issuecomment-71842046) is Shadow DOM support and applicability of `getBoundingClientRect` to disconnected elements. As I mentioned in #1784, I'd like to see this entire block (and all detection or special treatment of disconnection) removed.

NekR:
> If the patch needed for both branches is similar, we usually just do the cherry-picking with those minor necessary modifications ourselves. But the PR against the master branch should not have fixes for IE8.

Sure.

> The conflation of unrelated behaviors is Shadow DOM support and applicability of getBoundingClientRect to disconnected elements.

Yeah, I already commented this part in PR topic.

> As I mentioned in #1784, I'd like to see this entire block (and all detection or special treatment of disconnection) removed.

Ok, sure. I just was not clear about it in your previous comment. This would be better "We do not need IE8 support in master so please remove IE8 related code here and we will figure out about other way for IE8 in ```compat``` branch by ourselves".
I understand why it's unrelated to ```master``` -- because of no support of IE8. In all other cases it's very related unless you plan to use there ```try..catch``.

Will remove that block soon.

src/offset.js
```diff
@@ -82,27 +82,27 @@ jQuery.fn.extend({
...
 
-		docElem = doc.documentElement;
+		rect = elem.getBoundingClientRect();
 
-		// Make sure it's not a disconnected DOM node
-		if ( !jQuery.contains( docElem, elem ) ) {
-			return box;
+		// Make sure element is not hidden or disconnected
+		if ( !rect.width && !rect.height ) {
+			return { top: 0, left: 0 };
```
gibson042: I'd rather just `return;` here... `{ top: 0, left: 0 }` looks like a real value, when we know it cannot be.

src/offset.js
```diff
@@ -82,27 +82,27 @@ jQuery.fn.extend({
...
 		}
 
-		docElem = doc.documentElement;
+		rect = elem.getBoundingClientRect();
 
-		// Make sure it's not a disconnected DOM node
-		if ( !jQuery.contains( docElem, elem ) ) {
-			return box;
+		// Make sure element is not hidden or disconnected
+		if ( !rect.width && !rect.height ) {
```
staabm: should this be `||` instead? as soon as a element has either no width or height it is not visible, isn't it?

nazar-pc: What about 0 height and 0 width and border 1px? I'd remove that check completely.

gibson042: `getBoundingClientRect` returns border-boxes.

NekR: > should this be || instead? as soon as a element has neither width nor height it is not visible, isn't it?

Honestly, I think it should be:
```javascript
if ( !rect.width && !rect.height && !rect.top && !rect.left && !rect.bottom && !rect.right ) {
```
Because only this case matches truly hidden elements (```display: none;``` or disconnected). In all other cases real values might be useful.

src/offset.js
NekR: Sounds resonable. Will change it in next commit.

src/offset.js
```diff
@@ -82,27 +82,28 @@ jQuery.fn.extend({
...
 
-		docElem = doc.documentElement;
+		rect = elem.getBoundingClientRect();
 
-		// Make sure it's not a disconnected DOM node
-		if ( !jQuery.contains( docElem, elem ) ) {
-			return box;
+		// Make sure element is not hidden (display: none) or disconnected
+		if ( !rect.width && !rect.height && !rect.left &&
+			!rect.top && !rect.right && !rect.bottom ) {
```
gibson042: This is overkill; the only properties that matter for detecting visibility are width and height—and note that there is literally no way to differentiate no-content inline elements that happen to be at viewport origin from disconnected or display:none elements with gBCR, since they all have zero width/height/top/left. All are arguably "hidden" in the parlance of our API documentation, so our handling comes down to a matter of taste (which is responsible for the churn here; sorry about that @NekR):
* assume any data is valid and always apply scroll offset
  * inaccurate for disconnected/display:none
* assume zero width and zero height indicate disconnected/display:none; return `undefined` or `null` or `{ top: 0, left: 0 }`
  * inaccurate for rare zero-size edge cases
* consider zero width _or_ zero height "hidden"; return `undefined` or `null` or `{ top: 0, left: 0 }`
  * inaccurate for slightly less rare zero-size edge cases

All the "inaccurate" results are allowed by our documentation, but are probably not equally useful to consumers (and in fact, the `undefined` or `null` results may be beneficial). Option 2 is only relevant if we want to give normal treatment to _single_-dimension elements (e.g., explicitly-sized elements or no-content display:block), so what's more valuable: accurate position for zero-size elements (first option), or distinct output for probably-unpositionable input (third or second option, especially with non-rect return)?

@dmethvin, @mzgol, @timmywil: care to weigh in?

test/unit/offset.js
```diff
@@ -57,12 +57,11 @@ test("object without getBoundingClientRect", function() {
...
 
 test("disconnected node", function() {
-	expect(2);
+	expect(1);
 
 	var result = jQuery( document.createElement("div") ).offset();
 
-	equal( result.top, 0, "Check top" );
-	equal( result.left, 0, "Check left" );
+	equal( typeof result, "undefined", "Check result typeof" );
```
gibson042: A bit too precise, especially given that the call will throw on IE8 in the cherry-pick to `compat`... how about a try/catch and `ok( !result, "no position for disconnected element" )`?

And we'll want the same for a `display: none` element.

src/offset.js
```diff
@@ -82,27 +82,28 @@ jQuery.fn.extend({
 				});
 		}
 
-		var docElem, win,
+		var docElem, win, rect,
 			elem = this[ 0 ],
-			box = { top: 0, left: 0 },
 			doc = elem && elem.ownerDocument;
 
 		if ( !doc ) {
```
gibson042: I believe this can also go, along with the related "object without getBoundingClientRect" test, since all supported browsers hav gBCR.

src/offset.js
```diff
@@ -57,10 +57,10 @@ jQuery.offset = {
 		}
 
 		if ( options.top != null ) {
-			props.top = ( options.top - curOffset.top ) + curTop;
+			props.top = ( options.top - ( curOffset ? curOffset.top : 0 ) ) + curTop;
 		}
 		if ( options.left != null ) {
-			props.left = ( options.left - curOffset.left ) + curLeft;
+			props.left = ( options.left - ( curOffset ? curOffset.left : 0 ) ) + curLeft;
```
gibson042: Did you try leaving these lines alone in favor of adjusting the above to `curOffset = curElem.offset() || { top: 0, left: 0 }`? It strikes me as more straightforward, and possibly smaller.

src/offset.js
NekR: In my personal opinion, I do not think what visually hidden element is always hidden by perspective of developer or even browser. If element participates in browser layout -- then that element absolutely not hidden. Consider following situation:

Someone uses empty ```<div>``` like a placeholder for future content (ajax loading for example). By some situation -- user scroll page so that position ```<div>``` becomes in view (or other same situation). And then developer needs to load content in that ```<div>```.
So, to determine this situation developer absolutely need to know position of that ```<div>```, even if height of that ```<div>``` is 0.

Try to run this example line by line in console ($0 === some block in this page):
```javascript
var a = document.createElement('div');
$0.appendChild(a);
a.getBoundingClientRect(); // ClientRect {height: 0, width: 653.6363525390625, left: 124.90056610107422, bottom: 405.553955078125, right: 778.5369262695312…}
a.style.margin = '20px';
a.getBoundingClientRect(); // ClientRect {height: 0, width: 613.6363525390625, left: 144.9005584716797, bottom: 425.553955078125, right: 758.5369262695312…}
```

As you can see, practically that element is not hidden (even if it's not visible by eye on a page). Even more, it still participates in layout and when you will add margin to element, it still will be hidden, but will take 40px height on a screen.

If this is "rare" case, then okay, I will simplify that condition. But for now I will wait for other responses.

test/unit/offset.js
NekR: Okay, I can do that.

src/offset.js
NekR: I am not sure I got your message here. If you plan to remove that condition then you can remove it in separate thread. This PR is not about refacroting, but about support ShadowDOM.

src/offset.js
NekR: Yes, I thought about it and decided to not do that because it will create extra Object memory. Which is absolutely slower than simple check.

gibson042: But not slower in a way that matters... the cost of creating another object is dwarfed by the cost of layout, and these lines are not likely to be the bottleneck in any application.

src/offset.js
gibson042: Right; agreed.

src/offset.js
gibson042: I agree with you, which is why I'm advocating a test like `!rect.width && !rect.height` that will not trip up on single-dimension elements such as your no-content div. But at the same time, we must recognize that dimensionless (e.g., no-content _inline_) can be disconnected/display:none false positives (or vice versa)... I don't think the extra top/left checks are worth it, given that even they are imperfect.

src/offset.js
NekR: Invocation of ```getBoundingClientRect()``` does not mean that layout will be called each time when it's invoked, only if there are uncalculated changes (also I believe there is not real layout for truly hidden or disconnected elements).
It's okay and I can change that code if you prefer that. But it's strange to hear about that one slow operation (relatively slow to another which is faster) does not matter because there around are other slow operations.
Also I saw application of dozens of thousands simple calculations, where was no such things like layout, but they were incredibly slow just because every operation matters.

Okay, that is just thoughts and I will change that part of code later.

src/offset.js
NekR: > I don't think the extra top/left checks are worth it

Do you talk here about performance? 

There is only one possible false-positive situation when element is placed at top-left corner and screen is not scrolled (or simply position: fixed). 
If you really care about this situation we can use ```getClientRects()``` which returns 0 length list for elements which does not participate in layout, according to spec:
>1. If the element on which it was invoked does not have an associated layout box return an empty DOMRectList object and stop this algorithm.

> http://www.w3.org/TR/cssom-view/#dom-element-getclientrects

So this might be the case:
```javascript
if ( !element.getClientRects().length ) {
  // We are not interested
}
```
Also, if you do not like ```getClientRects()``` method, we can use something like ```.offsetParent``` which returns ```null``` almost in same situations (but ```getClientRects()``` is better of course):

> The **offsetParent** attribute must return the result of running these steps:
1. If any of the following holds true return null and terminate this algorithm:
* The element does not have an associated CSS layout box.
* The element is the root element.
* The element is the HTML body element.
* The element’s computed value of the position property is fixed.

> http://dev.w3.org/csswg/cssom-view/#dom-htmlelement-offsetparent

Anyway, we can be here accurate as much as we want.

NekR: > Yes to both.

Got it now, very good idea @gibson042.

Will make all these changes soon.

NekR:
Hi all, sorry for delay I had really busy days. I just made fixes as we discussed here. Also added test for hidden elements.
So if I understood all right -- we do not need special tests for ShadowDOM since we do not use special checks for it and just assume what it works for any visible and attached element in the tree, right?

nazar-pc:
Any updates here? Would be nice to finally close this issue.

src/offset.js
```diff
@@ -32,7 +32,7 @@ jQuery.offset = {
 			elem.style.position = "relative";
 		}
 
-		curOffset = curElem.offset();
+		curOffset = curElem.offset() || { top: 0, left: 0 };
```
gibson042: This relates to https://github.com/jquery/jquery/pull/2043#discussion_r24018843 ... do we _want_ to suppress exceptions when setting document-relative offset on no-layout elements? I'd like some other opinions here, but personally am inclined to drop this change and let the call throw.

test/unit/offset.js
```diff
@@ -56,13 +56,31 @@ test("object without getBoundingClientRect", function() {
...
+	} catch ( e ) {}
+
+	ok( !result, "no position for disconnected element" );
+});
+
+test("hidden (display: none) element", function() {
+	expect(1);
+
+	var result,
+		node = jQuery("<div style='display: none' />").appendTo(document.body);
```
gibson042: If appending is absolutely necessary, please do so under `#qunit-fixture` so the new element gets cleaned up.

src/offset.js
timmywil: Ditto. +1 for throwing.

markelog: Would need to mention it in the release notes

NekR: > This relates to #2043 (comment) ... do we want to suppress exceptions when setting document-relative offset on no-layout elements? I'd like some other opinions here, but personally am inclined to drop this change and let the call throw.

Do you suggest to just remove ``` || { top: 0, left: 0 }``` or guard for ```curOffset``` and throw manually custom error?

timmywil:
Marked as behavior change since we used to return `0,0` for detached nodes.

src/offset.js
timmywil: Per meeting, remove `|| { top: 0, left: 0 }`.

timmywil: Actually, we'll leave as is and make this change separately. Sorry to confuse!

gibson042:
Also a behavior change for throwing on `$disconnected.position()`: https://github.com/jquery/jquery/pull/2043#discussion_r24018843

src/offset.js
gibson042: Tracking in #2114.

src/offset.js
gibson042: Tracking in #2115.

NekR:
Hmm, CLA test says:
> Unfortunately, there was an error checking the CLA signatures.

> Error: Command failed: fatal: Invalid revision range 4fae91141c04dffb5094b5961eb5d7b2a1f3f423..2e7e1b1c1512dcf5c0f966deab114409d37a975f

What is wrong with it, should I rebase to latest jquery master?

dmethvin:
@scottgonzalez was this one of those spurious failures due to the Github API being too trigger-happy?

scottgonzalez:
Could be. The missing revision was the new one, so it would make sense that it's the result of the PR branch not being updated in time. Unfortunately that means that the retry logic won't fix this.

nazar-pc:
Ping again, what stops this merging this now?
As I see removing `|| { top: 0, left: 0 }` is the only thing here.

NekR:
@nazar-pc

They said:
> Actually, after further discussion, we'll leave as is and make this change separately. Sorry to confuse! Nothing further for you to do here @NekR

So actually nothing should prevent it (at least known).

timmywil:
This is ready to be merged. I was about to, but noticed that I'll need to spend a little time figuring out what is needed to cherry-pick this to the compat branch. I plan on getting to it this weekend.

nazar-pc:
@timmywil, 3 weeks later I want to ping you once more)

timmywil:
@nazar-pc Don't worry. I haven't forgotten. Other things take precedence right now.

nazar-pc:
Seriously, it s too long for already agreed change, almost 3 more weeks.

markelog:
@nazar-pc important thing that is will be released with 3.0, when it will be landed is that not relevant.

src/offset.js
timmywil: Actually, that test is failing with this PR. I'm going to include #2115 when landing.

NekR: @timmywil I am not sure about this, but as I remember that test did not worker even before I made any changes. Anyway, it's what all is resolved now, thanks!

PR #2061, opened 2015-02-04
CSS: Add a support test for the hack for .css('marginRight') etc.

mzgol:
This hack turns out to be needed by Android 4.0-4.3.

Add a support test so that the hack is invoked only where needed.

Refs 3747cc642a48d2a5a8ac83069f66bddd33bea301

Refs gh-1815
Refs gh-1820
Refs gh-1842


mzgol:
This is a `compat` version of PR #1842.

@gibson042 I'm not sure what to do with all these `// Minified:` comments, especially that this whole file will most likely need a refactor to save some size.

I'm submitting a PR instead of just landing because of doubts related to these comments.

mzgol:
@gibson042 This was added by you in https://github.com/jquery/sizzle/commit/36973e9340b9e072a2077125cfc442dc6a42b926; could we test for `document.documentMode` instead of `addEventListener`?

It'd be good to make sure it's fixed in Edge so that we don't have to apply an even more hacky workaround soon...

gibson042:
Just delete the affected "Minified" line(s) and we'll clean up the whole file later. Hopefully this doesn't add too much size, though.

mzgol:
> Hopefully this doesn't add too much size, though.

This used to add 86 bytes, I refactored it a little and now it's +74. Note, though, that it's just the test, the workaround was already present.

On `master` the workaround + support test combo added 127 bytes. Quite a lot.

mzgol:
PR updated. Down to +45 bytes.

src/css/support.js
```diff
@@ -94,9 +98,10 @@ define([
 
 		// Check for getComputedStyle so that this code is not run in IE<9.
 		if ( window.getComputedStyle ) {
-			pixelPositionVal = ( window.getComputedStyle( div, null ) || {} ).top !== "1%";
-			boxSizingReliableVal =
-				( window.getComputedStyle( div, null ) || { width: "4px" } ).width === "4px";
+			divStyle = window.getComputedStyle( div, null );
+			pixelPositionVal = ( divStyle || {} ).top !== "1%";
+			boxSizingReliableVal = ( divStyle || { height: "4px" } ).height === "4px";
```
mzgol: Who does remember why we need these `||`?

cc @jquery/core

gibson042: Hidden I frames in Firefox 4, of course: http://bugs.jquery.com/ticket/8763

I'm thinking it's safe to remove them now. :wink:

PR #2063, opened 2015-02-05
CSS: save 20 bytes in css/support

mzgol:
Refs gh-1842

cc @gibson042

### DIFF
--- a/src/css/curCSS.js
+++ b/src/css/curCSS.js
@@ -25,6 +25,11 @@ function curCSS( elem, name, computed ) {
 			ret = jQuery.style( elem, name );
 		}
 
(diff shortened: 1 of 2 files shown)


mzgol:
Note that while it saves some bytes, I've also moved the `== null` checks out of the support functions into the `computeStyleTests` one. I'm not sure that's desired as it penalizes every support test invocation by an additional function call.

Without this size optimization we're down 18 bytes, not 21.

mzgol:
Updated again. Savings down from 21 to 19 bytes but we're saving function calls.

gibson042:
:+1:

PR #2085, opened 2015-02-11
Manipulation: buildFragment always uses document of the context

piotrj:
buildFragment methods such as createDocumentFragment or createElement.
It called it on context, but the context could have been some random
node, not necesarilly document node. This commit fixes that method so it
always takes the ownerDocument of the context (or the context if it's
document itself) and callse the mentioned methods on that

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -203,7 +203,8 @@ jQuery.extend({
 
 	buildFragment: function( elems, context, scripts, selection ) {
 		var elem, tmp, tag, wrap, contains, j,
-			fragment = context.createDocumentFragment(),
+			doc = context.ownerDocument || context,
+			fragment = doc.createDocumentFragment(),
 			nodes = [],
 			i = 0,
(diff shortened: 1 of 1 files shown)


markelog:
`buildFragment` is called from two places, from `buildFragment` and `parseHTML`.

In first case, `ownerDocument` is already extracted.
In second case, second argument of `parseHTML` is used, which is documented to be a `document` element. Or, if not passed, default `document` is used. 

So `buildFragment` internally, always receives document as an argument. Whereas `buildFragment` should only be used internally, hence lack of documentation for it.

What problem you trying to solve?

piotrj:
@markelog I had a problem with buildFragment as I met a situation when it received a non-document context. Now I see that it was the problem of jquery-migrate library that was fixed here jquery/jquery-migrate@683b9b3ae40e3fe9e83f3dd39d8e5e958975f3fa. I guess I can close this PR.

PR #2087, opened 2015-02-13
Build: update load-grunt-tasks

DelmerGA:
The load-grunt-tasks v1.0.0 should specify a scope and pattern.


timmywil:
Thank you for submitting a PR, but this doesn't seem to be a requirement according to the documentation and the default options work fine for us.

PR #2089, opened 2015-02-13
Core: simplify "each" stylesheet iteration test

markelog:
This assertion always fail on me, in chrome, because of the extension that i enable even in incognito mode.  

This change would make this check generic and eliminate the possible problem in usual mode too.


gibson042:
LGTM

PR #2096, opened 2015-02-16
toggle: renamed argument 'state' to 'doShow'

burka:
To make it clear that passing true shows and passing false hides. Otherwise if you take the documentation with showOrHide at its word, it would do nothing if you pass false and would call the regular toggle() if you pass true. 
I created a pull request for the api.jquery.change as well.
https://github.com/jquery/api.jquery.com/pull/654

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -447,9 +447,9 @@ jQuery.fn.extend({
 	hide: function() {
 		return showHide( this );
 	},
-	toggle: function( state ) {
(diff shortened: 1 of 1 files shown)


timmywil:
Thank you for opening an issue, but it seems like bikeshedding to change the source. The docs issue is the right direction, tho.

burka:
Ok, I just thought it would clear things up a little.

Issue #1731, opened 2014-10-20
Run jQuery unit tests in XHTML


mgol:
Originally reported by dmethvin at: http://bugs.jquery.com/ticket/10198
<p>
We need to run the unit tests with an XHTML doctype to ensure that everything passes.
</p><p>Issue reported for jQuery 1.6.3</p>

anthonyryan1:
With the issue of XHTML regressions coming up again, it seems like an appropriate time to revisit this.

I think the optimal course of action would be:
1. Fixing QUnit for XHTML and ensuring they start regularly testing in an XHTML mode. We can't expect to keep the XHTML tests green if our test harness keeps breaking.
2. Requiring a webserver to serve the test suite so everything can be served with appropriate headers (we can't ensure all of the various `file:///` URLs are served with the correct content-type
3. Conditionally skip based on specific criteria (in this case `if (document.contentType === 'application/xhtml+xml') { /* skip */ }`). This would be a very significant effort as each XHTML failing test would need to be manually checked to understand why it's failing. Whether it's because of how it is being used (for example, appending an unclosed tag), or because of a bug in jQuery itself.
4. Get **all** of the tests passing or skipping under XHTML. Because while people might notice the difference between 0 and 1 failure. The difference between 17 and 18 will lead to people just ignoring these failures and we'll see no fewer regressions.

mzgol:
Rescheduling to 3.0.1 as it's non-trivial and we still have a lot to do before 3.0.0.

timmywil:
Moving to [Roadmap](https://github.com/jquery/jquery/wiki/Roadmap). We're not sure when we'll get to this.

mgol:
@timmywil When you're closing an issue while not fixing it please *always* make sure to remove a milestone, otherwise it will get picked up in a changelog totally incorrectly.

timmywil:
@mgol Good call, thanks for the reminder.

PR #2098, opened 2015-02-17
Make top of the HTML suite compliant with style guide

markelog:
Make top of the html suite compliant with <a href="http://contribute.jquery.org/style-guide/html/">our</a> html code style.

@dmethvin xhtml doctype was chosen for specific reason or it's fine like that?


dmethvin:
> @dmethvin xhtml doctype was chosen for specific reason or it's fine like that?

I think it was that way before I joined! Note htat we have #1731 open for a while to figure out how to avoid our XHTML regressions. If we aren't running the tests in XHTML I think it might be good to make it clearer by switching to the HTML5 doctype like you've done here. Then we would need a different main file for an XHTML test.

markelog:
Yeah, i distinctly remember that issue.

PR #2119, opened 2015-03-05
Data: move element cache to element[expando] (updated)

rwaldron:
- use Object.defineProperty
- remove unnecessary code path

Signed-off-by: Rick Waldron <waldron.rick@gmail.com>

### DIFF
--- a/src/data/Data.js
+++ b/src/data/Data.js
@@ -14,34 +14,23 @@ Data.accepts = jQuery.acceptData;
 Data.prototype = {
 
 	register: function( owner, initial ) {
-		var descriptor = {},
-			value = initial || {};
+		var value = initial || {};
 
-		try {
-			// If it is a node unlikely to be stringify-ed or looped over
-			// use plain assignment
-			if ( owner.nodeType ) {
-				owner[ this.expando ] = value;
+		// If it is a node unlikely to be stringify-ed or looped over
+		// use plain assignment
+		if ( owner.nodeType ) {
+			owner[ this.expando ] = value;
 
-			// Otherwise secure it in a non-enumerable, non-writable property
-			// configurability must be true to allow the property to be
-			// deleted with the delete operator
-			} else {
-				descriptor = {
-					value: value,
-					writable: true,
-					configurable: true
-				};
-				Object.defineProperty( owner, this.expando, descriptor );
-			}
-
-		// Support: Android < 4
-		// Fallback to a less secure definition
-		} catch ( e ) {
-			descriptor[ this.expando ] = value;
-			jQuery.extend( owner, descriptor );
+		// Otherwise secure it in a non-enumerable, non-writable property
+		// configurability must be true to allow the property to be
+		// deleted with the delete operator
+		} else {
+			Object.defineProperty( owner, this.expando, {
+				value: value,
+				writable: true,
+				configurable: true
+			});
 		}
-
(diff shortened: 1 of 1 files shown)


rwaldron:
cc @gibson042

src/data/Data.js
```diff
@@ -27,12 +27,12 @@ Data.prototype = {
...
 			// deleted with the delete operator
 			} else {
-				descriptor[ this.expando ] = {
+				descriptor = {
 					value: value,
 					writable: true,
 					configurable: true
 				};
-				Object.defineProperties( owner, descriptor );
+				Object.defineProperties( owner, this.expando, descriptor );
```
rwaldron: Derp.

src/data/Data.js
```diff
@@ -27,12 +27,12 @@ Data.prototype = {
...
 			// deleted with the delete operator
 			} else {
-				descriptor[ this.expando ] = {
+				descriptor = {
 					value: value,
 					writable: true,
 					configurable: true
 				};
-				Object.defineProperties( owner, descriptor );
+				Object.defineProperty( owner, this.expando, descriptor );
```
gibson042: This whole block can be simpler, and doesn't even need a `try`; see 95fb798980d7e404c413e29e20016db9052e2bf2.

src/data/Data.js
```diff
@@ -128,7 +122,7 @@ Data.prototype = {
 	},
 	remove: function( owner, key ) {
 		var i, name, camel,
-			cache = this.cache( owner );
+			cache = owner[ this.expando ];
```
gibson042: We need an early return when `cache` is nonexistent to avoid errors, right?

src/data/Data.js
jbedard: And do we want the `descriptor` variable when it is only assigned and used once? Currently it is also assigned twice but only used in this one spot...

rwaldron: > This whole block can be simpler, and doesn't even need a try; see 95fb798.

Simplified.

> And do we want the descriptor variable when it is only assigned and used once

Allocation removed.

src/data/Data.js
rwaldron: I didn't observe any such failures, but an early return is a nice optimization that doesn't appear to have any adverse side effects

src/data/Data.js
```diff
@@ -14,34 +14,25 @@ Data.accepts = jQuery.acceptData;
...
-		// Support: Android < 4
-		// Fallback to a less secure definition
-		} catch ( e ) {
-			descriptor[ this.expando ] = value;
-			jQuery.extend( owner, descriptor );
+		// Otherwise secure it in a non-enumerable, non-writable property
+		// configurability must be true to allow the property to be
+		// deleted with the delete operator
+		} else {
+			descriptor = {
```
jbedard: Is there a reason to keep the variable? Why not just
```javascript
Object.defineProperty( owner, this.expando, {
	value: value,
	writable: true,
	configurable: true
});
```
(which I see in a few other places in the code, not sure what the convention is though...)

rwaldron: Left over from when the same object was used in two places, but yes: I've just now removed that and changed to your suggestion.

gibson042:
LGTM :+1:

rwaldron:
@gibson042 Cool, thanks for the feedback/patience/attention ;)

PR #2121, opened 2015-03-05
Test against the current version of node as well as iojs

jonhester:
Currently travis only tests using node 0.10, so this adds 0.12 and iojs to the travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,5 @@ language: node_js
 sudo: false
 node_js:
 - '0.10'
+- '0.12'
+- 'iojs'


jonhester:
I just signed the CLA.

mzgol:
If this PR was submitted half a year ago I'd have said it wasn't necessary
(it's enough to run jsHint in one Node) but we're now smoke testing Node
and we'll soon be testing the Promises/A+ compliance there so it'd be good
to do it in all those versions. So +1 from me.

.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: node_js
 sudo: false
 node_js:
+- '0.12'
```
mzgol: It'd be better to put it after `0.10` to make the whole list chronological.

mzgol:
The commit message should start with a component name, i.e. `Build: Test...` etc.

jonhester:
@mzgol Amended my commit with the component name and changed the order of the node versions

PR #2124, opened 2015-03-07
fixed #2123 trigger Event called native DOM event called twice

### LINKED ISSUE #2123: trigger Event called native DOM event called twice
Native DOM event with on prefix can be called twice when the trigger function is being called.

https://jsfiddle.net/h131rmck/

andycall:
trigger Event called native DOM event called twice

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -316,7 +316,7 @@ jQuery.event = {
 			}
 
 			// Native handler
-			handle = ontype && cur[ ontype ];
+			handle = ontype && cur === elem &&  cur[ ontype ];
 			if ( handle && handle.apply && jQuery.acceptData( cur ) ) {
 				event.result = handle.apply( cur, data );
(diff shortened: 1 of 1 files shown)


src/event.js
```diff
@@ -316,7 +316,7 @@ jQuery.event = {
 			}
 
 			// Native handler
-			handle = ontype && cur[ ontype ];
+			handle = ontype && cur === elem &&  cur[ ontype ];
```
staabm: Nit: 1 space too much

andycall:
I had signed CLA

scottgonzalez:
You signed the CLA, but AndyCall isn't considered a valid name. Can you please sign with your full name?

andycall:
sure, 
This problem is if I bind an addEventListener("click") instead of onclick , this double trigger will be disappear

PR #2125, opened 2015-03-08
Remove old workarounds and simplify use of the native API 

markelog:
Simplifications of the `stopImmediatePropagation`, `stopPropagation` and `preventDefault` look like risky changes, but tests are passing and support in browsers already exist. 

/cc @dmethvin

### DIFF
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -10,9 +10,7 @@ jQuery.ajaxSettings.xhr = function() {
 	} catch ( e ) {}
 };
 
-var xhrId = 0,
-	xhrCallbacks = {},
-	xhrSuccessStatus = {
+var xhrSuccessStatus = {
 		// file protocol always yields status code 0, assume 200
 		0: 200,
 		// Support: IE9
@@ -20,17 +18,6 @@ var xhrId = 0,
 		1223: 204
 	},
 	xhrSupported = jQuery.ajaxSettings.xhr();
-
-// Support: IE9
-// Open requests must be manually aborted on unload (#5280)
-// See https://support.microsoft.com/kb/2856746 for more info
-if ( window.attachEvent ) {
-	window.attachEvent( "onunload", function() {
-		for ( var key in xhrCallbacks ) {
-			xhrCallbacks[ key ]();
-		}
-	});
-}
 
 support.cors = !!xhrSupported && ( "withCredentials" in xhrSupported );
 support.ajax = xhrSupported = !!xhrSupported;
@@ -43,8 +30,7 @@ jQuery.ajaxTransport(function( options ) {
 		return {
 			send: function( headers, complete ) {
 				var i,
-					xhr = options.xhr(),
-					id = ++xhrId;
+					xhr = options.xhr();
 
 				xhr.open(
 					options.type,
@@ -84,7 +70,6 @@ jQuery.ajaxTransport(function( options ) {
 				callback = function( type ) {
 					return function() {
 						if ( callback ) {
-							delete xhrCallbacks[ id ];
 							callback = xhr.onload = xhr.onerror = null;
 
 							if ( type === "abort" ) {
@@ -117,7 +102,7 @@ jQuery.ajaxTransport(function( options ) {
 				xhr.onerror = callback("error");
 
 				// Create the abort callback
-				callback = xhrCallbacks[ id ] = callback("abort");
+				callback = callback("abort");
 
 				try {
 					// Do send the request (this may raise an exception)
--- a/src/event.js
+++ b/src/event.js
@@ -688,7 +688,7 @@ jQuery.Event.prototype = {
 
 		this.isDefaultPrevented = returnTrue;
 
-		if ( e && e.preventDefault ) {
+		if ( e ) {
 			e.preventDefault();
 		}
 	},
@@ -697,7 +697,7 @@ jQuery.Event.prototype = {
 
 		this.isPropagationStopped = returnTrue;
 
-		if ( e && e.stopPropagation ) {
+		if ( e ) {
 			e.stopPropagation();
 		}
 	},
@@ -706,7 +706,7 @@ jQuery.Event.prototype = {
 
 		this.isImmediatePropagationStopped = returnTrue;
 
-		if ( e && e.stopImmediatePropagation ) {
+		if ( e ) {
 			e.stopImmediatePropagation();
 		}
 
--- a/src/event.js
+++ b/src/event.js
@@ -123,7 +123,7 @@ jQuery.event = {
 					special.setup.call( elem, data, namespaces, eventHandle ) === false ) {
 
 					if ( elem.addEventListener ) {
-						elem.addEventListener( type, eventHandle, false );
+						elem.addEventListener( type, eventHandle );
 					}
 				}
 			}
@@ -631,9 +631,10 @@ jQuery.event = {
 };
 
 jQuery.removeEvent = function( elem, type, handle ) {
+
 	// This "if" is needed for plain objects
 	if ( elem.removeEventListener ) {
-		elem.removeEventListener( type, handle, false );
+		elem.removeEventListener( type, handle );
 	}
 };
 
--- a/src/event.js
+++ b/src/event.js
@@ -26,6 +26,8 @@ function returnFalse() {
 	return false;
 }
 
+// Support: IE9
+// See #13393 for more info
 function safeActiveElement() {
 	try {
 		return document.activeElement;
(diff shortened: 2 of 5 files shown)


src/event.js
```diff
@@ -424,8 +426,11 @@ jQuery.event = {
 			delegateCount = handlers.delegateCount,
 			cur = event.target;
 
+		// Support (at least): Chrome, IE9
```
markelog: I wonder how we should formulate comment like this - we know problem exist in Chrome (in some version) and IE9 but it might also exist (and probably does like in Safari) in other browsers?

scottgonzalez: Well, the support comment as is already makes it confusing as to which parts of the conditional apply to which bug. I'd suggest making that clear and then testing the specific workarounds in the supported browsers to get accurate support comments.

markelog: Yes, that would be the best policy, but to get the accurate comments you would have to do a lot of tedious work without any glory.

dmethvin: The ticket reporting the original bug at http://bugs.jquery.com/ticket/13180 has a test case http://jsbin.com/welcome/71517/ that still blows up for IE11 but seems okay on Chrome and Firefox. For master at least, a `// Support: IE9+` would be correct?

src/event.js
```diff
@@ -424,8 +426,11 @@ jQuery.event = {
...
 		// Find delegate handlers
 		// Black-hole SVG <use> instance trees (#13180)
+		//
+		// Support: Firefox
```
dmethvin: This was from http://bugs.jquery.com/ticket/3861 and still around when I rewrote things in 1.8, but seems to have disappeared at some point: http://jsbin.com/qoqofa/1/edit?js,console,output

dmethvin: The Firefox ticket https://bugzilla.mozilla.org/show_bug.cgi?id=184051 for this is still open, so either my test case sucks or this is a dup of a ticket that was fixed.

src/event.js
gibson042: ` Support: IE 9 - 11` - https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-69379197

I'm OK with ditching the Chrome reference, since at this point it's just a historical curiosity.

src/event.js
markelog: @bzbarsky is it possible to prioritize this ticket?

src/event.js
markelog: Cool, will update the comment, but it doesn't answer my question, those situations are happen sometimes, most recently in https://github.com/jquery/jquery/commit/a5e1c9b44c971fd7046d9a95bd0810e50840b663.

In that case, we know problem exist in Chrome 34, but it might also be present in 33, but there is not actual reason to check them all, but we can very well mislead the person who would read such comment, perhaps we need "at least" notation for these kind of stoppers?

dmethvin: These cases predate our efforts to be very specific about browser bugs in `// Supports` comments, but they reinforce that being specific can save us a lot of work down the road.

I agree it is difficult and pointless to go back and see whether Chrome 18, for instance, has this bug. We're not planning to take any of this out for a while, so a labor-intensive investigation to get an accurate version may not be worth the effort until that point.

Still, there is a [growing problem with Chrome fragmentation](http://www.quirksmode.org/blog/archives/2015/02/chrome_continue.html) and big phone vendors like Samsung are stuck on Chrome 28. In the general case we can't just test on desktop Chrome 28 but infer that the problem also exists on Samsung's Chrome 28.

src/event.js
bzbarsky: Check with Olli (the bug assignee)?  He's the owner for events stuff.

src/event.js
gibson042: > we know problem exist in Chrome 34, but it might also be present in 33

That case is covered as well, assuming a fix immediately after version M: `Support: Chrome <=34 - M only`. If we don't know the fix version because that is _also_ too far back in the past, then I'd support either fudging it or extending the grammar to accept `Support: Chrome <=34 - <39 only`.

scottgonzalez: > f we don't know the fix version because that is also too far back in the past, then I'd support either fudging it or extending the grammar to accept Support: Chrome <=34 - <39 only.

I think if the fix is too far back in the past, the support comment would have been dropped already.

gibson042: Yes, that was my [original suggestion](https://github.com/jquery/jquery/pull/2125/files#r26094668).

src/ajax/xhr.js
```diff
@@ -21,17 +21,6 @@ var xhrId = 0,
 	},
 	xhrSupported = jQuery.ajaxSettings.xhr();
 
-// Support: IE9
```
gibson042: I'm for it. And if you look through the history of this block, you might also find some other code that only exists to support it and can now be removed as well.

markelog: Nice catch, updated

src/event.js
```diff
@@ -677,30 +683,24 @@ jQuery.Event.prototype = {
 	isImmediatePropagationStopped: returnFalse,
 
 	preventDefault: function() {
-		var e = this.originalEvent;
-
 		this.isDefaultPrevented = returnTrue;
 
-		if ( e && e.preventDefault ) {
-			e.preventDefault();
+		if ( this.originalEvent ) {
```
dmethvin: This loses the guard for `preventDefault` and `stopPropagation` existing before calling them. It makes me nervous but I cannot think of situations where that would be a problem. For example, if these are nested `jQuery.Event` instances (*really* common in jQuery UI) it should work fine.

src/ajax/xhr.js
mzgol: +1 to this change

src/event.js
mzgol: A version is needed here

dmethvin: Once we get the response from Firefox devs we can put one in. AFAICT it is fixed but their bug is still open so my repro may be bad.

mzgol: @dmethvin OK, good to know.

src/event.js
scottgonzalez: This will break for uses like: `elem.trigger( $.Event( { type: "custom" } ) );`. In this case, the original event will just be `{ type: "custom" }`, so it won't contain a `preventDefault()` method.

markelog: Yeah, i have same feelings, i thought maybe you could think of the case :-). Even mutation and history events works here too, but i guess you never know.

markelog: Okay, reverting then, thank you Scott

dmethvin: True but it would only break if someone returned `false` from their custom event handler or explicitly called `preventDefault` or `stopPropagation` on that custom event. Given the amount of code out there I could see it happening though.

markelog: Also it would be more consistent with native events, event in those do the nop when `preventDefault` is not possible, not error

src/event.js
```diff
@@ -626,8 +631,9 @@ jQuery.event = {
 };
 
 jQuery.removeEvent = function( elem, type, handle ) {
+	// This "if" is needed for plain objects
```
scottgonzalez: Blank line above the comment.

markelog: So <a href="https://contribute.jquery.org/style-guide/js/#comments">this</a> - 
```js
function foo( types, selector, data, fn, /* INTERNAL */ one ) {
    // Do stuff
}
```
Is a mistake in guide?

scottgonzalez: Yup, I'll fix that now. Read the first sentence in the Comments section.

mzgol: Is this a difference between a comment about the whole function and a comment targeting a specific instruction inside the function?

scottgonzalez: No, there are no differences.

markelog: > Read the first sentence in the Comments section.

Yeah, i have read it, that is why i asked :-), fix comming

markelog:
Pushed another commit that adds couple tests for the use-case mentioned by the @scottgonzalez

markelog:
> I was fine with leaving the guard in since there may be code out there that would do this (or some other case that would break) but if we're going to have a unit test for it I really think we should document it fully as well. If everyone is okay with that I can open a docs ticket.

@dmethvin so what are we want to do here?

dmethvin:
My preference would be to not land b1dbe6fea8abb3cf68f5652a5840017d1af4646b since `jQuery.Event({ type: "custom" });` is not supported. The docs say that already. The first arg should be either a native event object or a `jQuery.Event` and additional  properties go in the second arg.

scottgonzalez:
Yeah, sorry for throwing that in there and getting us off track for a bit :-/

markelog:
@dmethvin okay, so without support for it, sounds good to me, will bring back simplified guards then
@scottgonzalez i'd say it always good to rehash these kind of stuff

PR #2126, opened 2015-03-08
Ajax:Manipulation: preliminary implementation of async manip script load

markelog:
Also, remove undocumented "throws" property from ajax module

Ref #1895

WIP, needs more testing and it could be done better.


gibson042:
Is it wise to execute inline scripts synchronously and remote scripts asynchronously like this? For all the faults of `async: false`, it did maintain a behavior consistency that is lacking here (cf. http://bugs.jquery.com/ticket/11795 and discussion and e889134058232c5e19156353c5fc3bf3b4915a94, particularly the [inline-script tests](https://github.com/jquery/jquery/commit/e889134058232c5e19156353c5fc3bf3b4915a94#diff-328a9410239da61336e3662fdbfe9bf6R2114)).

Also, I note that the test changes are all relaxations... do we want one that strictly checks non-synchronicity? In other words, are we reversing our promises about when code will execute, or merely dropping them?

dmethvin:
I don't mind the sync/async inconsistency between inline and remote scripts. When the browser loads a page it runs inline scripts sync and external ones async, so we're aligning with native practice there at least.

For the unit tests we control the conditions so we can ensure that the code executes eventually. For the general case we've never been able to guarantee freedom from 404 or timeouts so dropping the assurances about when it happens is okay with me.

I did some perf work for a startup about a year ago that was doing its own template processing by pulling in an HTML fragment that included scripts via `<script>` tags and expected them to be sync. That was causing a horrible lag in load time, and just putting the script tag into the head helped a lot since it could be prefetched much earlier. Bad code will break with this change but I don't think that's a bad thing.

gibson042:
> When the browser loads a page it runs inline scripts sync and external ones async, so we're aligning with native practice there at least.

Huh? https://html.spec.whatwg.org/multipage/scripting.html#attr-script-async

markelog:
> Huh? https://html.spec.whatwg.org/multipage/scripting.html#attr-script-async

Since we're talking about dynamically created scripts `defer` and `async` attributes has no effect if script is inline. Inline scripts always executed synchronously.

>  The defer and async attributes must not be specified if the src attribute is not present.

and

> then the script is fetched and executed immediately, before the user agent continues parsing the page.

But that is beside the point, i used words "WIP" and "preliminary" for the reason, because i want your input but "Is it wise" not the phrase i was looking for :-). At the moment, i able to make existing tests to pass, but existing implementation could be different (not sure here), very much so, actually.

dmethvin:
Duh, yes. I don't know what I was thinking. The main issue is that jQuery and native currently disagree on what happens when you inject a `<script src>` tag into the page. That turns out to be a very large footgun for the average programmer.

gibson042:
There are three relevant characteristics of native browser script execution in the absence of `defer` and `async` attributes. Ordered from most valuable to least:

1. Strict ordering, so code like `<script src="library.js"></script><script src="use-library.js"></script><script>useLibraryAgain()</script>` is deterministic.
2. A known point when execution of any given script is finished (e.g., at any later position in the DOM). This is strictly stronger than the preceding, which would still hold if scripts were executed in order but asynchronously and without a "done" event/callback/etc.
3. A partial view of the DOM—nodes following the script are absent at execution time

Current jQuery behavior preserves the first two properties, and I don't think I'm willing to give up number 1. We can either chain asynchronous execution of all scripts (both inline and remote) with something like `Deferred`, or stop interpreting them altogether, or leave the current behavior unmodified, but we should not give up browser-like sequencing.

dmethvin:
Ugh, I wasn't thinking about the hard cases enough.

I can see your point on 1, and we could ensure that with a Promise but if we use `<script>` tags it seems like we'd need to actually wait for the first script to fetch and run its `onload` before we could even request the next one?

Really the whole mess with arbitrary script execution in HTML is a mess I'd love to see gone due to its potential for XSS. Normal `.innerHTML` doesn't run scripts, so this is a place where jQuery tried to make things easier with negative consequences.

Is there some way to make script injection like this still possible but more explicit? It seems like if blocking sync execution is desired, then we can't get rid of any of the performance or security problems at all and might as well leave things as-is.

gibson042:
Well, we never actually _use_ script tags... all the code is sent to `jQuery.globalEval`, with an intermediate `jQuery.ajax` via `jQuery._evalUrl` in the case of remote source. So it would be possible to establish strict ordering by revising the [relevant block of `domManip`](https://github.com/markelog/jquery/blob/async/src/manipulation.js#L534-L550) to something like
```js
chain = autoInvoke = {
	then: function( callback ) { return callback(); }
};
…
// Evaluate executable scripts on first document insertion
jQuery.each( scripts, function( _, node ) {
	if ( rscriptType.test( node.type || "" ) &&
		!dataPriv.access( node, "globalEval" ) &&
		jQuery.contains( doc, node ) ) {

		if ( node.src ) {
			if ( jQuery._evalUrl ) {
				chain = chain.then(function() {
					return jQuery._evalUrl( node.src );
				}, jQuery.error );
			}
		} else {
			chain = chain.then(function() {
				jQuery.globalEval( node.textContent.replace( rcleanScript, "" ) );
				return autoInvoke;
			}, jQuery.error );
		}
	}
});
```

dmethvin:
What I meant was that if we switched to script tags we'd need to do something trickier to guarantee script execution. The problem is that we need sync XHR inside `jQuery._evalUrl`  to fetch external scripts today with block-the-thread behavior (well, for *same-domain* script requests at least) and it would be really nice to get rid of that.

gibson042:
The above block (minus its completely redundant `jQuery.error`s) accomplishes that, maintaining ordering but keeping the `_evalUrl` change to asynchronous XHR.

markelog:
I guess there is a lot of confusion in this discussion, we're talking about *dynamically* inserted scripts, this change now follows native browser behaviour.

<blockquote> but we should not give up browser-like sequencing.</blockquote>It's not browser-like, current behaviour is the opposite of that.

For example, if you want to follow "browser-like" how would you interpret `defer` or `async` attributes?

This conversation, for some reason, now mixes two ways of script loading, there is static script inclusion and dynamic one, only one possible jQuery use-case of script handling is the dynamic one.

Why then we're talking about the static one?

gibson042:
> I guess there is a lot of confusion in this discussion, we're talking about dynamically inserted scripts, this change now follows native browser behaviour.

Precisely following native browser behavior would mean ignoring dynamically-added scripts completely, exempting them from execution. Assuming we won't be going that far, I'm concerned about conserving those _properties_ of native (load-time) behavior identified in https://github.com/jquery/jquery/pull/2126#issuecomment-78273289 .

> > but we should not give up browser-like sequencing.

> It's not browser-like, current behaviour is the opposite of that.

Browser-like in the sense of maintaining property 1 (pertaining to **sequencing**).

> For example, if you want to follow "browser-like" how would you interpret defer or async attributes?

By ignoring them, both because that's what we do now and and because they seem specific to load time: https://html.spec.whatwg.org/multipage/scripting.html#attr-script-async

dmethvin:
I apologize for making it even more confused! We are definitely talking about the dynamic injection of HTML fragments here, and the difference between what happens today when you use jQuery versus what will happen in the future when you use jQuery. 

Let's see if I can give a full description of the problem and get this right. Today if you run `$("#widget").load("/widgets/tooltip.html")` you can do this and it will work since the scripts are local and fetched with sync XHR. jQuery will *block the current thread* and serialize execution of all three scripts:
```html
<div class="tooltip" title="This is a tooltip">test</div>
<script src="tooltip.js"></script>
<script src="tooltip-extras.js"></script>
<script> $(".tooltip").tooltip();</script>
```
This is a pretty common jQuery pattern for injecting a self-contained template into a page, I've seen it several times in sites that I've worked on. (Although I always make

Since native browser HTML string injection doesn't run scripts, the same approach with `.innerHTML` would not run there; we don't have that behavior to compare ourselves to. Of course we always convert everything to DOM before injecting, which makes a *big* difference in semantics since it runs scripts even in the browser, but users aren't always aware of that. 

@gibson042's solution maintains the serialized nature of the three scripts (point 1 above) but doesn't make them blocking, so there could potentially be other code later in the thread before it yields that would die because the injected scripts had not yet run. Or, even later in the page, there could be timing issues where code expects that code to have been loaded. Those are the errors that could arise from giving up on points 2 and 3.

If backcompat wasn't an issue it seems like a new API that returned a Promise from loading an HTML fragment that included scripts would be a nice thing to have here. If the migration hassle to some new API was low, is there any chance we could fix the use case and eliminate running embedded scripts in general HTML injection? Or are we stuck with that?

gibson042:
> If backcompat wasn't an issue it seems like a new API that returned a Promise from loading an HTML fragment that included scripts would be a nice thing to have here.

I have thought about this as well, and the chief difficulty is lack of a good place to put the promise. The idea bouncing around in my head lately is instance-specific properties (i.e., independent of the constituent elements). Use would look something like
```js
// Can't use the return value because domManip and derivatives are chainable
$el.append( html ).removeClass( "empty" );

// ...but we CAN extend the collection object!
$el.scriptsDone.then( callback );
```

But this is both a sizable departure from our current surface area and strictly less important than point 1, which often allows you to manually (albeit _very_ messily) recover point 2:
```js
// Globally store the jQuery collection and the "scripts done" callback
var unique = getUniqueIdentifier();
globalObj.pending[ unique ] = { collection: $collection, afterScriptExecution: fn };

// Define code to invoke the callback after all scripts are loaded
var fireCallback = "(function( unique ) {\
	var globalObj = this,\
		$collection = globalObj.pending[ unique ].collection,\
		callback = globalObj.pending[ unique ].afterScriptExecution,\
		$sentinels = $collection.find( 'script.' + unique );\
\
	/* Remove a sentinel. If it wasn't the only one, terminate. */\
	$sentinels.eq( 0 ).remove();\
	if ( $sentinels.length > 1 ) { return; }\
\
	/* Run the callback and clean up */\
	delete globalObj.pending[ unique ];\
	callback( $collection );\
})('" + unique + "')";

// Insert our sentinel(s) _after_ the desired content
$collection.append(
	[].concat( input, "<script class='" + unique + "'>" + fireCallback + "</script>" )
);
```

> If the migration hassle to some new API was low, is there any chance we could fix the use case and eliminate running embedded scripts in general HTML injection?

It would be easy to add a `domManip` argument that controlled script execution, the challenge would be in bubbling it up to the outer manipulation methods. I suppose we could do things like `.safeAppend`/`.safeInsertBefore`/etc., but those don't look great to me.

timmywil:
This PR needs some love before 3.0 final.

markelog:
Totally, will resume work on it, will try this weekend

markelog:
Will bring this up in the next meeting

mgol:
@markelog This wasn't fixed so the milestone needs to be removed.

markelog:
Yeah and `wontfix` is not applicable for pulls

markelog:
Also i don't think that removal of milestone is requirement for pulls, there was a lot of work put in, by all involved, as you might notice by this thread, so not including this logic is still part of 3.0, in some sense.

And i think you overdoing this with all that bureaucracy.

mgol:
All issues/PRs that have a milestone applied end up in the generated changelog.

markelog:
Issues and commits are included in the changelog, not prs

mgol:
Oh, you're right! I thought PRs with milestones are treated like an issue+PR combined. So I guess it doesn't matter which PRs get a milestone and which don't.

PR #2128, opened 2015-03-08
Cleanup of doc (into docblock) in /data/accepts.js

NikV:
I think to better keep up with doc standard, we should turn the long comment doc into a better docblock.


timmywil:
Thank for submitting a PR, but we only use multi-line comments for file and function headers (http://contribute.jquery.org/style-guide/js/#comments).

Issue #2118, opened 2015-03-05
Return null from .attr when attributes don't exist


gibson042:
`.attr("nonexistent")` currently returns `undefined`, [probably](https://github.com/jquery/api.jquery.com/issues/523#issuecomment-49600962) from the days before `.prop`. But the native DOM `getAttribute` and `Sizzle.attr` both return `null` in such cases. Also, `.attr( name, null )` _removes_ attributes, and it would be very convenient to allow [universal round-tripping](https://github.com/jquery/api.jquery.com/issues/523#issuecomment-49554408).

Let's consider changing the return value to accommodate.

Note that _setter_ treatment of `undefined` is explicitly out-of-scope for this ticket, as are `.prop` and `.data` round-tripping, although those are certainly related topics.

timmywil:
@dmethvin We're leaning towards doing this for beta. Do you know of a use case that would make this catastrophic?

dmethvin:
It seems like returning `null` is a good idea, although who knows what it will break in the wild. The only way to find out is to change it! We should continue to return `undefined` for empty jQuery collections tho.

gibson042:
> We should continue to return undefined for empty jQuery collections tho.

Thanks; that reminded me to create #2134.

scottgonzalez:
> It seems like returning null is a good idea, although who knows what it will break in the wild.

Of course the answer turns out to be jQuery UI :-)

I'm working on updates right now.

timmywil:
The team decided to revert this due to breakages in UI and mobile.

dmethvin:
We backed this out for 1.12/2.2 which definitely seems right, but it's also excluded from 3.0. Should it be? That is, do we want a non-existent attribute to return `null` as proposed here? If not I want to remove the milestone and tag.

mgol:
I removed the milestone due to the revert and opened the issue & added the "Needs review" label to account for the Dave's question.

markelog:
> We backed this out for 1.12/2.2 which definitely seems right, but it's also excluded from 3.0. Should it be? 

I remember this discussion, we reverted cause it might be dangerous to do which proved the UI breakage, where is literally no gain from, except to align with standard, whereas behaviour in DOM spec is very questionable in the first place, since in all others API you get `undefined` value not `null` value

PR #2129, opened 2015-03-09
When no attribute exists return null instead of undefined

winhowes:
From issue #2118 returning `null` for non-existant attribute instead of `undefined` keeps in line with browser behavior for `getAttribute()`.

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -67,7 +67,7 @@ jQuery.extend({
 
 			// Non-existent attributes return null, we normalize to undefined
 			return ret == null ?
-				undefined :
+				null :
 				ret;
 		}
 	},
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -63,12 +63,7 @@ jQuery.extend({
 			return ret;
 
 		} else {
-			ret = jQuery.find.attr( elem, name );
-
-			// Non-existent attributes return null, we normalize to null
-			return ret == null ?
-				null :
-				ret;
+			return jQuery.find.attr( elem, name );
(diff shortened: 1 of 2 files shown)


winhowes:
Signed the CLA

winhowes:
Signed the CLA

gibson042:
We should be discussing #2118 at the next meeting, hopefully reaching a consensus on whether or not to accept it. In the meantime, this PR will need unit tests, and I also observe that `ret` is no longer necessary in that block... you can just directly return the value from `jQuery.find.attr`.

winhowes:
The reason I kept `ret` was in case some browser returned `undefined` instead of `null`. This would normalize those cases to `null` or is that not an issue?

timmywil:
It's not an issue. `jQuery.find.attr` will return null.

timmywil:
@winhowes Sorry for the stringency, but the CLA name and email must match the name and email in the commit. Could you sign again as "Winston" <winstonhowes@gmail.com>?

mzgol:
> @winhowes Sorry for the stringency, but the CLA name and email must match the name and email in the commit. Could you sign again as "Winston" winstonhowes@gmail.com?

The CLA should be signed with a full name. Instead of signing the CLA with just the first name, the OP should modify the commit author so that the full name is used. We can do it in this case manually, but please @winhowes change your Git settings for the sake of the future.

timmywil:
Ah, wasn't sure about that rule. Ok, in that case, I'll merge with the full name.

winhowes:
git settings changed, thanks for letting me know

Issue #2127, opened 2015-03-08
Data: new data-as-expando implementation causing olddisplay test failure


rwaldron:
cc @markelog 

https://github.com/jquery/jquery/commit/d702b7637a61e1973e08c27b8d8de2ed24a543e2#commitcomment-10087657

PR #2132, opened 2015-03-09
Data: restore explicit data removal of private data in cleanData.

rwaldron:
Fixes gh-2127

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -305,7 +305,7 @@ jQuery.extend({
 						}
 					}
 				}
-				delete data.events;
+				delete elem[ dataPriv.expando ];
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/manipulation.js
```diff
@@ -306,6 +306,7 @@ jQuery.extend({
 					}
 				}
 				delete data.events;
+				delete elem[dataPriv.expando];
```
gibson042: Missing spaces around the property. Also, can we drop the preceding `delete data.events`?

rwaldron: > Missing spaces around the property. 

I see that the circle of life is complete ;)

src/manipulation.js
```diff
@@ -305,7 +305,7 @@ jQuery.extend({
 						}
 					}
 				}
-				delete data.events;
+				delete elem[ dataPriv.expando ];
```
rwaldron: cc @gibson042

src/manipulation.js
gibson042: You said it. :bow:

PR #2141, opened 2015-03-12
iframe remove after callback

liyangready:
some  outer variable in callback function can't access if remove the 'iframe' before callback function execution in ie6/7 sometimes


mzgol:
Thanks but we don't support IE6/7 anymore. Also, this PR doesn't introduce any test confirming something was fixed and doesn't reference any issue.

Issue #2147, opened 2015-03-16
IE10+ fail the test for modifying offset values in a callback


mzgol:
See http://swarm.jquery.org/result/63457

PR #2148, opened 2015-03-16
Offset: Round offset value for the sake of floating errors

mzgol:
IE10+ may return not exactly the offset.top value set in an offset callback
if parent has fractional top offset itself. Checking for being close to the
desired result fixes the test error.

Fixes gh-2147

@markelog why is this test in `unit/dimensions.js`, not `unit/offset.js`?

BTW, this behavior may be surprising to a lot of people; maybe we should mention that in the API docs that setting an offset & getting it later may result in a slightly different value.

### DIFF
--- a/test/unit/dimensions.js
+++ b/test/unit/dimensions.js
@@ -453,7 +453,8 @@ testIframe( "dimensions/documentLarge", "window vs. large document", function( j
 });
 
 test( "allow modification of coordinates argument (gh-1848)", 1, function() {
-	var element = jQuery( "<div/>" ).appendTo( "#qunit-fixture" );
+	var offsetTop,
+		element = jQuery( "<div/>" ).appendTo( "#qunit-fixture" );
 
 	element.offset(function( index, coords ) {
 		coords.top = 100;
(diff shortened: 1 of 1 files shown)


test/unit/dimensions.js
```diff
@@ -461,7 +462,9 @@ test( "allow modification of coordinates argument (gh-1848)", 1, function() {
 		return coords;
 	});
 
-	equal( element.offset().top, 100, "coordinates are modified" );
+	offsetTop = element.offset().top;
+	ok( Math.abs(offsetTop - 100) < 0.02,
```
dmethvin: How did you decide on `0.02`? :smile:

mzgol: In IE11 it was `~0.011` so I just rounded. :P I'll happily take a more substantiated number.

dmethvin: I have no idea of a better number, so we might as well go with yours. It seems like this may be an issue we will see more often if we let more values be non-integer ones.

mzgol:
Landed at 62ae2d0fb7ac011bf2ad778f8158de408e785927 (I wrote what bug it fixes but forgot to write what PR...)

Issue #2153, opened 2015-03-20
Update jsdom & drop obsolete workarounds


mzgol:
We currently test on jsdom `1.5.x`. The latest version supporting Node.js is `3.1.2`, we should test on that. For example, in jsdom `3.1.2` a document created via `document.implementation.createHTMLDocument( "" )` has a body.

PR #2154, opened 2015-03-20
Core: Update tested jsdom, drop obsolete workarounds

mzgol:
The latest version supporting Node.js is 3.1.2; some workarounds are not needed
for this version. For example, in jsdom 3.1.2 a document created via
document.implementation.createHTMLDocument( "" ) has a body.

Fixes gh-2153

### DIFF
--- a/src/css/support.js
+++ b/src/css/support.js
@@ -46,61 +46,57 @@ define([
 		documentElement.removeChild( container );
 	}
 
-	// Support: node.js jsdom
-	// Don't assume that getComputedStyle is a property of the global object
-	if ( window.getComputedStyle ) {
-		jQuery.extend( support, {
-			pixelPosition: function() {
-				// This test is executed only once but we still do memoizing
-				// since we can use the boxSizingReliable pre-computing.
-				// No need to check if the test was already performed, though.
+	jQuery.extend( support, {
+		pixelPosition: function() {
+			// This test is executed only once but we still do memoizing
+			// since we can use the boxSizingReliable pre-computing.
+			// No need to check if the test was already performed, though.
+			computeStyleTests();
+			return pixelPositionVal;
+		},
+		boxSizingReliable: function() {
+			if ( boxSizingReliableVal == null ) {
 				computeStyleTests();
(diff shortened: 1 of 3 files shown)


src/css/support.js
```diff
@@ -46,61 +46,57 @@ define([
 		documentElement.removeChild( container );
 	}
 
-	// Support: node.js jsdom
-	// Don't assume that getComputedStyle is a property of the global object
```
mzgol: This comment didn't make sense even for jsdom 1.5. Were there any other reasons to hide these support tests? Are we worried about other envs that will expose `window` without `getComputedStyle`? If so, I'll revert but the comment will need to be changed completely.

gibson042: I'm not sure if it comes into play here, but `window` may be a non-Window global object: https://github.com/mzgol/jquery/blob/jsdom-update/src/intro.js#L38

In any case, the comment probably does need rewriting.

dmethvin: It used to say getComputedStyle assuming that was in the global scope. So it was more about assuming window and global were equivalent.

mzgol: Yup, `window` is not a global here but what was passed to the factory. Unless the global has a property `document` when we assume we're in a browser and assign `window = current global`.

This works without this `if` in jsdom because its `window` parameter has `getComputedStyle`. Do we want to support anything else here?

mzgol: I'd like to get input from more people here as I'm a little afraid of this change. cc @jquery/core

timmywil: As long as getComputedStyle is available, I'm fine with this.

mzgol: > As long as getComputedStyle is available, I'm fine with this.

Great. I'd like to have as many people test it as possible since I'm still worried a little so I'd like to get it into the first beta.

dmethvin: This should be safe, see 3722aef8be5d31a7dbf3de6998cdf249081f028c , dc3f7d3868e00fb318069b6f87dfc1fc4d10e56f and http://bugs.jquery.com/ticket/12235 for background.

PR #2158, opened 2015-03-23
Update node

supergray:


mzgol:
We should run tests in Node 0.10/0.12 & io.js, not only in one of those versions. This is especially important now when we smoke-test on Node.js & we run the Promises/A+ tests there.

mzgol:
BTW, this is a dupe of #2121

PR #2159, opened 2015-03-24
Build: Update native-promise-only

mzgol:
This should fix Android 4.0 jQuery.when tests issues.

@timmywil should we have issues for such PRs as well?


timmywil:
I don't think it's necessary.

mzgol:
If no one opposes, I'll merge this today or tommorow. The change is pretty trivial.

Issue #1817, opened 2014-10-29
Ignore source in bower


Martinspire:
When i'm retrieving jQuery with Bower i not only get the `/dist` folder, but also the `/src` folder. I'm not sure if that should be the case but i believe that source shouldn't be included in the bower installation. It adds 104 files to my project which i don't need. Its not very large (367kb at the moment) but unnecessary nonetheless.

PR #2161, opened 2015-03-24
npm ignore src directory

dgbeck:
The src directory is full of AMD modules that are clearly not intended to be used in common js environments. Including it in the npm install doubles the size of the install from aproximately 1MB to 2MB. These files should be npm ignored and the world will have a little less waste.

(On this same token, the `dist/cdn` directory also appears to be unnecessary in the npm install, and again effectively doubles the npm module's footprint, but I probably best to take this one step at a time.)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -7,6 +7,7 @@
 /.mailmap
 /.travis.yml
 
+/src
 /build
 /speed
 /test


dmethvin:
What about npm users who intend to use jQuery in an AMD environment? There is no CJS-format requirement for npm-published files. AMD can be converted to CJS using nodefy if you need it. The more common use case for jQuery by far is AMD since that's more popular on the client side.

On the existence of `dist/cdn` the problem is that the map file contains references to the names of the source and minified files. If we don't publish it by that name the person installing jQuery would need to somehow process internal file references in the map file or else it would be invalid.

dgbeck:
> What about npm users who intend to use jQuery in an AMD environment?

They just use `dist/jquery.js` or `dist/jquery.min.js` like the rest of npm users.

(For clarity let's leave exploration of `dist/cdn` issue for another time.)

mzgol:
This has been discussed at length. See e.g. #1526. We consider npm to be a JS package manager, not just Node.js/io.js package manager and [the npm team agrees](http://blog.npmjs.org/post/101775448305/npm-and-front-end-packaging). Thus, we need to publish the `src/` directory for people relying on our AMD modules. Telling those people to just use the full file is not an option. On the other side, if someone doesn't need those modules they may just ignore them.

The `dist/cdn` directory shouldn't be included, though. We can migrate this bug to say about `dist/cdn` or close it & open a new one.

mzgol:
OK, I opened #2164 for the `dist/cdn` issue. This one can be closed.

dgbeck:
Hi @mzgol, thanks for clarifying and for opening #2164. Since jquery is so popular and widely used, just a few hundred KB of saved space has a big impact.

On that note, returning to `src/` for a minute, after having reviewed all the other related issues like #2164, #1567, #1817, etc, I'm still not clear. What are the use case(s) for relying on the individual AMD modules instead of the full file? [This comment](https://github.com/jquery/jquery/issues/1817#issuecomment-60939518) from @dmethvin contains the only reference to a use case I could find, which is

> Some people want the AMD modules so that they can do a custom build or use requireJS

But people doing a custom build or tinkering with jquery internals will generally want / need to checkout the entire repo anyway, and then use that repo as the npm module. `npm link` is intended to be used for this purpose - to alias an npm module in development to a full repo.

I'm trying to think of a real life use case that is not naturally covered by `npm link` but coming up blank. If there is none, then it seems that there is an opportunity to cut down on waste, at least in npm land.

timmywil:
TJ Vantoll wrote a good article on uses of amd modules in jQuery. http://developer.telerik.com/featured/jquery-using-only-what-you-need/

mzgol:
> Since jquery is so popular and widely used, just a few hundred KB of saved space has a big impact.

Sure, but:
1. It affects only developers, not final users.
2. Many other packages, e.g. ones used by `npm` even publish their tests which we don't do. We're far from the most verbose packages and the whole npm package management encourages installation of many different versions of the same package if only one dependency needs an older/newer package version than another dependency.
3. Maintaining multiple packages would add us more work, make us more vulnerable to publishing mistakes etc. We already spend a lot of time on some releases and mistakes do happen (see http://blog.jquery.com/2014/12/18/jquery-1-11-2-and-2-1-3-released-safari-fail-safe-edition/, esp. the paragraph about a broken 2.1.2 release); we publish to npm, Bower, produce copies for CDNs (our and Google's/MS's/etc.). I don't think that the fact that many devs will download more files than they need to justifies making our release system even more complex, especially that our potential mistakes in this complex problem would affect way more people than use our `npm` package.

> But people doing a custom build (...) will generally want / need to checkout the entire repo anyway, and then use that repo as the npm module.

Why? They don't have to reuse our build system, everything is declared in AMD modules so it's consumable by every AMD build system as well. Using a repository as an `npm` module is not its main intended use case. In the future we'll export ES6 modules which have a chance of being natively understood by io.js (see e.g. https://github.com/iojs/NG/issues/9). We're slowly moving into that direction.

dgbeck:
I see. It's the people who are consuming individual pieces of jquery in production (via AMD, at the moment) who would need to jump through a hoop. At this point in time the cross section of these people and npm users has got to be extremely small, but I do see the dilemma. Thanks.

PR #2166, opened 2015-03-25
Data: Drop the tests relying on applets

mzgol:
BrowserStack regularly has outdated Java in their IE instances causing our
data tests to hang & give no output. Skip the tests that create applets,
applets are treated in the same way as an embed elements which we're already
testing.

Fixes gh-1938

(note to myself: if accepted, remember to cherry-pick this to `2.1-stable` & `1.11-stable` together with #2028)

Original description:

BrowserStack regularly has outdated Java in their IE instances causing our
data tests to hang & give no output. Skip the assertion if the test is run
via TestSwarm.

### DIFF
--- a/test/unit/data.js
+++ b/test/unit/data.js
@@ -159,10 +159,6 @@ test("jQuery.data(<embed>)", 25, function() {
 	dataTests( document.createElement("embed") );
 });
 
-test("jQuery.data(<applet>)", 25, function() {
-	dataTests( document.createElement("applet") );
-});
-
 test("jQuery.data(object/flash)", 25, function() {
 	var flash = document.createElement("object");
 	flash.setAttribute( "classid", "clsid:D27CDB6E-AE6D-11cf-96B8-444553540000" );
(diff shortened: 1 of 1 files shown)


test/unit/data.js
```diff
@@ -773,7 +773,12 @@ test("jQuery.acceptData", 11, function() {
 	ok( jQuery.acceptData( document.documentElement ), "documentElement" );
 	ok( jQuery.acceptData( {} ), "object" );
 	ok( jQuery.acceptData( document.createElement( "embed" ) ), "embed" );
-	ok( jQuery.acceptData( document.createElement( "applet" ) ), "applet" );
+
+	// BrowserStack regularly has outdated Java causing IE to hang on this line.
+	// Skip it on TestSwarm.
+	if ( location.search.indexOf( "swarmURL=" ) === -1 ) {
+		ok( jQuery.acceptData( document.createElement( "applet" ) ), "applet" );
+	}
```
gibson042: Don't do this; just drop the assertion unconditionally and let the preceding one cover it. See https://github.com/jquery/jquery/issues/1938#issuecomment-67330082 ("we don't need an **`applet`** or applet `object`") and observe that applet and embed get identical treatment: https://github.com/jquery/jquery/blob/7b111310977a30d02a8260a7ed6c9c437bb1b0a5/src/data.js#L242-L249

mzgol: Sounds good!

mzgol: Done.

test/unit/data.js
```diff
@@ -773,7 +773,6 @@ test("jQuery.acceptData", 11, function() {
 	ok( jQuery.acceptData( document.documentElement ), "documentElement" );
 	ok( jQuery.acceptData( {} ), "object" );
 	ok( jQuery.acceptData( document.createElement( "embed" ) ), "embed" );
-	ok( jQuery.acceptData( document.createElement( "applet" ) ), "applet" );
```
gibson042: Decrement the expected count to 10, then it's all set.

mzgol: OK, this time I run it and the tests pass. ;)

mzgol:
@gibson042 I had to drop one other test that was creating an applet, can you have a look? All those tests are run on embed as well so it should be fine.

I'm doing it all today in case of BrowserStack fixing this "Java out of date" bug so that I can be sure this time I've really fixed it.

mzgol:
I landed this; I didn't want to wait as it was causing TestSwarm to continuously restart IE data tests. It worked: http://swarm.jquery.org/job/216. Cherry-picking in progress...

Issue #2071, opened 2015-02-07
Animated toggle() turns display:inline into display:inline-block


cburschka:
There's an example here: http://jsfiddle.net/cburschka/ve5fbujq/

The non-animated `toggle()` appears to work normally. An animation like `toggle('slow')` will put a `display:inline-block` property on the `<span>` element, which in the above example causes the red text to drop to a new line instead of wrapping normally.

This happens whether or not the span was explicitly set to display:inline earlier, or was just inheriting the default value.

This bug seems similar to #1774 (apparently closed prior to the release of 2.1.3), but isn't quite the same.

dmethvin:
This looks correct to me. The `display: inline-block` is necessary to actually change the dimensions of the element during animation. The text wraps strangely because the `.hide()` animation changes the width, height, and opacity. When the width reduces to less than the length of the text, the text wraps.

What did you expect to happen?

cburschka:
Basically, the problem is that calling toggle() twice doesn't leave it in the state it was before - it's permanently turned into an inline-block element, even after the animation is over.

It can be worked around by explicitly resetting the element back to inline after the animation is done, though.

three3211:
@cburschka I am having the same issue. Whenever I call toggle, the hiding aspect works fine. However, when toggling back to show the hidden content, the inline style of `display:inline-block` is left and causes my content to appear not as it was originally.

As @cburschka, it can be worked around by explicitly resetting the element back to inline after the animation is done.

Issue #1767, opened 2014-10-21
$.show() can break responsive stylesheets


mzgol:
Originally reported by sten_paa@… at: http://bugs.jquery.com/ticket/15037
<p>
When $.show() is called the <em>computed</em> display value is cemented as an inline style.  This can cause inconsistent behavior.
</p>
<p>
Fiddle: <a class="ext-link" href="http://jsfiddle.net/jinglesthula/ZttLJ/" rel="nofollow"><span class="icon">&nbsp;</span>http://jsfiddle.net/jinglesthula/ZttLJ/</a>
</p>
<p>
Note that in the fiddle I use JavaScript to change the 'layout'.  This simulates what would happen if the <tt>display: inline</tt> style were applied conditionally based on a responsive media query.  That is, a given block of code could produce different results based on something like the orientation of a device and when the device orientation (really, the viewport width) changed with respect to when the show/hide calls are made.
</p>
<p>
I understand that based on the fact that an inline style is applied at the point when .show() is called that the browser is rendering the styles properly.  The 'bug' I suppose comes in the fact that a dynamic computed value is assumed to be ok to translate to a more fixed inline style.
</p>
<p>
Resolution may involve finding a way to 'clean up' the inline style at the end of .show() and allow all styles not provided by jQuery to continue doing exactly what they were before show/hide were invoked.
</p><p>Issue reported for jQuery 1.11.0</p>

jinglesthula:
On this page http://api.jquery.com/show/ a yellow Note callout was added indicating `show` can cause issues with responsive layout.  Now that 3.0 has been released and fixes the issue* I'm wondering if the note should be updated to indicate that <3.0 has the responsive issue, unless there are still other responsive layout issues with show even in 3.0.

\* at least in the fiddle above it fixed it when I switched it to 3.0 - I didn't do other testing regarding responsive layout + `show`/`hide`.

Issue #2062, opened 2015-02-04
Animation on Windows Store Apps


JCKodel:
In jquery-2.1.3.js, line 5498:

<code> * Try to determine the default display value of an element</code>

at line 5515:

<code>doc = iframe[ 0 ].contentDocument;</code>

always return undefined on Windows Store App, and is used in lines 5518 forward:

<code>doc.write();</code>

then crashes =\

Use case: trying to use jQuery BlockUI within a Windows Store App.

dmethvin:
As I understand the security model there, since Store apps have elevated privileges they don't take kindly to iframes that may have unknown content. Can you provide a jsfiddle that is the same as the smallest sample of code in your Windows Store app that crashes? The code path is [taking it through `defaultDisplay`](https://github.com/jquery/jquery/blob/cdfc2d092afad5a3e5b3978b04b402a1ee9dce79/src/css.js#L177-183) so I suspect it's the [use of `.show()`](https://github.com/malsup/blockui/blob/316f6e5d76a33266970778e80507149d9ef6a02d/jquery.blockUI.js#L402-430) inside BlockUI. But it's always a bad idea to guess.

It may turn out that we eliminate `defaultDisplay` in the process of dealing with gh-2057, but doing that is a breaking change and unlikely to land soon.

dmethvin:
If you can provide a more complete test case, we can reopen.

Issue #2057, opened 2015-02-03
Investigate possible performance improvements of show/hide methods


markelog:
Related:
https://github.com/jquery/jquery.com/issues/88
https://github.com/jquery/jquery/issues/1767

timmywil:
We could improve the case from https://github.com/jquery/jquery.com/issues/88 by simplifying – or avoiding – the check for visibility. It seems that webkit was taking a while getting that value. I'm not sure how that would work yet, though.

ivantodorovich:
The check for visibility could be optional by passing an argument and disabled by default.
It wouldn't sacrifice performance until it is really needed.

jamesarosen:
According to [this](http://tjvantoll.com/2013/01/09/html5-hidden-attribute-browser-support/), if jQuery *always* adds `[hidden] { display: none; }` to the page just before page-load, then it jQuery can use much simpler code:

```js
hide: function() {
  this.attr('hidden', true);
},
show: function() {
  this.attr('hidden', false);
}
```
That will support IE 7+ (browsers that obey the `[hidden]` selector).

I don't know the performance characteristics of `.attr` or `[hidden]`, though. Class selectors are probably faster than attribute selectors.

Another benefit of either class-based or attribute-based toggling is that it can be animated in CSS. That means a jQuery plugin can do `.show()` and I can decide show that show happens without changing the plugin's code.

zzzzBov:
@jamesarosen when showing/hidding elements, I used to think that the `[hidden]` attribute was a good choice, however [the HTML5 spec][1] includes a description that makes it inappropriate in a number of cases:

> The hidden attribute must not be used to hide content that could legitimately be shown in another presentation. For example, it is incorrect to use hidden to hide panels in a tabbed dialog, because the tabbed interface is merely a kind of overflow presentation — one could equally well just show all the form controls in one big page with a scrollbar. It is similarly incorrect to use this attribute to hide content just from one presentation — if something is marked hidden, it is hidden from all presentations, including, for instance, printers.

Because of this, I encourage the use of [`[aria-hidden]`][2], as it ties directly into the actual state of the element and also means that the element will be hidden in an accessible manner.

[1]: http://www.w3.org/TR/html51/single-page.html#the-hidden-attribute
[2]: http://www.w3.org/TR/wai-aria/states_and_properties#aria-hidden

niutech:
It may be a silly question, but why not just keep it simple?

```javascript
hide: function() {
  this.style.display = 'none';
[...]

PR #2180, opened 2015-03-30
CSS: Ignore the CSS cascade in show()/hide()/etc.

gibson042:
Fixes gh-1767
Fixes gh-2071

No more showing or animating stylesheet-hidden elements. And please note the new use of `jQuery.swap` (which interacts with #2182).

```
   raw     gz Compared to master @ 062b5267d0a3538f1f6dee3df16da536b73061ea    
 -1335   -425 dist/jquery.js                                                   
  -433   -204 dist/jquery.min.js
```


gibson042:
Ready for review. :fireworks:

src/css.js
```diff
@@ -10,17 +10,16 @@ define([
 	"./css/var/getStyles",
 	"./css/curCSS",
 	"./css/adjustCSS",
-	"./css/defaultDisplay",
 	"./css/addGetHookIf",
 	"./css/support",
-	"./data/var/dataPriv",
+	"./css/showHide",
```
dmethvin: Can we pull `.show()` and `.hide()` into this file and remove the dependency inside css.js completely? That way it becomes a separate module and can be removed in a custom build if you use classes for showing and hiding.

gibson042: So, introduce a new showHide module (`.show`, `.hide`, and `.toggle`) depending on data and depended upon by effects? That sounds reasonable, but out of scope for this ticket.

dmethvin: Yeah, it can go in a different ticket and pull request, I'll create an issue. Now that `showHide` is broken out it seems pretty clear we could do that, and in fact the new module only helps with size reductions if it can be excluded. As it is right now, it can't ever be excluded because of the dependency here.

dmethvin:
I love this PR, it simplifies things *so much*. The current code was really over-thinking so many edge cases and making life miserable.

It is a huge breaking change though, so it will need a lot of advance warning before 3.0 ships. All those edge cases are no doubt being used by people in various places, and we'll need to update the docs for these methods to clarify the cases we no longer support.

test/unit/css.js
```diff
@@ -475,14 +475,14 @@ test("show(); hide()", function() {
 
 	hiddendiv = jQuery("div.hidden");
 	hiddendiv.hide();
-	equal( hiddendiv.css("display"), "none", "Non-detached div hidden" );
+	equal( hiddendiv.css("display"), "none", "Cascade-hidden div after hide()" );
 	hiddendiv.show();
-	equal( hiddendiv.css("display"), "block", "Pre-hidden div shown" );
+	equal( hiddendiv.css("display"), "none", "Show does not trump CSS cascade" );
```
timmywil: This is the case that concerns me. I have used the stylesheet to hide elements initially and later called `.show()` on them. The new workflow – namely removing the `"hidden"` class – seems fine, but could require lots of changes by the user. We'll want to make sure this is covered in the blog post, docs, and migrate.

dmethvin: I've added a note to the draft blog post on it. Adding this to Migrate will be fun since it has to add back all the crap above.

gibson042: Yes, that is the crux of it, and the reason why #1767 is a blocker. It seems worth it to me, especially after seeing the code changes, but does require some substantial changes from everyone trying to show stylesheet-hidden elements.

dmethvin:
This should reference #2057 and would resolve it as far as I'm concerned.

timmywil:
@dmethvin Thanks Dave. I agree, and now it does. :)

Mottie:
I would like to suggest adding a boolean (e.g. `.show(true)`) to mimic the current behavior and show cascade-hidden elements. A named option would also be required if the developer is using an object. Sadly, this idea would not work with `.toggle()`.

gibson042:
I don't see that happening, for a few reasons:
* We're trying to avoid new boolean traps: https://github.com/jquery/jquery/wiki/Adding-new-features
* It's easy to handle with client code (e.g., `$el.removeClass("hidden").show()` or—at worst—`$el.css( "display", getVisibleDisplay ).show()`)
* We relied on [_**ugly**_ kludges](https://github.com/jquery/jquery/pull/2180/files#diff-aa200529dc721786f51c5f9e7af3d592) to determine substitute display values, [causing new problems](https://github.com/jquery/jquery/issues/2062) [without even guaranteeing correctness](http://bugs.jquery.com/search?q=defaultdisplay&go=&ticket=on)

gibson042:
However, I just looked over the code again, and `defaultDisplay` may be separable from post-animation display restoration (which were the primary impetus for this PR). If it is, and if we're willing to pay the size and maintenance costs, it should be possible to bring back showing cascade-hidden elements without re-breaking gh-1767 and gh-2071 (except for initially-hidden elements, which are SOL regardless).

I'm still against it, but thought the idea should be on the record.

Issue #2058, opened 2015-02-03
Don't expose jQuery.swap


mzgol:
Even if we want/need `jQuery.swap` internally, we still shouldn't expose it, there is no reason to do it now that we have a robust modular architecture. The method is undocumented so technically we could stop exposing it whenever we want; no better moment to do it than the 3.0 release!

I didn't see it in the code but just to make sure: @scottgonzalez & @arschmitz, I assume UI/Mobile doesn't depend on it?

scottgonzalez:
We definitely don't use it in UI.

arschmitz:
I didn't even know it existed until i saw the related discussion. It's definitely not in mobile either.

timmywil:
Sounds good to me! I think it was originally exposed so users could have control over when to retrieve useful values for hidden elements, but that use case is rare at best.

mzgol:
OK then, assigning to myself!

dmethvin:
If it turns out something needs it but we don't want it internally, we could put it in jQuery Migrate. I doubt it's used externally and we never documented it, plus it's really bad practice from a performance standpoint.

PR #2182, opened 2015-03-30
CSS: Don't expose jQuery.swap

mzgol:
jQuery.swap was an undocumented API used only internally. With the modular
AMD system we currently have it's not necessary to expose this function
publicly under the jQuery object.

Fixes gh-2058

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -8,6 +8,7 @@ define([
 	"./css/var/cssExpand",
 	"./css/var/isHidden",
 	"./css/var/getStyles",
+	"./css/var/swap",
 	"./css/curCSS",
 	"./css/adjustCSS",
 	"./css/defaultDisplay",
@@ -16,11 +17,10 @@ define([
 	"./data/var/dataPriv",
 
 	"./core/init",
-	"./css/swap",
 	"./core/ready",
 	"./selector" // contains
 ], function( jQuery, pnum, access, rmargin, rcssNum, rnumnonpx, cssExpand, isHidden,
-	getStyles, curCSS, adjustCSS, defaultDisplay, addGetHookIf, support, dataPriv ) {
+	getStyles, swap, curCSS, adjustCSS, defaultDisplay, addGetHookIf, support, dataPriv ) {
 
 var
 	// Swappable if display is none or starts with table
@@ -359,7 +359,7 @@ jQuery.each([ "height", "width" ], function( i, name ) {
 				// but it must have a current display style that would benefit
 				return rdisplayswap.test( jQuery.css( elem, "display" ) ) &&
 					elem.offsetWidth === 0 ?
-						jQuery.swap( elem, cssShow, function() {
+						swap( elem, cssShow, function() {
 							return getWidthOrHeight( elem, name, extra );
 						}) :
 						getWidthOrHeight( elem, name, extra );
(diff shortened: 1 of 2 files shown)


dmethvin:
Definitely a good idea. Now we get to find out who's using this!

arthurvr:
Sure! :+1:

mzgol:
This will also allow to change `swap` semantics. On `master` it's currently used in only one place so it might not need all its current quirks.

theodoreb:
It's being used in Drupal in a custom plugin we made to get the size of an horizontal menu: [jquery.intrisic.js](http://cgit.drupalcode.org/drupal/tree/core/misc/jquery.intrinsic.js#n43). It's used in all our admin pages for responsive menu purposes.

I haven't looked into what it'd take to replace it on our side, just noticed this issue. Totally fine with reworking the script. If you have in mind a easy way to replace this it'd be great, otherwise we'll do our homework.

dmethvin:
@theodoreb thanks for the heads-up, always better to know before than after. Besides the fact that this has been undocumented, it's a real performance killer because it can cause two reflows of the affected regions. Generally it's better to do the bookkeeping of visibility yourself (e.g. via a data item or class name) so the browser won't have to do all that work.

Sounds like we may want to move the `jQuery.swap()` code to jQuery Migrate then as a compat warning and polyfill?

markelog:
It might be very dangerous, without any gain, i would be more comfortable if we would "deprecate" it first by adding message in migrate but still exposing it until 4.0

mzgol:
This was never exposed. You cannot deprecate an API that was never documented - and what would be the gain? If someone didn't see it look into the docs to see there is no documentation, they wouldn't see it deprecated either. Also, it's trivial to add and will be easily handled by the Migrate plugin.

I think we should remove it.

markelog:
> This was never exposed.

It is exposed through API. You could also see the header of this pull if you don't believe me :-)

> You cannot deprecate an API that was never documented

That is why i put this word in quotes.

There is couple articles, as i recall, mentions `swap`, plugins that use it and real-life code depending on it.

> what would be the gain

Gain would be not to break users code.

> If someone didn't see it look into the docs to see there is no documentation, they wouldn't see it deprecated either

As i said above, we could add a warning to migrate first.

> I think we should remove it.

I don't think we should remove it, i think we should stop exposing it, but i don't see the rush, i think there wouldn't be any point to do it right now.

theodoreb:
For the record you can take Drupal out of the real-life code depending on it. Our code can do without it, it's merely by convenience we ended up using it. The version of Drupal using swap is still in beta, so on our end there is nothing we can't break at this level. 

I'd say putting it in jquery migrate is the right thing to do. Though I don't have anything in the fight since Drupal isn't and won't be using the migrate plugin.

dmethvin:
I've thought about this a bit more and am wondering if we are doing this backwards. Perhaps the better course would be to expose jQuery.swap and take out the "get dimensions of hidden elements" feature. That way, we let people continue to do slow things but don't hide the badness of it.

mzgol:
@dmethvin Would it really buy those users much? I think this method looks a little arcane, it's just an utility that made sense to save space internally by not repeating logic but I think it's not really intuitive to use. If someone needs it somewhere, they need to know what they want to swap and at this point it seems easier for such a person to do this swapping logic manually.

timmywil:
@mzgol agreed. This function doesn't have much utility. It makes more sense to me for the user to do it themselves.

dmethvin:
I see your point @mzgol, we're making a generic utility but someone who wanted to do this in their own code could do it with a lot less hassle. OK then let's carry on with the original plan :smiley_cat:

Issue #1684, opened 2014-10-08
CORS HTTP patch request is not working on IE10/11


tomchentw:
See https://github.com/jquery/jquery/issues/1684#issuecomment-65758579 for latest update.

mzgol:
@tomchentw Could you check if this works fine for you with https://dl.dropboxusercontent.com/u/10200888/jquery/jquery-cors.js? I submitted a PR with such a patch: #2183.

tomchentw:
Currently don't have an IE environment to test with...

mzgol:
jsFiddle broke jQuery Edge URLs; apparently jQuery Compat Edge & jQuery Edge are swapped. :/ See https://github.com/jsfiddle/jsfiddle-issues/issues/600

It's better to use jsbin now.

mzgol:
Closed via #2183, laded at 61f812b7e7b88dd6e0078c241e4c88905ea51562.

PR #2183, opened 2015-03-30
Ajax: Use the native XHR for all non-local requests in IE9+

mzgol:
IE throws an error on cross-domain PATCH requests if issued via the ActiveX
interface. This commit switches the logic to use the native XHR in all
non-local requests.

Fixes gh-1684

I'd like to test it but we'd need a *real* cross-domain request that has `Access-Control-Allow-Origin` set to `*` and will respond to `PATCH` requests. Can we have a page like that? cc @gnarf @ryanneufeld @aulvi

### DIFF
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -11,17 +11,31 @@ jQuery.ajaxSettings.xhr = window.ActiveXObject !== undefined ?
 	function() {
 
 		// XHR cannot access local files, always use ActiveX for that case
-		return !this.isLocal &&
+		if ( this.isLocal ) {
+			return createActiveXHR();
+		}
 
-			// Support: IE<9
-			// oldIE XHR does not support non-RFC2616 methods (#13240)
-			// See http://msdn.microsoft.com/en-us/library/ie/ms536648(v=vs.85).aspx
-			// and http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9
-			// Although this check for six methods instead of eight
-			// since IE also does not support "trace" and "connect"
-			/^(get|post|head|put|delete|options)$/i.test( this.type ) &&
+		// Support: IE 10-11
+		// IE seems to error on cross-domain PATCH requests when ActiveX XHR
+		// is used. In IE 9+ always use the native XHR.
+		// Note: this condition won't catch Spartan as it doesn't define
+		// document.documentMode but it also doesn't have ActiveX so it won't
+		// reach this code.
+		if ( document.documentMode > 8 ) {
+			return createStandardXHR();
+		}
 
-			createStandardXHR() || createActiveXHR();
+		// Support: IE<9
+		// oldIE XHR does not support non-RFC2616 methods (#13240)
+		// See http://msdn.microsoft.com/en-us/library/ie/ms536648(v=vs.85).aspx
+		// and http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9
+		// Although this check for six methods instead of eight
+		// since IE also does not support "trace" and "connect"
+		if ( /^(get|post|head|put|delete|options)$/i.test( this.type ) ) {
+			return createActiveXHR();
+		}
+
+		return createStandardXHR();
 	} :
 	// For all other browsers, use the standard XMLHttpRequest object
 	createStandardXHR;
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -1578,8 +1578,10 @@ module( "ajax", {
 		}
 	} );
 
-	// BrowserStack PATCH support sometimes breaks so on TestSwarm run the test in IE8 only.
-	if ( location.search.indexOf( "swarmURL=" ) === -1 || document.documentMode < 9 ) {
+	// BrowserStack PATCH support sometimes breaks so on TestSwarm run the test in IE only.
+	// Unfortunately, all IE versions gets special treatment in request object creation
+	// so we need to test in all supported IE versions to be sure.
+	if ( location.search.indexOf( "swarmURL=" ) === -1 || document.documentMode ) {
 		ajaxTest( "#13240 - jQuery.ajax() - support non-RFC2616 methods", 1, {
 			url: "data/echoQuery.php",
 			method: "PATCH",
--- a/test/integration/gh-1684-ajax.html
+++ b/test/integration/gh-1684-ajax.html
@@ -1,0 +1,52 @@
+<!DOCTYPE html>
+<html>
+<head lang="en">
+	<meta charset="UTF-8">
+	<meta http-equiv="X-UA-Compatible" content="IE=edge">
+	<title>Test for gh-1684</title>
+	<style>
+		#result {
+			font-size: 24px;
+			margin: 0.5em 0;
+		}
+		#response {
+			white-space: pre;
+		}
+		.error {
+			background-color: red;
+		}
+		.warn {
+			background-color: yellow;
+		}
+		.success {
+			background-color: lightgreen;
+		}
+	</style>
+</head>
+
+<body>
+	<div id="result"></div>
+	<div id="response"></div>
+	<script src="../../dist/jquery.js"></script>
+	<script>
+		if ( !jQuery.support.cors ) {
+			jQuery( "#result" )
+				.addClass( "success" )
+				.text( "CORS not supported in this browser. Test not run." );
+		} else {
+			jQuery.ajax( {
+				url: "http://httpbin.org/patch",
+				method: "PATCH",
+				success: function( data ) {
+					jQuery( "#result" ).addClass( "success" ).text( "Test passed." );
+					jQuery( "#response" ).text( "Response:\n" + JSON.stringify( data, null, 4 ) );
+				},
+				error: function( error ) {
+					jQuery( "#result" ).addClass( "error" ).text( "Test failed." );
+					jQuery( "#response" ).text( "Error:\n" + JSON.stringify( error, null, 4 ) );
+				}
+			} );
+		}
+	</script>
+</body>
+</html>


src/ajax/xhr.js
```diff
@@ -19,7 +19,19 @@ jQuery.ajaxSettings.xhr = window.ActiveXObject !== undefined ?
...
+			// Support: IE 10-11
+			// IE seems to error on cross-domain PATCH requests when ActiveX XHR
+			// is used. Use the following strategy:
+			// 1. If request is local, use the ActiveX XHR.
+			// 2. In IE9+ always use the native XHR.
+			// 3. If IE8 and the request type is not included in the following list,
+			//    use the ActiveX XHR.
+			// 4. Otherwise use the native XHR.
+			// jshint -W018
+			( !( document.documentMode < 9 ) ||
```
markelog: Look like a ua sniff, is that a really best way to do this?

mzgol: It's a browser sniff, not UA sniff (changes in UA won't make this code behave differently). Also, this will not catch any new browsers but only old, specific ones (i.e. only IE8 or newer IE in IE8 document mode) so this will not have any unwanted effect on new browsers.

But if you have any idea how to test for this specific issue, I'm listening. :) I just don't think it's possible to test for such weird cases in JS code, especially without seriously impacting performance.

timmywil: Yea, I think we'd all prefer feature tests, but we've resigned ourselves to do this in a few necessary places. This one seems legit.

src/ajax/xhr.js
```diff
@@ -19,7 +19,19 @@ jQuery.ajaxSettings.xhr = window.ActiveXObject !== undefined ?
...
+			// is used. Use the following strategy:
+			// 1. If request is local, use the ActiveX XHR.
+			// 2. Otherwise, in IE9+ always use the native XHR.
+			// 3. Otherwise, in IE8 if request type is *not* included in the following
+			//    list, use the ActiveX XHR.
+			// 4. Otherwise, use the native XHR.
+			// jshint -W018
+			( !( document.documentMode < 9 ) ||
+			// jshint +W018
+				/^(get|post|head|put|delete|options)$/i.test( this.type ) ) &&
```
gibson042: This all got quite convoluted. How about reversing the polarity?
```js
// Native XHR cannot access local files, always use ActiveX for that case
return this.isLocal ||

	// Also force ActiveX for non-RFC2616 methods with IE<9 (trac-13240)
	// http://msdn.microsoft.com/en-us/library/ie/ms536648(v=vs.85).aspx
	// ...but *ONLY* IE<9, since IE 10-11 cross-domain PATCH requires native XHR (gh-1684)
	document.documentMode < 9 && !/^(?:get|post|head|put|delete|options)$/i.test( this.type ) ?

	createActiveXHR() :
	createStandardXHR() || createActiveXHR();
```

mzgol: That seems more readable but what's the point of the last ` || createActiveXHR()` in your example?

markelog: Since we're talking about this, how about simple `if` or `else` here and there, ternaries might not be a good choice for conditions like this.

It shouldn't be hit bytes count too much or not at all

jaubourg: Why don't we specifically tests for "patch"?

```js
( document.documentMode < 9 && /^patch$/i.test( this.type ) || createStandardXHR() ) && createActiveXHR();
```

mzgol: > ```js
( document.documentMode < 9 && /^patch$/i.test( this.type ) || createStandardXHR() ) && createActiveXHR();
```

IE<9 doesn't support any non-RFC2616 method as indicated in the comment. I don't think testing for `PATCH` is enough.

Comment (mzgol) [inline src/ajax/xhr.js]: BTW, are we gaining anything by using the native XHR in IE8? Maybe we could simplify it to:

1. IE8 or a local request via IE9+ => ActiveX
2. A remote request on IE9+ => native XHR

Comment (gibson042) [inline src/ajax/xhr.js]: > That seems more readable but what's the point of the last ` || createActiveXHR()` in your example?

Because `createStandardXHR` has a `try` block and can return `undefined` when "[Enable native XMLHTTP support](https://msdn.microsoft.com/en-us/library/ms537505(v=vs.85).aspx)" is disabled. That's also the reason why this expression wasn't _already_ using a ternary conditional.

Comment (gibson042) [inline src/ajax/xhr.js]: > are we gaining anything by using the native XHR in IE8?

A good question. @jaubourg?

Comment (jaubourg) [inline src/ajax/xhr.js]: From the top of my head, I don't think there is a gain... but it may have been lost in the mysts of ajax refactoring.

The sad truth is you can disable XMLHttpRequest and/or ActiveX and there are lots of subtle differences. I'm not certain they are all documented and I'm not convinced there is a proper all-encompassing priority to be found here.

Seeing as it's getting to this kind of browser-sniffing / test-with-a-gigantic-regexp situation, I wonder if we wouldn't be better served by only providing a default implementation that creates a native XHR and then document that you can change the factory to use ActiveX on a per-request basis. That's the whole point of having the factory as a setting.

Comment (timmywil) [inline src/ajax/xhr.js]: @jaubourg Good point, but I think the solution that causes the least hassle for our users is a combination of both, where we do our best to cover the common cases (and a cross-domain patch seems ok to try and cover) and leave the extensibility to cover the rest.

Comment (dmethvin) [inline src/ajax/xhr.js]: At this point the problem only applies in compat anyway, since the ActiveX version won't work with the event-based implementation in master. With IE8 becoming totally unsupported early next year, there probably isn't a lot of reason to worry about this.

Comment (mzgol) [inline src/ajax/xhr.js]: @dmethvin @timmywil What do you propose? To keep the logic from this PR (modulo some modifications) or to always use ActiveX in IE8, essentially dropping support for IE8 with a disabled ActiveX XHR?

Comment (dmethvin) [inline src/ajax/xhr.js]: I'm okay with the `document.documentMode` check. Depending on when 3.0 ships I am starting to think we might just want to patch the 1.x line and not even have a compat branch, since we will drop it soon thereafter.

Comment (timmywil) [inline src/ajax/xhr.js]: Uhoh, can of worms opened.

Comment (timmywil) [inline src/ajax/xhr.js]: But for now, I think it's best to stick with the solution we have.

Comment (markelog) [inline src/ajax/xhr.js]: Only let's have an couple `if` here not ternary with `||` and `&&` abomination

Comment (mzgol) [inline src/ajax/xhr.js]: Sure. I'll modify the patch.

**Comment by mzgol** [comment]
PR updated, test added thanks to http://httpbin.org/ (thanks @tomchentw for the suggestion!). Unfortunately, if we want to have it tested we *have to* have working PATCH support on TestSwarm and it seems it got broken on BrowserStack *again*. I confirmed locally on IE10 & IE11 that the test fails on current `compat` & passes with my patch but it fails on BrowserStack in both cases.

We already have one PATCH test that's blacklisted on BrowserStack outside IE8. We could blacklist this one as well but it tests a bug in IE10-11 so I'd rather have them tested.

Can anyone reach to BrowserStack so that they *really* solve this issue so that it doesn't re-occur constantly? @jzaefferer?

File: test/unit/ajax.js [hunk 3]
```diff
@@ -1590,6 +1592,19 @@ module( "ajax", {
 				ok( false, "error" );
 			}
 		} );
+
+		if ( jQuery.support.cors ) {
+			ajaxTest( "gh-1684 - jQuery.ajax() - support cross-domain PATCH in IE10-11", 1, {
+				url: "http://httpbin.org/patch",
```
markelog: So now we need a internet connection to run the tests and rely on the third party?

mzgol: This test requires a *real* cross-domain PATCH request. So it's either an internet connection required or a separate server set up locally (which most people won't do, it's an additional hassle).

Apart from using a 3rd party service we may define a separate domain with a new nginx config just for this test. @gnarf doesn't have time to work on it now and even if I sent him an nginx config for this new server, this would still be the only externally visible server with CORS enabled that's responding to cross-domain requests. This is a security risk so has to be done carefully; that's why @gnarf suggested to use an external service.

Do you have any other ideas?

markelog: Best way would be to run http-server, but until we have php-dep with swarm we can't do it.

I'd say stubbing is another option to consider, i remember we talked about introducing these kind of testing for special cases, it will also resolve your browserstack issue.

markelog: Another way would be to test `jQuery.ajaxSettings.xhr`, not the whole `ajax` module.

mzgol: How would you test `jQuery.ajaxSettings.xhr`?

markelog: > How would you test jQuery.ajaxSettings.xhr?

In IE env set `type` to `patch` and call `jQuery.ajaxSettings.xhr` then test returned value.

> You cannot test if a native API bugs is worked around properly if you replace those native APIs with stubs that won't reproduce these bugs.

In technical terms, testing native API with conjunction of `ajax` module would be an integration test, testing through stubbing interface would be a unit test.

As an option, we can settle on unit test now and create issue for integration one and after we improve testing infrastructure introduce the intergration test.

mzgol: >> How would you test jQuery.ajaxSettings.xhr?
>
> In IE env set type to patch and call jQuery.ajaxSettings.xhr then test returned value.

This will only be a duplicate of the implementation from this patch. I don't want to check if I implemented what I think I did, this is pretty trivial. I want to ensure it works.

> In technical terms, testing native API with conjunction of ajax module would be an integration test, testing through stubbing interface would be a unit test.

We already have such tests - PATCH in IE8 is tested in this way.

> In technical terms, testing native API with conjunction of ajax module would be an integration test, testing through stubbing interface would be a unit test.

IMO a test with a stubbed API would be useless here, it would basically just repeat the code I wrote.

markelog: > This will only be a duplicate of the implementation from this patch. I don't want to check if I implemented what I think I did, this is pretty trivial.

This is point of testing as process, point of unit-testing, point of high-coverage tools. It is preferable to test everything.

> We already have such tests - PATCH in IE8 is tested in this way.

I'd say that is not enough.

> IMO a test with a stubbed API would be useless here, it would basically just repeat the code I wrote and wouldn't test anything.

I'm not sure who you come to this conclusion.

> I'd rather not wait for that, it may take a long time and in the meantime we'd have untested changes here.

Their shouldn't be untested, we would need to unit-test it, but your intergration tests in no way better. 

If, for example, Spartan browser wouldn't have a `documentMode` property which is a likely scenario, your tests will not caught anything. Since they are to specific and ua-oriented. 

Now we're introducing another third-party dependency and now requiring to have an internet connection, which we tried really hard to not to do.

Introducing new points of failure to the system is never a good thing.

mzgol: > If, for example, Spartan browser wouldn't have a `documentMode` property which is a likely scenario, your tests will not caught anything. Since they are to specific and ua-oriented.

I don't understand. Spartan doesn't have `document.documentMode` and doesn't have `window.ActiveXObject`. It will be treated as every non-IE browser so I don't see a problem here.

What's UA-oriented here besides the `documentMode` check that's only done to omit the test in non-IE browsers because of BrowserStack problems as it's obvious it works there as we're just using the native XHR always?

markelog: > I don't understand. Spartan doesn't have document.documentMode and doesn't have window.ActiveXObject. It will be treated as every non-IE browser so I don't see a problem here.

It should be like that, but until Spartan is officially released we can't know that, hence me noting this as a "likely scenario" not a definite fact, but that is beside the point, purely hypothetical - an example.

What i'm point that, is you know the conditions of the problematic situation and specificially reproducing it, this is a white box testing. Since these tests are so specific, they can't be preventive, only regressive. Which is not the point of integration assertions.

The same exact thing you could do with stubbing interface without introducing concerns above.

mzgol: >> I don't understand. Spartan doesn't have document.documentMode and doesn't have window.ActiveXObject. It will be treated as every non-IE browser so I don't see a problem here.
>
> It should be like that, but until Spartan is officially released we can't know that, hence me noting this as a "likely scenario" not a definite fact, but that is beside the point, purely hypothetical - an example.

Well, until Spartan is released theoretically anything may happen so we cannot test against it. I can swap the `documentMode` check here for the `window.ActiveXObject` one, this would be closer to what we're doing here as the whole logic is skipped in browsers without `ActiveXObject`.

> Since these tests are so specific, they can't be preventive, only regressive.

That's still important. Preventing testing here would be adding more cross-domain request types than just PATCH.

mzgol: > The same exact thing you could do with stubbing interface without introducing concerns above.

So what do you propose? To add a couple of stubs that represent IE8 (no cross-domain support, PATCH only via ActiveX) and IE10-11 (cross-domain PATCH only via native XHR) and test them? IMO it's incredibly hard to create such stubs and be sure everything is reproduced properly.

markelog: We could try, i'm fine with landing it as is, afterwards, i could try to do this myself if you found this to hard.

Would like others to agree with my view though.

mzgol: > i could try to do this myself if you found this to hard.

It's not hard to implement (just a little tedious), it's hard to be sure the whole behavior is replicated properly. Especially that IE11 up until recently was receiving more substantial patches so sth like that could have changed behind us and we wouldn't detect it if we just tested on stubs. Maybe now that's not an issue (@dmethvin probably know more) but I still feel uneasy about not testing this on real browsers instead of just our assumptions about how they work.

dmethvin: As I understand it, Spartan will have just about all the IE-isms removed. So it should be safe to use `.documentMode` because Spartan won't even have document modes. 

Also, a detection via `if ( window.ActiveXObject ) { ... }` fails in IE11 because it was used for bad browser sniffs. You could use it to your advantage in this case perhaps? Actually I guess not because IE10 should use the native XHR.

mzgol: > Also, a detection via if ( window.ActiveXObject ) { ... } fails in IE11 because it was used for bad browser sniffs.

`window.ActiveXObject !== undefined` will succeed, though since we use it ourselves. ;) But this would be stepping over our implementation, `document.documentMode` seems safer here.

timmywil:
In the meeting, we discussed the possibility of separating the integration test from our unit tests. I proposed this because I do agree that it's best that the unit tests do not require an internet connection. However, the test in this PR is worthwhile. So, I'd say place this test on a page that can be tested independently of the unit tests and we can at least run it before releases. In the future, I'd like to see more integration testing (specifically, to replace all of our mock mouse clicks and such in events).

mzgol:
PR updated. This is how the page looks like in IE11:

Before applying the patch:
screen shot 2015-05-11 at 19 24 24

After applying the patch:
screen shot 2015-05-11 at 19 23 58

If no one opposes, I'll merge it soon.

test/unit/ajax.js
```diff
@@ -1578,8 +1578,10 @@ module( "ajax", {
 		}
 	} );
 
-	// BrowserStack PATCH support sometimes breaks so on TestSwarm run the test in IE8 only.
-	if ( location.search.indexOf( "swarmURL=" ) === -1 || document.documentMode < 9 ) {
+	// BrowserStack PATCH support sometimes breaks so on TestSwarm run the test in IE only.
+	// Unfortunately, all IE versions gets special treatment in request object creation
+	// so we need to test in all supported IE versions to be sure.
+	if ( location.search.indexOf( "swarmURL=" ) === -1 || document.documentMode ) {
```
mzgol: Not sure about this change. We do have special logic for all IE versions now but PATCH is broken again on BrowserStack in some IEs. Removing it completely sounds dangerous, though until we have integration tests in place.

mzgol: Update: this wasn't working for me as I have the extension installed and then in the Live interface PATCH is not supported.

markelog:
LGTM

mzgol:
Landed at 61f812b7e7b88dd6e0078c241e4c88905ea51562

mzgol:
Removing the milestone since it's already on the issue and it should stay there.

Issue #2145, opened 2015-03-14
$.each fails intermittently on iOS due to Safari bug


osolo:
There is a timing bug in iOS8 that causes mobile Safari to incorrectly report a 'length' on objects that don't have one.

To the best of my knowledge, this happens on iOS8+, possibly only on 64-bit systems.  The bug is triggered for objects that have only numeric properties.  For example:
```javascript
foo = { 1: 'a', 2: 'b', 3: 'c' } 
```
In this case, if you query `foo.length` then mobile Safari will *sometimes* return 4 (the highest property + 1).

This causes functions like `$.each()` to treat objects such as foo above as arrays instead of objects, and when it tries to iterate them as such it fails since there is no foo[0].

The problem can be fixed in the function `isArrayLike()`.  Instead of just checking for 
```javascript
typeof length === "number"
```
you also need to check for
```javascript
obj.hasOwnProperty('length')
```

The latter check is immune to the iOS bug.

I realize this is a fix just for one browser, but it's a browser with a very large user base.

You can see more background and some repro steps at the following Stack Overflow discussion:

http://stackoverflow.com/questions/28155841/misterious-failure-of-jquery-each-and-underscore-each-on-ios

mzgol:
Thanks for the report! This is really weird and due to being a timing issue it seems impossible to write a test for it. :( Did you report it to Apple? And/or to bugs.webkit.org? We'd like an issue we'd be able to link to.

I agree we should work around it.

osolo:
@mzgol I agree that writing a test case would be impossible.  I didn't report the issue to Apple.  I think it would carry more weight coming from the jQuery team, but if you point me to the right direction I'd be happy to file with them myself.

mzgol:
The description you provided seems detailed enough. You could report a bug at https://bugs.webkit.org/, mention it affects jQuery & link to this issue.

osolo:
Link to WebKit bug: https://bugs.webkit.org/show_bug.cgi?id=142792

dmethvin:
Wow. I guess this must be rare enough that it doesn't occur that often? Plain object, only numeric indices, etc. There's a patch in https://github.com/jashkenas/underscore/pull/2094 that says it tries to work around this. We could do something similar but it seems like we're in the same situation as we were with the December release. Is it worth doing another minor-point release for this?

mzgol:
I'd like to get a response to this WebKit bug first to know what we're dealing with. @BenjaminPoulain, could you have a look?

[...]

PR #2185, opened 2015-04-04
Core: add workaround for iOS JIT error in isArrayLike

timmywil:
I was able to reproduce on an iPhone 5s and iPhone 6 plus in iOS 8.2. Additionally, I tested AMD and minified versions for posterity. Let me know if I missed something.

Perf difference seems to be [negligible](http://jsperf.com/jquery-each-vs-no-jit-error).

Fixes gh-2145

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -478,7 +478,12 @@ jQuery.each("Boolean Number String Function Array Date RegExp Object Error".spli
 });
 
 function isArraylike( obj ) {
-	var length = obj.length,
+
+	// Support: iOS 8.2 (not reproducible in simulator)
+	// `in` check used to prevent JIT error (gh-2145)
+	// hasOwn isn't used here due to false negatives
+	// regarding Nodelist length in IE
+	var length = "length" in obj && obj.length,
 		type = jQuery.type( obj );
 
 	if ( type === "function" || jQuery.isWindow( obj ) ) {
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -1217,6 +1217,27 @@ test("jQuery.each(Object,Function)", function() {
 	equal( i, 2, "Iteration over document.styleSheets" );
 });
 
+test( "JIT compilation does not interfere with length retrieval (gh-2145)", function() {
+	expect( 4 );
+
+	var i;
+
+	// Trigger JIT compilation of jQuery.each – and therefore isArraylike – in iOS.
+	// Convince JSC to use one of its optimizing compilers
+	// by providing code which can be LICM'd into nothing.
+	for ( i = 0; i < 1000; i++ ) {
+		jQuery.each( [] );
+	}
+
+	i = 0;
+	jQuery.each( { 1: "1", 2: "2", 3: "3" }, function( index ) {
+		equal( ++i, index, "Iteration over object with solely " +
+			"numeric indices (gh-2145 JIT iOS 8 bug)" );
+	});
+	equal( i, 3, "Iteration over object with solely " +
+		"numeric indices (gh-2145 JIT iOS 8 bug)" );
+});
+
 test("jQuery.makeArray", function(){
 	expect(15);
 


timmywil:
Saw note about closure compiler. Testing there.

timmywil:
After testing with closure compiler, changed solution to use hasOwnProperty.

mzgol:
@timmywil Just to make sure: did you test on a 64-bit (important!) iOS 8.2 device? My iPhone 5s got stolen and this is not reproducible in the simulator so I can't help you here currently, unfortunately.

mzgol:
But just looking at the code, it looks good.

timmywil:
@mzgol I did. I was able to reproduce in a test case and get it failing too.

test/unit/core.js
```diff
@@ -1215,6 +1215,19 @@ test("jQuery.each(Object,Function)", function() {
 		i++;
 	});
 	equal( i, 2, "Iteration over document.styleSheets" );
+
+	// Trigger JIT
+	for ( i = 0; i < 1000; i++ ) {
```
markelog: Maybe it would be better to put those assertions in separate test? In my experience, the lesser assertions in one test the better isolation when assertion(s) fails.

mzgol: @markelog +1, especially for something so specific as this bug.

timmywil: Good call.

timmywil:
@mzgol I just reread your comment. Sorry your iphone got stolen man! Will you get a replacement?

mzgol:
@timmywil I didn't get an insurance for it (my mistake) so I had to buy another one. :/

timmywil:
@mzgol ugh

markelog:
> iPhone 5s got stolen 

I feel you, same thing happened to me couple months back :-(

mzgol:
> I feel you, same thing happened to me couple months back :-(

&lt;bad joke&gt;Maybe they all badly needed 64-bit iOS devices to test their sites against this bug.&lt;/bad joke&gt;

markelog:
> \<bad joke\>

\<nerd\>That is incorrect tag!\</nerd\\\> so disappointed in you now :-)

snapwich:
@timmywil, I'm following your lead here as I'm trying to implement [a fix in Angular.js for its isArrayLike](https://github.com/angular/angular.js/pull/11508).  I like the solution you have using hasOwnProperty("length") as it is much less code than [the dynamic length lookup helper method](https://github.com/jridgewell/underscore/commit/086a56d04c0214054219d83eff25a4cbba524fc2) since we're only putting this fix in one place, and it does seem to fix the JIT bug. However, [it seems to break Internet Explorer for HTMLCollections](https://twitter.com/Snapwich/status/585568775203901441) which I assume should be considered array like.

It's clearly an IE bug since `typeof Object.getOwnPropertyNames(document.forms)["length"] !== undefined` is true in IE and all the evergreen browsers, so IE is obviously contradicting itself; but I'm not sure if that shoots a hole in using this method.

Any ideas on how to get around this?  Unfortunately Object.getOwnPropertyNames is only IE9+ so it's probably not useful here for anything other than making IE look silly.

markelog:
@snapwich would you mind posting a test-case? So there wouldn't be any assumptions?

gibson042:
http://jsfiddle.net/h2evpj2u/

We should probably align with [underscore](https://github.com/jashkenas/underscore/commit/3101e18a5818b158e9ffc18291be4c6a167e28f8)/[lodash](https://github.com/lodash/lodash/commit/4a8695caffa595d5eb2eca6d4b6f7fd899f54846) and use computed property access (`obj[ prop ]`, where `prop` has a value of "length").

timmywil:
@gibson042 lodash and underscore have since moved to the dynamic length solution that @snapwich referred to. It requires [adding another function](https://github.com/jquery/jquery/issues/2145#issuecomment-88650887) to retrieve length to avoid interference from minifiers – especially closure.

I would like to find a solution that doesn't go that route.

timmywil:
Although, I can't think of a better solution now that `hasOwn` seems debunked.

gibson042:
The important detail is computed property access; those functions are just a way to ensure that minification doesn't convert them into identifier property access. But jQuery always uses the same single minifier, and if we trust our unit tests then we can use a "length"-valued variable instead of introducing a new function.

timmywil:
That was my original route, but I'd like our code to work with the closure compiler as well so users can use their own minification.

markelog:
Are even sure it works now?

jdalton:
Have you tried a simple `in` check? As in `'length' in Object(value)`

caitp:
> It's clearly an IE bug since typeof Object.getOwnPropertyNames(document.forms)["length"] !== undefined is true in IE and all the evergreen browsers, so IE is obviously contradicting itself; but I'm not sure if that shoots a hole in using this method.

I don't think so --- IDL attributes are (in general) not (supposed to be) own properties, they're accessors on the prototype

timmywil:
I'm trying some stuff.

snapwich:
@jdalton `in`, why didn't I think of that.  brilliant :)

@timmywil `typeof obj === "object" && "length" in obj && obj.length` seems to be working for me in my testing and avoids the JIT bug.  I think the typeof check is now essential though as doing `in` on something like a string will throw an error.

timmywil:
@snapwich @jdalton Thanks! I was thinking the same thing. Doing testing now.

jdalton:
Careful `null` is `typeof` `"object"`

timmywil:
> Careful `null` is `typeof` `"object"`

In jQuery's case, that would either return or throw an error before `isArraylike` is called.

snapwich:
@jdalton the Angular isArrayLike I'm playing with does a null check first fortunately.  Although I seemed to have glazed over your cast to Object in your original comment, `"length" in Object(obj) && obj.length` seems to work just fine too and is shorter.

timmywil:
We had many tests failing already in IE due to the `hasOwn` + Nodelist problem. Thank you @jdalton for providing a good alternative.

test/unit/core.js
```diff
@@ -1217,6 +1217,25 @@ test("jQuery.each(Object,Function)", function() {
 	equal( i, 2, "Iteration over document.styleSheets" );
 });
 
+test( "JIT compilation does not interfere with length retrieval (gh-2145)", function() {
+	expect( 4 );
+
+	var i;
+
+	// Trigger JIT
```
gnarf: Could you be a bit more specific? This seems awkward.

caitp: "try to convince JSC to use one of its optimizing compilers by providing code which can be LICM'd into nothing", I think?

src/core.js
```diff
@@ -478,7 +478,13 @@ jQuery.each("Boolean Number String Function Array Date RegExp Object Error".spli
...
 
 function isArraylike( obj ) {
-	var length = obj.length,
+
+	// Support: iOS 8.2 (not reproducible in simulator)
+	// `in` check used to prevent JIT error (gh-2145)
+	// hasOwn isn't used here due to false negatives
+	// regarding Nodelist length in IE
+	// obj is casted to Object to avoid type errors (gh-2185)
+	var length = "length" in Object( obj ) && obj.length,
```
gibson042: I just double-checked, and removing this `Object` won't break anything. [`jQuery.makeArray`](https://github.com/timmywil/jquery/blob/2.1-stable/src/core.js#L353) already casts its input, and [`jQuery.each`](https://github.com/timmywil/jquery/blob/2.1-stable/src/core.js#L296) and [`jQuery.map`](https://github.com/timmywil/jquery/blob/2.1-stable/src/core.js#L408) already fail on strings: http://jsfiddle.net/j2qyduL9/ (which covers all uses of this function).

timmywil: Yea, good call. I assumed it worked with strings because I thought I had checked that a while back, but we don't have tests for it, so I'm not surprised it broke if it did work.

test/unit/core.js
timmywil: Do you mean more specific about how this triggers JIT compilation or which function is being JIT'd? Perhaps something like, "Trigger JIT compilation of `jQuery.each` – and therefore `isArraylike` – in iOS"?

timmywil:
I'll be merging this today.

mzgol:
This should land on `master` as well.

timmywil:
Done. 154166458284bcce7d6a86328b7fd13483232a1a

Issue #2204, opened 2015-04-13
Stealth-empty input to replaceWith leaves content in place


gibson042:
e.g., `$collection.replaceWith([""])` is a no-op: http://jsfiddle.net/w03Lertu/

markelog:
Is that a real use-case? Since in this docs i see - 

> May be an HTML string, DOM element, array of DOM elements

Since this would not be "array of DOM elements". What is byte-size and perf hit?

gibson042:
This is necessary for gh-1747, because intercepted HTML can be replaced with an empty string.

PR #2206, opened 2015-04-13
Manipulation: Detect sneaky no-content replaceWith input

gibson042:
Fixes gh-2204

```
   raw     gz Compared to master @ 892625b3c36eda6bb6ac226d15e0a158ba35cf21    
  +153    +39 dist/jquery.js                                                   
    +8     +7 dist/jquery.min.js
```

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -201,7 +201,7 @@ jQuery.extend({
 		return clone;
 	},
 
-	buildFragment: function( elems, context, scripts, selection ) {
+	buildFragment: function( elems, context, scripts, selection, ignored ) {
 		var elem, tmp, tag, wrap, contains, j,
 			fragment = context.createDocumentFragment(),
 			nodes = [],
@@ -257,9 +257,11 @@ jQuery.extend({
 		i = 0;
 		while ( (elem = nodes[ i++ ]) ) {
 
-			// #4087 - If origin and destination elements are the same, and this is
-			// that element, do not do anything
+			// Skip elements already in the context collection (trac-4087)
 			if ( selection && jQuery.inArray( elem, selection ) > -1 ) {
+				if ( ignored ) {
+					ignored.push( elem );
+				}
 				continue;
 			}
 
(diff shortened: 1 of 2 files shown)


dmethvin:
Wow, that was harder than it would have first appeared. Is there a reason to keep the list of elements versus just a flag?

gibson042:
I know. :disappointed: This is essentially implementing a second _output_ from `jQuery.buildFragment`—any flag would have to hang off of an existing input, but none of them looked palatable so I optimized on size.

dmethvin:
I looked at it for a while and couldn't figure out a better way.

src/manipulation.js
```diff
@@ -201,7 +201,7 @@ jQuery.extend({
 		return clone;
 	},
 
-	buildFragment: function( elems, context, scripts, selection ) {
+	buildFragment: function( elems, context, scripts, selection, ignored ) {
```
markelog: Should we make `buildFragment` private and change it signature, 5 arguments usually considered as bad API, even for helpers. One object as an argument?

mzgol: This is performance-critical code, I'm afraid wrapping it in an object might have a perf penalty (and it'll almost certainly be larger).

As for making it fully private, it seems like a good idea. We should do that with almost everything that's not documented IMO.

markelog: > This is performance-critical code, I'm afraid wrapping it in an object might have a perf penalty (and it'll almost certainly be larger).

Need proof for that statement

mzgol: The burden of proof is on the person that wants to introduce a change, i.e. here it's you. ;)

I just said it *might* introduce a perf penalty, I don't know.

dmethvin: I think the bigger issue is why `buildFragment` is so complex that it needs that many args, and why we had to add another one. I thought for sure there would be an easier way to deal with this crazy edge case, but couldn't find a reasonable way around it. It is a pretty hot path in our code.

markelog:
> I looked at it for a while and couldn't figure out a better way.

How about https://github.com/jquery/jquery/pull/2223

src/manipulation.js
gibson042: In short, because of self-manipulation. `.replaceWith` is the worst offender, but it's basically a problem any time there's a nonempty intersection between the input and the context collection, and exacerbated by the fact that `.domManip` lies in between the outer method and `buildFragment`.

That said, the `scripts` parameter is probably unnecessary and safe to remove when `buildFragment` is made private. In fact, that's an opportunity to look at more sweeping changes... why exactly do we pluck nodes from the DOM and into a fragment in `buildFragment` instead of `.domManip` anyway? It's an operation that `parseHTML` has to **undo**!

Issue #2094, opened 2015-02-16
Event handlers don't work like native ones


maciejmrozinski:
Event handler registered with addEventListener() can be instantly removed with removeEventListener(). on() and off() don't work like this. Example: http://jsfiddle.net/xf04www8/

dmethvin:
I think you may be relying on undefined behavior. Chrome and Firefox don't show `b` but IE11 does. I didn't try other browsers. I can find a [spec reference](http://www.w3.org/TR/DOM-Level-3-Events/#event-flow) that says the event *path* should not change during delivery, and jQuery honors that. I didn't find a discussion about what should happen if a listener is removed in the midst of dispatch. Can you find one?

maciejmrozinski:
I found something like this: http://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-EventTarget-removeEventListener

"This method allows the removal of event listeners from the event target. If an EventListener is removed from an EventTarget while it is processing an event, it will not be triggered by the current actions. EventListeners can never be invoked after being removed."

dmethvin:
Yes, it does seem like the spec intended that behavior, and there is [similar language on MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget.addEventListener). Note that none of the three browsers seem to call an event listener that is added during the process of handling an event. http://jsfiddle.net/xf04www8/1/  This seems oddly asymmetrical but there doesn't seem to be a special case laid out in `addEventListener` the way it is in `removeEventListener`.

This is the classic "modifying a collection while it's being enumerated" dilemma. 
jQuery has a single DOM event listener for all the jQuery handlers on an element. It takes a snapshot of those handlers that exist at the time that the event arrives. If it tried to use the live collection it would need to guard against the case where an already-called handler was removed or it could skip a handler that was still attached. 

This is a lot of mess for a really rare edge case. jQuery event handler semantics don't match DOM event listeners exactly. The fact that not even all browsers do this consistently tells you how rarely people depend on it. My suggested resolution is just to add a sentence to the [.on() docs](http://api.jquery.com/on/) clarifying that the handler list for an element is set when the event is first delivered.

As far as what you wanted to do here, would something like `event.stopImmediatePropagation()` be a solution?

maciejmrozinski:
The funny thing is that i'm not doing anything here with jQuery ;) I was just doing my own project and wanted to know how this case will work in native event handlers in browsers. Then i checked jQuery and found a mismatch.

[...]

PR #2207, opened 2015-04-13
Event: off() should delete all handlers from queue

kupriyanenko:
jQuery call deleted event handlers after handler removal. You can check native and jQuery behaviour in this example: http://jsfiddle.net/wwbne37b/.

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -388,7 +388,8 @@ jQuery.event = {
 
 			j = 0;
 			while ( (handleObj = matched.handlers[ j++ ]) &&
-				!event.isImmediatePropagationStopped() ) {
+				handlers.indexOf(handleObj) > -1 &&
+				!event.isImmediatePropagationStopped()) {
 
 				// Triggered event must either 1) have no namespace, or 2) have namespace(s)
 				// a subset or equal to those in the bound event (both can have no namespace).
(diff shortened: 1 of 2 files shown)


kupriyanenko:
Sorry, I forgot CLA. I added it and pushed commit with --force, but label doesn't changed.

dmethvin:
Is this in response to #2094?

The logic in this patch is wrong. it is dropping out of the dispatch loop if someone deletes one handler, even if there are more handlers on the list. The unit tests you added aren't complete enough to catch that.

From a performance standpoint, this is putting a lot of overhead on the dispatch code. If there are N handlers for an element we now traverse the handler list up to N-squared times inside the dispatcher via `.indexOf()`. Although we might be able to use that in master,we'd need to use `jQuery.inArray()` in compat, which would be even slower.

In #2094 we resolved to just document this because it was such a rare edge case. If there was a simple way to push the overhead to `.off()` where we'd rarely see it, we could fix it but I think that would involve exposing the dispatcher's handler list (or a deletion method). It doesn't seem worth the complexity to me but maybe there's a simpler solution I haven't seen.

As far as the CLA, be sure that your name and email on the CLA exactly match the one you used for your commits and it should work on your next commit.

timmywil:
Thank you for submitting a pull request, but given Dave's assessment, I'm inclined to close this.

Issue #2013, opened 2015-01-14
Progress doesn't fire after .then() for resolved promises


homm:
Suppose we have the Deferred object and try to set handlers on progress for both `.promise()` and `.then()` objects:

```js
var d = $.Deferred().notify();
d.promise().progress(function(){ console.log('promise') });
d.then().progress(function(){ console.log('then') });
```

This works as expected, both progress handlers will fire. But if we try to set progress handlers for resolved Deferred, the result will be different:

```js
var d = $.Deferred().notify().resolve(); // Look there, I'm resolved.
d.promise().progress(function(){ console.log('promise') });
d.then().progress(function(){ console.log('then') });
```

Only first handler will fire. The problem in [tuples order](https://github.com/jquery/jquery/blob/ab20d9d24f0a95df382c02e9ef3dcc4adb86e4f1/src/deferred.js#L12-L14). `Progress` tuple come last, and therefor, when [newDefer.notify()](https://github.com/jquery/jquery/blob/ab20d9d24f0a95df382c02e9ef3dcc4adb86e4f1/src/deferred.js#L39) is called, newDefer is already resolved and doesn't accept notifying.

This bug is similar to https://github.com/jquery/jquery/issues/1894, but slightly different.

dmethvin:
I think this a dup of #2010 but I'll let @gibson042 make the final call. 

I would say that this may not work in the near future with `.then` if we make it strictly Promise/A compatible, and you'd need to use Deferred-specific methods. Again, that hasn't been decided but as best practice it would be good to expect standard behavior from standard names.

gibson042:
Extremely similar, but not quite a duplicate... @homm, would you like to submit a PR? Both the code and new tests should be straightforward.

Issue #2010, opened 2015-01-13
pipe doesn't forward progress from already-resolved Deferred


gibson042:
http://jsfiddle.net/sp6a117y/

Analogous to #1894 in both [cause](https://github.com/jquery/jquery/blob/74ae5444832b2fb966768a97281d2ad8c088bc58/src/deferred.js#L33-L37) and solution. The new promise never sees progress at all because `.done` is called on it before `.progress`:
```js
jQuery.Deferred().resolve().pipe(function() {
    var dfd = jQuery.Deferred().notify( "a" );
    dfd.notify( "b" ).resolve( 1 );
    return dfd;
})
.progress( expectsInvocationWithB );
```

PR #2210, opened 2015-04-14
Deferred: Don't swallow progress of a resolved deferred in `then`

gibson042:
Fixes gh-2013
Fixes gh-2010

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -176,6 +176,18 @@ jQuery.extend({
 					}
 
 					return jQuery.Deferred(function( newDefer ) {
+						// progress_handlers.add( ... )
+						tuples[ 2 ][ 3 ].add(
+							resolve(
+								0,
+								newDefer,
+								jQuery.isFunction( onProgress ) ?
+									onProgress :
+									Identity,
+								newDefer.notifyWith
+							)
+						);
+
 						// fulfilled_handlers.add( ... )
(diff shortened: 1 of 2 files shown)


dmethvin:
Well that makes sense. LGTM.

gibson042:
Updated to fix #2010 as well.

src/deferred.js
```diff
@@ -240,29 +242,29 @@ jQuery.extend({
 
 					// rejected_callbacks.disable
 					// fulfilled_callbacks.disable
-					tuples[ i ^ 1 ][ 2 ].disable,
```
dmethvin: This looks like it was the only use of the xor operator in the whole codebase. I hope you're happy now.

src/deferred.js
```diff
@@ -16,12 +16,12 @@ jQuery.extend({
 	Deferred: function( func ) {
 		var tuples = [
 				// action, add listener, callbacks, .then handlers, final state
+				[ "notify", "progress", jQuery.Callbacks("memory"),
```
dmethvin: I don't know how much bigger the code would get if this was an array of props, but it would probably make the logic easier to follow. I can live with it tho.

src/deferred.js
gibson042: No regrets. :smiling_imp:

PR #2211, opened 2015-04-14
use getBoundingClientRect() for the hidden filter.

AsaAyers:
In my testing with Chrome (I didn't test other browsers) I found that all SVG elements seem to have an `offsetWidth` and `offsetHeight` of `0`, so they are all considered to be hidden. This solution isn't going to fix any browsers that don't support `getBoundingClientRect`.

In my project I've been able to monkey-patch my way around this:

```javascript
$.expr.filters.hidden = function(elem) {
	var rect = elem.getBoundingClientRect();
	return rect.height === 0 || rect.width === 0;
};
```

### DIFF
--- a/src/css/hiddenVisibleSelectors.js
+++ b/src/css/hiddenVisibleSelectors.js
@@ -1,9 +1,17 @@
 define([
 	"../core",
+	"../var/strundefined",
 	"../selector"
-], function( jQuery ) {
+], function( jQuery, strundefined ) {
 
 jQuery.expr.filters.hidden = function( elem ) {
+	// Support: BlackBerry 5, iOS 3 (original iPhone)
+	// If we don't have gBCR, just use 0,0 rather than error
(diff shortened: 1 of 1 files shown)


dmethvin:
In general we don't support operations on SVG elements. How did you come across this situation?

Also, the 2.1-stable branch isn't actively developed and new pull requests should be made against master. I think you'll find the code looks different there.

AsaAyers:
I have an app where I need to find every element that is [`$(':tabbable')`](https://api.jqueryui.com/tabbable-selector/) under a given element and flip `tabIndex` back and forth between `-1` and whatever it's set to. Part of the UI is drawn with an svg and contains a set of `<g tabIndex="0">` with click handlers, and keyup handlers to make them accessible to users of screen readers. Internally `:tabbable` uses `:visible`, which we see here is implemented by inverting `:hidden`.

I have submitted a new PR against `master` at https://github.com/jquery/jquery/pull/2212

Issue #2199, opened 2015-04-11
jQuery Class Operations on SVG DOM Attributes


artemist:
First of all, jQuery is hugely helpful and the whole community thanks to you all. It's SO helpful, in fact, that spaces that it lacks end up being a huge pain for developers.

Working with jQuery and detecting SVG hasClass, addClass, or any of the like on groups, path, or any other attributes is a huge pain. With the growing community and support for SVG, are there any plans in jQuery's roadmap to improve this workflow? 

Thanks again.

SaraSoueidan:
I second this request.

rwaldron:
Thanks for the report. More SVG support has long been listed on the [Won't Fix](http://contribute.jquery.org/wont-fix/) document, along with links to relevant discussions that elaborate on that position.

artemist:
I'm a little disappointed that this is the position taken, as SVG is gaining widespread interest and use, and though it isn't the HTML DOM, SVG attributes are still accessible in the DOM via vanilla js. Most of the tickets shown here are from 2+ years ago.

sergzbx:
2 years later it might be time to reconsider this @jeresig. Thanks.

dmethvin:
What exactly would you like to see added? Have you looked at libraries like [Raphael](http://raphaeljs.com/) that are designed to support SVG?

Keep in mind that SVG is not HTML, despite the fact that they look similar when represented as strings.

artemist:
Detecting hasClass, or adding/removing classes would be a hugely helpful UX interaction boon. Yes, I know Raphael and Snap.svg. But let's say you're using a different animation library. To then add Snap just to look for classes or add performant click events (particularly if you're already loading jQuery) feels painful if not redundant. 

I realize this request might be too large and out of scope, and yes, SVG isn't HTML, but if there's a possibility jQuery could add this feature, it's worth a discussion. Everyday usage of SVG on the web is fast-growing. Adding this kind of support has the potential for a large payoff for the community.

dmethvin:
The title of this request ("better SVG support") is still super vague, it would be pretty hard to make a pull request against. When you say "too large and out of scope" it sounds like you are asking for SVG DOM to be supported on all jQuery methods. Or are you just asking for class operations?

artemist:
[...]

PR #2212, opened 2015-04-15
use getBoundingClientRect() for the hidden filter

AsaAyers:
In my testing with Chrome (I didn't test other browsers) I found that all SVG elements seem to have an offsetWidth and offsetHeight of 0, so they are all considered to be hidden.

### DIFF
--- a/src/css/hiddenVisibleSelectors.js
+++ b/src/css/hiddenVisibleSelectors.js
@@ -6,7 +6,8 @@ define([
 jQuery.expr.filters.hidden = function( elem ) {
 	// Use OR instead of AND as the element is not visible if either is true
 	// See tickets #10406 and #13132
-	return !elem.offsetWidth || !elem.offsetHeight;
+	var rect = elem.getBoundingClientRect();
+	return !rect.height || !rect.width;
 };
 jQuery.expr.filters.visible = function( elem ) {
 	return !jQuery.expr.filters.hidden( elem );


dmethvin:
Any request with the letters "SVG" in it starts out with at least a quarter-can of worms, and it doesn't take long until the can is full. See #2199 for some discussion. SVG DOM is not HTML DOM and jQuery is primarily an HTML DOM library.

I also seem to recall that `getBoundingClientRect` is slower than the `offset*` properties, there was a benchmark in one ticket but I can't find it right now. 

In #2211 you said you are using this in a `:tabbable` query. What do you do with the SVG elements when you get them? Are you using jQuery to do it?

AsaAyers:
I need to restrict `tabIndex` to children of a specific element, so I first find `:tabbable` over the whole document, backup the `tabIndex` and `.attr('tabindex', -1)` to disable tabbing over the whole document. Then I find `:tabbable` for the one section I do need tabbing to work in and restore `tabindex` the same way.

If you think this is going to cause problems for others you're welcome to close it, I won't be offended. The problem is fixed in my project. I just think it's my responsibility as a user capable of contributing, to provide my issue and fix whenever I can.

shimonenator:
@AsaAyers Please review discussion in #2227, as there are some performance issues to be addressed in this issue/solution.

AsaAyers:
This is not worth the effort to contribute any more.

Issue #1766, opened 2014-10-21
contents and children selectors can return elements out of order


mzgol:
Originally reported by npben at: http://bugs.jquery.com/ticket/14960
<p>
This problem exists in version 1.11.0. 
</p>
<p>
I looked into source since the children and content selectors are just hardcoded not go into the logic that verifies uniqueness, it is probably just a situation that wasn't considered.
</p>
<p>
Examples are <a class="ext-link" href="http://jsfiddle.net/sfMr2/5/" rel="nofollow"><span class="icon">&nbsp;</span>here for children()</a> and <a class="ext-link" href="http://jsfiddle.net/sfMr2/6/" rel="nofollow"><span class="icon">&nbsp;</span>here for contents()</a>.
</p>
<p>
Removing these two tags from the guaranteedunique array fixes the issue, but I didn't make a patch because I'm not familiar with the source and there may be a better way to handle it.
</p>
<p>
Workaround: calling $().add(unorderedobject) causes the uniqueness check to sort the elements.
</p><p>Issue reported for jQuery 1.11.0</p>

gibson042:
As discussed in the [meeting](http://irc.jquery.org/%23jquery-meeting/default_%23jquery-meeting_20150420.log.html#t09:17:23), sorting a guaranteed-unique collection doesn't seem to be of high enough value to justify degrading performance without a real-world use case.

PR #2213, opened 2015-04-15
Traversing: Don't exempt .children and .contents from sorting

gibson042:
Fixes gh-1766

```
   raw     gz Compared to master @ 892625b3c36eda6bb6ac226d15e0a158ba35cf21    
   -20     -6 dist/jquery.js                                                   
   -23    -28 dist/jquery.min.js
```

### DIFF
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -7,14 +7,7 @@ define([
 	"./selector"
 ], function( jQuery, indexOf, rneedsContext ) {
 
-var rparentsprev = /^(?:parents|prev(?:Until|All))/,
-	// Methods guaranteed to produce a unique set when starting from a unique set
-	guaranteedUnique = {
-		children: true,
-		contents: true,
-		next: true,
-		prev: true
-	};
+var rparentsprev = /^(?:parents|prev(?:Until|All))/;
 
 jQuery.extend({
 	dir: function( elem, dir, until ) {
@@ -180,8 +173,8 @@ jQuery.each({
 		}
 
 		if ( this.length > 1 ) {
-			// Remove duplicates
-			if ( !guaranteedUnique[ name ] ) {
+			// Sorting (which implicitly de-dupes) is required for non-adjacent traversals
+			if ( name !== "prev" && name !== "next" ) {
 				jQuery.unique( matched );
 			}
 
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -7,8 +7,6 @@ define([
 	"./selector"
 ], function( jQuery, indexOf, rneedsContext ) {
 
-var rparentsprev = /^(?:parents|prev(?:Until|All))/;
-
 jQuery.extend({
 	dir: function( elem, dir, until ) {
 		var matched = [],
@@ -161,10 +159,18 @@ jQuery.each({
 		return elem.contentDocument || jQuery.merge( [], elem.childNodes );
 	}
 }, function( name, fn ) {
+	var // Sorting (which implicitly de-dupes) is required for non-adjacent traversals
+		sort = name !== "prev" && name !== "next",
+
+		// Reverse order for prev-derivatives and parents*
+		reverse = name !== "prev" && name.slice( 0, 4 ) === "prev" ||
+			name.slice( 0, 7 ) === "parents";
+
 	jQuery.fn[ name ] = function( until, selector ) {
 		var matched = jQuery.map( this, fn, until );
 
-		if ( name.slice( -5 ) !== "Until" ) {
+		// Only *Until methods support two parameters
+		if ( fn.length !== 3 ) {
 			selector = until;
 		}
 
@@ -173,13 +179,10 @@ jQuery.each({
 		}
 
 		if ( this.length > 1 ) {
-			// Sorting (which implicitly de-dupes) is required for non-adjacent traversals
-			if ( name !== "prev" && name !== "next" ) {
+			if ( sort ) {
 				jQuery.unique( matched );
 			}
-
-			// Reverse order for parents* and prev-derivatives
-			if ( rparentsprev.test( name ) ) {
+			if ( reverse ) {
 				matched.reverse();
 			}
 		}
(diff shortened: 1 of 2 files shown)


src/traversing.js
```diff
@@ -168,10 +159,18 @@ jQuery.each({
 		return elem.contentDocument || jQuery.merge( [], elem.childNodes );
 	}
 }, function( name, fn ) {
+	var // Sorting (which implicitly de-dupes) is required for non-adjacent traversals
```
dmethvin: Should the comment be on the line above?

src/traversing.js
```diff
@@ -168,10 +159,18 @@ jQuery.each({
...
+		sort = name !== "prev" && name !== "next",
```
dmethvin: Can we just do these inside the `if` below to avoid doing them if `this.length === 1`? We wouldn't even need the variables that way.

test/unit/traversing.js
```diff
@@ -508,9 +508,17 @@ test("siblings([String]) - jQuery only", function() {
...
-	deepEqual( jQuery("#foo").children().get(), q("sndp", "en", "sap"), "Check for children" );
-	deepEqual( jQuery("#foo").children("#en, #sap").get(), q("en", "sap"), "Check for multiple filters" );
+	expect( 3 );
+
+	deepEqual( jQuery( "#foo" ).children().get(), q( "sndp", "en", "sap" ),
+		"Check for children" );
+	deepEqual( jQuery( "#foo" ).children( "#en, #sap" ).get(), q( "en", "sap" ),
+		"Check for multiple filters" );
+	deepEqual(
+		jQuery( "#foo, #en" ).children().get(),
```
dmethvin: "...and we would have gotten away with it too, if it wasn't for that meddling comma operator!"

src/traversing.js
gibson042: I'm not sure. We have few examples of commenting the first var in a list of more than one, though only one of them [uses this style](https://github.com/jquery/jquery/blob/4cbf02df84dbcaa44b75a64ed832f7dbff2231dd/src/callbacks.js#L45). I'll change it if that's the consensus.

src/traversing.js
gibson042: Yes, but this way is smaller. And also imperceptibly faster, since the variables are defined along with the methods, rather than every time the methods run.

test/unit/traversing.js
gibson042: Jinkies! It was a child element under the mask the whole time! :squirrel:

src/traversing.js
markelog: Grey area, JSCS 1.13 wouldn't yell about it fyi

src/traversing.js
markelog: #180 is all about speed, i wonder if we wouldn't have this optimization, would we accept this pull for `next` and `prev` only now? Is it worth it?

dmethvin: Good point @markelog. I would guess that `.next()` and `.prev()` are rarely run on multiple elements. If that is really common, perhaps we can skip the `sort` variable and check for 'matched.length > 1` instead of `this.length > 1` below?

markelog: Sounds like a 
```js
$( "a lot of elements" ).(prev | next)()
```
is rare and weird use-case

scottgonzalez: Using these methods on a set of more than one is not uncommon.

As for large sets, I'm not sure how common it is, but cases like selecting all cells in the first column of a table and doing something, then selecting all cells in the next column via `.next()` seems sane (from a user's perspective) and could result in a large set.

The real question, though, is what performance difference are we talking about, and for what percentage of users?

dmethvin: Also, things may have changed in 4 years since you reported [trac-7964](http://bugs.jquery.com/ticket/7964). Do you recall if that ticket was based on some real-life perf issue? At this point we've reduced the optimization to just those two methods so it might be worth simplifying the code and seeing who complains.

scottgonzalez: It was based on real world usage of jQuery UI. The graphs from http://blog.jquery.com/2011/01/31/jquery-15-released/ (under Adjacent Traversal Performance) should still be relevant today, though the important data (IE performance improvement) is hard to see because of the crazy performance improvement in Chrome. The problem was that old IE (where we have to take a slow path in sorting, if I remember correctly) is noticeably slow with a few hundred elements.

Granted the specific scenario that prompted this was terrible UX (500 results in an autocomplete menu), but I do recall this impacting quite a few different scenarios that we had. However, this was for `.children()`, not `.next()` or `.prev()`. While we do have cases where we do `multipleElements.next()`, I'm not sure the numbers ever get very high. I'd have to look through all our uses of these methods to find out.

dmethvin: Since `.children()` is getting a hit here already, it sounds like the UI performance issue is going to return regardless of what we do to `.next()` or `.prev()`. 

We are ignoring hints about node order here and I suspect we could do an insertion or merge sort with de-duping more efficiently than collecting the entire set of partially-ordered nodes and passing them on to `unique()`. But that would require more code, probably a lot more.

gibson042: Sorting makes a _huge_ difference, even on relatively small input sets: http://jsperf.com/jquery-fn-next-sorting

To be honest, I'm upset about removing this exemption, and it may actually be worthwhile to squeeze something back in (although unfortunately, the operations required to detect a need for sorting are basically identical to those required to perform a sort). But I think that `$largeCollection.children()` is probably more rare than `$largeCollection.next()`, and would like to leave this PR as-is.

markelog: I think we all upset about this, https://github.com/jquery/jquery/pull/180 was a big win, right know, i'm not totally sure that https://github.com/jquery/jquery/issues/1766 is necessary, owner of this ticket found it by looking at the code, i'm not sure if it's real-life problem, whereas optimization was landed four years ago and know one notice the issue, but everybody notice the speed improvement.

Maybe we just document and forget about it?

gibson042:
As discussed in the meeting, sorting a guaranteed-unique collection doesn't seem to be of high enough value to justify degrading performance without a real-world use case.

PR #2217, opened 2015-04-16
Core: jQuery.grep throws exception when elem passed is undefined

gs-akhan:
.length access to undefined throws exception. Instead made length as zero so loops skips automatically.


gibson042:
Thank you for the pull request, but our practice is to assume valid input. `jQuery.grep` requires its first argument to be an array: http://api.jquery.com/jQuery.grep/

gs-akhan:
Ohh That's ok.. 
Thanks

PR #2218, opened 2015-04-16
Effects: easing._default

Mr21:
More info at the issue: #2219.

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -333,7 +333,10 @@ function Animation( elem, properties, options ) {
 		animation = deferred.promise({
 			elem: elem,
 			props: jQuery.extend( {}, properties ),
-			opts: jQuery.extend( true, { specialEasing: {} }, options ),
+			opts: jQuery.extend( true, {
+				specialEasing: {},
+				easing: jQuery.easing._default
+			}, options ),
 			originalProperties: properties,
 			originalOptions: options,
 			startTime: fxNow || createFxNow(),
--- a/src/effects.js
+++ b/src/effects.js
@@ -333,10 +333,7 @@ function Animation( elem, properties, options ) {
 		animation = deferred.promise({
 			elem: elem,
 			props: jQuery.extend( {}, properties ),
-			opts: jQuery.extend( true, {
-				specialEasing: {},
-				easing: jQuery.easing._default
-			}, options ),
+			opts: jQuery.extend( true, { specialEasing: {} }, options ),
 			originalProperties: properties,
 			originalOptions: options,
 			startTime: fxNow || createFxNow(),
--- a/test/unit/effects.js
+++ b/test/unit/effects.js
@@ -1178,45 +1178,45 @@ test("animate with per-property easing", function(){
 test("animate with CSS shorthand properties", function(){
 	expect(11);
 
-	var _default_count = 0,
-		_special_count = 0,
+	var default_count = 0,
+		special_count = 0,
 		propsBasic = { "padding": "10 20 30" },
-		propsSpecial = { "padding": [ "1 2 3", "_special" ] };
+		propsSpecial = { "padding": [ "1 2 3", "special" ] };
 
-	jQuery.easing._default = function(p) {
+	jQuery.easing.default = function(p) {
 		if ( p >= 1 ) {
-			_default_count++;
+			default_count++;
 		}
 		return p;
 	};
 
-	jQuery.easing._special = function(p) {
+	jQuery.easing.special = function(p) {
 		if ( p >= 1 ) {
-			_special_count++;
+			special_count++;
 		}
 		return p;
 	};
 
 	jQuery("#foo")
-		.animate( propsBasic, 200, "_default", function() {
+		.animate( propsBasic, 200, "default", function() {
 			equal( this.style.paddingTop, "10px", "padding-top was animated" );
 			equal( this.style.paddingLeft, "20px", "padding-left was animated" );
 			equal( this.style.paddingRight, "20px", "padding-right was animated" );
 			equal( this.style.paddingBottom, "30px", "padding-bottom was animated" );
-			equal( _default_count, 4, "per-animation default easing called for each property" );
-			_default_count = 0;
+			equal( default_count, 4, "per-animation default easing called for each property" );
+			default_count = 0;
 		})
-		.animate( propsSpecial, 200, "_default", function() {
+		.animate( propsSpecial, 200, "default", function() {
 			equal( this.style.paddingTop, "1px", "padding-top was animated again" );
 			equal( this.style.paddingLeft, "2px", "padding-left was animated again" );
 			equal( this.style.paddingRight, "2px", "padding-right was animated again" );
 			equal( this.style.paddingBottom, "3px", "padding-bottom was animated again" );
-			equal( _default_count, 0, "per-animation default easing not called" );
-			equal( _special_count, 4, "special easing called for each property" );
+			equal( default_count, 0, "per-animation default easing not called" );
+			equal( special_count, 4, "special easing called for each property" );
 
 			jQuery(this).css("padding", "0");
-			delete jQuery.easing._default;
-			delete jQuery.easing._special;
+			delete jQuery.easing.default;
+			delete jQuery.easing.special;
 		});
 		this.clock.tick( 400 );
 });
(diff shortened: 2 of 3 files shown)


src/effects.js
```diff
@@ -333,7 +333,10 @@ function Animation( elem, properties, options ) {
 		animation = deferred.promise({
 			elem: elem,
 			props: jQuery.extend( {}, properties ),
-			opts: jQuery.extend( true, { specialEasing: {} }, options ),
+			opts: jQuery.extend( true, {
+				specialEasing: {},
+				easing: jQuery.easing._default
+			}, options ),
```
gibson042: I think I'd prefer a default easing to stay at the Tween level, rather than here in Animate. And since @gnarf is back in this headspace for a bit, we can get a second opinion.

gibson042:
I like the direction here, but please create an issue so we can track it. Also, the CLA failure will need to be fixed by amending this commit—and all future ones—to use your full name (which we require and are now verifying).

Mr21:
hi gibson42 :) in my `.gitconfig` I have exactly:
```
[user]
	email = thomastortorini@gmail.com
	name = Mr21
[push]
	default = simple
```
I signed the CLA again yesterday with my nickname and my email. I don't understand what I have to do exactly. Do I have to write my real full name?

mzgol:
> Do I have to write my real full name?

Yes.

Mr21:
I didn't respect exactly the principle of the atomic commit, so I redo my commit.
Now, only 3 lines are changed.

Actually, If we don't precise the `easing` the `anim.originalOptions.easing` will be `undefined` BUT
the `anim.opts.easing` will be the default value: `"swing"`.

The duration don't respect the same idea. If I don't specify a `duration` the
`anim.opts.duration` and `anim.originalOptions.duration` will be equals to `jQuery.fx.speeds._default` (400). And I don't understand why there is also `anim.duration` directly.

src/effects.js
Mr21: I write a resume of the situation in another message.
But I like the way we put the default value **and after** the user rewrite them with the `.extend`.
I think (IMHO) we should do the same thing for the `duration` value.
And also for the `complete` callback with `jQuery.noop` maybe (it's only a suggestion).

Mr21:
Hi,

I read in your [meeting](http://irc.jquery.org/%23jquery-meeting/default_%23jquery-meeting_20150420.log.html#t09:17:23) nobody share my code design :cry: so I moved the `this.easing = easing || jQuery.easing._default;` in the **Tween.js** file.

But should I move the code of the `duration || _default` in the Tween file also?

Mr21:
I add two new tests for that but I have to change the variable's name of another test because it was using (and delete) the new `jQuery.easing._default`.

After you say it's better to see that in the Tween file I changed, but I didn't see the test `animate with per-property easing` (of the `test/unit/effects.js` file) failed.
It's concerne that:

``` javascript
jQuery(data).animate(
	{
		a: [ 100, "_test1" ],
		b: [ 100, "_test2" ],
		c: 100
	},
	400,
	"default_easing"
);
```

The property `c` should take the `"default_easing"` but this not work anymore with my 2nd commit.
But my first commit (when I change the `effects.js` file) work fine with this.

gibson042:
I'm having a hard time following this, so let me back up a bit and present a call stack as it appears in master:

Call | State | Notes
------------ | ------------- | -------------
`$elements.animate( { left: 0 }, null )` | `prop = { left: 0 }`<br>`speed = null`<br>`easing = undefined`<br>`callback = undefined`
► `optall = jQuery.speed( speed, easing, callback )` | `optall = { duration: 400, easing: undefined, queue: "fx", old: false, complete: function(){ … } }`
… (`doAnimation`) | `this = element0`
► `Animation( this, jQuery.extend( {}, prop ), optall )` | `elem = element0`<br>`properties = { left: 0 }`<br>`options = optall` | options passed to `Animation` are _not_ original, but have already been created/modified by `jQuery.speed`
► ► `animation = deferred.promise({ … })` | `animation = { originalProperties: properties, originalOptions: optall, duration: 400, props: { left: 0 }, opts: { specialEasing: {}, duration: 400, easing: undefined, … }, … }`<br>`props = animation.props` | …so likewise for `animation.originalOptions`. `animation.opts`, however, is a copy that can be modified independently (e.g., by `defaultPrefilter` or by this proposal)
► ► `propFilter( props, animation.opts.specialEasing )` |
► ► `animationPrefilters[ index ].call( animation, elem, props, animation.opts )` | 
► ► ► … (`defaultPrefilter`) | `this = animation`<br>`elem = element0`<br>`opts = animation.opts`
► ► `jQuery.map( props, createTween, animation )` |
► ► ► … (`createTween`) | `value = 0`<br>`prop = "left"`
► ► ► ► `collection[ index ].call( animation, prop, value )` |
► ► ► ► ► … (`tweeners[ "*" ]`) | `this = animation`<br>
► ► ► ► ► ► `tween = this.createTween( prop, value )` | `end = 0`
► ► ► ► ► ► ► `tween = jQuery.Tween( elem, animation.opts, prop, end, animation.opts.specialEasing[ prop ] || animation.opts.easing )` | 
► ► ► ► ► ► ► ► … (`jQuery.Tween.prototype.init`) | `tween.elem = element0`<br>`tween.prop = "left"`<br>`tween.easing = "swing"`<br>`tween.options = animation.opts`<br>`tween.start = this.now = 100`<br>`tween.end = 0`<br>`tween.unit = "px"` | tweens _require_ defined easing, so initialization applies a default if there's no explicit (e.g., animation-level) setting
► ► `jQuery.fx.timer(…)` |
► ► ► `jQuery.fx.start()` |

Having gone through all of that, I think I understand what you're after: a guarantee of animation-level `opts.easing`. There's a potential problem from `specialEasing` (since properties are allowed to ease independently), but that is admittedly a rare case, and would still tolerate the [animation-level default](https://github.com/Mr21/jquery/commit/df07ff709228726436b218592985d9e9e0e0ebca#diff-e24cc50c91c9bacee64ace3e2107bd8cR336) anyway. Note, however, that http://jsfiddle.net/Mr21/wh5sdgzu/ will not tolerate such variance because it looks for easing in `anim.opts` instead of a per-prop `anim.opts.specialEasing[<prop>]` or `anim.tweens[<N>]`.

So the bottom line here is that I would be fine with use of `jQuery.easing._default` in both `Animation` _and_ `Tween.prototype.init`. Tests for the former would assert that `$collection.animate( props )` creates an animation with `anim.opts.easing` matching a custom `jQuery.easing._default`, and tests for the latter would do the same against `anim.tweens[<N>].easing` (and not matching for `specialEasing` tweens). Does that make sense?

test/unit/effects.js
```diff
@@ -1178,45 +1178,45 @@ test("animate with per-property easing", function(){
...
-	var _default_count = 0,
-		_special_count = 0,
+	var default_count = 0,
+		special_count = 0,
 		propsBasic = { "padding": "10 20 30" },
-		propsSpecial = { "padding": [ "1 2 3", "_special" ] };
+		propsSpecial = { "padding": [ "1 2 3", "special" ] };
 
-	jQuery.easing._default = function(p) {
+	jQuery.easing.default = function(p) {
```
gibson042: Let's name this to better match its purpose, something like `jQuery.easing.animationScope` (and the other one to e.g. `jQuery.easing.propertyScope`).

Mr21: Yeah right, it's done.

Mr21:
@gibson042 nice summary!

> So the bottom line here is that I would be fine with use of jQuery.easing._default in both Animation and Tween.prototype.init.

I didn't understand exactly where do you think I have to put the `|| jQuery.easing._default`.
Because if I put it in the `opts : $.extend( ... )` (like my first commit) I don't have to repeat it in the Tween.init part. I'm confused :confused:

gibson042:
`jQuery.easing._default` will be referenced twice. You need it in `Tween.prototype.init` (as in [f4a3c787](https://github.com/jquery/jquery/commit/f4a3c787bd12bd0957a58327a7bf79916b2cd9bb#diff-ec4583af7089881c2eeab5e54a300ef6)) because tweens can be created outside of `Animation`, and you want it in `Animation` (as in [df07ff70](https://github.com/jquery/jquery/commit/df07ff709228726436b218592985d9e9e0e0ebca#diff-e24cc50c91c9bacee64ace3e2107bd8c)) so animation objects always have an "easing" property.

Mr21:
> because tweens can be created outside of Animation
Ahhh okay, I didn't know this. So I commit again.

In Qunit I read (on my test):

    3. No timers are still running
    Expected: 0
    Result: 1

Where is this test? And what I have to do to passe it?

gibson042:
That looks like one of our beforeEach/afterEach assertions, and indicative of an async issue (e.g., `jQuery.fx.stop` won't be called until the next tick).

Mr21:
Okay, I add `this.clock.tick( 25 );` and it's work fine.

PR #2223, opened 2015-04-20
Manipulation: Detect sneaky no-content replaceWith input

markelog:
Alternative to #2206 

-3 bytes to gzipped size

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -460,7 +460,11 @@ jQuery.fn.extend({
 		});
 
 		// Force removal if there was no new content (e.g., from empty arguments)
-		return arg && (arg.length || arg.nodeType) ? this : this.remove();
+		if ( !arg || !arg[ 0 ] || typeof arg === "string" ) {
+			return this.remove();
+		}
+
+		return this;
 	},
 
 	detach: function( selector ) {
--- a/test/unit/manipulation.js
+++ b/test/unit/manipulation.js
@@ -1279,15 +1279,19 @@ test( "replaceWith(string) for more than one element", function() {
 	equal(jQuery("#foo p").length, 0, "verify that all the three original element have been replaced");
 });
 
-test( "Empty replaceWith (#13401; #13596)", 8, function() {
+test( "Empty replaceWith (trac-13401; trac-13596; gh-2204)", function() {
+
+	expect( 10 );
+
 	var $el = jQuery( "<div/>" ),
 		tests = {
 			"empty string": "",
 			"empty array": [],
+			"array of empty string": [ "" ],
 			"empty collection": jQuery( "#nonexistent" ),
 
-       // in case of jQuery(...).replaceWith();
-			"empty undefined": undefined
+			// in case of jQuery(...).replaceWith();
+			"undefined": undefined
 		};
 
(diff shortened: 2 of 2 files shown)


src/manipulation.js
```diff
@@ -460,7 +460,11 @@ jQuery.fn.extend({
 		});
 
 		// Force removal if there was no new content (e.g., from empty arguments)
-		return arg && (arg.length || arg.nodeType) ? this : this.remove();
+		if ( !arg || !arg[ 0 ] || typeof arg === "string" ) {
```
markelog: Or

```js
if ( arg && (arg[ 0 ] || arg.nodeType) && typeof arg !== "string" ) {
	return this;
}

// Force removal if there was no new content (e.g., from empty arguments)
return this.remove();
```

Which i actually like better, since `remove` would not be called as much

gibson042:
I believe this will still fail on `$el.replaceWith( function( old ) { return textToBePrefiltered; } )`.

markelog:
@gibson042 could you post an jsfiddle?

markelog:
Perhaps i didn't get your example, since this -

```js
htmlPrefilter = jQuery.htmlPrefilter

jQuery.htmlPrefilter = function( html ) {
       // Remove <script> and <del> elements
       return htmlPrefilter.apply( this, arguments )
           .replace( /<(script|del)(?=[\s>])[\w\W]*?<\/\1\s*>/ig, "" );
   };

 $("div").replaceWith(function() {
    return "<script>alert(1)<\/script>";
 });
```

with conjunction with https://github.com/jquery/jquery/pull/2203 works for me

gibson042:
Your example works only because the replacer function returns non-empty. I put together a jsFiddle, and actually uncovered (and fixed) a bug in #2206: [http://jsfiddle.net/cdhxqwk4/](http://jsfiddle.net/cdhxqwk4/). Observe how **this** PR erroneously removes *all* the original elements.

I just want to stop being so damn clever with `.replaceWith` and get to a point where it does exactly what it says, which I think is achieved by #2206.

markelog:
@gibson042 this is a lot of code, could you provide simplified example? Just like we ask other people to do?

> I just want to stop being so damn clever with .replaceWith and get to a point where it does exactly what it says, which I think is achieved by #2206.

Yeah, your way is more robust and this logic is vague at best, but #2206 complicates a lot to fix a very-very edge case.

markelog:
Just a gist or PR with tests would do

gibson042:
Sorry; I thought the [fiddle](http://jsfiddle.net/cdhxqwk4/) was a clear comparison between the two PRs that demonstrated errors in this one. Straightforward tests are already at https://github.com/jquery/jquery/pull/2206/files#diff-328a9410239da61336e3662fdbfe9bf6

As for the rest, `.replaceWith` is **full** of sharp edges, but I believe that #2206 actually _simplifies_ the logic now, albeit by adding a new parameter to `buildFragment` and `domManip`. Trying to guess at possibly-dynamic `arguments` is just begging for problems, and requiring a `this.remove()` to clean up after the `domManip` callback is indicative of bad implementation.

markelog:
Okay, although those are really weird use-cases, i fully agree with your point.

PR #2232, opened 2015-04-26
Tests: Increase QUnit timeout

mzgol:
Android 2.3 is very slow & times out a lot in async tests, they have to be
restarted multiple times to settle. Long test execution is not a huge problem
as Android 2.3 is tested only periodically during the night, unstable tests
are a bigger problem. This might mitigate that.

In a regular scenario almost all tests should pass so increasing the timeout
for all browsers shouldn't have a huge impact on overall test time.

### DIFF
--- a/test/data/testrunner.js
+++ b/test/data/testrunner.js
@@ -15,7 +15,7 @@ var oldCacheLength = 0,
 
 // Max time for stop() and asyncTest() until it aborts test
 // and start()'s the next test.
-QUnit.config.testTimeout = 2e4; // 20 seconds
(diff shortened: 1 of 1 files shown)


mzgol:
I'm not sure if that's enough. I sometimes see timeouts in all suites from offset to traversing so it might be hitting some BrowserStack timeouts as well.

mzgol:
Per the meeting: it might be better to increase the timeout globally, especially that it sometimes hits Android 2.3 & IE as well.

mzgol:
PR updated.

Issue #2181, opened 2015-03-30
Test all factory use cases from intro.js


mzgol:
We have a lot of logic in [intro.js](https://github.com/jquery/jquery/blob/master/src/intro.js) and we only test two: the browser use case and the most common Node+jsdom. We should cover all scenarios so that freely changing this code actually breaks some tests.

mzgol:
We should do it before trying to fix #2177 so that it's easier to add a new test.

PR #2234, opened 2015-04-26
Core: Test all factory use cases from intro.js

mzgol:
Core: Test all factory use cases from intro.js

There is a lot of logic in intro.js; now we test four cases:
1. (implicitly, via QUnit tests) A real browser with window being the global
2. Browserify where there are both global & window variables.
3. Node with jsdom where window is passed manually to the jQuery factory.
4. Pure Node with incorrect window passed; jQuery should throw then.

Previously the second & fourth case was not tested and the third was tested
in a way that interfered with the main test environment.

We now also test if in the Browserify case we're not creating a jQuery global
by default.

Fixes gh-2181


dmethvin:
Great idea! We don't change this code very often but it seems every time we do it breaks.

mzgol:
Updated the commit message to include the 4th point.

build/tasks/lib/spawn_test.js
```diff
@@ -0,0 +1,16 @@
+/* jshint node: true */
+
+"use strict";
+
+// Run Node with provided parameters: the first one being the Grunt
+// done function and latter ones being files to be tested.
+// See the comment in ../node_smoke_tests.js for more information.
+module.exports = function spawnTest( done ) {
+	var testPaths = [].slice.call( arguments, 1),
```
gibson042: Missing space before `)`.

mzgol: Fixed here & in other couple of places.

test/node_smoke_tests/lib/ensure_jquery.js
```diff
@@ -0,0 +1,11 @@
+/* jshint node: true */
+
+"use strict";
+
+// Check if the object we got is the jQuery object by invoking a basic API.
+module.exports = function ensureJQuery( jQuery ) {
+	if ( typeof jQuery.trim !== "function" || jQuery.trim( " ab " ) !== "ab" ) {
```
gibson042: I'd like to see something simpler here for now; maybe `/^jQuery/.test( jQuery.expando )`?

test/promises_aplus_adapter.js
```diff
@@ -13,11 +14,9 @@ require( "jsdom" ).env( "", function ( errors, window ) {
 		var deferred = jQuery.Deferred();
 
 		return {
-			get promise() {
-				return deferred.promise();
-			},
+			promise: deferred.promise(),
```
gibson042: Is this acceptable? I thought something about promises-aplus-tests might have required a getter here.

gibson042:
I like the new structure here.

test/node_smoke_tests/lib/ensure_jquery.js
mzgol: Works for me.

test/promises_aplus_adapter.js
mzgol: It doesn't require a getter, it requires a property. Using a getter works as it de facto changes a method to a property but this works as well; tests pass (this is tested on Travis so it wouldn't be green if it failed in any tested Node/io.js).

mzgol: I mainly changed it here since I was touching this file as well and having a regular property here means we might be able to run at least a subset of tests in IE8 (not relevant for this branch but still).

test/node_smoke_tests/lib/ensure_jquery.js
mzgol: New version pushed.

mzgol:
@gibson042 all good now?

test/promises_aplus_adapter.js
gibson042: > tests pass (this is tested on Travis so it wouldn't be green if it failed in any tested Node/io.js)

It wouldn't manifest as failed tests, but as functionality that is now no longer being tested. Maybe I'm misremembering, though... this does look fine, so :+1:.

> I mainly changed it here since I was touching this file as well and having a regular property here means we might be able to run at least a subset of tests in IE8 (not relevant for this branch but still).

Understood and agreed.

gibson042:
LGTM.

mzgol:
I wonder what to do on `compat`. The [compat intro.js](https://github.com/jquery/jquery/blob/compat/src/intro.js) is almost identical to [the master one](https://github.com/jquery/jquery/blob/master/src/intro.js) but we don't support Node.js on `compat` (why do we have this window-passing logic there, then? @timmywil?). We do support Browserify, though so it'd be nice to test it.

timmywil:
> why do we have this window-passing logic there, then?

It's just so we didn't have to keep track of two intros (although, I'm having a flashback about some other environment, mozilla sdk or something?). I agree it would be nice to have Browserify tests in compat.

mzgol:
> It's just so we didn't have to keep track of two intros

Makes sense.

> (although, I'm having a flashback about some other environment, mozilla sdk or something?).

IMO we've said in various blog posts that `compat` only supports browsers; Node, Chrome extensions, Windows 8 apps etc. are handled in `master` only.

> I agree it would be nice to have Browserify tests in compat.

I'll think about something. The solution from this PR won't move over since I use Node here to test various situations and the Browserify one is "faked" by me via `global.window = window;`. In `compat` I'd most likely have to compile jQuery using Browserify and test the resulting bundle.

mzgol:
For now I'm just making `intro.js` files almost identical on both branches: #2294

Issue #2248, opened 2015-04-30
jQuery allows strings and objects as event handlers => error thrown


acasademont:
http://jsfiddle.net/t7da32og/

Click the button and that will make the browser throw an error. In the `on` function code is checked that the `fn` var is not `false` or whatever value the `if (!fn)` condition matches (0, null). But strings and objects are allowed and then the `handler.apply` function will of course fail.

My question is, is there a reason why these 2 types are not checked and the callback is not discarded? If not, I'll be more than happy to submit a PR to fix the issue!

timmywil:
Thank you for opening an issue, but if passing `"click"` as the first argument, our [docs](http://api.jquery.com/on/) say that a handler function is required. Passing an object there is not supported.

acasademont:
@timmywil  Yes I know that, but if there are guard checks for `false` `0` or `null` values why not for strings and objects? I'm seeing some random errors in my JS profiling tool caused by who-knows-what-random-external-code that could be avoided by these checks

acasademont:
I'm talking about this https://github.com/jquery/jquery/blob/master/src/event.js#L820-L824

acasademont:
IMHO a `typeof string || typeof object` check wouldn't hurt any performances as 99.99% of events are done right :)

timmywil:
In this case, I'd actually prefer an error be thrown. Why make the call at
all if it's not going to do anything? Anyway, our general policy in these
cases is, invalid input will produce unexpected results.

markelog:
@timmywil why do we check for falsy values? Perhaps we should remove that check?

acasademont:
And I agree 100% with you but there are tons of 3rd party rubbish scripts that browsers embed as extensions (viruses/trojans/malware...) that cause our JS profiling tools (we are using errorception) to be filled with this "faux" errors caused by all sorts of external hacks. I'm afraid we really can't control that code :(

For example in the `addClass` code there's a guard check for a string value https://github.com/jquery/jquery/blob/master/src/attributes/classes.js#L13

Instead of the current code a check for `function` would come handy.

acasademont:
instead of

```javascript
if ( fn === false ) {
    fn = returnFalse;
} else if ( !fn ) {
    return this;
}
```

this

```javascript
if ( typeof fn !== "function" ) {
    return this;
}
```
PS: still not sure why there's a separate case for `false`

markelog:
@acasademont 

[...]

PR #2249, opened 2015-04-30
Event: remove guard for falsy handler argument of jQuery#on method

markelog:
Since we don't have this in `off` method and its a common perception that this is a rudiment code.
See #2248

### DIFF
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -5,20 +5,6 @@ module( "event", {
 	teardown: moduleTeardown
 });
 
-test("null or undefined handler", function() {
-	expect(2);
-	// Supports Fixes bug #7229
-	try {
-		jQuery("#firstp").on( "click", null );
-		ok(true, "Passing a null handler will not throw an exception");
-	} catch ( e ) {}
-
-	try {
-		jQuery("#firstp").on( "click", undefined );
-		ok(true, "Passing an undefined handler will not throw an exception");
-	} catch ( e ) {}
-});
-
 test("on() with non-null,defined data", function() {
 
 	expect(2);
--- a/src/event.js
+++ b/src/event.js
@@ -819,8 +819,6 @@ jQuery.fn.extend({
 		}
 		if ( fn === false ) {
 			fn = returnFalse;
-		} else if ( !fn ) {
-			return this;
 		}
 
 		if ( one === 1 ) {


dmethvin:
I'm okay with doing this on a 3.0, but you can expect a lot of crappy code to explode. If we do this a shim should go into Migrate for it. Let's wait until a decision is made here (and a ticket created or #2248 renamed for it) before opening a Migrate request.

markelog:
kk

mzgol:
+1 from me, if not for the symmetry with `.off`

test/unit/event.js
```diff
@@ -5,20 +5,6 @@ module( "event", {
...
-		jQuery("#firstp").on( "click", null );
-		ok(true, "Passing a null handler will not throw an exception");
-	} catch ( e ) {}
-
-	try {
-		jQuery("#firstp").on( "click", undefined );
-		ok(true, "Passing an undefined handler will not throw an exception");
-	} catch ( e ) {}
-});
-
```
timmywil: What about changing these to `throws` tests?

markelog: It will throw otherwise

timmywil: Hmm?

timmywil: I mean something like this...

```js
throws(function() {
  jQuery("#firstp").on("click", null);
});

throws(function() {
  jQuery("#firstp").on("click", undefined);
});
```

markelog: code `jQuery("#firstp").on( "click", undefined );`  will immediately throw an exception

markelog: Oh, you want to add tests for undefined behaviour?

markelog: > What about changing these to throws tests?

Sorry, i thought you mean why i deleted them :-)

timmywil: I mean, we specifically want this to throw right? I figure, why not test for it to prevent us from changing back in the future?

dmethvin: The problem AIUI was that it throws when the event is triggered not when attached.

timmywil: No, in this case it will throw when attached.

markelog: @timmywil if throws or not is undefined/undocumented behaviour, so it seems we explicitly shouldn't add test for it.

@dmethvin i guess it's better to throw right away then later, but see above.

-- 
By "undefined behaviour" i mean <a href="http://en.wikipedia.org/wiki/Undefined_behavior">term</a> from the C world, so basically we free to do whatever we want.

"Knowning" @dmethvin though, i guess you more concern of how much yelling we would get if do this. I think not a lot, but that is my perception :-)

timmywil: Specifically, when it hits `handler.handler` in `jQuery.event.add`.

timmywil: @markelog The behavior is defined because we said so. We don't need to document that invalid input will throw. That is assumed. What I'm advocating for here is that we add tests to ensure the behavior that we want to happen actually happens. It prevents us from changing the behavior in the future unless we do it intentionally.

gibson042: I'm with @markelog on this removal, and would be fine with some hypothetical future behavior change to _stop_ throwing if that came with size/perf/clarity benefits. No assertions about invalid conditions.

timmywil: Fair enough.

dmethvin: So if we plan to land this, I will add a check in Migrate to warn if `fn` does not live up to its name? :smile_cat:

PR #2251, opened 2015-05-01
Update traversing.js

GregFrench:
Changed '==' to '===' to compare with 'null'.


dmethvin:
Hey @GregFrench, the comparison with `null` is in keeping with our [style guide](https://contribute.jquery.org/style-guide/js/#equality). Also, it would be pretty unusual for us to make a code change without an open issue explaining the reason for the change (bug? feature?) and some accompanying unit tests.

PR #2254, opened 2015-05-02
remove npm from dependencies

calvinmetcalf:
it is unclear why npm was included in the dependencies, it looks like it was accidentally included in this commit 2d5c5d213f09fa0205d07a2d60a36581058cc40a.  I think we can be fairly sure that they will already have npm at this point due to that list being used by npm :smile:

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "jsdom": "3.1.2",
     "load-grunt-tasks": "1.0.0",
     "native-promise-only": "0.7.8-a",
-    "npm": "2.1.12",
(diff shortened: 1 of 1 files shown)


mzgol:
Fortunately it's only a `devDependency` so it won't affect end users but still, unnecessary.

jzaefferer:
I have no idea why I "added" that, certainly wasn't on purpose.

markelog:
@calvinmetcalf could you sign <a href="http://contribute.jquery.org/CLA/">CLA</a> please?

calvinmetcalf:
Done

Issue #2264, opened 2015-05-05
Remove "ready" event


markelog:

timmywil:
PR ready for review.

PR #2265, opened 2015-05-06
Core: remove custom ready event

timmywil:
Fixes gh-2264

### DIFF
--- a/test/unit/ready.js
+++ b/test/unit/ready.js
@@ -1,73 +1,70 @@
-module("event");
+module( "event" );
 
-(function(){
+(function() {
 	var notYetReady, noEarlyExecution,
 		order = [],
 		args = {};
 
 	notYetReady = !jQuery.isReady;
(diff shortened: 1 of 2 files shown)


test/unit/ready.js
```diff
@@ -1,73 +1,70 @@
-module("event");
+module( "event" );
 
-(function(){
+(function() {
```
markelog: Generally, it was asked to not fix code-style issues if PR is not about a code-style

markelog:
io failures are related to https://github.com/jquery/jquery/issues/2266?

timmywil:
Yup

test/unit/ready.js
timmywil: Seems to me that the best opportunity to change code styles is while we're already in that code, but I'm not overly adamant.

markelog:
LGTM

PR #2268, opened 2015-05-06
Attributes: add SVG class manipulation

timmywil:
+22 bytes. You may or may not be surprised that this is actually [faster overall](http://jsperf.com/svg-class-manipulation).

Fixes gh-2199

### DIFF
--- a/src/attributes/classes.js
+++ b/src/attributes/classes.js
@@ -7,16 +7,20 @@ define([
 
 var rclass = /[\t\r\n\f]/g;
 
+function getClass( elem ) {
+	return elem.getAttribute && elem.getAttribute( "class" ) || "";
+}
+
 jQuery.fn.extend({
 	addClass: function( value ) {
-		var classes, elem, cur, clazz, j, finalValue,
+		var classes, elem, cur, curValue, clazz, j, finalValue,
 			proceed = typeof value === "string" && value,
 			i = 0,
 			len = this.length;
 
 		if ( jQuery.isFunction( value ) ) {
 			return this.each(function( j ) {
-				jQuery( this ).addClass( value.call( this, j, this.className ) );
+				jQuery( this ).addClass( value.call( this, j, getClass( this ) ) );
 			});
 		}
 
@@ -26,10 +30,9 @@ jQuery.fn.extend({
 
 			for ( ; i < len; i++ ) {
 				elem = this[ i ];
-				cur = elem.nodeType === 1 && ( elem.className ?
-					( " " + elem.className + " " ).replace( rclass, " " ) :
-					" "
-				);
+				curValue = getClass( elem );
+				cur = elem.nodeType === 1 &&
+					( " " + curValue + " " ).replace( rclass, " " );
 
 				if ( cur ) {
 					j = 0;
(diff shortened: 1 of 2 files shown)


mzgol:
It's not faster for me. Not slower either, though.
screen shot 2015-05-06 at 22 48 41
screen shot 2015-05-06 at 22 50 45

LGTM, a lot of people are asking for this.

timmywil:
Interesting you got different results than me. I'm glad it's not a perf hit, though.

timmywil:
@mzgol Oh, you got different results because I accidentally built on master and pushed master to dropbox. Please have a look again.

mzgol:
Indeed, now it looks faster. No +1 but +2 then ;)

timmywil:
@dmethvin Let's add this to the blog post! :smiley_cat:

src/attributes/classes.js
```diff
@@ -7,16 +7,20 @@ define([
 
 var rclass = /[\t\r\n\f]/g;
 
+function getClass( elem ) {
+	return elem.getAttribute && elem.getAttribute( "class" ) || "";
```
markelog: Why do we need `elem.getAttribute &&` bit?

markelog: Even if we need, maybe ternary would be better - 

```js
elem.getAttribute ? elem.getAttribute( "class" ) : ""
```
then `&&` & `||` magic

src/attributes/classes.js
```diff
@@ -26,10 +30,9 @@ jQuery.fn.extend({
 
 			for ( ; i < len; i++ ) {
 				elem = this[ i ];
-				cur = elem.nodeType === 1 && ( elem.className ?
-					( " " + elem.className + " " ).replace( rclass, " " ) :
-					" "
-				);
+				curValue = getClass( elem );
+				cur = elem.nodeType === 1 &&
```
markelog: So `cur` is undefined at the point, maybe simple

```js
if ( elem.nodeType === 1 ) {
  cur = ( " " + curValue + " " ).replace( rclass, " " );
}
```

Would suffice?

src/attributes/classes.js
timmywil: We need the check because we have tests supporting some silent fails. The proposed alternative differs in logic. This way still returns an empty string in the case where getAttribute does not exist.

src/attributes/classes.js
timmywil: Nice catch. I'll change that.

src/attributes/classes.js
markelog: > We need the check because we have tests supporting some silent fails

Could you post the link?

> This way still returns an empty string in the case where getAttribute does not exist.

Oh, same returned type, okay, cool

timmywil: I realised I said the wrong, but you got what I meant. getAttribute can return null and this ensures we return a string.

markelog: > getAttribute can return null and this ensures we return a string.

Aaam, if you said something wrong, i didn't notice :-), i meant it is good if some function always returns the same type.
I wondering about what tests we wouldn't pass though, if it would be just `elem.getAttribute( "class" ) || ""`?

timmywil: Here's the test that fails without the check https://github.com/timmywil/jquery/blob/master/test/unit/attributes.js#L1073

src/attributes/classes.js
timmywil: Actually, just tried this change and got a failure because cur retains its value from the previous loop and needs to be set to something falsey.

Issue #1760, opened 2014-10-21
.off() does not clean up empty cache entries


mzgol:
Originally reported by psquared at: http://bugs.jquery.com/ticket/14674
<p>
Even if all event handlers are removed with <tt>.off()</tt>, an empty cache entry is left in <tt>data_priv</tt>.  Unfortunately, those empty cache entries remain if elements are removed using some method other than <tt>.remove()</tt>, causing a memory leak.
</p>
<p>
<strong>Test Environment</strong>: <br>
Chrome 31.0.1650.63, jQuery 2.0.3
</p>
<p>
<strong>Steps to Reproduce</strong>:<br>
</p>
<ol><li>Attach an event handler to an element (e.g. using <tt>.on()</tt>)
</li><li>Remove the event handler with <tt>.off()</tt>
</li><li>Remove the element with <tt>Node.removeChild()</tt>
</li></ol><p>
<strong>Expected Results</strong>:<br>
No memory leaks.
</p>
<p>
<strong>Actual Results</strong>:<br>
An orphaned cache entry remains in <tt>data_priv</tt>.
</p>
<p>
<strong>Example</strong>:<br>
<a class="ext-link" href="http://jsfiddle.net/Rx3v3/" rel="nofollow"><span class="icon">&nbsp;</span>http://jsfiddle.net/Rx3v3/</a>
</p>
<p>
<strong>Additional info</strong>:<br>
Modifying the last block of <tt>jQuery.event.remove()</tt> as below seems to alleviate the problem.
</p>
<div class="code"><pre><span class="c1">// Remove the expando if it's no longer used
</span><span class="k">if</span> <span class="p">(</span> <span class="nx">jQuery</span><span class="p">.</span><span class="nx">isEmptyObject</span><span class="p">(</span> <span class="nx">events</span> <span class="p">)</span> <span class="p">)</span> <span class="p">{</span>
    <span class="k">delete</span> <span class="nx">elemData</span><span class="p">.</span><span class="nx">handle</span><span class="p">;</span>
    <span class="nx">data_priv</span><span class="p">.</span><span class="nx">remove</span><span class="p">(</span> <span class="nx">elem</span><span class="p">,</span> <span class="s2">"events"</span> <span class="p">);</span>

    <span class="c1">// Clean up empty cache entries:
[...]

PR #2271, opened 2015-05-06
Data: remove the expando when there's no more data; minor cleanup

timmywil:
- This also removes the `Data#discard()` method, which we never used, and the `initial` paramater in `Data#register()` – also never used.

Fixes gh-1760

### DIFF
--- a/test/unit/data.js
+++ b/test/unit/data.js
@@ -837,7 +837,24 @@ test("Check proper data removal of non-element descendants nodes (#8335)", 1, fu
 });
 
 testIframeWithCallback( "enumerate data attrs on body (#14894)", "data/dataAttrs.html", function( result ) {
-	expect(1);
+	expect( 1 );
 
-	equal(result, "ok", "enumeration of data- attrs on body" );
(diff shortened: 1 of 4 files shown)


rwaldron:
LGTM.

test/unit/data.js
```diff
@@ -837,7 +837,24 @@ test("Check proper data removal of non-element descendants nodes (#8335)", 1, fu
...
 
-	equal(result, "ok", "enumeration of data- attrs on body" );
+	var key,
+		div = jQuery( "<div/>" );
+	div.data( "some", "data" );
+	equal( div.data( "some" ), "data", "Data is added" );
+	div.removeData( "some" );
+
+	// Make sure the expando is gone
+	for ( key in div[ 0 ] ) {
```
mzgol: A similar test for on+off would be nice.

timmywil: Sure, I'll add that.

timmywil: Added

Issue #2266, opened 2015-05-06
jsdom dropped support for Node; node/iojs smoke test needs modification


timmywil:
We use [jsdom](https://github.com/tmpvar/jsdom) in our Node smoke tests. Unfortunately, the old version doesn't install on iojs and the new one doesn't support Node. We'll have to use something else.

timmywil:
Actually, another solution would be to install different versions of jsdom using the npm module before running the tests.

timmywil:
That would be the quicker path to success, but does anyone have suggestions for alternatives to jsdom? My quick searches haven't really turned up anything promising.

mzgol:
:/ I feared it would become an issue. jsdom 3.1.2 works just fine with io.js 1.8.1 but breaks with the newest breaking 2.0.0 release.

Installing it by ourselves unfortunately means skipping package.json but I don't see a better option.

I don't think anything else comes anywhere near close to jsdom in DOM support. @denis-sokolov, do you know about something?

@domenic is there anything else we could do here except to install different jsdom versions based on if we're on Node.js or io.js?

mzgol:
Using newest jsdom on io.js has the advantage of being easier for Windows users as it doesn't require compilation done by contextify; some users reported to me it's harder for them to do `npm install` in the jQuery repo because of that.

domenic:
If you get contextify patched to support io.js 2.0, then jsdom 3.x should continue to work everywhere. Someone has to do the work though.

timmywil:
contextify was already patched. We could install the latest version of contextify top-level so jsdom uses that, but that seems like a short-term solution.

mzgol:
> contextify was already patched. We could install the latest version of contextify top-level so jsdom uses that, but that seems like a short-term solution.

That's not needed. jsdom 3.1.2 depends on contextify `>= 0.1.9 < 0.2.0` so it will catch a newer vesion if it doesn't bump the minor. The newest contextify is 0.1.13, though, so even if contextify was patched, it hasn't been released yet.

In that case it seems it's best to wait.

timmywil:
Actually, I think I misunderstood the patch for [this issue](https://github.com/brianmcd/contextify/issues/149).

mzgol:
https://github.com/brianmcd/contextify/issues/151#issuecomment-99576382

denis-sokolov:
I am unaware of anything similar to `jsdom`.

[...]

PR #2272, opened 2015-05-06
Build: use different versions of jsdom for Node and iojs testing

timmywil:
Fixes gh-2266

### DIFF
--- a/build/tasks/install_jsdom.js
+++ b/build/tasks/install_jsdom.js
@@ -1,0 +1,27 @@
+module.exports = function( grunt ) {
+	grunt.registerTask( "jsdom", function() {
+		var current,
+			pkg = grunt.config( "pkg" ),
+			version = pkg.jsdomVersions[
+
+				// Unfortunately, this is currently the only
+				// way to tell the difference between Node and iojs
+				/^v0/.test( process.version ) ? "node" : "iojs"
+			];
+
+		try {
+			current = require( "jsdom/package.json" ).version;
+			if ( current === version ) {
+				return;
+			}
+		} catch ( e ) {}
+
+		// Use npm on the command-line
+		// There is no local npm
+		grunt.util.spawn( {
+			cmd: "npm",
+			args: [ "install", "jsdom@" + version ],
+			opts: { stdio: "inherit" }
+		}, this.async() );
+	});
+};
--- a/build/tasks/node_smoke_tests.js
+++ b/build/tasks/node_smoke_tests.js
@@ -5,7 +5,7 @@ module.exports = function( grunt ) {
 	var fs = require( "fs" ),
 		spawnTest = require( "./lib/spawn_test.js" ),
 		testsDir = "./test/node_smoke_tests/",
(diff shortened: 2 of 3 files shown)


build/tasks/install_jsdom.js
```diff
@@ -0,0 +1,29 @@
...
+		var done = this.async(),
+			pkg = grunt.config( "pkg" ),
+			version = pkg.jsdomVersions[
+
+				// Unfortunately, this is currently the only
+				// way to tell the difference between Node and iojs
+				/^v0/.test( process.version ) ? "node" : "iojs"
+			];
+
+		// Link npm so we can use the global one
```
mzgol: Why is that important? Can't we follow the usual pattern & use a local `npm`?

build/tasks/install_jsdom.js
```diff
@@ -0,0 +1,29 @@
...
+		grunt.util.spawn( {
+			cmd: "npm",
+			args: [ "link", "npm" ]
+		}, function() {
+			var npm = require( "npm" );
+			npm.load(function() {
+				npm.commands.ls( [ "jsdom" ], function( data ) {
```
mzgol: How fast is it? `npm ls` is slow enough that I built [check-dependencies](http://npmjs.com/package/check-dependencies) to have a quick check for installed deps. Manually doing `require('./node_modules/jsdom/package.json').version` (+ checking if the file exists first) would be faster, less code & I doubt it'll ever stop working.

build/tasks/install_jsdom.js
timmywil: There is no local npm. We don't include it in dependencies anymore.

build/tasks/install_jsdom.js
timmywil: Good idea. I'll do that instead.

build/tasks/install_jsdom.js
timmywil: We could use command-line npm with another spawn. Now that it's just one npm command, I don't mind that as much.

timmywil: Changed to use spawn.

build/tasks/install_jsdom.js
timmywil: Updated.

build/tasks/install_jsdom.js
```diff
@@ -0,0 +1,27 @@
...
+			current = require( "jsdom/package.json" ).version;
+			if ( current === version ) {
+				return;
+			}
+		} catch ( e ) {}
+
+		// Use npm on the command-line
+		// There is no local npm
+		grunt.util.spawn( {
+			cmd: "npm",
```
mzgol: Is the indentation OK here?

timmywil: Seems to be.

mzgol:
LGTM

timmywil:
Thanks @mzgol. I'm going to go ahead and merge this so our builds start passing.

Issue #2237, opened 2015-04-27
position() function reports incorrect left offset under Chrome


ReimuYome:
Test this
https://jsfiddle.net/Reimu/yu3c361g/2/
under Chrome and other browsers, you will see the <code>left</code> offset reported by <code>position()</code> method is incorrect under Chrome.

The value of  <code>left</code> offset under Chrome is the gap between &lt;html&gt; and &lt;body&gt;, not the real left offset of the horizontally centered &lt;div&gt;. 

This problem relates only with horizontally centered &lt;div&gt;s, and can be fixed by calculating manually with the <code>marginLeft</code> value of the element's computed style and the gap between &lt;html&gt; and &lt;body&gt;.

Maybe this is due to a possible bug / design choice of Chrome, as there are also some discussion about it on StackOverflow. But I think jQuery is supposed to deal with such difference between browsers.

alburthoffman:
position() does not include the margins. For me, it's kind of reasonable. Because margin is part of the element. For ur example, the victim div locates at left 8px and top 158px, but it has 160.5px left margin and 0px top margin. So the border of it has 168.5px from the left and 158px from the top.

Since you are not asking for position of the border of victim, so ......

ReimuYome:
@alburthoffman  Thank you for your explanation. Though still my concern is  the different behaviors of this method under different browser.

In my view, when using jQuery I shall be exempt of caring for what browser I am facing. As the behavior of this method shows asymmetry and it is not mentioned in the API documentation.

For this very case, I can obtain the correct result after switching to <code>offset</code> method. But that method has its own semantics and the act of switching itself shows some compatibility concern.

My suggestion is this method reporting same value under same browsers. Either subtracting margin value under Chrome, or follow the Chrome way in all cases.

dmethvin:
To report the same value though, we'd need to ignore the margin on the body in Chrome. There really is a margin on the body in Chrome: https://jsfiddle.net/yu3c361g/4/  So how do we rationalize ignoring a value set in CSS?

alburthoffman:
This depends on browsers. I tested it in chrome 41 and safari. And they are the same. But in firefox, the left value is different, because the marginLeft is set to 0 in it. 

So it will require lots of browser version checking in this function if it has to return a same value.

dmethvin:
It seems like the way most pages would fix this is through a CSS reset or normalize? It seems consistent in Firefox for me BTW.

ReimuYome:
[...]

PR #2276, opened 2015-05-08
CSS: Correct misrepresentation of "auto" horizontal margins as 0

gibson042:
Fixes gh-2237

But as expected, it's pretty big:
```
   raw     gz Compared to compat @ fb25bacf9b809a08c29e76decd8cda1579fdf192    
  +790   +196 dist/jquery.js                                                   
  +290    +78 dist/jquery.min.js
```

### DIFF
--- a/src/css/support.js
+++ b/src/css/support.js
@@ -5,7 +5,7 @@ define([
 
 (function() {
 	var div, container, style, a, pixelPositionVal, boxSizingReliableVal, pixelMarginRightVal,
-		reliableHiddenOffsetsVal, reliableMarginRightVal;
+		reliableHiddenOffsetsVal, reliableMarginRightVal, reliableMarginLeftVal;
 
 	// Setup
 	div = document.createElement( "div" );
@@ -56,7 +56,7 @@ define([
 		},
 
 		pixelMarginRight: function() {
-			// Support: Android 4.0-4.3
+			// Support: Android 4.0 - 4.3 only
 			if ( pixelPositionVal == null ) {
 				computeStyleTests();
 			}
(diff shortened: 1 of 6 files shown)


src/css/support.js
```diff
@@ -76,6 +76,14 @@ define([
 				computeStyleTests();
 			}
 			return reliableMarginRightVal;
+		},
+
+		reliableMarginLeft: function() {
+			// Support: IE <=8 only, Android 4.0 - 4.3 only, Firefox <=3 - 37
```
markelog: Usually, there is no spaces in ranges :-)

gibson042: jquery/contribute.jquery.org/issues/95

markelog: Yeah, it seems we didn't had consensus about this issue, it did come down to like/don't like, i'd say we should follow some standart here, if we so nit-picky :-).

http://practicaltypography.com/hyphens-and-dashes.html 
http://en.wikipedia.org/wiki/Dash#Ranges_of_values

markelog:
If consumer reset style it is Firefox only, right? Should we consider asking ff team to prioritize their ticket you mentioned in affiliated issue instead?

gibson042:
I'm not sure I understand your question... this affects [IE8, Android, and Firefox](https://github.com/jquery/jquery/pull/2276/files#diff-7a031a0a341e84edbc03348dea967f7cR81), but I guess it's true that Firefox is the only browser in which the behavior has yet to be fixed.

I probably should have already landed this, but since it's still open, it might as well still be open for discussion.

markelog:
I was referencing <a href="https://github.com/jquery/jquery/issues/2237#issuecomment-97281378">this</a> discussion.

timmywil:
@gibson042 Is this high enough priority to include in our 1.x release?

dmethvin:
Looks like IE8 started failing when this landed: http://swarm.jquery.org/result/720927

PR #2278, opened 2015-05-08
Replaced jQuery.each with forEach to iterate arrays.

jackmoore:
All of the browsers supported by jQuery 2.x should support ES5 array iteration, so I don't see any benefit to using jQuery.each.  Just my preference for using native moving forward, feel free to close as the perf and file size differences would be insignificant.

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -763,7 +763,7 @@ jQuery.extend({
 	}
 });
 
-jQuery.each( [ "get", "post" ], function( i, method ) {
+[ "get", "post" ].forEach(function( method ) {
 	jQuery[ method ] = function( url, data, callback, type ) {
 		// Shift arguments if data argument was omitted
 		if ( jQuery.isFunction( data ) ) {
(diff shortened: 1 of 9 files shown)


timmywil:
Thanks for contributing! However, we can't do this for backwards compatibility reasons.

markelog:
@timmywil proposal here is to use `Array#forEach` internally

timmywil:
So it is! However, haven't we also talked about that and decided it wasn't
worth it?

markelog:
Don't remember such discussion, maybe a missed it, i'd say it is only worth it, if it helps to reduce the size, otherwise eat your own dog food.

timmywil:
Ok, well I think I can recall the major points. 

1. The point @markelog brought up is a good one. We want to dog food our own methods.
2. IE8 doesn't have native `forEach` and we want to avoid diverging from the compat branch where it's not necessary or has no significant advantages.
3. It's not a perf issue.
4. Even if it is reduced size, the difference would not be enough to warrant ignoring the other points.

Just as a side note for future contributions, the 2.1-stable branch is not under active development and there are currently no plans to add any changes to that branch.

jackmoore:
Thanks, I appreciate the explanation.

Issue #2227, opened 2015-04-21
:visible/:hidden selectors unreliable


shimonenator:
# Issue

This issue is present in both version 1 and 2 of jQuery.

It is happening on inline elements wrapping block elements, which is valid in HTML5, such as a header and an image wrapped in a single anchor.

Two issues arise:

1. Different browsers report offsetWidth/Height of the anchor differently. For example Firefox would report one visible anchor in the example below, and Chrome will report 0. This also highly depends on the CSS styles applied, as on the page where this issue was discovered missing images are not displayed at all instead of the common missing image icon.
2. While a browser may report an anchor as invisible, child elements are still visible and clickable as they inherit anchor properties.

## Example

```html
<!doctype html>
<html>
<head>
	<meta charset="utf-8">
	<script src="http://code.jquery.com/jquery-2.1.3.min.js"></script>
</head>
<body>
	<a href=#>
		<h1>Header</h1>
	</a>
	<script>
		$(function () {
			alert('Visible anchors: ' + $('a:visible').length);
		})
	</script>
</body>
</html>
```

# Solution
There are 2 potential solutions that come to mind, but each has flaws

1. Loop through child elements and check each for visibility. Potentially harmful for performance, although the selector is already marked as such in the documentation. 
2. Check parent for visibility. Probably not enough on its own, as it will report incorrectly, if anchor doesn't have visible children.

timmywil:
Thank you for opening an issue. While you are correct that the behavior may be unexpected, the selector is working as documented. While the behavior differs between Firefox and Chrome, Chrome reports zero width and zero height, which technically means it is not visible according to our docs...

> Visible elements have a width or height that is greater than zero.

The fact the hidden elements can have visible content can seem strange, but it's the way it is in Chrome. But more importantly, the cost of working around this would be far greater than we are willing to deal with. I don't think this is the only edge case where the `:visible` selector behaves in a way that can be unexpected, but I would prefer to keep the selector as simple as possible and let the edge cases be covered by other means.

dmethvin:
This seems like a Chrome bug, does Safari have it as well?

mzgol:
> This seems like a Chrome bug, does Safari have it as well?

Yes.

dmethvin:
Here's a test case: http://jsfiddle.net/2tgw2yr3/

Both jsbin.com and browserstack.com are down for me at the moment.

shimonenator:
@timmywil thank you for a quick reply. 

[...]

PR #2281, opened 2015-05-08
CSS: fix :visible/:hidden selectors for inline element w/ content

timmywil:
- Reverts behavior from 10399dd, which we never released.
  BR and inline elements are considered visible.
- The possibility of dropping .offsetWidth and .offsetHeight
  was debunked by this perf:
  http://jsperf.com/visible-hidden-and-getclientrects

Fixes gh-2227

### DIFF
--- a/src/css/hiddenVisibleSelectors.js
+++ b/src/css/hiddenVisibleSelectors.js
@@ -4,12 +4,10 @@ define([
 ], function( jQuery ) {
 
 jQuery.expr.filters.hidden = function( elem ) {
-	// Use OR instead of AND as the element is not visible if either is true
-	// See tickets #10406 and #13132
-	return !elem.offsetWidth || !elem.offsetHeight;
+	return !jQuery.expr.filters.visible( elem );
 };
 jQuery.expr.filters.visible = function( elem ) {
-	return !jQuery.expr.filters.hidden( elem );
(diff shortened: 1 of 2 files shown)


gibson042:
LGTM

mzgol:
Does it mean we might not be able to use fractional values for .css('width') and friends due to perf reasons?

gibson042:
I don't think so. `.css` is explicitly asking for style computation, as opposed to these selectors which should be as fast as possible.

src/css/hiddenVisibleSelectors.js
```diff
@@ -4,12 +4,10 @@ define([
 ], function( jQuery ) {
 
 jQuery.expr.filters.hidden = function( elem ) {
-	// Use OR instead of AND as the element is not visible if either is true
-	// See tickets #10406 and #13132
-	return !elem.offsetWidth || !elem.offsetHeight;
+	return !jQuery.expr.filters.visible( elem );
```
staabm: Would it make sense to inline and reverse the logic from `jQuery.expr.filters.visible()` to save a func call?

timmywil: We haven't noticed a perf loss and this way guarantees hidden and visible are opposites.

dmethvin: The browser probably does that anyway, and even if it does take another function call the DOM queries will be much more cruel.

Issue #1768, opened 2014-10-21
"define" is not defined error. Broken build


mgol:
Originally reported by corbacho at: http://bugs.jquery.com/ticket/15061

<p>
When building a custom jQuery (master branch) with this command:
</p>

<pre class="wiki">grunt build:*:+core/ready
</pre><p>
It will output about the line 47:
</p>
<pre class="wiki">define("jquery", function(){});
</pre><p>
This produces the error
</p>
<pre class="wiki">ReferenceError: define is not defined
</pre><p>
<strong>Why I'm using grunt build?</strong>
I know that "grunt custom" is the recommended alias, but I'm trying to build a minimum possible jQuery that contains the necessary code for a reliable $(document).ready() .
I want to propose this solution to be part of the upcoming Drupal 8 core, as an alternative to domready solution <a class="ext-link" href="https://github.com/ded/domready" rel="nofollow"><span class="icon">&nbsp;</span>https://github.com/ded/domready</a> that we are using now
</p>

<p>
The whole command is :
</p>

<pre class="wiki">grunt build:*:+core/ready:+exports/global:-exports/amd:-ajax:-ajax/xhr:-ajax/script:-ajax/jsonp:-css:-deprecated:-dimensions:-effects:-event:-event/alias:-offset:-wrap:-sizzle:-queue:-data:-css:-selector:-traversing:-serialize
</pre><p>
This builds a 11Kb size domready solution. 5Kb gzipped
</p>

<p>
Removing the buggy line mentioned before, it works:
<a class="ext-link" href="http://jsfiddle.net/WnhWc/2/" rel="nofollow"><span class="icon">&nbsp;</span>http://jsfiddle.net/WnhWc/2/</a>
</p><p>Issue reported for jQuery 2.1.0</p>

PR #2284, opened 2015-05-09
Build: Remove empty define({}) from build output

Gerhut:
Fixes gh-1768
Closes gh-1569


mzgol:
This has already been closed on `master` and will arrive in jQuery 3.0.0 and we don't expect a new patch 1.11 release currently.

PR #2285, opened 2015-05-09
Build: Remove empty define({}) from build output

Gerhut:
Fixes gh-1768
Closes gh-1569


mzgol:
See https://github.com/jquery/jquery/pull/2284#issuecomment-100518492

Issue #2286, opened 2015-05-09
Remove deprecated event aliases


markelog:
`load`, `unload` and `error` - deprecated <a href="http://blog.jquery.com/2012/08/09/jquery-1-8-released/">since</a> 1.8. 

Ref http://bugs.jquery.com/ticket/11733

markelog:
If we decide to do this, we also need a migrate and docs ticket

dmethvin:
https://github.com/jquery/jquery-migrate/pull/94 takes care of the Migrate part.

zowers:
[upgrade guide](https://jquery.com/upgrade-guide/3.0/) says 
>  If you need to attach **errors** by these names, use the .on() method

should be 

>  If you need to attach **events** by these names, use the .on() method

mgol:
@zowers Thanks for the report, fixed!

If you find more mistakes, please report them at https://github.com/jquery/jquery.com/issues, this issue tracker is for the jQuery library.

The upgrade guide itself is written in Markdown in https://github.com/jquery/jquery.com/blob/master/pages/upgrade-guide/3.0.md.

darwindlrd:
Now that I use function that allows do the same job?

PR #2287, opened 2015-05-09
Event: remove deprecated event aliases

markelog:
Fixes gh-2286
Ref trac-11733


dmethvin:
LGTM, not a lot of testing was going on there.

markelog:
I guess we have the same issue for remaining the aliases, i will propose PR for it

mzgol:
LGTM

Issue #2283, opened 2015-05-09
Effects: finish() should call progress(1)


Mr21:
Hi :)

On this [jsfiddle.net/Mr21/cr5391w2](http://jsfiddle.net/Mr21/cr5391w2/) we can see the difference between `.stop` and `.finish`. The darkest blue block is modified only with the `progress` callback.
But the callback `progress` is not called a last time with the value `1` when we force the end of the animation.

If you are agree on this problem, I found where we should put a line to fix this problem.

dmethvin:
Since we're using a `jQuery.Deferred` for the animation we would need to call the deferred's `.notifyWith()` method which should do the rest for us, but I agree it seems like animations should do this on `.finish()` or `.stop(true)`.

Mr21:
@dmethvin okay it's done :)

PR #2292, opened 2015-05-11
Effects: finish should call progress

Mr21:
A pull request for #2283.


dmethvin:
Thanks! We will need some unit tests for this change to ensure that it actually works. The closest existing test we have is [here](https://github.com/jquery/jquery/blob/5c3101fee60046fa7976b3131fada8dfe9fbd53e/test/unit/effects.js#L1772-1851) but you can probably create a simpler one just to check for `.finish()` calling the callback and the `.deferred()`.

src/effects.js
```diff
@@ -319,6 +319,7 @@ function Animation( elem, properties, options ) {
 
 				// Resolve when we played the last frame; otherwise, reject
 				if ( gotoEnd ) {
+					deferred.notifyWith( elem, [ animation, 1, 0 ]);
```
gibson042: Missing space between `]` and `)`.

gibson042:
I was on the fence about changing when we notify animations, but this makes sense. :+1:

timmywil:
Agreed :+1: Thanks @Mr21!

src/effects.js
Mr21: Oops, it's okay.

test/unit/effects.js
```diff
@@ -1780,7 +1781,8 @@ test("Animation callbacks (#11797)", 15, function() {
 			ok( true, "empty: start" );
 		},
 		progress: function( anim, percent ) {
-			equal( percent, 0, "empty: progress 0" );
+			equal( percent, prog, "empty: progress " + prog );
```
Mr21: I change something here, because in jQuery if there is no property to animate (here it's empty with the `{}`) `.stop( true )` is called. So, progress is now called twice.

Mr21:
@dmethvin, yep I wrote a new test and modify another one.

test/unit/effects.js
```diff
@@ -1780,7 +1781,8 @@ test("Animation callbacks (#11797)", 15, function() {
...
+			prog = +!percent;
```
gibson042: Just `prog = 1`, please. There's no need to be overly clever in known environments like unit tests.

test/unit/effects.js
```diff
@@ -1850,6 +1852,35 @@ test("Animation callbacks (#11797)", 15, function() {
...
+			cb += "[p," + p + "," + ms + "]";
+		},
+		done: function() {
+			cb += "[d]";
+		},
+		fail: function() {
+			cb += "[f]";
+		},
+		always: function() {
+			equal( cb, "[s][p,0," + dur + "][p,1,0][d]", "callbacks order" );
```
gibson042: This is also too opaque. How about pulling in [qunit-assert-step](https://github.com/JamesMGreene/qunit-assert-step)?

test/unit/effects.js
Mr21: Yeah right.

test/unit/effects.js
Mr21: Oh nice, I did that because I didn't another way.
What I have to do? install a new dependency (in the `package.json`) etc.?

test/unit/effects.js
```diff
@@ -1850,6 +1852,35 @@ test("Animation callbacks (#11797)", 15, function() {
 	this.clock.tick( 10 );
 });
 
+test("Animation callbacks (#2292)", 1, function() {
+	var jq_elem = jQuery("#foo"),
```
arthurvr: why aren't we using a camel cased variable name here, as always?

Mr21: It's because I'm used to add `jq_` for every jQuery object, but I removed the variable anyway, thanks.

test/unit/effects.js
gibson042: Yep.

Mr21:
@gibson042, I've used *qunit-assert-step* (like you said) to test if the callbacks are called in order. It's works fine :)
I put the plugin in the `external` directory with qunit, is it the good way?

gibson042:
Yep; looks like the plugin was added correctly. :+1:

test/unit/effects.js
```diff
@@ -1850,6 +1852,44 @@ test("Animation callbacks (#11797)", 15, function() {
...
+		},
+		done: function() {
+			assert.step( 4 );
+		},
+		fail: function() {
+			ok( false, "Animation failed" );
+		},
+		always: function() {
+			assert.step( 5 );
+			ok( true, "The callbacks were called in the correct order" );
```
gibson042: This assertion adds no value.

mzgol:
@Mr21 
>> @gibson042, I've used *qunit-assert-step* (like you said) to test if the callbacks are called in order. It's works fine :)
>> I put the plugin in the `external` directory with qunit, is it the good way?
>
> Yep; looks like the plugin was added correctly. :+1:

Actually, it wasn't. ;) You shouldn't copy files to `external/` by yourself, you should add proper files to the [npmcopy config](https://github.com/jquery/jquery/blob/2d715940b9b6fdeed005cd006c8bf63951cf7fb2/Gruntfile.js#L49-L69) and then invoke `grunt npmcopy` which will do the copying. Then you need to commit those copied files.

test/unit/effects.js
```diff
@@ -1850,6 +1852,44 @@ test("Animation callbacks (#11797)", 15, function() {
...
+		},
+		progress: function( anim, p, ms ) {
+			if ( !( step++ ) ) {
+				assert.step( 2 );
+				equal( p,    0, "first call of progress: percent != 0" );
+				equal( ms, dur, "first call of progress: ms != duration (" + dur + ")" );
+			} else {
+				assert.step( 3 );
+				equal( p,  1, "last call of progress: percent != 1" );
+				equal( ms, 0, "last call of progress: ms != 0" );
```
gibson042: These `equal`s should be `strictEqual`s, and the messages ought to be more clear (e.g., `"(first|last) progress callback: (progress ratio|remaining ms)"`).

Mr21:
@mzgol, okay thanks, I modified the `Gruntfile.js`, it's works.

test/unit/effects.js
Mr21: I renamed these 4 tests, thanks

test/unit/effects.js
Mr21: Yeah right, I removed this

Mr21:
No more conflict here

mzgol:
Landed, thanks!

PR #2294, opened 2015-05-11
Core: Align code in intro.js with master

mzgol:
Since we're testing the factory code from intro.js on master only it's best
to have those two files almost identical.

This commit leaves only one difference between them: "jQuery Compat" vs.
"jQuery" in the second line.

### DIFF
--- a/src/intro.js
+++ b/src/intro.js
@@ -15,13 +15,13 @@
 (function( global, factory ) {
 
 	if ( typeof module === "object" && typeof module.exports === "object" ) {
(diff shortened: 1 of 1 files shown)


markelog:
lgtm

mzgol:
Landed at fe2a584ed662ad15e0cfb2e806652e2a8724f0c8

mgol:
This was a `compat`-only PR so I'm removing the `3.0.0` milestone.

PR #2296, opened 2015-05-11
CSS: Collapse a double if statement into one

mzgol:
Saves 3 bytes gzipped

This seems extremely obvious... Am I missing sth?


gibson042:
Huh. Must have come out smaller at one point: a339096d6a9d0313c2184de0d626e2c7e127955b

This LGTM, though.

Issue #2015, opened 2015-01-16
css vendor prefix error


pltnkv:
$().css automaticly add vendor prefixes, but if firstly call "$().css" function with prefixes styles, auto-adding break. Run example in Safari
http://jsfiddle.net/8cf0hyhj/

dmethvin:
Yes, trying this on Safari 8 via Browserstack I agree it's failing. The problem seems to be that if *you* explicitly set the vendor-prefixed property on a style then Safari starts responding to that property on any `CSSStyleDeclaration`. Totally bizarre. Can someone else verify they see this happening in `vendorPropName()` as well?

I don't know that this is something we can fix. Is there a reason you just don't let us do the prefixes for you?

mzgol:
This bug is caused by:
https://github.com/jquery/jquery/blob/740e190223d19a114d5373758127285d14d6b71e/src/css.js#L265

If we don't find the original style property *and* don't find a prefixed one, we return the original one from `vendorPropName()`. This is then saved to `jQuery.cssProps[propertyName]` in the line linked above. The first `.css` setter checks for `msTransform` first and since Safari doesn't recognize `msTransform`, `msMsTransform`, `webkitMsTransform` nor `mozMsTransform`, `jQuery.cssProps.transform` is saved to be `"msTransform"`.

Later when we try to set `transform` to something, we have it cached that it maps to `msTransform` which obviously won't work.

We should fix it.

mzgol:
PR: #2298

PR #2298, opened 2015-05-11
CSS: Don't cache unrecognized CSS property names

mzgol:
This prevents jQuery from caching a prefixed property name if provided
directly by the user, e.g. the following code:

	elem.css( "msTransform", "translate(5px, 2px)" );

should not prevent one from from later setting the transition directly:

	elem.css( "transform", "translate(5px, 2px)" );

on a browser not understanding the unprefixed version which is the case
for Safari 8 & transform.

Fixes gh-2015

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -3,6 +3,7 @@ define([
 	"./var/pnum",
 	"./core/access",
 	"./css/var/rmargin",
+	"./var/document",
 	"./var/rcssNum",
 	"./css/var/rnumnonpx",
 	"./css/var/cssExpand",
@@ -18,8 +19,8 @@ define([
 	"./core/init",
 	"./core/ready",
 	"./selector" // contains
-], function( jQuery, pnum, access, rmargin, rcssNum, rnumnonpx, cssExpand, isHidden,
-	getStyles, swap, curCSS, adjustCSS, addGetHookIf, support, showHide ) {
+], function( jQuery, pnum, access, rmargin, document, rcssNum, rnumnonpx, cssExpand,
+	isHidden, getStyles, swap, curCSS, adjustCSS, addGetHookIf, support, showHide ) {
 
 var
 	// Swappable if display is none or starts with table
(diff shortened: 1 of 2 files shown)


mzgol:
When writing this patch I discovered that we have no tests that our prefixing mechanism works. :O The commit that introduced them didn't add them:
https://github.com/jquery/jquery/commit/5376a809c0d2bee4b7872847c2821e458dfdcc3b

mzgol:
BTW, this both fixes the bug & reduces the size by 1 byte. ;)

mzgol:
I found some issues with this patch, marking as WIP.

mzgol:
Should be ready now.

src/css.js
```diff
@@ -203,7 +202,7 @@ jQuery.extend({
 			style = elem.style;
 
 		name = jQuery.cssProps[ origName ] ||
-			( jQuery.cssProps[ origName ] = vendorPropName( style, origName ) );
+			( jQuery.cssProps[ origName ] = vendorPropName( origName ) || origName );
```
mzgol: Without this change we'd set `node.style.undefined`; I added a test for that.

src/css.js
```diff
@@ -261,7 +260,7 @@ jQuery.extend({
 
 		// Make sure that we're working with the right name
 		name = jQuery.cssProps[ origName ] ||
-			( jQuery.cssProps[ origName ] = vendorPropName( elem.style, origName ) );
+			( jQuery.cssProps[ origName ] = vendorPropName( origName ) || origName );
```
mzgol: No test will fail without this change but we'll be needlessly checking `getComputedStyle( node ).undefined` so I added it. I assume no test is necessary for this change.

mzgol:
Should be ready now. -5 bytes. :)

mzgol:
I added one more test.

test/unit/css.js
```diff
@@ -1108,4 +1108,84 @@ test( "Do not throw on frame elements from css method (#15098)", 1, function() {
...
+		} else {
+			jQuery.each( vendorPrefixes, function( index, prefix ) {
+				var prefixedProp = prefix + "Appearance";
+				if ( prefixedProp in emptyStyle ) {
+					appearanceName = prefixedProp;
+				}
+			} );
+		}
+
+		if ( "transform" in emptyStyle ) {
```
mzgol: `transform` is unprefixed in WebKit Nightly but it's prefixed in all Androids < 5 so this should not become a useless test any time soon.

src/css.js
```diff
@@ -34,29 +35,27 @@ var
 		fontWeight: "400"
 	},
 
-	cssPrefixes = [ "Webkit", "Moz", "ms" ];
+	cssPrefixes = [ "Webkit", "Moz", "ms" ],
+	emptyStyle = document.createElement( "div" ).style;
```
dmethvin: Nice!

dmethvin:
LGTM, nice unit test. :smile:

timmywil:
:+1:

mzgol:
I'll merge it tomorrow if no one opposes.

Issue #2300, opened 2015-05-12
Events: originalEvent must not be an empty object


gabrielschulhof:
Since https://github.com/jquery/jquery/commit/a873558436383beea7a05fd07db7070a30420100 we must make sure that all objects that are supposed to be events are events. Thus, code like [originalEvent: {}](https://github.com/jquery/jquery/blob/7855a1a7d8a5c959d4ef2f951b4eb971908ac96f/src/event.js#L619) must be replaced with ```originalEvent: new jQuery.Event()```

timmywil:
I took the liberty of assigning you, @markelog. :)

markelog:
@timmywil Sure thing,

> Since a873558 we must make sure that all objects that are supposed to be events are events.

Why?

timmywil:
Because our simulated focus event doesn't have those methods.

gabrielschulhof:
@gibson042 [mentioned](http://irc.jquery.org/%23jquery-dev/default_%23jquery-dev_20150512.log.html#t05:54:11) that we may be able to get rid of the fake ```originalEvent``` altogether.

PR #2303, opened 2015-05-12
Event: Remove fake originalEvent from jQuery.Event.simulate

gabrielschulhof:
Fixes gh-2300

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -610,10 +610,14 @@ jQuery.event = {
 			event,
 			{
 				type: type,
-				isSimulated: true,
-				originalEvent: {}
+				isSimulated: true
 			}
 		);
+
+		// This prevents stopPropagation(), stopImmediatePropagation(), and preventDefault() from
+		// preventing default on the donor event.
+		delete e.originalEvent;
+
 		if ( bubble ) {
 			jQuery.event.trigger( e, null, elem );
 		} else {
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -2682,20 +2682,20 @@ test( "preventDefault() on focusin does not throw exception", function( assert )
 	expect( 1 );
 
 	var done = assert.async(),
-		input = jQuery( "<input>" ).appendTo( "#form" );
+		input = jQuery( "<input/>" ).appendTo( "#form" );
 		input
 			.on( "focusin", function( event ) {
-				var flawlessExecution = true;
+				var exceptionCaught;
 
 				try {
 					event.preventDefault();
-				} catch( theException ) {
-					flawlessExecution = false;
+				} catch ( theException ) {
+					exceptionCaught = theException;
 				}
 
-				ok( flawlessExecution, "Preventing default on focusin throws no exception" );
+				assert.strictEqual( exceptionCaught, undefined,
+					"Preventing default on focusin throws no exception" );
 
-				input.remove();
 				done();
 			} )
 			.focus();
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -2678,6 +2678,60 @@ test( ".off() removes the expando when there's no more data", function() {
 	}
 });
 
+test( "preventDefault() on focusin does not throw exception", function( assert ) {
+	expect( 1 );
+
+	var done = assert.async(),
+		input = jQuery( "<input>" ).appendTo( "#form" );
+		input
+			.on( "focusin", function( event ) {
+				var flawlessExecution = true;
+
+				try {
+					event.preventDefault();
+				} catch( theException ) {
+					flawlessExecution = false;
+				}
+
+				ok( flawlessExecution, "Preventing default on focusin throws no exception" );
+
+				input.remove();
+				done();
+			} )
+			.focus();
+} );
+
+test( "jQuery.event.simulate() event has no originalEvent", function( assert ) {
+	expect( 1 );
+
+	var done = assert.async(),
+		input = jQuery( "<input>" )
+		.on( "click", function( event ) {
+			assert.strictEqual( "originalEvent" in event, false,
+				"originalEvent not present on simulated event" );
+			done();
+		} );
+
+	jQuery.event.simulate( "click", input[ 0 ], new jQuery.Event(), true );
+} );
+
+test( "Donor event interference", function( assert ) {
+	assert.expect( 4 );
+
+	jQuery( "#donor-outer" ).on( "click", function() {
+		assert.ok( true, "click bubbled to outer div" );
+	} );
+	jQuery( "#donor-input" ).on( "click", function( event ) {
+		assert.ok( true, "got a click event from the input" );
+		assert.ok( !event.isPropagationStopped(), "propagation says it's not stopped" );
+	} );
+	jQuery( "#donor-input" ).on( "change", function( event ) {
+		assert.ok( true, "got a change event from the input" );
+		event.stopPropagation();
+	} );
+	jQuery( "#donor-input" )[0].click();
+} );
+
 // This tests are unreliable in Firefox
 if ( !(/firefox/i.test( window.navigator.userAgent )) ) {
 	test( "Check order of focusin/focusout events", 2, function() {
--- a/test/index.html
+++ b/test/index.html
@@ -42,6 +42,11 @@
 	<!-- this iframe is outside the #qunit-fixture so it won't reload constantly wasting time, but it means the tests must be "safe" and clean up after themselves -->
 	<iframe id="loadediframe" name="loadediframe" style="display:none;" src="data/iframe.html"></iframe>
 	<dl id="dl" style="position:absolute;top:-32767px;left:-32767px;width:1px;">
+	<div id="donor-outer">
+		<form id="donor-form">
+			<input id="donor-input" type="radio" />
+		</form>
+	</div>
 	<div id="qunit-fixture">
 		<p id="firstp">See <a id="simon1" href="http://simon.incutio.com/archive/2003/03/25/#getElementsBySelector" rel="bookmark">this blog entry</a> for more information.</p>
 		<p id="ap">


test/unit/event.js
```diff
@@ -2678,6 +2678,43 @@ test( ".off() removes the expando when there's no more data", function() {
 	}
 });
 
+test( "preventDefault() on focusin does not throw exception", function( assert ) {
+	expect( 1 );
+
+	var done = assert.async(),
+		input = jQuery( "<input>" ).appendTo( "#form" );
```
gibson042: Let's self-close the input for that XHTML support that we keep telling ourselves will be verified someday.

test/unit/event.js
```diff
@@ -2678,6 +2678,43 @@ test( ".off() removes the expando when there's no more data", function() {
...
+		input
+			.on( "focusin", function( event ) {
+				var flawlessExecution = true;
+
+				try {
+					event.preventDefault();
+				} catch( theException ) {
```
gibson042: Missing space before `(`. Also, what would you think about capturing the exception and asserting strict equality with `undefined` (for a more valuable diff than `Expected: true Result: false` if this ever fails)?

test/unit/event.js
```diff
@@ -2678,6 +2678,43 @@ test( ".off() removes the expando when there's no more data", function() {
...
+					flawlessExecution = false;
+				}
+
+				ok( flawlessExecution, "Preventing default on focusin throws no exception" );
+
+				input.remove();
```
gibson042: No need for this; `#form` is inside the auto-resetting fixture.

test/unit/event.js
dmethvin: Agreed for sanitation and sanity reasons, although this goes through the `rsingleTag` path anyway.

src/event.js
```diff
@@ -610,8 +610,7 @@ jQuery.event = {
 			event,
 			{
 				type: type,
-				isSimulated: true,
-				originalEvent: {}
```
dmethvin: The comment above this says it was done to avoid a `.stopPropagation()` being in the simulated event and affecting the donor event. There still seem to be places in compat that would be broken by this, for example https://github.com/jquery/jquery/blob/compat/src/event.js#L897-903 when the donor event is `click` and the simulated event is `change`, with a change handler calling `event.stopPropagation()` actually stopping propagation on the original click event.

timmywil: But if there is no originalEvent attached, that wouldn't happen, right?

timmywil: Or are you saying that should be allowed?

dmethvin: For the case of a synthetic event created via `jQuery.Event`, there won't be an `originalEvent` property. For a native event created that way, [one will be there](https://github.com/gabrielschulhof/jquery/blob/2826dfc65b6c9b9968be3297e6ae01cfced521c8/src/event.js#L643) and it might get used when it shouldn't.

That said, I don't see a fail in the compat branch for this so I'm going to try and create one.

dmethvin: OK, here's a test that fails in IE8: https://jsfiddle.net/74eutos8/embedded/result/

timmywil: Ah, I see, we want to override the original change event. What about removing it instead of replacing it?

gabrielschulhof: Could we not simply ```e.originalEvent = undefined;```? after doing the extend? That would explicitly sever the connection to the donor event's ```originalEvent```.

OTOH, considering all the changes involved, it may simply be easier to revert https://github.com/jquery/jquery/commit/a873558436383beea7a05fd07db7070a30420100

dmethvin: Removing `.originalEvent` should do the trick I think, or the previous proposal of using a new `jQuery Event` as well. It would depend on whether the event handler functions that are attached by jQuery Mobile or its clients expect an `event.originalEvent` to be there when they are called.

Note that the whole special event system has very few user-serviceable parts and `jQuery.event.simulate` isn't documented. It was primarily built as a way to normalize browser behaviors inside jQuery core, not as a public extension mechanism.

gibson042: `null`ing `originalEvent` like @gabrielschulhof suggests fixes the test: https://jsfiddle.net/74eutos8/1/embedded/result/

timmywil: I like removing `.originalEvent`. If we can use `delete` without any browsers spitting up, I'd like that best.

timmywil: e.g. http://jsfiddle.net/timmywil/74eutos8/2/show/

gabrielschulhof: My gut tells me we should completely eradicate ```originalEvent``` rather than leaving it part of the structure, i.e.,

```JS
if ( "originalEvent" in e ) {
  delete e.originalEvent;
}
```

because that would be consistent with the way a native event would look. Nevertheless, I don't have enough experience to be able to decide which way of getting rid of it is best. Either way, it looks like something reasonable (like null or undefined) and falsy works.

dmethvin: I'm for the `delete` route as well. I don't think we'd need the guard as it shoudl just be a no-op there.

timmywil: You shouldn't need the check. `delete` should still work fine even when the property doesn't exist. That said, it will always exist in this case.

test/unit/event.js
gabrielschulhof: Good idea!

src/event.js
markelog: This property is a "documented" one - http://api.jquery.com/on/, also there might be some use-cases when `originalEvent` could be useful. 

I'd say commit that started this whole thing is just to risky.

timmywil: > there might be some use-cases when originalEvent could be useful

Like what, tho? This won't remove it for the majority of cases.

dmethvin: The only cases where this method is even used is inside special event hooks, so very few events are fired using it. I agree that there could be cases where a realistic `originalEvent` might be useful, for example when simulating Pointer events from Mouse events. But the old behavior of creating an empty object wouldn't be good enough for that case anyway.

markelog: I'm afraid there could be more cases that we unaware of, never used `originalEvent` though, so i can't judge here.
It is a documented property, plus having it in one event and not in the other, seems inconsistent.

Can we consider reverting problematic commit instead? Seem less breaky.

dmethvin: I'm fine with doing that as well. @timmywil ?

timmywil: > Can we consider reverting problematic commit instead? Seem less breaky.

But we don't even know what it's breaking. There's not enough utility to keep it, especially as an empty plain object.

timmywil: I don't know. I guess I'd be fine with reverting the commit, but I wish we had more information.

markelog: >  I guess I'd be fine with reverting the commit, but I wish we had more information.

Yeah, me too, but we can say consistency could be an argument for reverting it.

dmethvin: So the commit from #2125 that brought this up wasn't a critical thing, it was just @markelog finding a guard he thought wasn't needed. We discussed it quite a bit there and *thought* there weren't any cases where it was needed, but we were wrong. So we either need to back out the original change or make further changes that we think would *really* prevent it from being needed.

scottgonzalez: I don't see any need to keep it. `originalEvent` is useful and should definitely stay around, but I can't think of any existing use cases for `originalEvent` in simulated events since they've been empty objects for the past four years.

timmywil: > but we can say consistency could be an argument for reverting it.

We can make a lot of unnecessary changes for the sake of consistency. I still think I'd prefer to delete here.

markelog: Sounds like judgement call to me :-), in any case, we will have beta first, if something breaks we will know about.

gabrielschulhof: Fundamentally (in my eyes, at least): Are ```.preventDefault()```, ```.stopPropagation()```, and ```.stopImmediatePropagation()``` defining properties of events? Are we as jQuery specifying that they are? If so, then everywhere an event appears it must be a well-formed event.

Equally fundamentally: Must every event.originalEvent. ... .originalEvent chain end in a native event? Currently, the answer would seem to be "no", since such a chain, constructed on top of a simulated event, is rooted in an empty object.

timmywil: To the answer the first question, yes, but I don't know of another place that isn't true. An instance of jQuery.Event is a normalized event object with those methods.

As for the second, if all goes well, maybe all we need is a clarification in the docs. Something like, "The native event object is available at `event.originalEvent`, if the original event is not a simulated event (e.g. "focus")."

scottgonzalez: > Are .preventDefault(), .stopPropagation(), and .stopImmediatePropagation() defining properties of events?

I'd say yes.

> Must every event.originalEvent. ... .originalEvent chain end in a native event?

Nope. But not because it can end in an empty object right now. It's because you can call `jQuery.Event()` with any event, even one that started from jQuery.

scottgonzalez: > As for the second, if all goes well, maybe all we need is a clarification in the docs. Something like, "The native event object is available at event.originalEvent, if the original event is not a simulated event (e.g. "focus")."

But that's never true for any widget events in jQuery UI. A native event *may* exist, but if it does, it's at least two levels deep. For example, `menufocus` could have an original event of `focus` (via `jQuery.Event()`) which itself has an original event of `focus` (via a native event).

timmywil: Ok, but we can change the language.

markelog: I think we really should, so if choose this direction we would need a corresponded docs ticket

dmethvin: As far as docs, the page for `.on()` already says that for native events there's an `originalEvent` property. The only docs clarification there I can think of is to add an `Event.originalEvent` page and make it explicitly clear that the property may or may not be there. The concept of an "event chain" isn't something that jQuery Core depends upon, it's just an accident of the way `jQuery.Event` works.

If you look at the problem from the standpoint of how well `.simulate()`d a *native* event is, then the best approach seems to use `new jQuery.Event(type)` because no natively generated event would be missing an `originalEvent`. However, the only place we're simulating events in master is for focusin/focusout, we've never gotten a report of problems the way it's being done now, and this whole simulation mess is hopefully not going to be used a lot in the future. These are all really rare edge cases so I doubt it really matters what we do here--or if it does matter there is no way we'll find out except to change the code.

gabrielschulhof: Maybe I'm losing track of the discussion, but we do need to make at least one code change, since, as it stands, preventing default on a simple ```focusin``` event causes an exception. IIUC we can make one of three changes:
  0. Revert @markelog's commit - this would cause us to violate the principle that all events must be well-formed, but that's OK, because it's never caused a problem before. OTOH, if we go this route, we cannot update the docs page as @dmethvin suggested, because it won't be that that "the property may or may not be there", but that it may or may not be a well-formed event.
  1. ```e.originalEvent = new jQuery.Event();``` - IINM this does not require a docs update, but degrades event performance, as @gibson042 [pointed out](http://irc.jquery.org/%23jquery-dev/default_%23jquery-dev_20150512.log.html#t05:49:49).
  2. ```delete e.originalEvent;``` - This requires the docs update mentioned by @dmethvin.

markelog: > As far as docs, the page for .on() already says that for native events there's an originalEvent property.

I see only 

> This object is a normalized subset of data provided by the browser; the browser's unmodified native event object is available in event.originalEvent.

Don't see "native" there, not sure if that phrase suppose to warn user about simulated event. In any case, focus(in | out) do exist in most browsers and in IE it would exist for real. User might not even know we shim it.

> but degrades event performance, as @gibson042 pointed out.

I doubt it, event processing is pretty hot already, event creation should not reflect on it.

gibson042: Given our current uses of simulate, it's true that creating instantiation would not have substantial performance impact. But I still don't see the value in creating an object with no useful properties. How is `jQuery.Event{ type: undefined, timeStamp: <Date>, [jQuery.expando]: true }` better than `{}`, and how is `{}` better than `undefined`?

markelog: > How is jQuery.Event{ type: undefined, timeStamp: <Date>, [jQuery.expando]: true } better than {}, and how is {} better than undefined?

Same way, i guess, it is not better then removing that `originalEvent` property, i'm for the reverting, but only because of the consistency, even if it is pointless, which i'm not totally sure it is.

For example, in standart event model of the DOM you can `preventDefault` or `stopPropagation` focus event, but it is not preventable nor it bubbles, but functions are still there as per spec - http://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-Event-preventDefault

So "If an event is cancelable, the preventDefault method is used to signify that the event is to be canceled...  Calling this method for a non-cancelable event has no effect".

You know it always there, you know it always be consistent no matter what type of event it is, for me this is sounds as a good policy.  

Although this situation is different since user might actually need access to the real event. And as i said - 

>  In any case, focus(in | out) do exist in most browsers and in IE it would exist for real. User might not even know we shim it.

gibson042: I pretty much agree, and that looks like a solid argument against `originalEvent: {}`. So now we're left with:
* `originalEvent: new jQuery.Event()`
* `originalEvent: null`
* `e.originalEvent = undefined`
* `delete e.originalEvent`

All are compatible with a873558436383beea7a05fd07db7070a30420100, and the first is compatible with existing documentation.

dmethvin: I think if the property exists at all it should be valid, so that rules the middle two. The best fidelity would be to go with creating a new empty event, just in case some user event handler saw our simulated native event and expected originalEvent to be there. In reality it's not likely to matter and the stubby little event won't have any useful properties so it won't look much like a native event. It's almost as if we're putting one there just to avoid errors.

markelog: Oh, i thought it was clear where i'm going with this -

```js
orignalEvent: event.originalEvent
```

I think this is a best course of action.

gabrielschulhof:
@arschmitz @timmywil the event-related breakage in the mobile tests ensues from the fact that we
```JS
// mock originalEvent
$.Event.prototype.originalEvent = {
  touches: [ { pageX: 0, pageY: 0 } ]
};
```
... and stuff like that. If we mock ```originalEvent```, it is our responsibility to mock it correctly.

src/event.js
dmethvin: I think that brings us back to the problem we were trying to avoid, which is that the "donor" event gets affected. That is why we wanted either a different empty event or no `originalEvent`.

timmywil: I think if it's not a native event, or even represents a native event, it
probably shouldn't be there, or we get into how we're going to simulate
native events. Consistency is not a strong enough argument to do that.

dmethvin: The only place we use it--only one place in master--is with native events. It's an undocumented non-public API so our concern was just with internal use cases.

timmywil: Yea, that's my point. In this case, we're trying to stub something that does not exist natively and, as far as we know, no one uses. We used a plain object before, but it seems to me that nothing we add would be a truly accurate stub, so we may as well `delete` it.

markelog: Yep, but on with trajectory we don't need to do anything, since native `preventDefault` and `stopPropagation` for focus/blur event are noop. So nothing to see here.

In compat we have simulated bubble `change` and `submit`, but we can add guard for `isSimulated`  property in our `stopPropagation` method (didn't test that).

> We used a plain object before, but it seems to me that nothing we add would be a truly accurate stub, so we may as well delete it.

Not sure why we would delete it, property explicitly called `originalEvent` so it seems putting there original event is right way to do it.

Why not be consistent and add more bytes to this?

markelog: > I think if it's not a native event, or even represents a native event

With my proposal it would be a native event.

timmywil: > With my proposal it would be a native event.

Yes, but wouldn't that allow for this...

> with a change handler calling event.stopPropagation() actually stopping propagation on the original click event.

markelog: Nuh, that is the thing it wouldn't, see - https://github.com/jquery/jquery/pull/2303/files#r30607068

timmywil: Oh, I see. Add an `isSimulated` property?

markelog: Basically, it should look something like this - 
```js
stopPropagation: function() {
...
if (this.isSimulated) return;
}
```

need a good test for it though

markelog: But only in compat

gabrielschulhof: ... but after the simulated event processing is done, you would have to remove the ```isSimulated``` property, otherwise the donor event processing will be done the same way as the original event processing. So, like:

1. Receive native FocusEvent ```nativeFocusEvent```
2. ```nativeFocusEvent.isSimulated = true;``` as part of ```jQuery.event.simulate()```
3. Process the simulated event.
4. Trigger the actual focus event. If we don't ```delete nativeFocusEvent.isSimulated```, then, during the processing of the actual focus event, ```jqueryFocusEvent.originalEvent.isSimulated``` will also be true, because we didn't remove the isSimulated property after the processing of the simulated event.

markelog: Not sure if i follow, could you show some code, test-case?

gabrielschulhof: @markelog

https://gist.github.com/gabrielschulhof/b09c9e34bfc0904916ef

scottgonzalez: This discussion seems to just keep dragging on, but as far as I can tell, nobody has provided a single reason that the property needs to, or even should exist, based on a real use case.

markelog: @gabrielschulhof i guess code worth i thousands words - https://github.com/jquery/jquery/pull/2335, https://github.com/jquery/jquery/pull/2336

@scottgonzalez 
1. We wouldn't need to change docs
2. We should be back-compat if possible
3. We shouldn't add needless logic & bytes
4. User could explore new properties of the original event when they comes up, without for us to do anything or change our stub
5. There is already possible use-cases, users don't use them because they can't which doens't mean they aren't useful - http://jsfiddle.net/rs5jn1bq/
6. Consistency, this should be important and it correlates with basic concepts of the DOM events described in the spec

scottgonzalez: > 1. We wouldn't need to change docs

What docs? This is for a simulated event, I don't believe there's anything special in the docs about this.

> 2. We should be back-compat if possible

There's no possible use case to be back compatible with. Did someone provide an actual scenario that I missed?

> 3. We shouldn't add needless logic & bytes

You're concerned about the delete over setting to an empty object in terms of logic and bytes?

> 4. User could explore new properties of the original event when they comes up, without for us to do anything or change our stub

That's impossible, considering it's been an empty object forever.

> 5. There is already possible use-cases, users don't use them because they can't which doens't mean they aren't useful - http://jsfiddle.net/rs5jn1bq/

And are the properties legitimate considering they're not even from the correct event type?

> 6. Consistency, this should be important and it correlates with basic concepts of the DOM events described in the spec

Consistency with what?

markelog: > What docs? This is for a simulated event, I don't believe there's anything special in the docs about this.

See above.

> There's no possible use case to be back compatible with. Did someone provide an actual scenario that I missed?

This is exposed documented API, according to the semver this would be a breakage, therefore it could break real code even if it is an empty object.

> You're concerned about the delete over setting to an empty object in terms of logic and bytes?

See alternative pulls.

> That's impossible, considering it's been an empty object forever.

My argument not about past but about future - see alternative pulls

> And are the properties legitimate considering they're not even from the correct event type?

Why not?

> Consistency with what?

With other events and with other environments where we don't need to shim those events

Issue #2192, opened 2015-04-08
.toggleClass( className, function )


redexp:
How about to add new api for `toggleClass` where first argument is class name and second is function which should return boolean? 

For example
```html
<span data-tab="1">One</span>
<span data-tab="2">Two</span>
<span data-tab="3">Three</span>
```

```javascript
var currentTab = 2;
$('[data-tab]').toggleClass('hidden', function (i, span) {
  return $(span).data('tab') !== currentTab;
});
```

I was surprised why this is not implemented yet? Almost all methods do accept second argument as function, but `toggleClass` not.

AurelioDeRosa:
I understand your position but I'm against it if this will increase the weight of jQuery too much. I haven't seen other people complaining about this feature, so I guess this is something not many people wanted.

dmethvin:
Function args for every method don't really make code a lot shorter or easier to understand. You can just use `.each()` for those cases.
```js
var currentTab = 2;
$('[data-tab]').each(function (i, span) {
  $(this).toggleClass("hidden", $(this).data('tab') !== currentTab);
});
```

redexp:
@dmethvin we can use `.each()` everywhere, but why than we need `.attr|prop|css(string, function)`? Also I think `.toggleClass(className, function)` is more useful than `.toggleClass(function, state)`

@AurelioDeRosa I think to implement this we need just 3 lines of code, not too much

dmethvin:
@redexp I don't think we *do* need it, and wish it wasn't there because it requires more code inside jQuery than it is worth. It also can be a performance killer because of the way it was defined, for example, [`.html(function)`](http://api.jquery.com/html/#html-function) is passed the current HTML content which may not be needed at all but can be very expensive to retrieve.

sqqihao:
@redexp, you can rewrite the jquery with you favor;

redexp:
I know guys, I can do all that stuff without any problems, I'm just talking about API *consistency* with attr, prop and css

dmethvin:
API consistency is how we got into this mess. Something should only be added to the API if it's making some operation easier, not just because some other API has it.

redexp:
Well if everybody think like that, then we can close the issue

Mr21:
For the defense of @redexp, the function `access` (in the core part) don't do that? normally this function save lines everywhere in many case where the user give a function etc.?

dmethvin:
[...]

PR #2306, opened 2015-05-13
Attributes: toggleClass(string [,function])

jdphenix:
Implementation for #2192

### DIFF
--- a/src/attributes/classes.js
+++ b/src/attributes/classes.js
@@ -114,22 +114,31 @@ jQuery.fn.extend({
 		}
 
 		return this.each(function() {
-			var className, i, self, classNames;
+			var className, i, self, classNames, predicate;
 
 			if ( type === "string" ) {
 
 				// Toggle individual class names
 				i = 0;
 				self = jQuery( this );
+				predicate = jQuery.isFunction( stateVal ) ? stateVal.call( self, i, self) : null;
 				classNames = value.match( rnotwhite ) || [];
 
 				while ( ( className = classNames[ i++ ] ) ) {
-
-					// Check each className given, space separated list
-					if ( self.hasClass( className ) ) {
-						self.removeClass( className );
+					if ( typeof predicate === "boolean" ) {
+						if ( predicate ) {
+							self.addClass( className );
+						} else {
+							self.removeClass( className );
+						}
 					} else {
-						self.addClass( className );
+
+						// Check each className given, space separated list
+						if ( self.hasClass( className ) ) {
+							self.removeClass( className );
+						} else {
+							self.addClass( className );
+						}
 					}
 				}
 
--- a/test/unit/attributes.js
+++ b/test/unit/attributes.js
@@ -1328,6 +1328,34 @@ test( "toggleClass(Function[, boolean]) with incoming value", function() {
 	ok( !e.is(".test"), "Assert class not present" );
 });
 
+test( "toggleClass(String[, Function])", function() {
+	expect( 9 );
+
+	var el = jQuery("<div><span data-tab='1'>One</span><span data-tab='2'>Two</span><span data-tab='3'>Three</span></div>"),
+		elChildren = el.children(),
+		currentTab = 2,
+		testFn = function( i, e ) {
+			return jQuery(e).data("tab") !== currentTab;
+		};
+
+	ok( !elChildren.is(".test"), "Assert class not present" );
+	ok( !testFn(0, jQuery("<span data-tab='2'></span>")), "Test function true when expected false" );
+	ok( testFn(0, jQuery("<span data-tab='1'></span>")), "Test function false when expected true" );
+
+	elChildren.toggleClass( "test", testFn );
+
+	ok ( el.children(":nth-child(1)").hasClass("test"), "Assert first child has test class" );
+	ok ( !el.children(":nth-child(2)").hasClass("test"), "Assert second child doesn't have test class" );
+	ok ( el.children(":nth-child(3)").hasClass("test"), "Assert third child has test class" );
+
+	currentTab = 1;
+	elChildren.toggleClass( "test", testFn );
+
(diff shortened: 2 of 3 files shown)


jdphenix:
I clearly completely messed this up initially, fixing it

jdphenix:
Okay, git shenanigans all done with

src/attributes/classes.js
```diff
@@ -114,22 +114,31 @@ jQuery.fn.extend({
...
-
-					// Check each className given, space separated list
-					if ( self.hasClass( className ) ) {
-						self.removeClass( className );
+					if ( typeof predicate === "boolean" ) {
+						if ( predicate ) {
+							self.addClass( className );
+						} else {
+							self.removeClass( className );
+						}
```
mzgol: You can use `self.toggleClass( className, predicate )` here; similar modification can be done below. This will make the whole change smaller.

test/unit/attributes.js
```diff
@@ -1328,6 +1328,34 @@ test( "toggleClass(Function[, boolean]) with incoming value", function() {
 	ok( !e.is(".test"), "Assert class not present" );
 });
 
+test( "toggleClass(String[, Function])", function() {
+	expect( 9 );
+
+	var el = jQuery("<div><span data-tab='1'>One</span><span data-tab='2'>Two</span><span data-tab='3'>Three</span></div>"),
```
mzgol: Add a space after `(` and before `)` please.

test/unit/attributes.js
```diff
@@ -1328,6 +1328,34 @@ test( "toggleClass(Function[, boolean]) with incoming value", function() {
...
+		elChildren = el.children(),
+		currentTab = 2,
+		testFn = function( i, e ) {
+			return jQuery(e).data("tab") !== currentTab;
```
mzgol: more spaces needed (as above)

test/unit/attributes.js
mzgol: Add a space after `(` and before `)` please. This applies to most of the test.

test/unit/attributes.js
mzgol: This test is dependent on the `data` module now. I think it should be independent; please switch this to a regular `getAttribute('data-tab')`.

test/unit/attributes.js
```diff
@@ -1328,6 +1328,34 @@ test( "toggleClass(Function[, boolean]) with incoming value", function() {
...
+		};
+
+	ok( !elChildren.is(".test"), "Assert class not present" );
```
mzgol: `hasClass( ".test" )` would be more explicit. Also, `is` is defined in the `traversing` module, we should try to stay in `attributes`.

jdphenix:
@mzgol Okay, seems straightforward enough. I'll have an updated commit here in a couple of days. Thanks for the feedback.

mzgol:
Hi @jdphenix, do you think you'll have time to finish it in the next few weeks? :) Just checking.

jdphenix:
Sure. Got a little busier than I expected.

mzgol:
@jdphenix Thanks. This is nothing urgent, we don't have to make it ready for the first alpha.

jdphenix:
Just checking in, still alive here.

mzgol:
No worries. Whenever you have time.

src/attributes/classes.js
```diff
@@ -114,22 +114,31 @@ jQuery.fn.extend({
...
 		return this.each(function() {
-			var className, i, self, classNames;
+			var className, i, self, classNames, predicate;
 
 			if ( type === "string" ) {
 
 				// Toggle individual class names
 				i = 0;
 				self = jQuery( this );
+				predicate = jQuery.isFunction( stateVal ) ? stateVal.call( self, i, self) : null;
```
Mr21: I didn't exactly understood this line but normally jQuery send the DOM element itself, not a jQuery object.
And `i` will be always `0` no?

dmethvin: To the point @Mr21 makes about `i = 0`, that does seem wrong. In other places where we allow functions, the function is called for each element.  Here we are only calling it once and saving the result.

dmethvin: Can we write the docs for this before landing it? I can see several different interpretations. Just for review, [we already have these](http://api.jquery.com/toggleClass/):
```js
toggleClass(String classNames)
toggleClass(String classNames, Boolean state)
toggleClass(Boolean state)
toggleClass(Function(index, currentClassNames, state), Boolean state)
```
In all cases the `state` is optional. so the proposal is to add `.toggleClass(String classNames, Function)` to the list. I can see that working several ways:

1. Function called once for the whole collection, given the incoming `jQuery` object collection as `this` and the String list of classes as an arg, returning a Boolean. In essence the function is just returning the `state` that `toggleClass(String classNames, Boolean state)` uses, so this option has that going for it. I think we want a return of `undefined` to specify the default toggling behavior?
2. Function called once for each element, given the element as `this` and the index as an arg. This is close to what the code above does, assuming we use the index from the `.each()` above it instead of the `i`. However, that means the function doesn't see the classes being toggled, wouldn't we also want to pass them in? The current PR already allows `undefined` to specify the default toggling behavior.
3. Function called for each combination of element and className, given the element as `this` and the index and className as args. More function calls but avoids the need for an explicit loop inside the called function if iterating over the classes is a common need. Again with the `undefined` for toggling.

Also note that undocumented variants like `toggleClass(Function, Function)` are going to start doing something now due to the way we look at `state` in the code, I don't know if those should be cause for concern or documentation.

mzgol: @dmethvin Good points. @jdphenix, would you want to try to document this thing? The docs live in https://github.com/jquery/api.jquery.com/

markelog:
@jdphenix friendly ping?

dmethvin:
I'm going to close this PR, the `.toggleClass()` method has changed a lot since we removed the save/restore Boolean signature so at minimum it will need some rewriting. I think we're still open to a proposal here but need to know what the method will be documented to do, per my comment above.

PR #2325, opened 2015-05-18
Attributes: Disable SVG testing on IE 8

mzgol:
Refs 20aaed367f993f3c2aa204183d82d0d73efa114f
Refs gh-2268

cc @timmywil 

See http://swarm.jquery.org/result/303863


mzgol:
OK, I see https://github.com/jquery/jquery/commit/57fb2dc02ec286c51b456f0e460ad258f332b6ac landed and is more to the point, closing.

PR #2339, opened 2015-05-21
Tests: Change quotes according to style guidelines

64lines:
Modifies the quotes in the test files according to the style guidelines explained here: http://contribute.jquery.org/style-guide/js/#quotes

### DIFF
--- a/test/data/offset/absolute.html
+++ b/test/data/offset/absolute.html
@@ -18,8 +18,8 @@
 		<script src="../../jquery.js"></script>
 		<script type="text/javascript" charset="utf-8">
 			jQuery(function($) {
-				$('.absolute').click(function() {
-					$('#marker').css( $(this).offset() );
+				$(".absolute").click(function() {
+					$("#marker").css( $(this).offset() );
 					var pos = $(this).position();
 					$(this).css({ top: pos.top, left: pos.left });
 					return false;
--- a/test/data/offset/body.html
+++ b/test/data/offset/body.html
@@ -12,8 +12,8 @@
 		<script src="../../jquery.js"></script>
 		<script type="text/javascript" charset="utf-8">
 			jQuery(function($) {
-				$('body').click(function() {
-					$('#marker').css( $(this).offset() );
+				$("body").click(function() {
+					$("marker").css( $(this).offset() );
 					return false;
 				});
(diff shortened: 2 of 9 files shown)


markelog:
It seems we can `rm` those files.

/cc @timmywil

timmywil:
Sure! Are they not used?

markelog:
As i see they not used in testsuite, it looks like their function is to open them manually, but honestly, i didn't even know we had them.

timmywil:
What about moving them to an `integration` folder? Eventually, we can automate running all pages in that folder using something like webdriver. The tests would still need adjustment, but we can get to that later.

mzgol:
Note that `compat` already has the `test/integration` directory, currently containing just one test: https://github.com/jquery/jquery/tree/42ea7468250f1b612684a71b8643ae25367ff469/test/integration

markelog:
> What about moving them to an integration folder?

Would still need to rewrite them, for example, there is page that uses `live()` in there. I'm not sure what we support in there.

timmywil:
Good point, but, like I said, we can save the rewrites/removals for later.

markelog:
So @lexinerus would like to do that? Make sure those tests are still relevant and if they are, rewrite them and move to `integration` folder?

jzaefferer:
The CLA check is fine now.

mzgol:
@markelog @timmywil I'd rather not have files not working with current jQuery in the `integration` folder. Perhaps in `integration/old` or `integration/legacy`, or `integration/to-rewrite` so that it's clear they're not expected to work currently?

timmywil:
We don't run these tests very often. I don't think we need the reminder
with a folder name, just with an issue that tells us to rewrite these tests
and run them before final release.

markelog:
I guess we need a .md doc for the release procedure, or we already have one? Since we doing more then just running one command

mzgol:
> I guess we need a .md doc for the release procedure (or we already have one?). Since we doing more then just running one command

https://github.com/jquery/jquery/wiki/Release-process

markelog:
Ah, thanks, okay so we would need to add another item there when we deal with update these tests

timmywil:
Item added.

Issue #2037, opened 2015-01-26
Feature request: parse url querystring


rasca:
I'm sure I'm not alone here, there are dozens of questions asking how to parse the url querystring with jQuery. There are a lot of answers but most of them don't accept multi valued keys or keys without values e.g. `?a=1&a=2&c&d`.

I end repeating code in several of my projects, I think this could be included in jQuery core.

matryer:
Maybe embedding something like: https://github.com/stretchr/arg.js

dmethvin:
This is a pretty simple plugin. Look at the size of arg.js, we definitely wouldn't want to increase the size of jQuery that much for something that already has a solution available for those who need it.

PR #2352, opened 2015-05-27
add url module first

wenber:
we want to jquery has a module to solve functions about url


mzgol:
Thanks for the PR but this would better work as a plugin, we don't want to add such new functionality to Core currently. See #2037 for a previous request.

wenber:
thank you!

PR #2358, opened 2015-05-28
Event: remove preDispatch hook & simplify "simulate" signature

markelog:
I bet there is more like that in there

/cc @dmethvin

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -370,11 +370,6 @@ jQuery.event = {
 		args[0] = event;
 		event.delegateTarget = this;
 
-		// Call the preDispatch hook for the mapped type, and let it bail if desired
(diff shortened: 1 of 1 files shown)


timmywil:
It doesn't take up much room. Maybe we should document?

markelog:
> It doesn't take up much room. Maybe we should document?

Okay, so to not wiggle about with this pull, i will remove changes of `event.special` structure and just simplify the `simulate` method and create api ticket.

Sounds good?

timmywil:
:+1:

dmethvin:
Sounds good to me.

timmywil:
Yup, +1

markelog:
Yeah, yeah i got it :-), will land it soon

Issue #1965, opened 2014-12-22
Code coverage


markelog:
Evaluate and prototype implementation of code coverage tool. Maybe not even on Core at first, Sizzle?

/cc @timmywil, @mzgol, @gibson042, @jaubourg, @dmethvin

listepo:
+1

stevepeak:
> Full disclosure I'm the founder of Codecov

We would love to see jQuery use [Codecov](https://codecov.io) as its coverage provider. I would love to share about the many reasons why I believe Codecov are great fit and industry leading in code coverage. Anything I can do to help this transition I would make myself very available. Thank you!

gnarf:
> Anything I can do to help this transition I would make myself very available.

Help implement it in a PR and I'm sure we can work together on it!  A couple of questions:

* What does it take for us to use it? (we don't use travis)
* Can we use it in the jenkins / testswarm / qunit setup that we have?
* Do you have some example reports to look at, could you run it against jQuery yourself and link us to a report?

I'm sure others will have some more questions :)

stevepeak:
> What does it take for us to use it? (we don't use travis)

Any CI method works. We have two primary uploading methods for javascript: the [native method](https://github.com/cainus/codecov.io) and the [global method](https://github.com/codecov/codecov-bash). I'll look into creating a pull request, I'll mention you when I create it. Thank you @gnarf

> Can we use it in the jenkins / testswarm / qunit setup that we have?

Jenkins: yes. Qunit: yes. Testswarm: should (I have not used this before). We do merge coverage reports automatically and keep record of each one. [See here for example](https://codecov.io/gh/pyca/cryptography) (Click on the **Unified** dropdown).

> Do you have some example reports to look at, could you run it against jQuery yourself and link us to a report?

I can reference our customer repos, but I have not yet ran coverage against jQuery yet. I will try to attempt this during my pull request.

> I'm sure others will have some more questions :)

I'm very happy to help out!

gnarf:
> I'm very happy to help out!

Feel free to ping me (gnarf) in freenode IRC #jquery-dev with any questions/help you'd need from me here.

timmywil:
Closing in favor of tracking in the [Roadmap](https://github.com/jquery/jquery/wiki/Roadmap).

PR #2362, opened 2015-05-30
Ajax: Remove remnants of the load event alias handling

mzgol:
Refs 0705be475092aede1eddae01319ec931fb9c65fc
Refs gh-2287

cc @markelog

### DIFF
--- a/src/ajax/load.js
+++ b/src/ajax/load.js
@@ -4,22 +4,13 @@ define([
 	"../ajax",
 	"../traversing",
 	"../manipulation",
-	"../selector",
-	// Optional event/alias dependency
-	"../event/alias"
+	"../selector"
 ], function( jQuery ) {
-
-// Keep a copy of the old load method
-var _load = jQuery.fn.load;
 
 /**
  * Load a url into a page
  */
 jQuery.fn.load = function( url, params, callback ) {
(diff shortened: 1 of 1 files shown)


markelog:
How did you find this?

mzgol:
I just reminded myself that we had this duplicate logic and that I didn't see it removed in your PR so I found it. :)

mzgol:
Hmm, I guess the `"../event/alias"` dependency can be removed as well?

markelog:
> I just reminded myself that we had this duplicate logic

After a week it was landed? Your memory is an interesting thing :-)

mzgol:
> After a week it was landed? Your memory is an interesting thing :-)

Actually, it was because jsFiddle broke on Git jQuery as they used the `.load()` shortcut and I searched for your commit to present it to them as the reason. And then it hit me when I looked at the diff. :)

markelog:
> Actually, it was because jsFiddle broke on Git jQuery as they used the .load() shortcut and I searched for your commit to present it to them as the reason. And then it hit me when I looked at the diff. :)

:heart: :rage: both emotions there, we really need some tests for that kind of thing, i'd say we really need a code coverage tool here.

/cc @timmywil

timmywil:
Yup, it's on the list! #1965

Issue #2267, opened 2015-05-06
`isArrayLike` no longer fails gracefully for null and undefined from 1.11.1 to 1.11.3


ScottLNorvell:
in 1.11.1, isArrayLike looks like this:

```javascript
function isArraylike( obj ) {
	var length = obj.length,
		type = jQuery.type( obj );

	if ( type === "function" || jQuery.isWindow( obj ) ) {
		return false;
	}

	if ( obj.nodeType === 1 && length ) {
		return true;
	}

	return type === "array" || length === 0 ||
		typeof length === "number" && length > 0 && ( length - 1 ) in obj;
}
```

in 1.11.3, it looks like this:
```javascript
function isArraylike( obj ) {

	// Support: iOS 8.2 (not reproducible in simulator)
	// `in` check used to prevent JIT error (gh-2145)
	// hasOwn isn't used here due to false negatives
	// regarding Nodelist length in IE
	var length = "length" in obj && obj.length,
		type = jQuery.type( obj );

	if ( type === "function" || jQuery.isWindow( obj ) ) {
		return false;
	}

	if ( obj.nodeType === 1 && length ) {
		return true;
	}

	return type === "array" || length === 0 ||
		typeof length === "number" && length > 0 && ( length - 1 ) in obj;
}
```

`"length" in obj` produces the error `TypeError: Cannot use 'in' operator to search for 'length' in 6` when `obj === 6`.  Similarly when `obj` is a `undefined` or `null`. This causes, specifically `$.each` to fail when you accidentally put in a numerical value for `obj`. Not sure if this is an issue, but in 1.11.1, nothing failed when `obj` was accidentally a number. This caused some tests to fail that we've fixed. Thought I'd mention it.

dmethvin:
This is basically a dup of #2242, we don't guarantee that invalid inputs will be silently ignored.

gibson042:
Agreed.

timmywil:
@dmethvin is right, but I'm starting to wonder if it would be nice if we failed silently here. The string case was more obscure, but I would expect that changing the behavior for `null` and `undefined` will be much more problematic.

mzgol:
@timmywil agreed. Also, other implementations like lodash's `_.each` silently ignore things like `null` or `undefined`.

gibson042:
[...]

PR #2363, opened 2015-05-31
Core: .each should accept an `undefined` value

Mr21:
Hi :)

I found in JS we have this:
``` javascript
var n = 0;
n.length === undefined;
false.length === undefined;
true.length === undefined;
NaN.length === undefined;

// There is an exception here:
undefined.length; // crash
```

And with the actual jQuery, sometime we have to do this:
``` javascript
if (obj)
    $.each(obj, function() {
        // ...
    });
```

If we reorder few instructions in the `$.each` function (without more bytes) we can remove the necessity of the `if` above.

Do you agree?


timmywil:
Thank you for opening a PR. We've discussed this issue [here](https://github.com/jquery/jquery/issues/2267). It also affects `.map`. Would you be up for opening a PR for that too? I think `.each` and `.map` do need adjustments, since they check length before the call to `isArraylike`, but if it would save anything to make changes there as well, feel free.

mzgol:
Such a change requires a test, though.

Mr21:
Oh right, cool, I will do this (the other PR and the test), this evening.
Thanks :)

Mr21:
I added a test, but this test fail actually, because modifying `$.each` is not enough... I have to modified the `isArraylike` function, I change this in the PR of `.map` #2368.
Maybe you can merge the other before this one.

test/unit/core.js
```diff
@@ -871,6 +871,15 @@ test("jQuery.map", function() {
 	equal( result.join(""), "00012223", "Array results flattened (#2616)" );
 });
 
+test("jQuery.each(undefined, function) (PR #2363)", 1, function() {
+	try {
+		jQuery.each( undefined, jQuery.noop );
```
markelog: If this PR should fix #2267 you probably should add tests for all those values

mzgol:
@Mr21 a PR should close an issue if it changes behavior so perhaps you could consolidate the changes in one PR that would fix #2267?

markelog:
One PR is preferable, but we could do two atomic commit, one would ref, other would close

timmywil:
Sorry, I should have suggested one PR.

Mr21:
So, I closed the other PR about `.map`.
I restart this PR too, can you review this first (atomic) commit about the fix of `isArraylike`?

src/core.js
```diff
@@ -432,17 +430,20 @@ function(i, name) {
 });
 
 function isArraylike( obj ) {
+	var type, length;
+
+	if ( obj == null ||
+		( type = jQuery.type( obj ) ) === "function" ||
+		type === "number" ||
+		jQuery.isWindow( obj ) ) {
+		return false;
```
Mr21: Here the code said `false` for `null`, `undefined`, `"number"`, `NaN`, `"function"` only.

src/core.js
```diff
@@ -432,17 +430,20 @@ function(i, name) {
...
 	// `in` check used to prevent JIT error (gh-2145)
 	// hasOwn isn't used here due to false negatives
 	// regarding Nodelist length in IE
-	var length = "length" in obj && obj.length,
-		type = jQuery.type( obj );
-
-	if ( type === "function" || jQuery.isWindow( obj ) ) {
-		return false;
-	}
+	length = "length" in obj && obj.length;
```
Mr21: But here, the code fail with: `""`, `"string"` and `false`.
I don't really know if you want to considere the string like an array.

src/core.js
```diff
@@ -300,15 +300,13 @@ jQuery.extend({
...
-			} else {
-				push.call( ret, arr );
-			}
+		if ( isArraylike( Object(arr) ) ) {
+			jQuery.merge( ret,
+				typeof arr === "string" ?
+				[ arr ] : arr
+			);
+		} else if ( arr != null ) {
+			push.call( ret, arr );
```
gibson042: Was this rearranged only for smaller size?

src/core.js
```diff
@@ -300,15 +300,13 @@ jQuery.extend({
 	makeArray: function( arr, results ) {
 		var ret = results || [];
 
-		if ( arr != null ) {
-			if ( isArraylike( Object(arr) ) ) {
```
gibson042: Please add spaces around `arr`.

src/core.js
```diff
@@ -300,15 +300,13 @@ jQuery.extend({
...
-				jQuery.merge( ret,
-					typeof arr === "string" ?
-					[ arr ] : arr
-				);
-			} else {
-				push.call( ret, arr );
-			}
+		if ( isArraylike( Object(arr) ) ) {
```
gibson042: Please add spaces around `arr`.

src/core.js
gibson042: This is very hard to read. Please assign `type` in its declaration, and remove the `type === "number"` check completely. Also, it may be smaller to assign `length` in its declaration as well (e.g., `length = obj != null && "length" in obj && obj.length`).

src/core.js
gibson042: Those failures are acceptable, because strings and booleans would never end up here from valid function input.

gibson042:
This PR needs a reference to #2267 and unit tests for passing `null`/`undefined` to `jQuery.each` and `jQuery.map` (and corresponding changes to those methods so they don't throw on accessing `.length`).

src/core.js
Mr21: Hmm yes but also because the `arr != null` is checked already in `isArraylike` now.

gibson042: That's irrelevant here, though, because `isArrayLike` will see `Object(arr)` (which is never `null` or `undefined`) instead of `arr`.

Mr21: Oh right, it's because yesterday I tried to remove the call of `Object(...)` but it failed with `"string"` after :/

Mr21:
@gibson042, yeah I know I did it yesterday but **markelog** spoke about two atomic commits.
Have I to make only two commits here or it's you after who are reorganize all the code of the PR?
It's the reason why I didn't add my unit test today.

gibson042:
Just add relevant commits for both `jQuery.each` and `jQuery.map` to this branch, which will close an updated #2267 when it lands. `isArraylike` will only need changes as necessary to support the newly-valid `null` and `undefined` input for those public methods.

Mr21:
@gibson042 all is done :)

mzgol:
@Mr21 Could you squash your commits and add `Fixes gh-2267` to the commit message? I modified its title so that it doesn't mention numbers which are not officially handled. See d471842b3e3af83c9a1be06b5d16f75bfa96af8c for an example commit message.

Mr21:
@mzgol, right, it's done :)

src/core.js
mzgol: @Mr21 did you check this rearrangement decreases the gzipped minified size?

test/unit/core.js
```diff
@@ -1190,6 +1190,18 @@ test("jQuery.each(Object,Function)", function() {
 	equal( i, document.styleSheets.length, "Iteration over document.styleSheets" );
 });
 
+test("jQuery.each/map(undefined/null,Function)", 1, function() {
```
mzgol: You mention `Function` here but the test is only about `undefined/null`.

Mr21: Oh it's because I am testing :
``` javascript
$.each( null || undefined, function() {
});
```

Like the title of the precedent test ( *jQuery.each(Object,Function)* )

mzgol: Ah, it's a comma, not a slash. OK, sorry. :)

src/core.js
Mr21: Nop I didn't check (I can't now :( ), but in resume the changement is:
``` javascript
if(){if(){}else{}}  // before
if(){}else if(){}   // after
```

mzgol: You can invoke `grunt` on the commit before your PR and then on your PR; at the end you'll see the summary with byte difference for the gzipped case.

mzgol: You can invoke `grunt` on the your PR and then on your PR with block order swapped back. You'll see byte difference summary at the end.

Mr21: @mzgol I will do it asap, I discover sometimes the gzipped version is bigger while there are few less caracteres, *magic gzip*  ^^

Mr21: @mzgol, so with the actual change we have:
```
   raw     gz Sizes
255363  75554 dist/jquery.js
 84483  29366 dist/jquery.min.js

   raw     gz Compared to master @ 8f13997e89ca325a49f7581fad7e85fe37bad166
   -33     +3 dist/jquery.js
    -3     +6 dist/jquery.min.js
```
After, if I re-add the additionnal `if` in the `makeArray` function we have:
```
   raw     gz Sizes
255379  75556 dist/jquery.js
 84485  29364 dist/jquery.min.js

   raw     gz Compared to master @ 8f13997e89ca325a49f7581fad7e85fe37bad166
   -17     +5 dist/jquery.js
    -1     +4 dist/jquery.min.js

   raw     gz Compared to each-undefined @ 2d1766f83adb5f9032e2645f5ac55268c566e75b
   +16     +2 dist/jquery.js
    +2     -2 dist/jquery.min.js

   +16     +2 dist/jquery.js
    +2     -2 dist/jquery.min.js
```

mzgol: @Mr21 OK, so it seems it was slower in the previous version, wasn't it? So changing it back would decrease the final size and would make the diff smaller - it seems like a good thing to do.

Mr21: @mzgol Okay, I've rollback this change

mzgol:
LGTM now.

mzgol:
Landed, thanks!

test/unit/core.js
FagnerMartinsBrack: This landed and nobody noticed that it should have used `expect()` instead of the deprecated argument =/

markelog: Oh the humanity! How can we possible recover from this horrible-horrible mistake?! Sorry, could resist. 

We will fix all of that occurrences at some point.

Issue #2301, opened 2015-05-12
Remove internal args to .on & .remove


mzgol:
We shouldn't be exposing our private APIs, even if it's just via undocumented additional params.

I found 2 such occurrences: `.on` & `.remove`.

dmethvin:
I agree. We got rid of several of these earlier, not sure why these stayed around.

mzgol:
PR: #2366

PR #2366, opened 2015-06-01
Remove internal args to on & remove methods

mzgol:
Fixes gh-2301

+24 bytes, most of which come from event changes. The manipulation change adds just 5 bytes.

I've also tried to move `on` to a proper `/var/` directory but it only increased the size further.

Any ideas on how to reduce the size increase?

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -196,6 +196,27 @@ function domManip( collection, args, callback, ignored ) {
 	return collection;
 }
 
+function remove( elem, selector, keepData ) {
+	var node,
+		nodes = selector ? jQuery.filter( selector, elem ) : elem,
+		i = 0;
+
+	for ( ; (node = nodes[i]) != null; i++ ) {
+		if ( !keepData && node.nodeType === 1 ) {
+			jQuery.cleanData( getAll( node ) );
+		}
+
+		if ( node.parentNode ) {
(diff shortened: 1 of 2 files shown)


timmywil:
Nice :+1:

timmywil:
I wonder if it would gzip better if we did `on.call(this, ...`

mzgol:
> I wonder if it would gzip better if we did on.call(this, ...

In `manipulation` switching to `.call` adds 11 bytes, in `event` it adds 3 bytes. I tried that at the beginning but I discovered it's actually larger.

dmethvin:
:+1: thanks for doing this.

mzgol:
I'll land this shortly barring any new remarks.

src/manipulation.js
```diff
@@ -420,7 +424,7 @@ jQuery.fn.extend({
 	},
 
 	detach: function( selector ) {
-		return this.remove( selector, true );
+		return remove( this, selector, true );
```
gibson042: There's small savings (4 btyes min+gzip) to defining `detach` immediately before `remove` and moving both of them to the top of the `jQuery.fn.extend` block, but that's about it. I'm in favor of this.

mzgol: Currently it's only -3 bytes but I'll do it.

PR #2368, opened 2015-06-01
Core: .map should accept an `undefined` value

Mr21:
Hi :)

I modified the order of the instruction in `isArraylike`, because, I don't understand why but `"a" in undefined` crash but not `for (var i in undefined) {}`...

More info at #2363.


Mr21:
I closed this one, this change will be in this one #2363

Issue #1693, opened 2014-10-10
Make jQuery collections for-of-able


simevidas:
Chrome 38 added support for ES6 `for-of`. I'm proposing that jQuery collections are made compatible with this statement, which can be achieved with this line (from [discussion on Twitter](https://twitter.com/domenic/status/520384658555809792)):

```js
jQuery.prototype[Symbol.iterator] = Array.prototype.values;
```

**Live demo:** http://jsbin.com/movuxu/1/edit (view in Chrome)

spicyj:
I'd love to see this. Would make using jQuery with Babel more palatable.

timmywil:
You know, I'm not actually against this.

mzgol:
It'd be cool to have this natively in 3.0; even if simple. :)

mzgol:
The biggest problem is size, I guess. The following:
```js
if ( typeof Symbol !== "undefined" && Symbol.iterator ) {
	jQuery.fn[Symbol.iterator] = [][Symbol.iterator];
}
```
increases the gzipped size by 32 bytes. Other forms seem larger.

mzgol:
Since there is no definite decision yet, let's reopen for now.

jrencz:
The concept has already been described by @jakearchibald in Dec 2004 for NodeList: http://jakearchibald.com/2014/iterators-gonna-iterate/#nodelist-iteration

mzgol:
We've decided we're going to do it in 3.0.0. Taking it.

mzgol:
PR: #2369

PR #2369, opened 2015-06-01
Core: Make jQuery objects iterable

mzgol:
Make iterating over jQuery objects possible using ES 2015 for-of:

    for ( node of $( "<div id=narwhal>" ) ) {
        console.log( node.id ); // "narwhal"
    }

Fixes gh-1693

TODO:
* [x] Add a Node.js test checking if the iterator is defined even for `core-js`-polyfilled `Symbol` & `Symbol.iterator`.
* [x] Add tests ensuring that the iterator works with babel + core-js polyfills.
* [x] Rename jsHint to JSHint in comments.
* [x] Drop the `Symbol.iterator` check, leave just the `typeof Symbol` one.
* [x] Switch `[][Symbol.iterator]` to `arr[Symbol.iterator]`.
* [x] ~~Use `QUnit.skip` to skip the test in non-ES6 browsers.~~ Just use the `ok` assertions in non-ES6 browsers.
* [ ] Split the JSHint upgrade to a separate commit (to do just before landing).

@rwaldron, could you have a look if it's safe to include this in 3.0.0?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@
 
 /dist
 /node_modules
+
+/test/node_smoke_tests/lib/ensure_iterability.js
--- a/package.json
+++ b/package.json
@@ -26,7 +26,9 @@
   "dependencies": {},
   "devDependencies": {
     "commitplease": "2.0.0",
+    "core-js": "0.9.17",
     "grunt": "0.4.5",
+    "grunt-babel": "5.0.1",
     "grunt-cli": "0.1.13",
     "grunt-compare-size": "0.4.0",
     "grunt-contrib-jshint": "0.11.2",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "grunt": "0.4.5",
     "grunt-cli": "0.1.13",
     "grunt-compare-size": "0.4.0",
-    "grunt-contrib-jshint": "0.10.0",
+    "grunt-contrib-jshint": "0.11.2",
     "grunt-contrib-uglify": "0.7.0",
     "grunt-contrib-watch": "0.6.1",
     "grunt-git-authors": "2.0.1",
--- a/src/core.js
+++ b/src/core.js
@@ -425,6 +425,16 @@ jQuery.extend({
 	support: support
 });
 
+// JSHint would error on this code due to the Symbol not being defined in ES5.
+// Defining this global in .jshintrc would create a danger of using the global
+// unguarded in another place, it seems safer to just disable JSHint for these
+// three lines.
+/* jshint ignore: start */
+if ( typeof Symbol === "function" ) {
+	jQuery.fn[ Symbol.iterator ] = arr[ Symbol.iterator ];
+}
+/* jshint ignore: end */
+
 // Populate the class2type map
 jQuery.each("Boolean Number String Function Array Date RegExp Object Error".split(" "),
 function(i, name) {
--- a/test/node_smoke_tests/lib/ensure_iterability_es6.js
+++ b/test/node_smoke_tests/lib/ensure_iterability_es6.js
@@ -1,0 +1,25 @@
+/* jshint esnext: true */
+
+"use strict";
+
+var assert = require( "assert" );
+
+module.exports = function ensureIterability() {
+	require( "jsdom" ).env( "", function( errors, window ) {
+		assert.ifError( errors );
+
+		var i,
+			ensureJQuery = require( "./ensure_jquery" ),
+			jQuery = require( "../../../dist/jquery.js" )( window ),
+			elem = jQuery( "<div></div><span></span><a></a>" ),
+			result = "";
+
+		ensureJQuery( jQuery );
+
+		for ( i of elem ) {
+			result += i.nodeName;
+		}
+
+		assert.strictEqual( result, "DIVSPANA", "for-of doesn't work on jQuery objects" );
+	} );
+};
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -1538,3 +1538,23 @@ testIframeWithCallback( "Don't call window.onready (#14802)", "core/onready.html
 			equal( error, false, "no call to user-defined onready" );
 	}
 );
+
+test( "Iterability of jQuery objects (gh-1693)", function() {
+	/* jshint unused: false */
+	expect( 1 );
+
+	var i, elem, result;
+
+	if ( typeof Symbol === "function" ) {
+
+		elem = jQuery( "<div></div><span></span><a></a>" );
+		result = "";
+
+		try {
+			eval( "for ( i of elem ) { result += i.nodeName; }" );
+		} catch ( e ) {}
+		equal( result, "DIVSPANA", "for-of works on jQuery objects" );
+	} else {
+		ok( true, "The browser doesn't support Symbols" );
+	}
+} );
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,5 +1,6 @@
 {
 	"preset": "jquery",
 
-	"excludeFiles": [ "external", "src/intro.js", "src/outro.js" ]
+	"excludeFiles": [ "external", "src/intro.js", "src/outro.js",
+		"test/node_smoke_tests/lib/ensure_iterability.js" ]
 }
--- a/.jshintignore
+++ b/.jshintignore
@@ -9,3 +9,4 @@ test/data/readywaitasset.js
 test/data/readywaitloader.js
 test/data/support/csp.js
 test/data/support/getComputedSupport.js
+test/node_smoke_tests/lib/ensure_iterability.js
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -30,6 +30,18 @@ module.exports = function( grunt ) {
 				cache: "build/.sizecache.json"
 			}
 		},
+		babel: {
+			options: {
+				sourceMap: "inline",
+				retainLines: true
+			},
+			nodeSmokeTests: {
+				files: {
+					"test/node_smoke_tests/lib/ensure_iterability.js":
+						"test/node_smoke_tests/lib/ensure_iterability_es6.js"
+				}
+			}
+		},
 		build: {
 			all: {
 				dest: "dist/jquery.js",
--- a/build/tasks/node_smoke_tests.js
+++ b/build/tasks/node_smoke_tests.js
@@ -15,7 +15,8 @@ module.exports = function( grunt ) {
 
 	fs.readdirSync( testsDir )
 		.filter( function( testFilePath ) {
-			return fs.statSync( testsDir + testFilePath ).isFile();
+			return fs.statSync( testsDir + testFilePath ).isFile() &&
+				/\.js$/.test( testFilePath );
 		} )
 		.forEach( function( testFilePath ) {
 			var taskName = "node_" + testFilePath.replace( /\.js$/, "" );
--- a/build/tasks/node_smoke_tests.js
+++ b/build/tasks/node_smoke_tests.js
@@ -5,7 +5,7 @@ module.exports = function( grunt ) {
 	var fs = require( "fs" ),
 		spawnTest = require( "./lib/spawn_test.js" ),
 		testsDir = "./test/node_smoke_tests/",
-		nodeSmokeTests = [ "jsdom" ];
+		nodeSmokeTests = [ "jsdom", "babel:nodeSmokeTests" ];
 
 	// Fire up all tests defined in test/node_smoke_tests/*.js in spawned sub-processes.
 	// All the files under test/node_smoke_tests/*.js are supposed to exit with 0 code
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -22,6 +22,7 @@
 		"define": false,
 		"DOMParser": false,
 		"Promise": false,
+		"Symbol": false,
 		"QUnit": false,
 		"ok": false,
 		"equal": false,
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -21,6 +21,7 @@
 		"require": false,
 		"define": false,
 		"DOMParser": false,
+		"Promise": false,
 		"QUnit": false,
 		"ok": false,
 		"equal": false,
--- a/test/node_smoke_tests/.jshintrc
+++ b/test/node_smoke_tests/.jshintrc
@@ -1,0 +1,14 @@
+{
+	"boss": true,
+	"curly": true,
+	"eqeqeq": true,
+	"eqnull": true,
+	"expr": true,
+	"immed": true,
+	"noarg": true,
+	"quotmark": "double",
+	"undef": true,
+	"unused": true,
+
+	"node": true
+}
--- a/test/node_smoke_tests/document_missing.js
+++ b/test/node_smoke_tests/document_missing.js
@@ -1,19 +1,11 @@
-/* jshint node: true */
-
 "use strict";
 
-var ensureGlobalNotCreated = require( "./lib/ensure_global_not_created" ),
+var assert = require( "assert" ),
+	ensureGlobalNotCreated = require( "./lib/ensure_global_not_created" ),
 	jQueryFactory = require( "../../dist/jquery.js" );
 
-try {
+assert.throws( function () {
 	jQueryFactory( {} );
-	console.error( "The jQuery factory should reject window without a document" );
-	process.exit( 1 );
-} catch ( e ) {
-	if ( e.message ===  "jQuery requires a window with a document" ) {
-		ensureGlobalNotCreated( module.exports );
-		process.exit( 0 );
-	}
-	console.error( "An unexpected error thrown; message: ", e.message );
-	process.exit( 1 );
-}
+}, /jQuery requires a window with a document/ );
+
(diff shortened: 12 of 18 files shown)


src/core.js
```diff
@@ -425,6 +425,16 @@ jQuery.extend({
 	support: support
 });
 
+// jsHint would error on this code due to the Symbol not being defined in ES5.
+// Defining this global in .jshintrc would create a danger of using the global
+// unguarded in another place, it seems safer to just disable jsHint for these
+// three lines.
+/* jshint ignore: start */
+if ( typeof Symbol === "function" && Symbol.iterator ) {
```
scottgonzalez: Why not just check for `Symbol` on the global instead of directly as a global?

mzgol: Because we don't have a reference to the global, we only have reference to `window`. In the browser it's equivalent but we also support running jQuery on Node (in a very basic sense but still) where `window` is a fake window object created by a library like jsdom.

To reference a global we'd need to pass an additional parameter to the jQuery factory... but that wouldn't work since if used with Browserify `this` references the module and not the global. So we'd basically need to run a cascade test - check for `global`, then `this`. This wouldn't be shorter and would certainly be more fragile.

scottgonzalez: Thanks. I didn't realize we didn't have a reference to the global.

ljharb: Should this check `typeof Symbol.iterator === 'symbol'`?

mzgol: Do we need to worry about cases where a global Symbol is defined as a
function and Symbol.iterator doesn't have a typeof equal to "symbol"? I
imagine this would break using this code with Symbol polyfills that can't
override typeof.

-- 
Michał Gołębiowski

src/core.js
```diff
@@ -425,6 +425,16 @@ jQuery.extend({
 	support: support
 });
 
+// jsHint would error on this code due to the Symbol not being defined in ES5.
```
markelog: I think we are being overcautious here

test/unit/core.js
```diff
@@ -1538,3 +1538,19 @@ testIframeWithCallback( "Don't call window.onready (#14802)", "core/onready.html
 			equal( error, false, "no call to user-defined onready" );
 	}
 );
+
+if ( typeof Symbol === "function" && Symbol.iterator ) {
```
markelog: It is better to put this `if` inside the test, so amount of tests would be the same for each browser, it would also be easier to filter it out if needed

test/unit/core.js
```diff
@@ -1538,3 +1538,19 @@ testIframeWithCallback( "Don't call window.onready (#14802)", "core/onready.html
...
+	test( "Iterability of jQuery objects (gh-1693)", function() {
+		/* jshint unused: false */
+		expect( 1 );
+
+		var i,
+			elem = jQuery( "<div></div><span></span><a></a>" ),
+			result = "";
+
+		try {
+			eval( "for ( i of elem ) { result += i.nodeName; }" );
```
markelog: It seems `jQuery.globalEval` would be easier?

src/core.js
mzgol: Maybe. But one of the ways we currently ensure we don't completely blow up in environments like Node.js is to restrict the list of defined globals to a very small list (mostly `window` & timer functions). So I preferred to not add another one.

src/core.js
mzgol: @ljharb I've just checked it and if you use Babel's `browser-polyfill.js` (which is mostly compiled [core-js](http://npmjs.com/package/core-js) & [regenerator](https://www.npmjs.com/package/regenerator)) then the global `Symbol` is defined but `typeof Symbol.iterator === "object"`. I think it's good to work in such an environment as well.

I should probably add a test for that.

test/unit/core.js
mzgol: Why global? I'd like it to stay contained in this test. A global script wouldn't e.g. have access to the `result` variable.

test/unit/core.js
mzgol: Maybe I can `QUnit.skip` the test in browsers without iterators?

markelog: Worth of try :-)

test/unit/core.js
markelog: You could register it as global, we have mechanism for that, see ajax tests

src/core.js
rwaldron: > jsHint would error on this code due to the Symbol not being defined in ES5.

How problematic is it for jQuery to switch on `esnext: true`?

src/core.js
rwaldron: This is tough because I agree with both @ljharb and @mzgol on this point. On one hand I'm not convinced that half-shimming features is a good idea; on the other, I think as long as this allows for the widest net possible, it's totally ok. It's hard to say how or if it will break in the wild, but I don't think you're going to know until you try. It's a bold move and I like it.

rwaldron:
:clap: This is so _awesome_.

src/core.js
gibson042: I'd be fine with removing the truthy `Symbol.iterator` guard completely, simplifying this condition to `if ( typeof Symbol === "function" )`. AFAICT, the worst outcome of such would be a useless property on the jQuery prototype.

src/core.js
```diff
@@ -425,6 +425,16 @@ jQuery.extend({
...
+// Defining this global in .jshintrc would create a danger of using the global
+// unguarded in another place, it seems safer to just disable jsHint for these
+// three lines.
+/* jshint ignore: start */
+if ( typeof Symbol === "function" && Symbol.iterator ) {
+	jQuery.fn[ Symbol.iterator ] = [][ Symbol.iterator ];
```
gibson042: This should use [`arr`](https://github.com/mzgol/jquery/blob/for-of/src/var/arr.js) instead of `[]`.

src/core.js
ljharb: This, and no less, is what I currently consider to be test a valid "enough" Symbol polyfill: https://github.com/ljharb/object.assign/blob/master/hasSymbols.js

The "worst outcome" here could include a discoverable string property on the jQuery prototype, that people may rely on in browsers without native `Symbol`, and thus their code will break in newer browsers.

That said, every browser with native symbols also has `Symbol.iterator` so far, from my testing.

src/core.js
ljharb: @mzgol fwiw, node has `Symbol` as of `v0.12`, and all `io.js` versions have it as well

src/core.js
mzgol: This test is definitely too large for jQuery.

I'd share your concerns a year ago but now latest stable versions of Chrome & Firefox support symbols so I doubt any code using jQuery 3.0.0 (note that we're not releasing tomorrow, it'll take at least 1-2 months) will rely on behavior in browsers without symbols, rather the opposite.

I don't have a definite opinion about removing the `Symbol.iterator` check.

ljharb: Many people still support IE 8, so people will be using `Symbol` polyfills for a very long time, including with jQuery v3.0.0.

I definitely understand that the test might be too large. I do believe though that jQuery should do as many checks as it feels comfortable doing to ensure that `Symbol`s are well-behaved in the engine.

dmethvin: What is the specific problem with this line as-is? Perhaps we should stay with it until there's a proven problem in the wild that requires a stricter and more complex check.

mzgol: > Many people still support IE 8, so people will be using Symbol polyfills for a very long time, including with jQuery v3.0.0.

I realise that; however, people test on modern Chrome/Firefox more than they test on IE8 so I don't think people will write code expecting a polyfill and not discovering it breaks in browsers with native `Symbol`, rather the opposite.

Bad polyfills can do a lot of damage even if we try to protect against them so I'd try to go this route and see if we get any feedback about possible problems with this approach rather than lock down the code to strictly native symbols - in that latter case people would rely on the iterability only in Chrome extensions, Firefox OS apps, Windows 10 apps, Node/io.js etc. but not on the Web. Or even worse - they may use a polyfill but test only in Chrome/Firefox not knowing that the polyfill is explicitly excluded by our strict logic.

So I'd follow @dmethvin's advice and just see where the proposed approach leads us. :)

rwaldron: > What is the specific problem with this line as-is? Perhaps we should stay with it until there's a proven problem in the wild that requires a stricter and more complex check.

+ 1

This is basically what I was getting at above: you'll never know until you try!

> Many people still support IE 8

jQuery 3 requires the jQuery 3 Compat shim to support IE8. Surely that shim can do some extra protection work if necessary?

gibson042: > The "worst outcome" here could include a discoverable string property on the jQuery prototype, that people may rely on in browsers without native `Symbol`, and thus their code will break in newer browsers.

@ljharb What would such a reliance look like? `for ( el of $collection )` won't work without a valid `Symbol.iterator`, but even if someone loads jQuery on top of an incomplete polyfill, there's nothing for them to do with a property named `"undefined"` or `"Symbol (iterator)"` or `"@@iterator"` if it has an `undefined` or other falsy value.

markelog: > jQuery 3 requires the jQuery 3 Compat shim to support IE8. Surely that shim can do some extra protection work if necessary?

What do you mean by that? Did i miss something? What shim?

dmethvin: I think he means the Compat branch or jquery-compat package which is I guess what we'll call it up on npm. It's a full jQuery with all the IE8 crap in it tho, not a shim. I guess you could consider jquery-migrate to be a shim but I don't think it would have a role here.

src/core.js
```diff
@@ -425,6 +425,16 @@ jQuery.extend({
 	support: support
 });
 
+// jsHint would error on this code due to the Symbol not being defined in ES5.
+// Defining this global in .jshintrc would create a danger of using the global
+// unguarded in another place, it seems safer to just disable jsHint for these
```
arthurvr: > `jsHint`

I always thought the correct way to capitalize it is `JSHint`.

timmywil: I'm not really that scared of polyfills in this case. In fact, I wouldn't
mind checking to see that it works with Babel.

src/core.js
ljharb: @gibson042 `for..of` is by no means the only way to use iterators - `Array.from($collection)` will use it too, and that can be polyfilled (and is, in `es6-shim`)

rwaldron: @markelog 

> What do you mean by that? Did i miss something? What shim?

http://blog.jquery.com/2014/10/29/jquery-3-0-the-next-generations/ <-- Is this still legit?


> for..of is by no means the only way to use iterators - Array.from($collection) will use it too, and that can be polyfilled (and is, in es6-shim)

Yes, but `Array.from` also works with any array-like that doesn't have Symbol.iterator.

gibson042: Right. I'm saying that our incorporation of a broken or incomplete `Symbol` polyfill would be their problem, not ours. jQuery wouldn't break, it would just lack this particular enhancement.

rwaldron: Are you responding to me?

gibson042: @rwaldron: I was responding to both you and @ljharb. I acknowledge that `for…of` is not the only use of iterators, and advocate for letting jQuery be exactly as iterable as arrays sharing its environment (and hence not to check for `Symbol.iterator` at all).

rwaldron: > and advocate for letting jQuery be exactly as iterable as arrays sharing its environment (and hence not to check for Symbol.iterator at all).

Almost agree. If there were an environment that implemented `Symbol`, but hadn't implemented `Symbol.iterator` yet, you'd end up with `jQuery.fn.undefined = undefined`. If you don't care about that, then it's all good and I agree. Also, I don't know of any runtimes that match that, but for a few months both Chrome and Firefox did (obviously not anymore).

gibson042: > If there were an environment that implemented `Symbol`, but hadn't implemented `Symbol.iterator` yet, you'd end up with `jQuery.fn.undefined = undefined`.

Correct. Or perhaps even something a little weirder, if `Symbol.iterator` is partially implemented (e.g., by incomplete polyfill).

> If you don't care about that, then it's all good and I agree.

Correct. I don't care about that, and here's why (edit mine):

> I don't know of any runtimes [supported by jQuery] that match that

rwaldron: Excellent.

mzgol: Current situation is as follows:

1. Latest stable Firefox, Chrome & Opera support `Symbol`, `Symbol.iterator` & `for-of` on any collections.
2. IE11 support neither `Symbol` nor `for-of` but Edge supports both.
3. Safari 8 doesn't support `Symbol`; it supports `for-of` only on Arrays. WebKit Nightly supports `Symbol`, `Symbol.iterator` & `for-of` on any collections.

I don't think we're in a lot of danger that any browser will arrive with that much broken `Symbol` global judging by the above info.

src/core.js
mzgol: @timmywil Agreed, added to the list. In fact it'll be testing against `core-js`, not Babel as that's the default polyfill that Babel advises; technically speaking Babel transforms the source code but doesn't define any polyfills by itself.

src/core.js
mzgol: We still support Node 0.10, though.

mzgol: @rwaldron 
>> jsHint would error on this code due to the Symbol not being defined in ES5.
>
> How problematic is it for jQuery to switch on `esnext: true`?

This would be going even further; we'd get automatically defined all ES6 globals & syntax which not all our supported browsers implement.

On the `compat` branch we even explicitly define `es3: true` so that jsHint won't e.g. allow us to introduce trailing commas in array/object literals (yes, it did happen).

I think turning `esnext` on could be dangerous here.

src/core.js
mzgol: @arthurvr 
>> `jsHint`
>
> I always thought the correct way to capitalize it is `JSHint`.

Indeed.

src/core.js
rwaldron: @mzgol as always, thank you for the thorough follow up

src/core.js
rwaldron: @mzgol again, thank you for the follow up

> This would be going even further; we'd get automatically defined all ES6 globals & syntax which not all our supported browsers implement.

Just to clarify, are you referring to jQuery 3 or all versions?

> On the compat branch we even explicitly define es3: true so that jsHint won't e.g. allow us to introduce trailing commas in array/object literals (yes, it did happen).
> I think turning esnext on could be dangerous here.

Yeah, this is very compelling and I hadn't even considered it when I first asked—thanks for the reality check :D

mzgol: >> This would be going even further; we'd get automatically defined all ES6 globals & syntax which not all our supported browsers implement.
>
> Just to clarify, are you referring to jQuery 3 or all versions?

To all jQuery versions currently in development, i.e. `jQuery 3.0.0` & `jQuery Compat 3.0.0`.

rwaldron: @mzgol thanks (for your patience) again!

test/unit/core.js
mzgol: OK, but I don't really think that would be easier to write and I prefer to not create globals (even temporary) if that's not necessary.

mzgol:
PR updated, please re-review.

.gitignore
```diff
@@ -10,3 +10,5 @@
 
 /dist
 /node_modules
+
+/test/node_smoke_tests/lib/ensure_iterability.js
```
mzgol: I could put it all in `dist` and not add anything to `.gitignore` or `.jshintignore` but then I'd have to copy the entire `test/node_smoke_tests` directory over as the `ensure_iterability_es6.js` file requires files relative to its original location so transpiling to a different directory wouldn't fly.

gibson042: What's going on here?

package.json
```diff
@@ -26,10 +26,12 @@
   "dependencies": {},
   "devDependencies": {
     "commitplease": "2.0.0",
+    "core-js": "0.9.14",
     "grunt": "0.4.5",
+    "grunt-babel": "5.0.0",
     "grunt-cli": "0.1.13",
     "grunt-compare-size": "0.4.0",
-    "grunt-contrib-jshint": "0.10.0",
+    "grunt-contrib-jshint": "0.11.2",
```
mzgol: This is required so that `typeof x === "symbol"` works in `esnext` mode.

test/node_smoke_tests/lib/ensure_iterability_es6.js
```diff
@@ -0,0 +1,30 @@
...
+
+		ensureJQuery( jQuery );
+
+		for ( i of elem ) {
+			result += i.nodeName;
+		}
+
+		if ( result !== "DIVSPANA" ) {
+			console.error( "for-of doesn't work on jQuery objects; " +
+				"expected: 'DIVSPANA'; got: '" + result + "'" );
```
gibson042: We should probably use `assert`, but I'm fine with a later sweep of `node_smoke_tests` for that.

gibson042:
LGTM :+1:

test/unit/core.js
ljharb: Using the `Function` constructor and invoking it is probably a bit nicer/safer than using `eval`, fwiw

mzgol:
I'll land this shortly barring any new remarks.

Issue #2308, opened 2015-05-13
slideDown, show and more stopped working for `display: none` elements


phistuck:
Test case -
```
<!doctype html>
<script src="http://code.jquery.com/jquery-git2.js"></script>
<style>
div, p { display: none }
</style>
<div>stuff<br/>that<br/>takes<br/>a<br/>height</div>
<p>stuff<br/>that<br/>takes<br/>a<br/>height</p>
<script>
$("p").show();
$("div").slideDown();
</script>
```

Changing `jquery-git2` to `jquery-2.1.1` fixes the issue. This is a regression.

This is caused by https://github.com/jquery/jquery/commit/86419b10bfa5e3b71a7d416288ab806d47a31d1f

matthaywardwebdesign:
@phistuck Can confirm, did break a client's site which happened to be using jquery-git.js. Seeing identical behaviour to above.

mzgol:
Wow, don't use jquery-git.js in production, ever.

matthaywardwebdesign:
@mzgol Haha yep,  never have since. This site happened to be my first paid job many a year ago. Definitely a bad idea, I'm surprised it survived as long as it did.

timmywil:
This is an intended behavior change for jQuery 3.0. It fixes a few long-standing issues.

scottgonzalez:
So for the extremely common case that was presented, what is the recommended solution?

jzaefferer:
Somewhat related, if its supposed to go into 3.0, why is it changing jquery-gi2t?

mzgol:
@jzaefferer `jquery-git2` currently mirrors `jquery-git`, `jquery-git1` mirrors `jquery-git-compat`, i.e. both development lines. I've set it up this way some time ago so that some tools (like jsFiddle/jsBin) that pull the newest jQuery for testing still works. jsFiddle & jsBin now use the new URLs so we should just remove `jquery-git2` & `jquery-git1`.

jzaefferer:
I see. Well, you should start with permanent redirects before deleting them. For jQuery UI we use `jquery-git` and `jquery-git2` for testing. Other projects might be doing the same, so some kind of official information about these changes would be quite useful.

mzgol:
@jzaefferer `jquery-git` & `jquery-git2` point to the same file now so you're effectively testing `master` twice & `compat` not at all.

scottgonzalez:
He meant that we test against `jquery-git` and `jquery-git1`.

[...]

Issue #1774, opened 2014-10-21
Show animations, like $.show() $.toggle() and $.slideDown(), don't respect css display property if initial state is display:none;


mgol:
Originally reported by nicholasbg at: http://bugs.jquery.com/ticket/15159

<p>
When an element that has a class with a css display property that is not its default (for example a &lt;span&gt; with "display:block"), and has an inline style (like "display:none"), setting it to be displayed with something like $.toggle(200) will cause it to not use the class's display property, while $.toggle() will work properly. Even $.toggle(0) will exhibit this behaviour.
</p>

<p>
From what I can tell this bug was introduced in 1.11.0/2.1.0
</p>

<p>
This jsFiddle will demonstrate the bug...
<a class="ext-link" href="http://jsfiddle.net/e8kpg/" rel="nofollow"><span class="icon">&nbsp;</span>http://jsfiddle.net/e8kpg/</a>
</p>

<p>
You can see it working normally if you change the jQuery version to something before 1.11.0/2.1.0 or remove the 200 from .toggle(200).
</p>

<p>
This is a very similar case to...
</p>

<p>
Ticket <a class="closed ticket" href="http://bugs.jquery.com/ticket/14824" title="bug: jQuery v1.11 and v2.1 slideUp and slideDown not respecting css display ... (closed: fixed)">#14824</a>
<a href="http://bugs.jquery.com/ticket/14824">http://bugs.jquery.com/ticket/14824</a>
</p><p>Issue reported for jQuery 1.11.1</p>

nicholasbg:
Sorry for coming late but this bug was not fixed on any 1.x or 1.x version of jQuery. As stated in the bug report, you need to make the inline style display none and to add a duration to see the bug in action.

Here is a fork of robertmaxrees 1.11.1 fiddle with the bug (with the above changes) still happening.
https://jsfiddle.net/e8kpg/6/

The bug is fixed in the edge (currently  v3.0.0-pre) and edge compat version though.

PR #2374, opened 2015-06-02
CSS: try more simplified approach to show/hide methods

markelog:
This is WIP for the feedback.
Ref gh-2308

Okay, this is first option that i mentioned at the meeting.

Basically, we don't care about "responsive stylesheets", detached elements - all that stuff.
We just take a most simplified and expected approach - default display or `none` that is it.
`show()` - will show element, `hide()` - will hide it.

If this is custom or non-standart element - `display: block`, not spec default `display:inline`.

I remember @paulirish advocated for something like this, if you still interested in this issue, please see the full story <a href="https://github.com/jquery/jquery/issues/2308">here</a>.

--
Breaking -

```js
// div { display: inline; }
$( "div" ).hide().show(); // block
// or
$( "div" ).slideUp(function() {
	$( this ).slideDown();  // block
});
```

Since perf issue is only about `show/hide` methods we could preserve above behaviour for the those aliases by dividing render logic (`showHide` method) between simple `show/hide` and animation.

### DIFF
--- a/src/css/defaultDisplay.js
+++ b/src/css/defaultDisplay.js
@@ -1,0 +1,32 @@
+define([
+	"../core"
+], function( jQuery ) {
+
+var displays = {
+	li: "list-item",
+	table: "table",
+	caption: "table-caption",
+	colgroup: "table-column-group",
+	thead: "table-header-group",
+	tfoot: "table-footer-group",
+	tbody: "table-row-group",
+	col: "table-column",
+	td: "table-cell",
+	th: "table-cell",
+	tr: "table-row"
+};
+
+jQuery.map([ "b", "big", "i", "small", "tt",
+	"abbr", "acronym", "cite", "code", "dfn", "em", "kbd", "strong", "samp", "var",
+	"a", "bdo", "br", "img", "map", "object", "q", "script", "span", "sub", "sup",
+	"button", "input", "label", "select", "textarea" ],
+	function( value ) {
+		displays[ value ] = "inline";
+	});
+
+function defaultDisplay( elem ) {
+	return displays[ elem.nodeName.toLowerCase() ] || "block";
+}
+
+return defaultDisplay;
+});
--- a/src/effects.js
+++ b/src/effects.js
@@ -4,10 +4,10 @@ define([
 	"./var/rcssNum",
 	"./css/var/cssExpand",
 	"./css/var/isHidden",
-	"./css/var/swap",
 	"./css/adjustCSS",
 	"./data/var/dataPriv",
 	"./css/showHide",
+	"./css/defaultDisplay",
 
 	"./core/init",
 	"./queue",
@@ -16,7 +16,8 @@ define([
 	"./manipulation",
 	"./css",
 	"./effects/Tween"
-], function( jQuery, document, rcssNum, cssExpand, isHidden, swap, adjustCSS, dataPriv, showHide ) {
+], function( jQuery, document,
+            rcssNum, cssExpand, isHidden, adjustCSS, dataPriv, showHide, defaultDisplay ) {
 
 var
 	fxNow, timerId,
@@ -153,14 +154,11 @@ function defaultPrefilter( elem, props, opts ) {
 
 		// Identify a display type, preferring old show/hide data over the CSS cascade
 		restoreDisplay = dataShow && dataShow.display;
-		if ( restoreDisplay == null ) {
-			restoreDisplay = dataPriv.get( elem, "display" );
-		}
+
 		display = jQuery.css( elem, "display" );
+
 		if ( display === "none" ) {
-			display = restoreDisplay || swap( elem, { "display": "" }, function() {
-				return jQuery.css( elem, "display" );
-			} );
+			display = defaultDisplay( elem );
 		}
 
 		// Animate inline elements as inline-block
@@ -177,7 +175,9 @@ function defaultPrefilter( elem, props, opts ) {
 						restoreDisplay = display === "none" ? "" : display;
 					}
 				}
+
 				style.display = "inline-block";
+				hidden = false;
 			}
 		}
 	}
--- a/src/css/showHide.js
+++ b/src/css/showHide.js
@@ -1,41 +1,18 @@
 define([
-	"../data/var/dataPriv"
-], function( dataPriv ) {
+	"./defaultDisplay"
+], function( defaultDisplay ) {
 
 function showHide( elements, show ) {
-	var display, elem,
-		values = [],
+	var elem,
 		index = 0,
 		length = elements.length;
 
-	// Determine new display value for elements that need to change
 	for ( ; index < length; index++ ) {
 		elem = elements[ index ];
-		if ( !elem.style ) {
-			continue;
-		}
 
-		display = elem.style.display;
-		if ( show ) {
-			if ( display === "none" ) {
-				// Restore a pre-hide() value if we have one
-				values[ index ] = dataPriv.get( elem, "display" ) || "";
-			}
-		} else {
-			if ( display !== "none" ) {
-				values[ index ] = "none";
-
-				// Remember the value we're replacing
-				dataPriv.set( elem, "display", display );
-			}
-		}
-	}
-
-	// Set the display of the elements in a second loop
-	// to avoid the constant reflow
-	for ( index = 0; index < length; index++ ) {
-		if ( values[ index ] != null ) {
-			elements[ index ].style.display = values[ index ];
+		// In case there is a non-element node
+		if ( elem.style ) {
+			elem.style.display = show ? defaultDisplay( elem ) : "none";
 		}
 	}
 
--- a/test/data/testsuite.css
+++ b/test/data/testsuite.css
@@ -68,8 +68,7 @@ div.noopacity {
 	opacity: 0;
 }
 
-div.hidden,
-span.hidden {
+div.hidden {
 	display: none;
 }
 
@@ -117,9 +116,18 @@ div#fx-tests div.noback {
 	display: none;
 }
 
+/* tests to ensure jQuery can determine the native display mode of elements
+   that have been set as display: none in stylesheets */
+div#show-tests * { display: none; }
+
 #nothiddendiv { font-size: 16px; }
 #nothiddendivchild.em { font-size: 2em; }
 #nothiddendivchild.prct { font-size: 150%; }
+
+/* 8099 changes to default styles are read correctly */
+tt { display: none; }
+sup { display: none; }
+dfn { display: none; }
 
 /* #9239 Attach a background to the body( avoid crashes in removing the test element in support ) */
 body, div { background: url(http://static.jquery.com/files/rocker/images/logo_jquery_215x53.gif) no-repeat -1000px 0; }
--- a/test/unit/css.js
+++ b/test/unit/css.js
@@ -475,14 +475,14 @@ test("show(); hide()", function() {
 
 	hiddendiv = jQuery("div.hidden");
 	hiddendiv.hide();
-	equal( hiddendiv.css("display"), "none", "Cascade-hidden div after hide()" );
+	equal( hiddendiv.css("display"), "none", "Non-detached div hidden" );
 	hiddendiv.show();
-	equal( hiddendiv.css("display"), "none", "Show does not trump CSS cascade" );
+	equal( hiddendiv.css("display"), "block", "Pre-hidden div shown" );
 
 	div = jQuery("<div>").hide();
 	equal( div.css("display"), "none", "Detached div hidden" );
 	div.appendTo("#qunit-fixture").show();
-	equal( div.css("display"), "block", "Initially-detached div after show()" );
+	equal( div.css("display"), "block", "Pre-hidden div shown" );
 
 });
 
@@ -490,8 +490,8 @@ test("show();", function() {
 
 	expect( 18 );
 
-	var hiddendiv, div, pass, old, test;
-		hiddendiv = jQuery("div.hidden");
+  var hiddendiv, div, pass, old, test;
+	hiddendiv = jQuery("div.hidden");
 
 	equal(jQuery.css( hiddendiv[0], "display"), "none", "hiddendiv is display: none");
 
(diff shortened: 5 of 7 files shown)


src/css/defaultDisplay.js
```diff
@@ -0,0 +1,32 @@
...
+jQuery.map([ "b", "big", "i", "small", "tt",
+	"abbr", "acronym", "cite", "code", "dfn", "em", "kbd", "strong", "samp", "var",
+	"a", "bdo", "br", "img", "map", "object", "q", "script", "span", "sub", "sup",
+	"button", "input", "label", "select", "textarea" ],
+	function( value ) {
+		displays[ value ] = "inline";
+	});
+
+function defaultDisplay( elem ) {
+	return displays[ elem.nodeName.toLowerCase() ] || "block";
```
dcherman: Would this interfere with the default display type of custom elements?  I believe those default to `inline`

markelog: Please read pull description 

Regards,
Oleg

src/effects.js
```diff
@@ -153,14 +154,11 @@ function defaultPrefilter( elem, props, opts ) {
 
 		// Identify a display type, preferring old show/hide data over the CSS cascade
 		restoreDisplay = dataShow && dataShow.display;
-		if ( restoreDisplay == null ) {
-			restoreDisplay = dataPriv.get( elem, "display" );
-		}
+
 		display = jQuery.css( elem, "display" );
```
timmywil: There seems to be some confusion about the perf issue. Retrieving display from computed CSS is the problem. Webkit takes a long time retrieving that value. Our defaultDisplay implementation was not the problem.

timmywil: There are certainly other problems with it, but the PR as it stands does not address the perf issue.

timmywil: Edited comment above to clarify.

timmywil: Wait, I can't read filenames, apparently. Ignore all that.

markelog: `defaultDisplay` could be a problem for certain cases, but here, it was added mostly for the back-compat.

Please check this pull with different angle.

timmywil: Did you my most recent comment? I'm liking this PR.

timmywil:
Argument in favor: I think if users were to explain what they think these methods already do, this would be the closest to that.

gibson042:
This strikes me as a big step backwards on at least two fronts, both of which I consider important:
* interrupted animations
* respecting cascaded styles

I don't like `defaultDisplay`, but could see reintroducing it, especially in the hard-coded capacity presented here. Just not at the expense of the above.

timmywil:
I'm blanking on the interrupted animations problem (could you explain that one?), but there's no way to respect cascaded styles without retrieving computed display. I'm fine letting that one go. If someone did `div { display: inline; }`, we could point them to `.addClass()` or `.css()`. I think the performance issue takes priority because it hurts all use cases, not just a fraction.

rwaldron:
I'd like to know more about the intention: specifically, why does jQuery want to break from behaviors that were implemented as a result of bugs that were reported over the years?

timmywil:
@rwaldron A couple issues started all this. These are the related ones with relevant discussions...

https://github.com/jquery/jquery.com/issues/88, which led to https://github.com/jquery/jquery/issues/2057
https://github.com/jquery/jquery/issues/1767

rwaldron:
@timmywil this is exactly the background I needed—thanks!

timmywil:
After looking over the issues again while grabbing those links, I realized that not respecting cascading styles could be a bigger issue than I thought. Every approach we take breaks something; we just need to figure out what our priorities are.

I know it would help me if we had a compiled list of all of the characteristics we'd like show/hide and related methods that would make up the holy grail. Here's a start...

- shows elements that are hidden by stylesheets
- retrieves a proper display value that follows the spec
- retrieves a proper display value that respects the CSS cascade
- I'm still not sure what @gibson042 is referring to with interrupted animations, but that too
- avoid reflows when setting display values on multiple elements

gibson042:
> I'm blanking on the interrupted animations problem (could you explain that one?)

https://github.com/jquery/jquery/pull/2374/files#diff-c11dc311e64f99c637c0ef4fc61054a6L2147
```js
// Start and abort an animation, ending with elements in their initial state
$collection.toggle( N );
setTimeout(function() {
	$collection.stop().toggle( N/2 );
}, N/2 );
```

> there's no way to respect cascaded styles without retrieving computed display

That's not my primary point here. What I'm saying is that we should not "jump" elements before animating them, or leave an inline `display` style at the end of a "show" animation unless it was there before we started (in which case we should restore it).

> I think the performance issue takes priority because it hurts all use cases

Can you describe the "performance issue" **exactly**? [Current tip behavior](https://github.com/jquery/jquery/blob/86419b10bfa5e3b71a7d416288ab806d47a31d1f/src/css/showHide.js) implements "hide" as caching the inline (_not_ computed) `display` style and overwriting it with "none", and "show" as replacing `display: none` with the cached value, falling back on `""` (which clears the inline style).

gibson042:
> I know it would help me if we had a compiled list of all of the characteristics we'd like show/hide and related methods that would make up the holy grail. Here's a start...
>
- shows elements that are hidden by stylesheets
- retrieves a proper display value that follows the spec
- retrieves a proper display value that respects the CSS cascade
- I'm still not sure what @gibson042 is referring to with interrupted animations, but that too
- avoid reflows when setting display values on multiple elements

Your first three points are intertwined, right? Current tip doesn't do that, but we could reintroduce it (where "it" is essentially `defaultDisplay`) at whatever level of thoroughness we want _without_ sacrificing the recent "responsive stylesheet" improvements. If we do so, I'd want it to be simple, perhaps with @markelog's hard-coding here or perhaps with the previous more robust (but potentially non-performant) "clear an inline display and check again"—just not all the way to the iframe nonsense that was removed by 86419b10bfa5e3b71a7d416288ab806d47a31d1f.

timmywil:
> Can you describe the "performance issue" exactly?

Sorry, I didn't meant to imply that the current tip behavior still had the performance issue. In fact, it fixes it. I'd just like to avoid reintroducing it, which might happen if we read the cascade.

> Your first three points are intertwined, right?

They are, but supporting the third may reintroduce the performance problem.

> Current tip doesn't do that, but we could reintroduce it (where "it" is essentially defaultDisplay) at whatever level of thoroughness we want without sacrificing the recent "responsive stylesheet" improvements.

I would like this, but I just don't see a way to do it without a computed read. I think we can do everything on the list except "be fast" (list was edited), which is unfortunate. So, I'm trying to figure out if there's anything else that we can drop that wouldn't have a major impact on the community.

timmywil:
> Can you describe the "performance issue" exactly?

These were my findings... https://github.com/jquery/jquery.com/issues/88#issuecomment-72479215

gibson042:
Thanks @timmywil. I guess it comes down to `.show()`ing cascade-hidden elements, which I'd prefer to abandon but everyone else seems to want saved. So we'll have to get the computed `display` when there's none defined inline (which is slow) and invent a value with `defaultDisplay` when that comes back "none" (which is large and unreliable). Oh, and the whole thing will fail when an element with inline `display: none` is _also_ cascade-hidden. So much for progress.

markelog:
For me there is two ideas we can try - 

<ol>
<li> Divide logic of animated element and element that needs to be shown/hidden. For the first case we can remember the initial value, with full fledged <code>getComputedStyle</code> call. For the animated elements we already do that, browser should optimize those calls anyhow. Besides, there is no known perf issues with animation.
<br/><br/>
For the elements that only needs to be shown/hidden use logic expressed in this PR.</li>
<br/><br/>
<li>Introduce the new methods only for show/hide case, with logic expressed in this PR.</li>
</ol>

markelog:
btw, @rwaldron since you are up to speed now, could share your opinion?

gibson042:
Any behavioral distinction between e.g. `.show()` and `.show( 0 )` would be unacceptably confusing and error-prone for users.

markelog:
> would be unacceptably confusing and error-prone for users

This is open for interpretation, as i said at the meeting it would better to distance ourself from peremptory statements on this subject.

This is a serious behaviour change, perhaps most serious we have for 3.0 even following promise standart in my opinion is nothing comparing to this.

I like second option, since nothing would break and perf would be fixed.

gibson042:
> > would be unacceptably confusing and error-prone for users

> This is open for interpretation, as i said at the meeting it would better to distance ourself from peremptory statements on this subject.

The following issues all complain about a distinction between animated and non-animated `show`/`hide`/`toggle`:
* http://bugs.jquery.com/ticket/7543
* http://bugs.jquery.com/ticket/10335
* http://bugs.jquery.com/ticket/14750
* http://bugs.jquery.com/ticket/14848
* https://github.com/jquery/jquery/issues/1774
* https://github.com/jquery/jquery/issues/2071

markelog:
List of tickets is better, more objective, but as i said, i believe this is open for interpretation, for example, we faced a lot of tickets when jquery introduced `prop` interface and changed behaviour of `attr` method but we still decided that such change was justifiable.

Would be or wouldn't be first option is "unacceptable" should be under question - i think we have to weight all pros and cons and be objective as possible

dmethvin:
I think we're all frustrated here because reality sucks. I see no good solutions that are both performant and back-compatible.

Creating a new API doesn't solve the performance issue if people ignore the new API and continue to use the old one because they are using old code that continues to work, or because 10 years of StackOverflow answers mention show/hide as a solution. Evangelizing a new API is hard, seems like it would be easier to just evangelize classes as an alternative to show/hide if that gives us the same result.

Following @markelog for a minute, if `.show()` did a new fast thing and `.show(0)` did the old compatible slow thing, there would be a way to get the old behavior back easily in plugins needing it.  However, that old logic would have to remain in core, albeit moved to effects and removable via custom build. Plus the perf issue might still show up as a stutter when people did a lot of animations at once. 

That said, I am not a fan of this PR because it seems even less intuitive and more difficult for users to override than letting the cascade style come through on `.show()`. If we had some crystal ball to know what cases are most common in the world it would be easier to decide.

markelog:
@dmethvin so you suggesting revert?

>  Evangelizing a new API is hard, seems like it would be easier to just evangelize classes as an alternative to show/hide if that gives us the same result.

true, although for the plugins that need to show something fast there is no better way (there is a way of course, but it is not better :-) then use `show/hide` or replacement methods.

dmethvin:
I thought our last plan was to do a blog post explaining the trade-offs and try the current master approach of respecting cascade for an alpha to see how it broke people's code. The main argument against that seemed to be that it might look like we didn't know what we are doing. Asking the community for some input doesn't seem like a bad thing to me though.

markelog:
For me, this #2308 issue is a killer, basically, well, not even "basically" - all animation aliases would stop working, how can we justify that?

Mr21:
Since this time, has the subject been discussed at a higher level? Like W3C or something about integrate a new CSS property only for a display boolean?
We could easily imagine a `display: true/false` and a `displayType: "block";` as a new feature of a HTML5.
It's a silly question but I'm curious :S

markelog:
There discussions like that, but even if this was implemented in all supported browsers it still would not fix the performance issue

gibson042:
@Mr21: https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute

markelog:
hidden attribute works differently and wouldn't help us, i was referring to analog of the `display` `initial` value which would work as described `display: true/false`

Mr21:
For a better understanding, if we imagine a `jquery.css` next to every `jquery.js` with:
`.jquery-hidden { display:none !important; }` in it, and the `.hide/show()` should only some aliases of `.add/removeClass("jquery-hidden")` is this solution would help you?

markelog:
> This attribute must not be used to hide content that could legitimately be shown.

This is what i mean by "differently", see https://github.com/jquery/jquery/issues/2057 and lets not detour this discussion

scottgonzalez:
> Evangelizing a new API is hard, seems like it would be easier to just evangelize classes as an alternative to show/hide if that gives us the same result.

Every time you say this, I will repeat my response: This is not a viable solution for any plugin developers that care about content that already exists on the page.

scottgonzalez:
> For a better understanding, if we imagine a jquery.css next to every jquery.js with:
> .jquery-hidden { display:none !important; } in it, and the .hide/show() should only some aliases of .add/removeClass("jquery-hidden") is this solution would help you?

Not unless you can magically make the millions of sites that use jQuery also never use any other method of hiding content.

dmethvin:
> Not unless you can magically make the millions of sites that use jQuery also never use any other method of hiding content.

That philosophy got us here, "We must be able to handle all methods of hiding and showing content, no matter the cost in our complexity or their performance," Perhaps we *are* stuck with what we had before, but it will be for back-compat reasons.

@mr21 one case we have to deal with is the one where elements are initially hidden by a stylesheet. So although we *could* add a stylesheet rule to show content we would need to know what `display` value to put there, which is the same problem we already have today with the inline style.

Anyway, this PR is about #2308, are there other solutions? If we kept master as-is, how would users and plugin writers change their code to work around it? Seems like the impacts could be everywhere.

scottgonzalez:
> That philosophy got us here, "We must be able to handle all methods of hiding and showing content, no matter the cost in our complexity or their performance," Perhaps we are stuck with what we had before, but it will be for back-compat reasons.

I've never once said "we must be able to handle all methods of hiding and showing content." You're looking at the two most extreme scenarios, and that's not fair. I don't think I've ever mentioned back compat in these discussions, just extremely common, and perfectly sane, use cases.

Mr21:
Yeah thanks @dmethvin, I tried today to modify my `jquery.js` and pass the CSS tests to be confronted of the problem... and I realize there was two differentes user's conception about `.show`.
It's like 50% of the users see the `.show` method like a solution to *forceShow* something, and the others 50% see the `.show` like *unhide* to get back from a previous call of `.hide` without affecting the cascade.
(I have no idea how you will do)

gibson042:
Posted on IRC, but copied here to provide a reasonable template (e.g., for a `forceShow` method):
```js
$collection.css( "display", function( idx, computed ) {
	return (computed || "none") === "none" ?
		// In the general case, this function must be user-defined
		// ...but reasonable approximations can be made, as in PR #2374
		getVisibleDisplay( this ) :
		undefined;
} );
```

lazd:
I know I'm joining this discussion late, but why not advocate the use of HTML’s `[hidden]` attribute and have `show()`/`hide()` toggle that instead of [encouraging people to use classes](https://github.com/jquery/jquery/issues/1767#issuecomment-59862308)? This would unify the [DOM API for `element.hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) with jQuery's `show()`/`hide()` and would [embrace the HTML specification](http://www.w3.org/TR/html-markup/global-attributes.html#common.attrs.hidden).

Yes, this would need a simple `*[hidden] { display: none; }` in IE < 11 and Android < 4, but we're already [asking folks to add a `.hidden { display: none; }`](http://blog.jquery.com/2015/07/13/jquery-3-0-and-jquery-compat-3-0-alpha-versions-released/comment-page-1/#comment-551839) when they want an element hidden by default and to use `addClass()`/`removeClass()` to toggle it.

dmethvin:
@lazd https://github.com/jquery/jquery/issues/2308#issuecomment-101865495

lazd:
@dmethvin thanks, I commented on that issue with my thoughts. https://github.com/jquery/jquery/issues/2308#issuecomment-121297905

timmywil:
As we discussed, we're going to be taking a different direction here and revert to the old behavior with a minor difference: only cache the display value if one exists inline that we did not put there ourselves.

Issue #2320, opened 2015-05-15
Document "invocation correctness" development principle


gibson042:
Ref https://github.com/jquery/jquery/issues/2134#issuecomment-102510442

We now (modulo particularly hairy backwards compatibility issues) consider it safe to assume that all function input matches a documented signature, and therefore allow functions to throw exceptions rather than fail silently. This should be captured somewhere, and my first thought was [CONTRIBUTING.md](https://github.com/jquery/jquery/blob/master/CONTRIBUTING.md).

timmywil:
That seems like a good place.

markelog:
So boo plain objects, right?

dmethvin:
I started a wiki page that we can use to iterate, it might be a little easier to do there than on a PR. Or maybe harder, I dunno, but you can just start editing with your ideas if you want. People don't generally read any of this, so usually it will be us linking to it to justify not changing (or changing) some reported behavior.

https://github.com/jquery/jquery/wiki/API-design-guidelines

PR #2384, opened 2015-06-10
Core: organize prop & attr code to be similar

pgilad:


src/attributes/attr.js
```diff
@@ -23,11 +23,11 @@ jQuery.fn.extend({
...
 jQuery.extend({
 	attr: function( elem, name, value ) {
-		var hooks, ret,
+		var ret, hooks, notxml,
 			nType = elem.nodeType;
 
-		// don't get/set attributes on text, comment and attribute nodes
-		if ( nType === 3 || nType === 8 || nType === 2 ) {
+		// Don't get/set attributes on text, comment and attribute nodes
+		if ( !elem || nType === 3 || nType === 8 || nType === 2 ) {
```
campersau: The `!elem` check is unnecessary see https://github.com/jquery/jquery/pull/2201

pgilad:
@campersau Good catch.

1. In prop (https://github.com/jquery/jquery/blob/master/src/attributes/prop.js#L32) it does happen. Should that be removed too?
2. Regarding the comment in #2201 - I think `!elem` should be then checked before we access the `nodeType` property.

Thoughts?

src/attributes/attr.js
gibson042: The lack of this guard in `jQuery.attr` is a behavior difference with respect to `jQuery.prop` since 55ac56aeda786a4d1b677aefd3f5bb134ecb02ad, and if anything the latter should be brought in line (since we read `elem.nodeType` prior to it, we've already thrown if `elem` is `null` or `undefined`). See also #2320.

src/attributes/attr.js
```diff
@@ -38,7 +38,8 @@ jQuery.extend({
 
 		// All attributes are lowercase
 		// Grab necessary hook if one is defined
-		if ( nType !== 1 || !jQuery.isXMLDoc( elem ) ) {
+		notxml = nType !== 1 || !jQuery.isXMLDoc( elem );
+		if ( notxml ) {
```
gibson042: This variable might not be benefiting us in either method. I'd check the impact of removing it.

gibson042:
This is good (-35 min+gz). Can you also look at increasing the return similarities (from `value !== undefined` onward)?

pgilad:
@gibson042 No problem.

Regarding checking if `elem` exists (in either prop or attr). What are the thoughts?

gibson042:
> Regarding checking if `elem` exists (in either prop or attr). What are the thoughts?

Drop it from both.

pgilad:
Ok rebased what was noted so far... Will continue trying to create similarities...

pgilad:
@gibson042 I think the module `selector-sizzle` is missing in `prop.js`  (but found in `attr.js` )- we use `jQuery.isXMLDoc` (https://github.com/jquery/jquery/blob/master/src/attributes/prop.js#L36)

pgilad:
There is also an inconsistency where in `prop( someKey, null )` - the null will get assigned and in `attr( someKey, null )` is translated to `jQuery.removeAttr`.

Is there are reason for that?

gibson042:
> I think the module `selector-sizzle` is missing in `prop.js`  (but found in `attr.js` )- we use `jQuery.isXMLDoc` (https://github.com/jquery/jquery/blob/master/src/attributes/prop.js#L36)

What do you mean? That method is provided by `selector`, which is part of the [minimum build set](https://github.com/jquery/jquery/blob/a644101ed04d0beacea864ce805e0c4f86ba1cd1/Gruntfile.js#L36-L39).

pgilad:
I mean `"../selector"` is imported in `attr.js` but not in `prop.js` even though they both use the method `isXMLDoc` found in `selector-sizzle`.

gibson042:
> There is also an inconsistency where in `prop( someKey, null )` - the null will get assigned and in `attr( someKey, null )` is translated to `jQuery.removeAttr`.

> Is there are reason for that?

It is intentional, but surprisingly [_not_ documented](http://api.jquery.com/attr/#attr2)! That inconsistency should be preserved by this pull request, as should the string coercion in `attr`. DOM attributes are an inherently less powerful data store than Javascript properties.

gibson042:
Ah. prop.js should also import `../selector`.

pgilad:
@gibson042 Great... I think I implemented all of the changes. Also flattened the multiple nested ifs (I highly dislike the nested ifs and prefer the early returns.

markelog:
@pgilad we're ready to land this, but could you also provide a pull for the `compat` branch? Otherwise it would be hard for us to port these changes in there

pgilad:
@markelog Done - https://github.com/jquery/jquery/pull/2426
Also rebased this PR against master

markelog:
Thank you!

Issue #2177, opened 2015-03-27
'setTimeout is not defined' ReferenceError in CommonJS-like environment (Firefox SDK)


andyoxus:
This is an unusual use-case of jQuery (2.1.3) - an SDK addon has no obvious window object, and no reason to manipulate the DOM (injected 'content scripts' do this). In fact there *is* a 'hidden' window, which means jQuery should be usable for things like Deferreds, etc. Here's the 'main' index.js (the first three lines are the Firefox 'magic' to get hold of the hidden window, and are not important):

```javascript
var {Cc, Ci} = require("chrome");
var hiddenDOMWindow = Cc["@mozilla.org/appshell/appShellService;1"].
    getService(Ci.nsIAppShellService).hiddenDOMWindow;

var j = require('./jquery')(hiddenDOMWindow);

var deferred = j.Deferred().done(function(value) {
   console.log('Done', value);
});

deferred.resolve("hello world");
```
This throws a 'setTimeout is not defined' ReferenceError while `require()`ing jQuery, with the following stack trace

```
jQuery.ready.promise@resource://jq-issue/jquery.js:3441:4
resource://jq-issue/jquery.js:3456:1
module.exports@resource://jq-issue/jquery.js:31:12
resource://jq-issue/index.js:5:9
run@resource://gre/modules/commonjs/sdk/addon/runner.js:145:19
[...snip...]
```
Line 3441 of jquery.js is

```javascript
setTimeout( jQuery.ready );
```
In this environment `this` is `Sandbox`, which doesn't have a `setTimeout`. `window`, however, does, and the fix (for me) is just

```javascript
window.setTimeout( jQuery.ready );
```

There are other instances of this usage of `setTimeout` in jQuery.

mzgol:
The problem in changing `setTimeout` to `window.setTimeout` is that if you want to use pieces of jQuery in another environment that doesn't have `window`, i.e. Node.js then it won't work. So I don't think we can do that. The only way I can see that'd cover both scenarios would be to create a hook for users to provide their own `setTimeout`/etc. implementations and just default to `global.setTimeout`.

The list of assumed globals can be found at:
https://github.com/jquery/jquery/blob/bd9a1385feea349f48de658875c3f1bfc75daf74/src/.jshintrc#L17-L20
(`window` & `noGlobal` are taken from the factory, `define` & `module` are for AMD/CommonJS). So this problem affects only timers.

andyoxus:
[...]

PR #2391, opened 2015-06-14
Core: Use window.setTimeout & friends when provided

mzgol:
Fixes gh-2177

Unfortunately, this adds 60 bytes gzipped.

PR #2392, opened 2015-06-14
Core: Use window.setTimeout & friends when provided

mzgol:
Fixes gh-2177

More convoluted than #2391 but +51 bytes instead of +60.


src/intro.js
```diff
@@ -12,30 +12,38 @@
...
-(function( global, factory ) {
+(function() {
+	/* jshint -W117 */
+	var globalSetTimeout = setTimeout,
+		globalClearTimeout = clearTimeout,
+		globalSetInterval = setInterval,
+		globalClearInterval = clearInterval;
+	/* jshint +W117 */
 
-	if ( typeof module === "object" && typeof module.exports === "object" ) {
```
mzgol: This file is best viewed with `?w=1` in the URL. ;)

src/intro.js
```diff
@@ -12,30 +12,38 @@
  * Date: @DATE
  */
 
-(function( global, factory ) {
+(function() {
+	/* jshint -W117 */
+	var globalSetTimeout = setTimeout,
```
mzgol: Of course we'd be able to reduce the size significantly by avoiding this additional scope to save the globals but then we have to rename `setTimeout` & friends in the code.

src/var/setInterval.js
```diff
@@ -0,0 +1,4 @@
+define(function() {
+	/* global globalSetInterval: false */
+	return window.setIntarval || globalSetInterval;
```
mzgol: `s/setIntarval/setInterval/`...

mzgol:
I guess this won't work in the AMD mode...

mzgol:
Closing in favor of #2397.

Issue #2341, opened 2015-05-22
getStyles of an element inside the shadow dom


edoardocavazza:
Hi,

Using Polymer in my application, I have found out that the function 
```javascript
var getStyles = function( elem ) {
	// Support: IE<=11+, Firefox<=30+ (#15098, #14150)
	// IE throws on elements created in popups
	// FF meanwhile throws on frame elements through "defaultView.getComputedStyle"
	if ( elem.ownerDocument.defaultView.opener ) {
		return elem.ownerDocument.defaultView.getComputedStyle( elem, null );
	}

	return window.getComputedStyle( elem, null );
};
```

throws an exception when `elem` is a shadow dom element, because `elem.ownerDocument.defaultView` is undefined.

Adding the control, all works well.
```javascript
if ( elem.ownerDocument.defaultView && elem.ownerDocument.defaultView.opener ) {
```

dmethvin:
@edoardocavazza Can you provide a relatively simple test case we could use to detect this? Ideally it should not include Polymer since our unit test will not use it. In what cases is your code trying to get styles on a shadow root?

dmethvin:
We haven't heard back on this so I'm closing it. We'd need a test case and use case.

PR #2393, opened 2015-06-14
CSS: save 7 bytes on the getStyles function

Mr21:
Hi,

I saw few bytes to save after reading #2341, but I didn't change the logic.

### DIFF
--- a/src/css/var/getStyles.js
+++ b/src/css/var/getStyles.js
@@ -6,6 +6,10 @@ define(function() {
 		// FF meanwhile throws on frame elements through "defaultView.getComputedStyle"
 		var view = elem.ownerDocument.defaultView;
 
-		return ( view.opener && view || window ).getComputedStyle( elem );
+		if ( !view.opener ) {
+			view = window;
+		}
+
+		return view.getComputedStyle( elem );
 	};
 });
--- a/src/css/var/getStyles.js
+++ b/src/css/var/getStyles.js
@@ -1,12 +1,11 @@
 define(function() {
 	return function( elem ) {
+
 		// Support: IE<=11+, Firefox<=30+ (#15098, #14150)
 		// IE throws on elements created in popups
 		// FF meanwhile throws on frame elements through "defaultView.getComputedStyle"
-		if ( elem.ownerDocument.defaultView.opener ) {
-			return elem.ownerDocument.defaultView.getComputedStyle( elem );
-		}
+		var view = elem.ownerDocument.defaultView;
 
-		return window.getComputedStyle( elem );
+		return ( view.opener && view || window ).getComputedStyle( elem );
 	};
 });


markelog:
Actually, i see only -7 bytes between `8e111df...569123` i really dislike those `return ... && ... ||` makes the logic vague and more complex, especially when we have so much divergence in browser behaviour. 

This is why i specifically didn't used any tricks to make it smaller. If it was 35 bytes it could be a close call, but it seems 7 bytes not worth it

arthurvr:
> Actually, i see only -7 bytes between 8e111df...569123 i really dislike those return ... && ... || makes the logic vague and more complex, especially when we have so much divergence with browsers behaviour.


+1. I prefer to keep code more readable and less complex, even if you're able to save a few bytes. 7 bytes isn't worth it IMHO.

Mr21:
How did you compare the size? Me, I count how many caracteres there are in the jquery.min.js before and after my change. Do I have to do something more than just start grunt?

So, the main gain is not the return, it's the double of `ownerDocument.defaultView.getComputedStyle(e)`.

AurelioDeRosa:
I think that @markelog is comparing the gzipped version while you're comparing the minified-only version.

Mr21:
@AurelioDeRosa, thanks I thought the minified version was also the *gzipped* version...
I was not informed of the compression for the file transfert in webserver Oo

I did that version because I saw many return with ternary in the code, etc.
I rewrote my change (the minified version doesn't change, it detects the pattern).

staabm:
there is a grunt task to compare the size diff:
https://github.com/jquery/jquery/blob/master/Gruntfile.js#L22

markelog:
> I did that version because I saw many return with ternary in the code, etc.

Yeah, i'm usually advocating against that style if there is a better way. And yes, we usually care about gzipped size.

Now it looks much better, so if no one objects and test would run smoothly i will land this

mzgol:
LGTM.

arthurvr:
:+1:

mzgol:
@Mr21 If you run `grunt` you'll see the summary at the end:
```
Running "compare_size:files" (compare_size) task
   raw     gz Sizes                                                            
254280  75219 dist/jquery.js                                                   
 84215  29307 dist/jquery.min.js                                               

   raw     gz Compared to master @ 8e111df641cca3e1b75b31a1971bfc89014b4ded    
  +223    +62 dist/jquery.js                                                   
   +58    +40 dist/jquery.min.js                                               
```
You're interested in the gz min number compared to master, in this example it's +40 bytes.

Mr21:
@mzgol, so simple.
My grunt is not perfectly installed (i'm on Windows)... I will see what I have to do.

Now I see that every time... thanks for your help :)
``` bash
~/GitHub/jquery>grunt
>> Local Npm module "grunt-babel" not found. Is it installed?

Running "build:all:*" (build) task
>> File 'dist/jquery.js' created.

Running "jsonlint:pkg" (jsonlint) task
>> 1 file lint free.

Running "jshint:all" (jshint) task
>> 137 files lint free.

Running "jshint:dist" (jshint) task
>> 1 file lint free.

Running "jscs:src" (jscs) task
>> 90 files without code style errors.

Running "jscs:gruntfile" (jscs) task
>> 1 files without code style errors.

Running "jscs:test" (jscs) task
>> 1 files without code style errors.

Running "jscs:release" (jscs) task
>> 1 files without code style errors.

Running "jscs:tasks" (jscs) task
>> 7 files without code style errors.

Running "uglify:all" (uglify) task
>> 1 sourcemap created.
>> 1 file created.

Running "remove_map_comment" task

Running "dist:*" (dist) task
Warning: Task "babel:nodeSmokeTests" not found. Use --force to continue.

Aborted due to warnings.
```

mzgol:
Landed, thanks! In the end, it saved 6 bytes on `master` (3a0d582cf63b6e8f77150d9c38d2bf34d0c7790e) & 0 on `compat` (bf282ea8e20892af2d1d2c76b93b0c1f2141db97) but is still worthy due to making the code more readable.

PR #2394, opened 2015-06-15
Core: Use window.setTimeout & friends when provided

mzgol:
Fixes gh-2177

Alternative to #2391 & #2392. +40 bytes.

### DIFF
--- a/src/var/intervalClear.js
+++ b/src/var/intervalClear.js
@@ -1,0 +1,4 @@
+define(function() {
+	/* global clearInterval: false */
+	return window.clearInterval || clearInterval;
+});
--- a/src/var/intervalSet.js
+++ b/src/var/intervalSet.js
@@ -1,0 +1,4 @@
+define(function() {
+	/* global setInterval: false */
+	return window.setInterval || setInterval;
+});
--- a/src/var/timeoutSet.js
+++ b/src/var/timeoutSet.js
@@ -1,0 +1,4 @@
+define(function() {
+	/* global setTimeout: false */
+	return window.setTimeout || setTimeout;
+});
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -15,6 +15,10 @@ module.exports = function( grunt ) {
 	// The concatenated file won't pass onevar
 	// But our modules can
 	delete srcHintOptions.onevar;
+	srcHintOptions.globals.setTimeout = false;
+	srcHintOptions.globals.clearTimeout = false;
+	srcHintOptions.globals.setInterval = false;
+	srcHintOptions.globals.clearInterval = false;
 
 	grunt.initConfig({
 		pkg: grunt.file.readJSON( "package.json" ),
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -14,10 +14,6 @@
 
 	"globals": {
 		"window": true,
-		"setTimeout": true,
-		"clearTimeout": true,
-		"setInterval": true,
-		"clearInterval": true,
 
 		"jQuery": true,
 		"define": true,
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -2,6 +2,8 @@ define([
 	"./core",
 	"./var/document",
 	"./var/rnotwhite",
+	"./var/timeoutSet",
+	"./var/timeoutClear",
 	"./ajax/var/location",
 	"./ajax/var/nonce",
 	"./ajax/var/rquery",
@@ -9,7 +11,8 @@ define([
 	"./ajax/parseJSON",
 	"./ajax/parseXML",
 	"./deferred"
-], function( jQuery, document, rnotwhite, location, nonce, rquery ) {
+], function( jQuery, document, rnotwhite, timeoutSet, timeoutClear,
+	location, nonce, rquery ) {
 
 var
 	rhash = /#.*$/,
@@ -642,7 +645,7 @@ jQuery.extend({
 
 			// Timeout
 			if ( s.async && s.timeout > 0 ) {
-				timeoutTimer = setTimeout(function() {
+				timeoutTimer = timeoutSet(function() {
 					jqXHR.abort("timeout");
 				}, s.timeout );
 			}
@@ -676,7 +679,7 @@ jQuery.extend({
 
 			// Clear timeout if it exists
 			if ( timeoutTimer ) {
-				clearTimeout( timeoutTimer );
+				timeoutClear( timeoutTimer );
 			}
 
 			// Dereference transport for early garbage collection
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -1,8 +1,9 @@
 define([
 	"../core",
 	"../var/document",
+	"../var/timeoutSet",
 	"../deferred"
-], function( jQuery, document ) {
+], function( jQuery, document, timeoutSet ) {
 
 // The deferred used on DOM ready
 var readyList;
@@ -73,7 +74,7 @@ jQuery.ready.promise = function( obj ) {
 		// discovered by ChrisS here: http://bugs.jquery.com/ticket/12282#comment:15
 		if ( document.readyState === "complete" ) {
 			// Handle it asynchronously to allow scripts the opportunity to delay ready
-			setTimeout( jQuery.ready );
+			timeoutSet( jQuery.ready );
 
 		} else {
 
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -1,8 +1,9 @@
 define([
 	"./core",
 	"./var/slice",
+	"./var/timeoutSet",
 	"./callbacks"
-], function( jQuery, slice ) {
+], function( jQuery, slice, timeoutSet ) {
 
 function Identity( v ) {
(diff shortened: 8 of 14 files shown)


src/var/intervalClear.js
```diff
@@ -0,0 +1,4 @@
+define(function() {
+	/* global clearInterval: false */
```
timmywil: Is this necessary since the globals are only removed for the built file?

mzgol: I removed `clearInterval` from JSHint globals but this is the one place where it has to appear.

src/var/intervalSet.js
```diff
@@ -0,0 +1,4 @@
+define(function() {
```
arthurvr: One day we should really decide if we're going to use those new spacing rules or not: `define( function() {`. We're inconsistent at the moment. It's the latest style guide but nobody on the team seems to like it.

src/var/intervalClear.js
timmywil: Of course. I was reading it all backwards.

src/var/intervalSet.js
mzgol: Yeah, I did it incorrectly, I'll fix. Thanks for noticing.

timmywil: shh

mzgol: You may notice I did it correctly in other places, though. ;) Here I just copied a template from another file, hence the mistake.

Once JSCS starts checking this, the problem will go away. ;)

mzgol: PR updated.

mzgol: OK, I can't change it because the regex for matching defines in `/var/*` modules expects no space: https://github.com/jquery/jquery/blob/8e111df641cca3e1b75b31a1971bfc89014b4ded/build/tasks/build.js#L14 (cc @timmywil)

markelog: > Once JSCS starts checking this, the problem will go away. ;)

Checking? Huh, this is past century talk! It will fix cases like that.

src/var/timeoutSet.js
```diff
@@ -0,0 +1,4 @@
+define(function() {
+	/* global setTimeout: false */
+	return window.setTimeout || setTimeout;
```
markelog: It looks weird without context, i would suggest add descriptive comments to each of those new files

mzgol: This is a long comment, see https://github.com/mzgol/jquery/blob/7eef60f491be191e5844f3a3420bee238c7d73f3/src/intro.js#L19-L25. Should I repeat this comment in every of those 4 files?

src/var/intervalClear.js
```diff
@@ -0,0 +1,4 @@
...
+	return window.clearInterval || clearInterval;
```
markelog: For which exactly `window.clearInterval` and just `clearInterval`? First is for jsdom and browser env, second for node?

src/var/timeoutSet.js
markelog: I'd say link should suffice

mzgol: But link to what exactly? I linked to my first PR, this is the third where there is no central place to put it.

src/var/intervalClear.js
mzgol: Yup.

markelog:
I still question legitimacy of that ticket, we didn't receive any real issues about that and it seems only reason why we need to do that is

> However, some environments provide their own window.setTimeout etc.

Those environments should create browser-like conditions, not that other way around. If you think about it, it seems really weird, for us to do this, it is like for Firefox to change some their behaviour to keep consistency with the jsdom, which, actually should create Firefox like environment. Sounds like absurd, i bet no browser lib would do that.

In other words, jsdom should create hacks and not us, and again to emphasize - we didn't received real issues which should provoke this PR. 

Also, @domenic could you please check it out, since it seems, we doing that just for you.

Otherwise LGTM :-).

src/var/timeoutSet.js
markelog: Oh, then it would duplicate it everywhere :-(, would be really awkward though, or maybe just link to this pull would be enough?

domenic:
There's no reason to do this. Scripts that run in jsdom can use `setTimeout` and friends just fine, with no prefix.

If you run into any other "do this to support jsdom" issues, let me know and we can fix jsdom.

mzgol:
@markelog 
>> However, some environments provide their own window.setTimeout etc.
>
> Those environments should create browser-like conditions, not that other way around.

The thing is those environments *do* provide those conditions. The current situation is that we support both scenarios:

1. The global behaves like a browser window (web browsers, browserify): we return the jQuery object.
2. The global doesn't behave like a browser window: we then return a function to which you can provide your own `window` instance and it'll return the jQuery object using that `window`.

In the (2) case we're cheating a little as we generally take stuff off this `window` but with exception of `setTimeout` & friends which we always take from the global.

We could decide that this is an edge case that's not worth supporting but it will mean a little divergence from the (2) use case.

@domenic This mostly isn't about jsdom but about running in an environment where you can provide most needed stuff via your own `window` instance but without assuming you have anything on a global. Though if you were creating & closing (`window.close()`) a lot of windows in jsdom and using `setTimeout` etc. in them you'd leak memory, wouldn't you? `jsdom` has its own `setTimeout` etc. for a reason, doesn't it?

mzgol:
> we didn't received any issues which should provoke this PR

We did, the exact one I mentioned in the description of this PR... i.e. #2177.

domenic:
Oh, I don't really know anything about the use case of pasing in a window to a factory function. jsdom is geared around the use case of running jquery inside the jsdom environment itself, not that thing.

markelog:
@mzgol well, in there, owner of that ticket tries to use Deferreds (in the environment that doesn't have setTimeout) which is better to do through https://github.com/jaubourg/jquery-deferred-for-node package or through custom build, so this is really out there from my perspective and doesn't worth 40 bytes and code comlication.

So i as i see it, since jsdom could work with jQuery just fine, the reason to do it is to be more consistent but a lot more hacky?

--
Thank you for the replay @domenic, you helped a lot.

mzgol:
@domenic Hm. It was all added in 1f67d07c60c37e60052db37fc03d42af482c2d03. You even posted some comments under the ticket. ;)

mzgol:
Well, if we just want to assume that `window` is available then the whole factory function with all this complicated logic when to return the object and when the function accepting `window` as a parameter is unnecessary... So how did we end up here? :P @timmywil?

mzgol:
Also, if we want to go full-in in requiring a proper window to be present
then I'd prefer to remove the JSHint exception for setTimeout etc. and just
use window.setTimeout without a fallback to the global one.

-- 
Michał Gołębiowski

timmywil:
Actually, I'd be fine with requiring a proper window and moving to window.setTimeout everywhere. The problems that would cause in Node would be mitigated by using tools like jsdom along with jQuery, which is probably the only way jQuery should be used in Node anyway.

Side note: we probably shouldn't recommend https://github.com/jaubourg/jquery-deferred-for-node after 3.0 comes out. It hasn't received updates in a long time and does not support Promises/A+.

mzgol:
@timmywil @markelog OK, I created yet another attempt: #2397.

markelog:
> Side note: we probably shouldn't recommend https://github.com/jaubourg/jquery-deferred-for-node after 

Let's ask @jaubourg, i thought he is still commited to supporting this module

markelog:
> I'd be fine with requiring a proper window and moving to window.setTimeout everywhere

Why we need to window "prefix" for global vars?

src/var/intervalSet.js
markelog: @mzgol btw, coud you create a ticket about that?

mzgol:
> Why we need a window "prefix" for global vars?

For the same reason we prefix `getComputedStyle`, `document`, `document.documentElement` etc. - so that we're consistent in taking them from the provided `window` instance instead of the global. If we e.g. assumed a global `getComputedStyle`, jsdom wouldn't work. It would work without scoping `setTimeout` but there are other use cases. I think we should be consistent.

dmethvin:
It took me a while to catch up. In this particular case on #2177 it seems like the methods would normally be on `global` but are only on `window` for this unusual Firefox env, so we reference them from there?

src/var/intervalSet.js
mzgol: @markelog #2399

mzgol:
@dmethvin This comment of mine is a good summary: https://github.com/jquery/jquery/pull/2394#issuecomment-112192426

We take most globals from `window` but we take `setTimeout` etc. from the global which works for jsdom but not necessarily for other envs and is inconsistent. This PR tries to support both but the simpler one (the 4th my PR trying to fix this issue ;)) that just takes them from `window` is at #2397.

markelog:
Now i remember, okay, i agree, but i still would just write `window.setTimeout`

mzgol:
Closing in favor of #2397.

Issue #2310, opened 2015-05-13
Changes to offset break jQuery UI


arschmitz:
https://github.com/jquery/jquery/commit/1617479fcf7cbdaf33dc9334ed10a0f30bf14687 contains several breaking changes for jQuery UI

The first is the switch to return undefined for hidden or disconnected elements which has broken all of our interactions tests. 

The second is to allow offest to throw `getBoundingRect not a function` when `offset` is called on the window. While i agree this is probably not a valid use to begin with this may break a lot of other code expecting `undefined`

timmywil:
For the first case, I think `undefined` is more accurate. Retrieving offset on a disconnected element doesn't make much sense when you think about it.

For the second, I'd be in support of failing silently and returning undefined.

dmethvin:
The nice thing about returning `undefined` is that a naive caller will be expecting an object and die in their own code, which will be easier to debug. We don't document that it returns anything but an object and I'm okay with that since the unhandled cases are mentioned in the docs.

gibson042:
Are the failures all from stuff like [`$( element || window ).offset()`](http://irc.jquery.org/%23jquery-dev/latest.log.html#t08:03:38)? Because if so, just use `$( element ).offset()` instead—the method will return `undefined` if its context collection is empty.

As part of the larger story, though, these are the kinds of things that I _expect_ to see: failures from mild API misuse that are easy for consumers to fix. If we're not serious about valid vs. invalid input for our code, then I want to know now, because we'll just end up backing out every other change.

timmywil:
> If we're not serious about valid vs. invalid input for our code, then I want to know now, because we'll just end up backing out every other change.

In this case, returning undefined is still one way to handle invalid input. Not everything needs to throw an error, right?

arschmitz:
@gibson042 yes it is from things like that and i do agree its easy to fix we will fix it regardless of the outcome. However the problem is this can be pretty buried in code and hard to track down.

arschmitz:
Also you return `undefined` on disconnected or hidden nodes why would you throw on window but return undefined in those cases returning undefined in any "invalid" case would seem better.

gibson042:
> In this case, returning undefined is still one way to handle invalid input. Not everything needs to throw an error, right?

I don't want to "handle" invalid input. We're not creating errors, we're _exposing_ them by assuming that the input is valid. Why check for conditions that we tell consumers to avoid creating? Or in other words, how can input be invalid if we explicitly check for it?

[...]

PR #2396, opened 2015-06-15
Offset: return zeros for disconnected elements and allow setting

timmywil:
Fixes gh-2310

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -32,7 +32,7 @@ jQuery.offset = {
 			elem.style.position = "relative";
 		}
 
-		curOffset = curElem.offset() || { top: 0, left: 0 };
+		curOffset = curElem.offset();
 		curCSSTop = jQuery.css( elem, "top" );
 		curCSSLeft = jQuery.css( elem, "left" );
 		calculatePosition = ( position === "absolute" || position === "fixed" ) &&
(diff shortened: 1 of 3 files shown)


test/unit/core.js
```diff
@@ -101,12 +105,16 @@ test("jQuery()", function() {
 	elem = jQuery("\n\n<em>world</em>")[0];
 	equal( elem.nodeName.toLowerCase(), "em", "leading newlines" );
 
-	elem = jQuery( "<div/>", attrObj );
+	elem = jQuery("<div/>", attrObj );
```
gibson042: ?

timmywil: That was from the revert commit.

src/offset.js
```diff
@@ -92,7 +92,12 @@ jQuery.fn.extend({
 
 		rect = elem.getBoundingClientRect();
 
-		// Make sure element is not hidden (display: none) or disconnected
+		// Return zeros for disconnected elements (gh-2310)
+		if ( !jQuery.contains( document, elem ) ) {
```
gibson042: We don't need or want to check ancestry, since the below width/height/getClientRects check already excludes these cases. Returning `rect` at the end of the function in place of an implicit `undefined` should be all that's required.

timmywil: I tried that first, but I'm trying to separate hidden elements from disconnected ones. Should we return zeros for hidden elements too?

gibson042: I see no reason to treat them differently. Neither have layout, and therefore neither have "real" offset values, yet either could appear in the context collection.

PR #2397, opened 2015-06-16
Core: Use window.setTimeout & friends instead of global equivalents

mzgol:
Fixes gh-2177

+28 bytes.

An alternative to #2391 (+60), #2392 (+51) & #2394 (+40).

### DIFF
--- a/src/var/clearInterval.js
+++ b/src/var/clearInterval.js
@@ -1,0 +1,3 @@
+define(function() {
+	return window.clearInterval;
+});
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -14,10 +14,6 @@
 
 	"globals": {
 		"window": true,
-		"setTimeout": true,
-		"clearTimeout": true,
-		"setInterval": true,
-		"clearInterval": true,
 
 		"jQuery": true,
 		"define": true,
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -2,6 +2,8 @@ define([
 	"./core",
 	"./var/document",
 	"./var/rnotwhite",
+	"./var/setTimeout",
+	"./var/clearTimeout",
 	"./ajax/var/location",
 	"./ajax/var/nonce",
 	"./ajax/var/rquery",
@@ -9,7 +11,8 @@ define([
 	"./ajax/parseJSON",
 	"./ajax/parseXML",
 	"./deferred"
-], function( jQuery, document, rnotwhite, location, nonce, rquery ) {
+], function( jQuery, document, rnotwhite, setTimeout, clearTimeout,
+	location, nonce, rquery ) {
 
 var
 	rhash = /#.*$/,
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -1,8 +1,9 @@
 define([
 	"../core",
 	"../var/document",
+	"../var/setTimeout",
 	"../deferred"
-], function( jQuery, document ) {
+], function( jQuery, document, setTimeout ) {
 
 // The deferred used on DOM ready
 var readyList;
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -1,8 +1,9 @@
 define([
 	"./core",
 	"./var/slice",
+	"./var/setTimeout",
 	"./callbacks"
-], function( jQuery, slice ) {
+], function( jQuery, slice, setTimeout ) {
 
 function Identity( v ) {
 	return v;
--- a/src/effects.js
+++ b/src/effects.js
@@ -2,6 +2,9 @@ define([
 	"./core",
 	"./var/document",
 	"./var/rcssNum",
+	"./var/setInterval",
+	"./var/clearInterval",
+	"./var/setTimeout",
 	"./css/var/cssExpand",
 	"./css/var/isHidden",
 	"./css/var/swap",
@@ -16,7 +19,8 @@ define([
 	"./manipulation",
 	"./css",
 	"./effects/Tween"
-], function( jQuery, document, rcssNum, cssExpand, isHidden, swap, adjustCSS, dataPriv, showHide ) {
+], function( jQuery, document, rcssNum, setInterval, clearInterval, setTimeout,
+	cssExpand, isHidden, swap, adjustCSS, dataPriv, showHide ) {
 
 var
 	fxNow, timerId,
--- a/src/queue/delay.js
+++ b/src/queue/delay.js
@@ -1,8 +1,10 @@
 define([
 	"../core",
+	"../var/setTimeout",
+	"../var/clearTimeout",
 	"../queue",
 	"../effects" // Delay is optional because of this dependency
-], function( jQuery ) {
+], function( jQuery, setTimeout, clearTimeout ) {
 
 // Based off of the plugin by Clint Helfers, with permission.
 // http://web.archive.org/web/20100324014747/http://blindsignals.com/index.php/2009/07/jquery-delay/
--- a/src/var/clearTimeout.js
+++ b/src/var/clearTimeout.js
@@ -1,0 +1,3 @@
+define(function() {
+	return window.clearTimeout;
+});
--- a/src/var/setInterval.js
+++ b/src/var/setInterval.js
@@ -1,0 +1,3 @@
+define(function() {
+	return window.setInterval;
+});
--- a/src/var/setTimeout.js
+++ b/src/var/setTimeout.js
@@ -1,0 +1,3 @@
+define(function() {
+	return window.setTimeout;
+});


src/var/clearInterval.js
```diff
@@ -0,0 +1,3 @@
+define(function() {
+	return window.clearInterval;
```
markelog: Aaam, why do we put those files in different modules?

mzgol: Because they're variables so they have to be in different modules. It's the same with other window globals like `document`, `documentElement` as well as methods like `toString`, `indexOf` etc.

mzgol: Also, not sure how you'd otherwise reference `setTimeout` etc. in other modules if it wasn't a value returned from a module but just attached to it somehow.

dmethvin: If they were always used together you could put them in a single module and reference them from an object it returned. However that is probably less compressible and slower since the property name won't be minified the way a variable is?

markelog: > However that is probably less compressible and slower since the property name won't be minified the way a variable is?

Easy to check :)

@timmywil could provide a consult on amd here, i understand why we do this for `toString` or `indexOf` but defining `document` and `documentElement` is that some sort of convention or convenience?

mzgol: Apart from most likely being larger and differ from other vars we define (so it would be harder to comprehend) it would require people to write sth like `timers.setTimeout` instead of `setTimeout`.

So it might be larger and it would be harder to use. I see no gain in grouping those modules into one, only drawbacks.

mzgol: > but defining document and documentElement

It's so that we don't need to repeat `window.document` multiple times, risking we'll make a mistake our linter won't catch.

markelog: So convianiance, so why it is convenient to write four files include them in big argument list of the function call and use alias instead of just doing `window.setTimeout`?

mzgol: @markelog I've already replied to your question:
> risking we'll make a mistake our linter won't catch.

markelog: Could you elaborate?

dmethvin: It wouldn't catch `window.setTimmeout` for example, our previous use of `globals` in the `.jshintrc` prevented such spelling errors. It would seem that even some basic unit tests would prevent those kinds of issues though.

mzgol: Most of these changes landed in 76df9e4e389d80bff410a9e5f08b848de1d21a2f. AFAIK @gibson042 advocated creating variables for every one of them so maybe he'll want to say sth here.

markelog: > It would seem that even some basic unit tests would prevent those kinds of issues though.

I have same train of thought, i'm not sure such complication and size increase worth it

mzgol: Maybe. But we should be consistent here with other window globals, either we define modules for all of them or none. And since my 76df9e4e389d80bff410a9e5f08b848de1d21a2f was going in the "more var micro-modules" direction, I'd like to hear what @gibson042 & @timmywil have to say here.

markelog: > But we should be consistent here with other window globals, either we define modules for all of them or none. 

Well, as we discovered above we did this for the convenience, modulization of every host objects seems strange

markelog: If we decide that this in fact overkill, that we should definitely rethink <a href="https://github.com/jquery/jquery/commit/76df9e4e389d80bff410a9e5f08b848de1d21a2f#diff-6">this</a> change, since this var is used only in one place

timmywil: I think it's a good idea to use var modules here and for things like document and documentElement. It doesn't affect the built file – it just becomes a var statement – and it helps us ensure we're using the right document/documentElement/timeout function. Dave suggested this is taken care of by basic unit tests, but I don't think that's what we've seen historically. We've accidentally used the wrong document over and over and over until we did this.

markelog: So every host object should be modulized? Or just those that are used more than in one module?

mzgol: > It doesn't affect the built file

It does affect it by creating the variable, in some cases it increases the gzipped size a little.

> We've accidentally used the wrong document over and over and over until we did this.

I believe this was `location` that was breaking Node use? This has been taken care of in 76df9e4e389d80bff410a9e5f08b848de1d21a2f by removing the `browser` environment from `.jshintrc` and just putting `window` as a global there.

gibson042: > Most of these changes landed in 76df9e4e389d80bff410a9e5f08b848de1d21a2f. AFAIK @gibson042 advocated creating variables for every one of them so maybe he'll want to say sth here.

[Specifically](https://github.com/jquery/jquery/pull/1949#discussion_r22042857), "Since we're defining `docElem` here and in offset.js, let's make it a variable too". IIRC, it was just a redundancy concern as @markelog suggests:

> So every host object should be modulized? Or just those that are used more than in one module?

(we actually _still_ have a similar situation in the form of `div = document.createElement( "div" )`: [attributes/support](https://github.com/jquery/jquery/blob/63f19a95b9bfffdc7b8800f89a9def768cc3aebd/src/attributes/support.js), [css/support](https://github.com/jquery/jquery/blob/63f19a95b9bfffdc7b8800f89a9def768cc3aebd/src/css/support.js), [data/support](https://github.com/jquery/jquery/blob/63f19a95b9bfffdc7b8800f89a9def768cc3aebd/src/data/support.js), [event/support](https://github.com/jquery/jquery/blob/63f19a95b9bfffdc7b8800f89a9def768cc3aebd/src/event/support.js), [manipulation/support](https://github.com/jquery/jquery/blob/63f19a95b9bfffdc7b8800f89a9def768cc3aebd/src/manipulation/support.js))

Provided naked use of the globals will fail linting, though, I personally don't care. Come to think of it, I suppose it would be fine to drop _all_ the modules and prefix everything with `window.`—whatever's smaller.

mzgol: OK, I'm fine with this change as well, I just don't want to go round in circles. Because this PR goes in the same direction as currently code has, I'd like to land it as-is. Removing all variable modules removal of which decreases the built size can be done separately later.

If there are no other remarks to this PR, I'd like to land it & close other ones.

gibson042: > I'd like to land it & close other ones.

:+1:

timmywil: > It does affect it by creating the variable, in some cases it increases the gzipped size a little. Maybe in others it has the opposite effect, I didn't measure.

Sure, I just meant there's no size difference between putting `var document = window.document` at the top of core.js like we used to or use a var module. The latter is better for AMD loading. For this, if it's smaller to use `window.`, that's fine. But I agree we can do that separately.

mzgol: @gibson042 
> (we actually _still_ have a similar situation in the form of `div = document.createElement( "div" )`: [attributes/support](https://github.com/jquery/jquery/blob/63f19a95b9bfffdc7b8800f89a9def768cc3aebd/src/attributes/support.js), [css/support](https://github.com/jquery/jquery/blob/63f19a95b9bfffdc7b8800f89a9def768cc3aebd/src/css/support.js), [data/support](https://github.com/jquery/jquery/blob/63f19a95b9bfffdc7b8800f89a9def768cc3aebd/src/data/support.js), [event/support](https://github.com/jquery/jquery/blob/63f19a95b9bfffdc7b8800f89a9def768cc3aebd/src/event/support.js), [manipulation/support](https://github.com/jquery/jquery/blob/63f19a95b9bfffdc7b8800f89a9def768cc3aebd/src/manipulation/support.js))

These are all code pieces from `compat` where we assume the browser environment and don't support Node, browser extensions etc. All uses of `document` on `master` are via the variable or a property of `window`, this is enforced via JSHint.

mzgol:
If no one opposes, I'll land it tomorrow & create a separate PR for removing some `var` modules.

src/var/clearInterval.js
gibson042: > These are all code pieces from `compat` where we assume the browser environment and don't support Node, browser extensions etc. All uses of `document` on `master` are via a module or a property of `window`, this is enforced via JSHint.

You misunderstand me. What I mean is that we create `<div>` elements multiple times during initialization (yes, [in](https://github.com/jquery/jquery/blob/0e4477c676db0427bb9b0bf39df8631501e62f24/src/css.js#L39) [master](https://github.com/jquery/jquery/blob/0e4477c676db0427bb9b0bf39df8631501e62f24/src/manipulation/support.js) [too](https://github.com/jquery/jquery/blob/0e4477c676db0427bb9b0bf39df8631501e62f24/src/css/support.js#L11)) with IIFEs that wouldn't even have to _be_ wrapped if they reused variables. I don't know the impact on file size, but I do know that those IIFEs survive minification and interrupt `var` consolidation. Some of which is properly laid at the feet of UglifyJS2, but there are still steps we could take to build a more compressible file. That's what I was referencing.

mzgol: OK, I see. This is a delicate situation; re-using elements that had various operations applied to them before and using them for support tests may break support tests if done in a particular order; I'd rather take the slightly larger size and not risk introducing hard to track race conditions.

At least when we had one big `support.js` file it all happened in a stable order; now it depends on the methods used by a jQuery user.

gibson042: That's true, but mitigable by two means: resetting style & innerHTML before use (which we actually do, but can't enforce), and using distinct variable names. At any rate, it's off topic for this discussion so I'll either forget about it or open a PR eventually.

mzgol:
I had to change those modules to regular `window.*` uses after all as using modules caches those methods forever, preventing Sinon from mocking them & breaking many many effects tests.

dmethvin:
Ugh.

src/var/clearInterval.js
mzgol: @markelog 
> If we decide that this in fact overkill, that we should definitely rethink <a href="https://github.com/jquery/jquery/commit/76df9e4e389d80bff410a9e5f08b848de1d21a2f#diff-6">this</a> change, since this var is used only in one place

It's used in just one file but in many places. Removing the var saves 7 bytes, though...

mzgol: Removing the `document` module would increase the size by 26 bytes, though so I'm not really into it.

markelog: > it's off topic for this discussion so I'll either forget about it or open a PR eventually.

No-no, lets not forget! If we don't do it, somebody else might - https://github.com/jquery/jquery/issues/2405

markelog: > It's used in just one file but in many places

By "place" i mean "file" :-), created - https://github.com/jquery/jquery/issues/2407

markelog:
> I had to change those modules to regular window.* uses after all as using modules caches those methods forever, preventing Sinon from mocking them & breaking many many effects tests.

I bet we can workaround that, but i also bet, that we don't want to workaround that. 

That is what i meant by "complication" of code, when we use such convenience style of variable initialization.

Issue #2399, opened 2015-06-16
The build process requires no space between closing `}` and `)`


mzgol:
Our build process requires no space between closing `}` and `)` for variable modules (i.e. those with paths containing `/var/`). This makes it impossible to follow our style guide and write those modules in this way:
```js
define( function() {
	/* code */
} );
```
instead of currently required:
```js
define( function() {
	/* code */
} );
```
There may be other cases like that in used regexes.

I'll assign this to @timmywil since it's his code. I hope you don't mind, Timmy. :-)

timmywil:
Yea, I'll make it optional.

mzgol:
Why wasn't this auto-closed?

PR #2400, opened 2015-06-16
Build: space between curly and paren is optional

timmywil:
Fixes gh-2399

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -11,7 +11,7 @@ module.exports = function( grunt ) {
 	var fs = require( "fs" ),
 		requirejs = require( "requirejs" ),
 		srcFolder = __dirname + "/../../src/",
-		rdefineEnd = /\}\);[^}\w]*$/,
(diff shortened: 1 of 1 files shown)


build/tasks/build.js
```diff
@@ -11,7 +11,7 @@ module.exports = function( grunt ) {
 	var fs = require( "fs" ),
 		requirejs = require( "requirejs" ),
 		srcFolder = __dirname + "/../../src/",
-		rdefineEnd = /\}\);[^}\w]*$/,
+		rdefineEnd = /\}\s*?\);[^}\w]*$/,
```
mzgol: A very minor nit: the non-greediness doesn't matter here as `\)` doesn't match `\s` anyway, other regexps in this file don't use it.

timmywil: I generally use non-greedy unless I explicitly want greediness.

mzgol: OK, doesn't matter.

mzgol:
LGTM.

Issue #2329, opened 2015-05-19
Create e2e tests


markelog:
So we would stop using `dispatchEvent`, `createEvent` methods and such, like focus tests.

 - [ ]  focus tests - https://github.com/jquery/jquery/blob/a644101ed04d0beacea864ce805e0c4f86ba1cd1/test/unit/event.js#L2703, https://github.com/jquery/jquery/blob/a644101ed04d0beacea864ce805e0c4f86ba1cd1/test/unit/event.js#L2676

- [ ] isDefaultPrevented, stopImmediatePropagation - https://github.com/jquery/jquery/blob/a644101ed04d0beacea864ce805e0c4f86ba1cd1/test/unit/event.js#L375

scottgonzalez:
See also https://github.com/jquery/qunit/issues/741

markelog:
I'm gonna log in ticket desc what e2e-tests we need

timmywil:
Closing in favor of tracking in the [Roadmap](https://github.com/jquery/jquery/wiki/Roadmap).

Issue #1724, opened 2014-10-20
width(), height(), etc. shouldn't round


mzgol:
Originally reported by john at: http://bugs.jquery.com/ticket/9628
<p>
Right now we're rounding off the answers from width(), height(), etc. to the nearest pixel - this makes it hard to do good positioning of elements (especially for animations). For example if you have 3 items that are width 33% inside of a 100px element we're returning 33px for each element. Theoretically an animation method might be able to return a more interesting result if they know that there are extra fractional-numbers getting rounded off.
</p>
<p>
I got a report on this directly from Mozilla but I agree with them that this particular change makes it hard to get to good numbers.
</p>
<p>
See also: <a class="closed ticket" href="http://bugs.jquery.com/ticket/3239" title="bug: innerHeight/outerHeight does not round up (closed: fixed)">#3239</a>.
</p><p>Issue reported for jQuery 1.6.1</p>

MarventusWP:
I would like to add, since I don't see it mentioned anywhere in here, that .css("height") and .css("width") are also rounding values. Here's a fiddle: http://jsfiddle.net/Marventus/L16gmjuz/
Hope this  helps.

mzgol:
> I would like to add, since I don't see it mentioned anywhere in here, that .css("height") and .css("width") are also rounding values.

Yes, that's caused by the same code. `.css('width')` is going through a special hook instead of `getComputedStyle` so that it works on disconnected nodes, see:
https://github.com/jquery/jquery/blob/002240a6eb1cee2fcd886d5cf44893eb67f246f1/src/css.js#L71-L152
and:
https://github.com/jquery/jquery/blob/002240a6eb1cee2fcd886d5cf44893eb67f246f1/src/css.js#L353-L382

That's quite a lot of code, btw...

AurelioDeRosa:
Hi guys. What's the state of this issue? Will it be fixed in jQuery 3?

arthurvr:
@AurelioDeRosa Well, it's on the v3.0 milestone at least. :)

AurelioDeRosa:
@arthurvr I saw the label as well but this doesn't reply to my question.

mzgol:
@AurelioDeRosa It's planned for 3.0 and I'm assigned. I just need to get to it. :) I'll try to do it soon, I'd like it to land before we tag the first beta.

mzgol:
PR: #2439. Feedback welcome.

mzgol:
[...]

PR #2401, opened 2015-06-16
CSS: dimensions workaround for IE11 fullscreen quirk

### LINKED ISSUE #1764: IE 11 fullscreen inside iframe needs workaround
Originally reported by christopher@… at: http://bugs.jquery.com/ticket/14903
<p>
IE 11 has a bug in its fullscreen support. If the element being made fullscreen is inside of an iframe, the values for offsetWidth/offsetHeight and the results of getBoundingClientRect are 100x too small.
</p>
<p>
Details of the issue are here:
</p>
<p>
<a class="ext-link" href="http://christophercurrie.github.io/technology/2014/03/20/internet-explorer-11-fullscreen-bug.html" rel="nofollow"><span class="icon">&nbsp;</span>http://christophercurrie.github.io/technology/2014/03/20/internet-explorer-11-fullscreen-bug.html</a>
</p>
<p>
It's also been noticed that click positions *may* be off by 100x in the opposite direction. I'll update the ticket if we find a repro for that issue as well.
</p><p>Issue reported for jQuery 2.1.0</p>

AVGP:
Fixes #1764

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -113,6 +113,17 @@ function getWidthOrHeight( elem, name, extra ) {
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
+	// Support: IE11
+	// Fix for edge case in IE 11. See gh-1764
+	if ( document.msFullscreenElement && window.top !== window ) {
+		// Support: IE11
+		// Running getBoundingClientRect on a
+		// disconnected node in IE throws an error
+		if ( elem.getClientRects().length ) {
+			val = Math.round( elem.getBoundingClientRect()[ name ] * 100 );
+		}
+	}
+
 	// Some non-html elements return undefined for offsetWidth, so check for null/undefined
 	// svg - https://bugzilla.mozilla.org/show_bug.cgi?id=649285
 	// MathML - https://bugzilla.mozilla.org/show_bug.cgi?id=491668


mzgol:
This fix will not work if IE/Edge fixes this bug so I don't think it'll work. We'd need to have a way to feature detect this bug but this would require going to fullscreen which is impossible (and we wouldn't want to do such an invasive test anyway).

Is there any non-intrusive way to detect the issue?

mzgol:
I guess it would work if we could get a confirmation from the MS Edge team that this will *not* get fixed in IE11 as Edge has unprefixed the fullscreen API. We'd workaround it only for IE11 and just hope Edge fixes it soon.

Thoughts, @dmethvin?

AVGP:
@mzgol Valid point. Do you have any info regarding this bug for Edge?
But I'm sure there's gonna be a fix at some point.

There actually is one other possibility as pointed out by http://christophercurrie.github.io/technology/2014/03/20/internet-explorer-11-fullscreen-bug.html which is checking if clientWidth / clientHeight are smaller than offsetWidth / offsetHeight.

Should I try that approach instead?

AVGP:
@mzgol I have now narrowed down the fix to only kick in if dimensions are obviously wrong (clientrect > offset)

mzgol:
@AVGP A minor difference in output of these two methods may happen because of fractional values etc. I'd feel safer if you checked that it's much larger, e.g. 99 larger (that's the bug, isn't it?) - maybe not 100x to account for minor differences.

Ideally we'd do it via a support test (see https://github.com/jquery/jquery/blob/master/src/css/support.js and how `support.*` is used in the code) & cache the result but here the difference exists only in fullscreen mode so I guess we do have to check it every time.

We could cache the result on its first failure but I'm not sure if it's worth it as most people will use the non-fullscreen mode and we couldn't cache the result there.

Unfortunately, we have no way of automatically testing it so it will have to stay without a test for now.

It's starting to look good! :+1:

src/css.js
```diff
@@ -113,6 +113,14 @@ function getWidthOrHeight( elem, name, extra ) {
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
+	// Fix for edge case in IE 11. See https://github.com/jquery/jquery/issues/1764
+	if ( window.top !== window.self && document.msFullscreenElement &&
+		elem.getBoundingClientRect().width > elem.offsetWidth ) {
```
mzgol: Shouldn't this be the other way round?

src/css.js
```diff
@@ -113,6 +113,14 @@ function getWidthOrHeight( elem, name, extra ) {
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
+	// Fix for edge case in IE 11. See https://github.com/jquery/jquery/issues/1764
```
mzgol: Add:
```
// Support: IE11+
```
over this line. We document all browser workarounds in this way.

src/css.js
mzgol: Hmm, wait a moment. From the description at https://connect.microsoft.com/IE/feedback/details/838286/ie-11-incorrectly-reports-dom-element-sizes-in-fullscreen-mode-when-fullscreened-element-is-within-an-iframe it seems that both values, `elem.getBoundingClientRect().width` & `elem.offsetWidth` are 100 times too small. Is this test correct?

AVGP: Interesting. I have tested against [IE 11 on Windows 7](https://az412801.vo.msecnd.net/vhd/VMBuild_20141027/VirtualBox/IE11/Windows/IE11.Win7.For.Windows.VirtualBox.zip) using the VM that Microsoft provides and found getBoundingClientRect to return correct values, while offsetWidth/offsetHeight returned values 100x smaller (but rounded, so I couldn't just multiply those with 100).

I observed this behaviour consistently throughout my tests...

src/css.js
AVGP: Thanks, sorry! Didn't know that :-)

src/css.js
mzgol: I have a Windows 10 VM around, I'll check if the problem exists in IE11 & Edge and to what extent.

mzgol: BTW, I'll be working soon on fixing #1724 so there will be more `getBoundingClientRect` around. It would be cool if it was correct in IE11 because this whole workaround wouldn't be needed then. ;)

AVGP: thanks!

AVGP: Nevermind! I'm an idiot. I mixed up clientWidth / clientHeight with the bounding rect. Dang! So that check is not going to work :/

AVGP: But that's not a big problem. So I'll check if the ratio clientWidth / offsetWidth is larger than 90. But that's a bit of a danger, as somebody might use a giant border...

AVGP:
@mzgol Alright, I fixed the detection to use the ratio of clientWidth and offsetWidth and the workaround only becomes effective if that's larger than 90 (as there is some rounding error involved).

I don't think using the support.* method and its caching is useful here, it might rather introduce issues me thinks (especially as it's such a nasty edge case).

The code works reliably in my IE 11 / Win7 VM, what does your IE 11 / Edge / Win 10 VM say?

mzgol:
@AVGP I see the issue in IE11 with both `offsetWidth` & `getBoundingClientRect`:
fullscreen-bug-ie11

Edge doesn't experience this bug (\o/):
fullscreen-ok-edge

AVGP:
@mzgol Yup, as I said: I mixed up clientWidth vs. getBoundingRect().width - I fixed that in my latest change.

And hooray for Edge! \o/ :rainbow:

mzgol:
Fortunately, Edge not only supports the unprefixed fullscreen API but it also *doesn't* support the prefixed one. So if `document.msFullscreenElement !== undefined` (if the browser is not in fullscreen mode this will equal `null`) then we know we're in IE11 land and this bug exists.

I think comparing with `clientWidth` is risky as I'm sure there are cases where people create elements with giant borders (some icons are often emulated this way).

So, judging that we cannot test it reliably I think we could construct a support test like:
```js
support.widthReliableInFullscreenMode = 'msFullscreenElement' in document;
```
with a comment about why we don't have a "true" test and then in the code if this value is `true` then check if `document.msFullscreenElement` is truthy and then apply your workaround.

I'll reach out to MS to make sure they don't intend to fix it in IE11, if that's true I think we're good to go.

AVGP:
@mzgol Coool! Alright, that makes things much more predictable!

Now one last question: shouldn't it be

```javascript
support.widthUnreliableInFullscreenMode = 'msFullscreenElement' in document;
```
so the inverse (check name) as the width is *unreliable* in fullscreen if we're having the IE 11?

AVGP:
@mzgol by the way: Thanks for onboarding me so patiently, really appreciate your support :bow:

mzgol:
> Now one last question: shouldn't it be
> 
> support.widthUnreliableInFullscreenMode = 'msFullscreenElement' in document;

I'd prefer to not introduce a negation in the name, it's harder to comprehend what `widthUnreliableInFullscreenMode` being `false` means. Most our support tests follow this advice - if they're `true`, the browser passes the test (this isn't true for all the tests but shhh ;)). it just means that you need to negate the condition as I made the mistake in the test. :)

> @mzgol by the way: Thanks for onboarding me so patiently, really appreciate your support :bow: 

It's great to have dedicated contributors. :) Pure pleasure.

AVGP:
@mzgol I went with `reliableDimensionsInFullscreenMode` to follow the convention in `support.*` and to express that it's actually both width and height that's affected.

mzgol:
> @mzgol I went with reliableDimensionsInFullscreenMode to follow the convention in support.* and to express that it's actually both width and height that's affected.

Sounds good. :+1:

src/css/support.js
```diff
@@ -96,7 +96,11 @@ define([
 			div.removeChild( marginDiv );
 
 			return ret;
-		}
+		},
+		// Support: IE 11
+		// IE 11 has a bug with offsetWidth / offsetHeight / getClientBoundingRect()
+		// in fullscreen mode inside an iframe
```
mzgol: I'd describe what exactly the bug is about and link to the MS Connect bug about it.

src/css.js
```diff
@@ -113,6 +113,15 @@ function getWidthOrHeight( elem, name, extra ) {
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
+	// Support: IE11+
+	// Fix for edge case in IE 11. See https://github.com/jquery/jquery/issues/1764
+	if (!support.reliableDimensionsInFullscreenMode && window.top !== window.self &&
+		document.msFullscreenElement) {
+		val = name === "width" ?
+			Math.round(elem.getBoundingClientRect().width  * 100) :
+			Math.round(elem.getBoundingClientRect().height * 100);
```
mzgol: You can shorten it:
```js
val = Math.round( elem.getBoundingClientRect()[ name ] * 100 );
```

src/css.js
```diff
@@ -113,6 +113,15 @@ function getWidthOrHeight( elem, name, extra ) {
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
+	// Support: IE11+
+	// Fix for edge case in IE 11. See https://github.com/jquery/jquery/issues/1764
+	if (!support.reliableDimensionsInFullscreenMode && window.top !== window.self &&
```
mzgol: We put spaces after the opening paren and before the closing one, same below when using `Math.round`. For more details, see https://contribute.jquery.org/style-guide/js/#spacing

src/css/support.js
```diff
@@ -96,7 +96,11 @@ define([
...
+		reliableDimensionsInFullscreenMode: !("msFullscreenElement" in document)
```
mzgol: Spaces after `(` and before `)`.

src/css/support.js
AVGP: The next line describes it in more detail. Added a link to the MS connect bug now.

src/css.js
AVGP: Oh yeah, why didn't I notice :D Thanks!

src/css.js
AVGP: I see. I wonder why jscs didn't catch that...

dmethvin:
Just to summarize, it looks like Edge fixed this so the use of `"msFullscreenElement" in document` is reasonable to catch the problem in IE11?

I think the `reliableDimensionsInFullscreenMode` name is fine. Years ago most of the feature detects were structured so that the normal/passing situation was `false` and the failing case (generally old IE) was `true`. That led to a lot of confusing negation.

src/css/support.js
```diff
@@ -96,7 +96,12 @@ define([
 			div.removeChild( marginDiv );
 
 			return ret;
-		}
+		},
+		// Support: IE 11
+		// IE 11 has a bug with offsetWidth / offsetHeight / getClientBoundingRect()
+		// in fullscreen mode inside an iframe the values are 100x smaller than they should be
+		// See https://connect.microsoft.com/IE/feedback/details/838286/ie-11-incorrectly-reports-dom-element-sizes-in-fullscreen-mode-when-fullscreened-element-is-within-an-iframe
```
mzgol: Maybe shorten the URL to just https://connect.microsoft.com/IE/feedback/details/838286 because it's taking a lot of editor columns. :)

src/css.js
```diff
@@ -113,6 +113,13 @@ function getWidthOrHeight( elem, name, extra ) {
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
+	// Support: IE11+
+	// Fix for edge case in IE 11. See https://github.com/jquery/jquery/issues/1764
+	if ( !support.reliableDimensionsInFullscreenMode && window.top !== window.self &&
+		document.msFullscreenElement ) {
+		val = Math.round( elem.getBoundingClientRect()[name] * 100 );
```
mzgol: Spaces are needed after `[` and before `]` as well. We should get this JSCS update ASAP because it's hard to catch all these cases manually...

AVGP:
@dmethvin Thanks for the summary.

Yes, the bug is fixed in Edge, according to the test @mzgol did in https://github.com/jquery/jquery/pull/2401#issuecomment-112757460 - so all that's left is to use `msFullscreenElement` to figure out if we're in IE11 where this problem will need the workaround.

mzgol:
@AVGP One more thing: we unit-test all support test results in https://github.com/jquery/jquery/blob/master/test/unit/support.js to be sure we don't break those tests later; could you add your new test there?

AVGP:
@mzgol Hm, to be fair I found these pretty unwieldy and I've got no idea how to add my test there.

AVGP:
@mzgol Nevermind, I *think* I know how to add it to the tests.
What's the UA for MS Edge (vs. IE 11)

mzgol:
@AVGP Edge is already added to the list there. :) https://github.com/jquery/jquery/blob/842958e7aecd0d75a7ee9e2aaec83457701aa2f3/test/unit/support.js#L54

AVGP:
@mzgol Yes, but it was missing an IE11 clause, I think.
Check latest changes -- I hope I did it correctly.

I took the UA info from https://msdn.microsoft.com/en-us/library/hh869301(v=vs.85).aspx

test/unit/support.js
```diff
@@ -67,7 +67,27 @@ testIframeWithCallback( "Check CSP (https://developer.mozilla.org/en-US/docs/Sec
 			"pixelMarginRight": true,
 			"pixelPosition": true,
 			"radioValue": true,
-			"reliableMarginRight": true
+			"reliableMarginRight": true,
+			"reliableDimensionsInFullscreenMode": true
+		};
+	} else if ( /rv:11\.0/i.test( userAgent ) ) {
```
mzgol: IE 11 was already captured by the `trident\/7\.0` part in the regex below; they were grouped as all test results were the same, now they have to be divided.

I'd use `trident\/7\.0`, though and not `rv:11\.0` as that would conflict with Firefox 11. We don't test on Firefox 11 but it's always better to prevent confusion if possible. Also, looking at `trident` it's more clear we're dealing with IE than when looking at a generic `rv`.

test/unit/support.js
```diff
@@ -67,7 +67,27 @@ testIframeWithCallback( "Check CSP (https://developer.mozilla.org/en-US/docs/Sec
...
+		expected = {
+			"ajax": true,
+			"boxSizingReliable": true,
```
mzgol: This should be `false`. Generally, you can copy over the shared test results of IE10/11 and just add your test, the only one that will differ.

src/css.js
```diff
@@ -113,6 +113,13 @@ function getWidthOrHeight( elem, name, extra ) {
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
+	// Support: IE11+
+	// Fix for edge case in IE 11. See https://github.com/jquery/jquery/issues/1764
+	if ( !support.reliableDimensionsInFullscreenMode && window.top !== window.self &&
+		document.msFullscreenElement ) {
+		val = Math.round( elem.getBoundingClientRect()[ name ] * 100 );
```
mzgol: Hmm, now that I think about it... Since we currently use `elem.offsetWidth` & `elem.offsetHeigh` that are non-fractional in all browsers, we should use the same methods here. IOW, you can just assign:
````js
val *= 100;
```
as you already have the value, it's just 100 times too small. I'll most likely change it to `getBoundingClientRect` later to get fractional values but I'll do it for all browsers at once, it should consistently be integers before that happens.

test/unit/support.js
```diff
@@ -231,7 +259,8 @@ testIframeWithCallback( "Check CSP (https://developer.mozilla.org/en-US/docs/Sec
 			"pixelMarginRight": false,
 			"pixelPosition": false,
 			"radioValue": true,
-			"reliableMarginRight": true
+			"reliableMarginRight": true,
+			"reliableDimensionsInFullscreenMode": true
```
mzgol: Can you put it before `reliableMarginRight` everywhere? We're trying to keep the list alphabetically sorted. :)

src/css.js
AVGP: That'll give incorrect values, as `elem.offset*` will for instance be "3" where `getClientBoundingRect().width` would be 3.14 -- that's why I'm not using `offset*` here.

mzgol: Yes, that's intended, it has worked this way in all browsers since almost forever. ;) We don't support fractional width/height anywhere for now, #1724 is supposed to change that but until that happens we need to be consistent.

dmethvin: Yes, unfortunately we should return the integer values until we jump off the cliff at some point (if ever) and use `gBCR` everywhere.

gibson042:
Recommendation: `reliableFullscreenBox` (or, if you _must_, `reliableFullscreenDimensions`) instead of `reliableDimensionsInFullscreenMode`.

src/css.js
AVGP: Uhm, no - just multiplying `offsetWidth` by 100 is incorrect and this function *is* returning integer values, Gentlemen :-)

Look - if the clientWidth is `314` and the IE 11 bug happens, we'll get the following values:

```javascript
elem.offsetWidth == 3
elem.getBoundingClientRect().width == 3.14
Math.round( elem.getBoundingClientRect().width * 100 ) == 314
elem.offsetWidth * 100 == 300
```
which makes this value still integer but actually correct (`314` as expected versus `300` which your suggestion would return)

mzgol:
> Years ago most of the feature detects were structured so that the normal/passing situation was false and the failing case (generally old IE) was true. That led to a lot of confusing negation.

On `master` in all support tests `true` means "passed". On `compat` there is one exception: [ownLast](https://github.com/jquery/jquery/blob/a0a5c0be2def09f8d380a69d8e6f000169123990/test/unit/support.js#L224). Perhaps we should change it to be consistent.

mzgol:
> Recommendation: reliableFullscreenBox

Works for me, especially that `support` properties are not mangled so they affect the final build size.

AVGP:
@gibson042 done! Thanks for the suggestion, muuuch better!

src/css.js
mzgol: Ahh, stupid me, you're right. :) Looks good then.

src/css.js
```diff
@@ -113,6 +113,13 @@ function getWidthOrHeight( elem, name, extra ) {
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
+	// Support: IE11+
+	// Fix for edge case in IE 11. See https://github.com/jquery/jquery/issues/1764
+	if ( !support.reliableFullscreenBox && window.top !== window.self &&
+		document.msFullscreenElement ) {
+		val = Math.round( elem.getBoundingClientRect()[ name ] * 100 );
```
mzgol: One other thing: as you can see in [src/offset.js#L93-L98](https://github.com/jquery/jquery/blob/842958e7aecd0d75a7ee9e2aaec83457701aa2f3/src/offset.js#L93-L98), in IE (all versions) running `getBoundingClientRect()` on disconnected nodes throws an error and we support `elem.css('width')` on disconnected elements. So you need to set the value only for non-disconnected node, you can use a similar `getClientRects()` check as in the offset module.

I guess we could abstract this function and put in a separate module but this optimization may as well be done later, for now we just need to ensure this won't throw.

AVGP: Alright, good catch!

For now I'll check if we can use `getClientBoundingRects` using the method that's used in https://github.com/jquery/jquery/blob/842958e7aecd0d75a7ee9e2aaec83457701aa2f3/src/offset.js#L93-L98 - if we cant, I'll approximate the dimension using `val *= 100` which is better than a value that's off by a factor of 100, I think.

src/css.js
```diff
@@ -113,6 +113,20 @@ function getWidthOrHeight( elem, name, extra ) {
...
 
+	// Support: IE11+
+	// Fix for edge case in IE 11. See https://github.com/jquery/jquery/issues/1764
+	if ( !support.reliableFullscreenBox && window.top !== window.self &&
+		document.msFullscreenElement ) {
+		// Support: IE<=11+
+		// Running getBoundingClientRect on a
+		// disconnected node in IE throws an error
+		if ( !elem.getClientRects().length ) {
+			val *= 100;
```
mzgol: You skip this assignment and just do anything in the case where `.length` is truthy. Disconnected elements have `offsetWidth === 0` (see [the spec](http://dev.w3.org/csswg/cssom-view/#dom-htmlelement-offsetwidth)) so `val === 0` at this point for such elements; multiplying it by 100 won't change anything.

mzgol: We just need to make sure disconnected elements will enter the `if ( val <= 0 || val == null )` condition below.

AVGP: Ahhh, I see :-) Thanks for clearing that up!

jacobrossi:
Yep, we don't intend to fix this in IE11. So this fix LGTM!

mzgol:
@jacobrossi Thanks for the info!

@AVGP LGTM. :)

src/css.js
```diff
@@ -113,6 +113,18 @@ function getWidthOrHeight( elem, name, extra ) {
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
+	// Support: IE11+
+	// Fix for edge case in IE 11. See https://github.com/jquery/jquery/issues/1764
```
markelog: We usually do the `gh-<issue>` then providing the full link

src/css.js
```diff
@@ -113,6 +113,18 @@ function getWidthOrHeight( elem, name, extra ) {
...
+	if ( !support.reliableFullscreenBox && window.top !== window.self &&
+		document.msFullscreenElement ) {
```
markelog: I don't get it, `reliableFullscreenBox` checks `msFullscreenElement` property and here you do the same?

src/css.js
```diff
@@ -113,6 +113,18 @@ function getWidthOrHeight( elem, name, extra ) {
...
+		// Support: IE<=11+
+		// Running getBoundingClientRect on a
+		// disconnected node in IE throws an error
+		if ( elem.getClientRects().length ) {
```
markelog: To check if node is connected or not, we use `jQuery.contains( elem.ownerDocument, elem )`, perf would be interesting to see though

src/css.js
mzgol: Yes, the support test just checks if `msFullscreenElement` is supported, here we check if we're in fullscreen mode as only there the bug exists.

markelog:
Hm, fix this issue without tests is dangerous, we can't unit-test it, but we can e2e-test?  I would like to see analogue of <a href="https://github.com/jquery/jquery/tree/compat/test/integration">`test/integration`</a> for the `master` branch here, since after we do https://github.com/jquery/jquery/issues/2329 we could this test in our workflow

src/css.js
mzgol: We seem to be inconsistent now, e.g.: 0e4477c676db0427bb9b0bf39df8631501e62f24.

We have to use `jQuery.contains` on `compat`, not necessarily on `master`.

Note that here we don't really care if the node is disconnected but we care for `elem. getBoundingClientRect()` to not throw. These situations are equivalent, though.

This code is restricted to one browser, anyway so whatever works in IE11 is fine.

mzgol:
> Hm, fixing this issue without tests is dangerous, we can't unit-test it, but we can e2e-test it?

I guess we could port the test case from http://christophercurrie.github.io/technology/2014/03/20/internet-explorer-11-fullscreen-bug.html.

I wouldn't call not adding it dangerous, though. As long as we don't automate integration testing, we can't really say we're testing the issue. This is more a reference than a test that ensures we don't regress.

src/css.js
markelog: Created

mzgol: @markelog Thanks. Since we have the issue now, I wouldn't block this PR over it.

markelog:
@mzgol absence of automatization of those tests is dangerous, just like leaving fix without the test

src/css.js
markelog: Should it be a property then?

markelog: Like `typeof support.reliableFullscreenBox === function`

src/css.js
markelog: Old jscs, update will follow after 2.0

src/css.js
mzgol: I don't understand what you propose.

`document.fullscreenElement` is the element put into fullscreen mode if any is, otherwise `null` in browsers that support the fullscreen API. IE11 is the only browser to have the bug we're working around here and the only browser that supports the API in the form of `document.msFullscreenElement`. Therefore, in the support test we check if this property exists at all and here we test if any element is in fullscreen mode.

mzgol:
> On compat there is one exception: ownLast. Perhaps we should change it to be consistent.

I created an issue to track it: https://github.com/jquery/jquery/issues/2406

src/css.js
markelog: Yes, but this support method could return information about both i.e. if fix should be applied or not

markelog: It can even store the `window.top !== window.self` bit

mzgol: Try testing that on TestSwarm then. :-) This would be needlessly complicated IMO.

Our support tests currently are either a boolean property or a function that returns a boolean (and that result is stable, i.e. it always returns the same value in the same browser). Complicating that logic would be confusing.

I think the separation to:

1. a test checking if the bug exists
2. its application where necessary

is a good thing to have.

dmethvin: I had the same confusion about why `msFullscreenElement` was being (re)checked but then I [read the docs](https://msdn.microsoft.com/en-us/library/dn254937\(v=vs.85\).aspx) and I assume this variable is there but `null` if we're not *currently* in fullscreen. Their docs aren't very clear but checking on the IE11 console that seems to be the case.

> It can even store the `window.top !== window.self` bit

Hmm, I wonder if we need to handle the case where a jQuery in one frame is manipulating elements in another frame. We do get occasional bug reports about those use cases. If we wanted to handle that we would need to use `elem.ownerDocument` and `elem.document.defaultView` each time?

markelog: > Try testing that on TestSwarm then. :-) 

We can't right now, right now we do only ua-testing, which is bad substitution. We can check this with e2e-tests. You might not remember this, but those ua tests were created as temp solution.

> Complicating that logic would be confusing.

I don't think it would complicate stuff, the opposite - you have function that checks if fix should be applied, that is it, nothing else, nothing more - all encapsulated. Now this is all united, dependant, tma and all that, which is bad.

you could do
 `if (isNeeded()) { "there you go" }`

instead of doing
 `if ("maybe-browser-has-this" || "or-maybe-that" && "but-if-doesn't-then") { "wtf was that?!" }`

markelog: >  We can check this with e2e-tests. You might not remember this, but those ua-tests were created as temp solution.

Also, that would be another good step of removing of exposure of the `jQuery.support`

christophercurrie:
Many thanks to @AVGP and @mzgol for working so hard to get this fix in. :+1:

src/css.js
mzgol: @markelog 
>> Try testing that on TestSwarm then. :-)
>
> We can't right now, right now we do only ua-testing, which is bad substitution.

I meant that TestSwarm wraps everything in an additional wrapper iframe which changes results here.

> We can check this with e2e-tests. You might not remember this, but those ua-tests were created as temp solution.

I had an impression they exist so that we're sure we don't regress support tests results in our supported browsers. Many of those results are already implicitly tested via testing APIs that depend on them working.

Anyway, any such change is a plan reaching far further than this PR as it'd require much larger effort. We may add a test page in the integration folder but otherwise I don't think there's anything else to do in this PR at this point.

mzgol:
> @mzgol absence of automatization of those tests is dangerous, just like leaving fix without the test

Right; I just having a test in `test/integration/` that's not run automatically doesn't really decrease the danger level. ;)

src/css.js
```diff
@@ -113,6 +113,18 @@ function getWidthOrHeight( elem, name, extra ) {
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
+	// Support: IE11+
```
mzgol: `// Support: IE11 only` as there is not IE12 and it's fixed in Edge.

src/css.js
gibson042: > > It can even store the `window.top !== window.self` bit

> Hmm, I wonder if we need to handle the case where a jQuery in one frame is manipulating elements in another frame. We do get occasional bug reports about those use cases. If we wanted to handle that we would need to use `elem.ownerDocument` and `elem.document.defaultView` each time?

To maximize internal consistency, I recommend matching [Sizzle](https://github.com/jquery/sizzle/blob/master/src/sizzle.js#L509): `(parent = document.defaultView) && parent.top !== parent`, where `document` is `elem.ownerDocument`. Unless, of course, this bug also seeps into child frames, in which case it should be `window.top !== window`. In either case, though, `self` is redundant.

markelog: @mzgol 

> Anyway, any such change is a plan reaching far further than this PR as it'd require much larger effort

For the whole `jQuery.support` yes, but we discussing here encapsulation of this specific method.

@gibson042 Well here, we need to change some things, like not `parent` but `window.parent`

mzgol: Encapsulation of this method currently would make every supported browser return `true` on this support test in `test/unit/support.js` where run outside of an iframe except on TestSwarm where it would return `false` in IE11.

Also, if we want to consider what @dmethvin said:
> I wonder if we need to handle the case where a jQuery in one frame is manipulating elements in another frame. 

then it would complicate our TestSwarm situation even further.

I feel changing this one property in a way you propose is complicated enough on its own from the testing point that it warrants a separate issue, we shouldn't require the OP to deal with all that.

markelog: I don't think assigning property value as `!( "msFullscreenElement" in document )` and then checking it in IE is real test for this bug.

As i said above, we could deal with `support` properties latter, right now, having test in integration folder while leaving comment that suggest to check specific test if someone would like to change relevant lines would be enough for me.

markelog: > I wonder if we need to handle the case where a jQuery in one frame is manipulating elements in another frame.

That sounds as an edge case of an edge case :-), i would suggest to deal with that situation when/if user actually face it

markelog:
> Right; I just having a test in test/integration/ that's not run automatically doesn't really decrease the danger level. ;)

Well, that is two different issues, one is to follow XP covenants and other is to have test for the specific issue.

For our predecessors from the seventies second, was only one thing that mattered, provided that they tested their software it before shipping.

src/css.js
mzgol: @markelog
> As i said above, we could deal with support properties later, right now, having test in integration folder while leaving comment that suggest to check specific test if someone would like to change relevant lines would be enough for me.

Sounds good to me as well.

markelog:
> For our predecessors, from the seventies, second, was only one thing that mattered, provided that they tested their software before shipping.

If you know what i'm hinting to you about :-)

src/css.js
markelog: Just checking, we agree that we want to encapsulated all circumstances (like `window.top !== window.self`) of the described bug to `reliableFullscreenBox ` (or different a name), add integration test and don't check `support` property that relates to it?

mzgol: @markelog In this PR I think adding the integration test would be enough. I don't like the idea of creating a support test function that works differently than all others; there is a lot more code that maybe could be refactored this way ([checkClone](https://github.com/jquery/jquery/blob/842958e7aecd0d75a7ee9e2aaec83457701aa2f3/src/manipulation.js#L126), [clearCloneStyle](https://github.com/jquery/jquery/blob/842958e7aecd0d75a7ee9e2aaec83457701aa2f3/src/css.js#L233), [pixelMarginRight](https://github.com/jquery/jquery/blob/842958e7aecd0d75a7ee9e2aaec83457701aa2f3/src/css/curCSS.js#L30) etc.) and it would be inconsistent to change just one (I'm not sure I like this direction either way but I don't think the discussion belongs to this PR/issue).

The only additional things I'm OK with doing in this PR are adding a HTML page with an integration test and optionally removing this support test altogether and just checking `if ( document.msFullscreenElement && window.top !== window )` as the test doesn't buy us a lot, indeed. Everything else I'd like to discuss separately as they seem to be refactorings that affect way more than just code in this PR.

mzgol: @dmethvin 
> I had the same confusion about why msFullscreenElement was being (re)checked but then I read the docs and I assume this variable is there but null if we're not currently in fullscreen.

That's the usual idiom for DOM stuff. :) A property always exists and is just set to `null` (not `undefined`) if not defined. The same applies for e.g. `history.state`, you can just check if it's `!== undefined` to be sure it's supported.

AVGP: Now I've lost track. What do I need to do to land this fix?
Am 17.06.2015 22:30 schrieb "Michał Gołębiowski" <notifications@github.com>:

markelog: Okay, i hear you, created https://github.com/jquery/jquery/issues/2409

mzgol: The thread got long and a lot of it is unrelated to the fullscreen issue.
:) I'll write a summary for you when I get home (I'm writing from a mobile
right now).

-- 
Michał Gołębiowski

gibson042: > removing this support test altogether and just checking `if ( document.msFullscreenElement && window.top !== window )` as the test doesn't buy us a lot, indeed

Seconded.

markelog: @AVGP yeah, sorry about that, we get like this sometimes :-)

AVGP:
Right, in summary: I'm lost. I need a short, actionable summary of what I have to do to make this land.

@mzgol is going to put one together for me, so stay tuned.

P.S.: Data roaming su-ucks!

AVGP:
@christophercurrie Thanks for the nice, isolated test case :-)

mzgol:
@AVGP OK, there's one thing needed to finish this PR: remove `support.reliableFullscreenBox`, i.e. revert all changes in `src/css/support.js` & `test/unit/support.js` and change the condition `if ( document.msFullscreenElement  && window.top !== window )`. Sorry for guiding you into this rabbit hole, that was my mistake as the test doesn't buy us anything here, we have to check the element later anyway. Also, note the change from `window.self` to `window` as they're equivalent. This is not a major change, though so I can finish it up if you want.

One other thing is the integration test page but there are a couple of things needed to remember there so I can do it myself in a separate commit and not bother you about it. :) But I'll describe: one should create an HTML page similar to https://github.com/jquery/jquery/blob/compat/test/integration/gh-1684-ajax.html (note that this one is present only in the `compat` branch so you'll need to create a `test/integration` directory). This comment says how it looks like: https://github.com/jquery/jquery/pull/2183#issuecomment-100984961. I wouldn't just copy  http://christophercurrie.github.io/technology/2014/03/20/internet-explorer-11-fullscreen-bug.html but create a simple page that will say "Test failed" or "Test passed" as the page I linked to. If a person is not in fullscreen mode instead of "Test passed" it should probably say "Go to fullscreen mode", there should also be a button allowing to go to fullscreen mode & back.

AVGP:
@mzgol Alright. I reverted the changes and adjusted the condition as desired.

I'm not sure if I will give the integration test a shot. It does sound interesting, but I simply don't know if I come around doing it.

If you take that upon you, that'd be great - if I have some time, I may give it a go (and will push here for you to see)... whatever gets us this fix landed :-)

mzgol:
OK, I'll add the test. The rest looks ready. Thanks @AVGP!

AVGP:
@mzgol Well thanks to you, @markelog, @christophercurrie and @gibson042 for helping me making it this far :)

mzgol:
@AVGP Landed on `master` & `compat`, thanks! I added the integration test in a separate PR, see #2425 (screenshots attached).

AVGP:
@mzgol Thanks for making it so! I'm looking forward to the next release then :-)

Issue #2406, opened 2015-06-17
Reverse support.ownLast


mzgol:
`jQuery.support.ownLast` is the only support test that's supposed to be `false` when it succeeds. It's confusing, we should reverse its value.

mzgol:
PR: #2408

mzgol:
Closed via #2408.

mzgol:
Clearing the milestone as we're not releasing `compat` after all and this change won't be applicable to code that has to support IE6/7.

markelog:
@mzgol fyi, since we are releasing another compat, `wonfix` removed

mzgol:
Not compat, but regular 1.x. I added wontfix here because this issue doesn't matter for code soon to be dead. If it's included in your 1.12-stable branch, though, then removing wontfix is fine.

mzgol:
But, if you're removing the wontfix label, please remember to open the issue and assign the proper milestone...

markelog:
> Not compat, but regular 1.x.

It is just a name change, so for me, it is a same thing.

> But, if you're removing the wontfix label, please remember to open the issue and assign the proper milestone...

should be in general

markelog:
> Not compat, but regular 1.x.

Apparently, for the team too, since we still use `1.x` label

PR #2408, opened 2015-06-17
Core: Change support.ownLast to support.ownFirst

mzgol:
jQuery.support.ownLast was the only support test that was supposed
to be false when it succeeded. It was confusing.

Fixes gh-2406


timmywil:
:+1:

mzgol:
Landed at 523de777f933d3f50cdf1fe7473a23a368aedcad.

PR #2423, opened 2015-06-21
Ajax: Retrieve GET variables from url

Jakemichaeldrew:


dmethvin:
Hi @Jakemichaeldrew, hold on there! Stop making new pull requests for now.

dmethvin:
If you  need to make changes to a pull request, just push new commits to the existing PR. Don't create a new PR. However, this isn't a PR we'd accept so just give us a chance to catch up wtih all the activity you've created here.

PR #2425, opened 2015-06-22
CSS: Add an integration test for issue gh-1764

mzgol:
Refs gh-1764
Refs gh-2401

Screenshots in IE11:
1. not fullscreen:
screen shot 2015-06-23 at 01 42 16
2. fullscreen without the patch: 
screen shot 2015-06-23 at 01 42 34
3. fullscreen with the patch: 
screen shot 2015-06-23 at 01 53 59

### DIFF
--- a/test/integration/data/gh-1764-fullscreen-iframe.css
+++ b/test/integration/data/gh-1764-fullscreen-iframe.css
@@ -1,0 +1,18 @@
+.result {
+	font-size: 24px;
+	margin: 0.5em 0;
+	width: 700px;
+	height: 56px;
+}
+
+.error {
+	background-color: red;
+}
+
+.warn {
+	background-color: yellow;
+}
+
+.success {
+	background-color: lightgreen;
+}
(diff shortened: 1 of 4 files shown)


markelog:
This is ported from http://christophercurrie.github.io/technology/2014/03/20/internet-explorer-11-fullscreen-bug.html?

mzgol:
No, written from scratch. I wanted to make it as similar to the `compat` PATCH test as possible.

mzgol:
@timmywil @markelog is it OK to land?

markelog:
This probably would need to be changed anyway if/when we integrate e2e-testing, so if it works now it is LGTM.

mzgol:
I forgot: I should add another div on the page that's not in an iframe to
ensure we're correcting the output only when it's wrong. The script needs
to be extracted to avoid the duplication. I'll update the PR tomorrow.

-- 
Michał Gołębiowski

mzgol:
I refactored the test; JS & CSS are now in separate files, shared by the main HTML & the iframe one.

mzgol:
I'll land it tomorrow around noon European time or slightly later.

PR #2427, opened 2015-06-23
Fix: organize prop & attr code to be similar

pgilad:
This fixes the previous faulty commit of d0388e9e806ca3b30e7bbaaa1c336b7c98dc5f88 on the `compat` branch.

The changes of that commit were reverted and the new (hopefully) correct changes were added again.

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -1,11 +1,11 @@
 define([
 	"../core",
-	"../var/rnotwhite",
 	"../core/access",
 	"./support",
+	"../var/rnotwhite",
 	"./val",
 	"../selector"
-], function( jQuery, rnotwhite, access, support ) {
+], function( jQuery, access, support, rnotwhite ) {
 
 var boolHook,
 	attrHandle = jQuery.expr.attrHandle,
@@ -18,7 +18,7 @@ jQuery.fn.extend({
 	},
 
 	removeAttr: function( name ) {
-		return this.each(function() {
+		return this.each( function() {
 			jQuery.removeAttr( this, name );
 		});
 	}
@@ -26,10 +26,10 @@ jQuery.fn.extend({
 
 jQuery.extend({
 	attr: function( elem, name, value ) {
-		var hooks, ret,
+		var ret, hooks,
 			nType = elem.nodeType;
 
-		// don't get/set attributes on text, comment and attribute nodes
+		// Don't get/set attributes on text, comment and attribute nodes
 		if ( nType === 3 || nType === 8 || nType === 2 ) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


src/attributes/attr.js
```diff
@@ -3,11 +3,14 @@ define([
 	"../core/access",
 	"./support",
 	"../var/rnotwhite",
+	"./val",
```
markelog: Why?

src/attributes/attr.js
```diff
@@ -74,6 +77,8 @@ jQuery.extend({
 			set: function( elem, value ) {
 				if ( !support.radioValue && value === "radio" &&
 					jQuery.nodeName( elem, "input" ) ) {
+					// Setting the type on a radio button after the value resets the value in IE8-9
```
markelog: Need a blank line here

src/attributes/attr.js
```diff
@@ -113,8 +126,13 @@ boolHook = {
 		if ( value === false ) {
 			// Remove boolean attributes when set to false
 			jQuery.removeAttr( elem, name );
+		} else if ( getSetInput || !ruseDefault.test( name ) ) {
+			elem.setAttribute( jQuery.propFix[ name ] || name, name );
+
 		} else {
-			elem.setAttribute( name, name );
+			// Support: IE<9
```
markelog: See https://contribute.jquery.org/style-guide/js/#comments, we will fix them all eventually, but i guess new code should adhere to it already

markelog:
I would like to merge it, waiting couple hours for @gibson042 response

src/attributes/attr.js
pgilad: This existed in the code before my previous commit. See https://github.com/jquery/jquery/blob/0de798d6c2277a286c90e08de669a23e6e12a4a2/src/attributes/attr.js#L6

mzgol: It doesn't exist in current `compat`: https://github.com/jquery/jquery/blob/d0388e9e806ca3b30e7bbaaa1c336b7c98dc5f88/src/attributes/attr.js#L6

pgilad:
Hey @markelog thanks for the review. I added some extra blank spaces where needed.

src/attributes/attr.js
pgilad: Yes because that was a bad merge from master by yours truly :blush:

mzgol: Ah, ok :)

pgilad: This is the file prior to the change: https://github.com/jquery/jquery/blob/0de798d6c2277a286c90e08de669a23e6e12a4a2/src/attributes/attr.js

mzgol: I guess the interesting diff is at https://github.com/pgilad/jquery/commit/ab692d396542b1a65a93460359fe1d8c6ca6ccf6.

mzgol: @markelog Since this PR contains the revert anyway and it makes sense to land it as 2 commits, maybe we should just merge the revert now (& fix the build this way) and then just review the proper patch? In this way we also wouldn't be in a hurry to quickly fix the build as it will be fixed already.

Now we're commenting on a Frankenstein diff...

markelog: I was thinking to unite those commits so revert one would go away and from history stand point, it would be just support commit

> Now we're commenting on a Frankenstein diff...

nicely put, lol

src/attributes/attr.js
```diff
@@ -122,19 +143,56 @@ boolHook = {
...
-		var ret, handle;
-		if ( !isXML ) {
-			// Avoid an infinite loop by temporarily removing this function from the getter
-			handle = attrHandle[ name ];
-			attrHandle[ name ] = ret;
-			ret = getter( elem, name, isXML ) != null ?
-				name.toLowerCase() :
-				null;
-			attrHandle[ name ] = handle;
+	attrHandle[ name ] = getSetInput || !ruseDefault.test( name ) ?
```
markelog: Would love to change to simple `if`...`else`, but i guess this story for another day

src/attributes/attr.js
mzgol: IMO it's harder to see the logic of the change when it's modifying the previous bad merge.

But however your prefer, just my opinion.

markelog: We can compare it like that https://github.com/pgilad/jquery/compare/6df1bf9be818b8955f94113aecc689fdf7a2e172...ab692d396542b1a65a93460359fe1d8c6ca6ccf6

Although, i'm also not pushing for specific land strategy

src/attributes/attr.js
pgilad: Done

src/attributes/attr.js
markelog: I wonder, we don't have it in `master` plus it seems we don't use this declaration, test doesn't fail without it. Maybe it is needed to build this module and that is indirect dependency?

@timmywil? We can deal with this later though

markelog:
Closed via https://github.com/jquery/jquery/commit/f2bcf87c25155fd8361c744c684a0d526bd184a5

markelog:
It seems we're passed on first try, thank you!

pgilad:
:clap:

src/attributes/attr.js
timmywil: My guess is it's leftover from a prop hook for retrieving button value or something. It doesn't seem necessary anymore.

PR #2428, opened 2015-06-25
Core:Suggest remove unnecessary parameter

mixed:
I removed unnecessary parameter for constructor of "end" method. It`s because jQuery constructor with no parameter is the same as jQuery constructor with null parameter.

[Review on Reviewable](https://reviewable.io/reviews/jquery/jquery/2428)

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -103,7 +103,7 @@ jQuery.fn = jQuery.prototype = {
 	},
 
 	end: function() {
-		return this.prevObject || this.constructor(null);
+		return this.prevObject || this.constructor();
 	},
 
 	// For internal use only.


mzgol:
LGTM, especially that the `jQuery()` signature is documented at http://api.jquery.com/jQuery/ and `jQuery( null )` is not.

Could you sign our [CLA](https://contribute.jquery.org/CLA/)? This is why the build fails.

mixed:
@mzgol 
Thank you. I singed CLA just. :)

mzgol:
@mixed You need to sign the CLA with your full name; the commit needs to have your full name as an author as well (it'd be best if you modified your Git config so that this problem doesn't happen in the future).

mixed:
OK. Again PR. :)

mzgol:
You didn't need to close the PR; it was enough to amend the commit & change the author info & then "git push --force" to the branch. In this way we'd have the whole discussion contained in one place, not spread accross multiple attempts.

mixed:
@mzgol 
I'm sorry...  My mistake. I am not comfortable "github".  I found this way. but I send PR(#2441) already.
I changed name and email. but I unsinged CLA still. Is it my problem?
https://github.com/jquery/jquery/pull/2441#issuecomment-117936736

mgol:
This issue was duplicated in #2441 so I removed the milestone from this one.

PR #2436, opened 2015-06-29
Serialize: handle arrays with null values

DanielNill:
Arrays with null values are treated as nested objects.  This causes something like `{foo: ["bar", null, "buzz"]}` to be serialized at `foo[]=bar&foo[1]=&foo[]=buzz` when really it should be serialized as `foo[]=bar&foo[]=&foo[]=buzz`

This error is caused because javascript's typeof operator treats null as an object.

It might be worth reviewing http://www.ecma-international.org/ecma-262/5.1/#sec-11.4.3 to make sure there aren't other cases of other poorly typed entities causing similar problems.


mzgol:
Thanks for the PR. If we want to change it, we should just use `jQuery.type` that already returns `"null"` for `null`, it was created specifically to workaround faulty `typeof` behavior.

Besides that, this sounds fine to me barring any back compat problems. Could you sign our [CLA](https://contribute.jquery.org/CLA/)?

test/unit/serialize.js
```diff
@@ -77,6 +77,9 @@ test("jQuery.param()", function() {
 	params = {"test": {"length": 3, "foo": "bar"} };
 	equal( jQuery.param( params, false ), "test%5Blength%5D=3&test%5Bfoo%5D=bar", "Sub-object with a length property" );
 
+  params = {"test": [1,2,null]};
```
mzgol: Please use spaces between elements, as well as after `{` & `[`, before `}` & `]` etc. Yes, the surrounding style is wrong, we'll fix it but it doesn't seem worth to make it even worse. ;) For more, read https://contribute.jquery.org/style-guide/js/#spacing.

dmethvin:
This change looks good to me too. There are only a handful of places where we treat `undefined` and `null` equally and we do that via an `== null` comparison. In a quick scan of the jQuery code there seem to be some simple tests like `.data(null, ...)` that also pass the object test but those are already outside the domain of legal values and we don't guarantee behavior there. If you find cases that need work though, please open another issue.

This issue has probably existed so long because it's a rare case not defined by any standard. Serializing JavaScript objects is best done with JSON and not form encoding, this case never occurs with HTML forms. Might as well make it do something reasonable though! It seems unlikely the current behavior is being depended upon by anyone.

DanielNill:
I've signed the CLA with full name DanielNill and email daniellnill@gmail.com
and my git config shows:
```
[user]
	name = DanielNill
	email = daniellnill@gmail.com
```

Not sure why the bot is still marking this unsigned.

test/unit/serialize.js
```diff
@@ -77,6 +77,9 @@ test("jQuery.param()", function() {
 	params = {"test": {"length": 3, "foo": "bar"} };
 	equal( jQuery.param( params, false ), "test%5Blength%5D=3&test%5Bfoo%5D=bar", "Sub-object with a length property" );
 
+  params = { "test": [ 1, 2, null ] };
+  equal( jQuery.param( params, false ), "test%5B%5D=1&test%5B%5D=2&test%5B%5D=", "object with array property with null value" );
```
Mr21: Maybe, you should put a tab instead of spaces to match the other indentation.

mzgol:
You're missing a space between your first & last name. That's probably why the CLA chech still fsils.

DanielNill:
@mzgol a little confused, does the bot check for a space?  I assumed that as long as my name signed on the cla matched my name in my git config it would authorize it.  Since, at this point I'd have multiple commits that would need to be amended what is the best way to handle that?

mzgol:
> @mzgol a little confused, does the bot check for a space?

I think it does. It definitely doesn't accept everything as a name.

> Since, at this point I'd have multiple commits that would need to be amended what is the best way to handle that?

You can squash your commits first, then amend the remaining commit via:
```
git commit --amend --author="FirstName LastName <login@domain>"
```
and:
```
git push --force
```

jzaefferer:
The CLA check is fine now.

jzaefferer:
Sorry, that was a false positive. We don't consider "DanielNill" a valid name; that should probably be "Daniel Nill"?

mzgol:
Landed, thanks!

PR #2441, opened 2015-07-02
Core:Suggest remove unnecessary parameter

mixed:
Sorry. Again PR #2428
I removed unnecessary parameter for constructor of "end" method. It`s because jQuery constructor with no parameter is the same as jQuery constructor with null parameter.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -103,7 +103,7 @@ jQuery.fn = jQuery.prototype = {
 	},
 
 	end: function() {
-		return this.prevObject || this.constructor(null);
+		return this.prevObject || this.constructor();
 	},
 
 	// For internal use only.


mixed:
@mzgol 
I have a question. I singed CLA.
The name is "yongwoo jeon".
The email is "yongwoo.jeon@navercorp.com".
And Git config is it.
"
user.name=yongwoo jeon
user.email=yongwoo.jeon@navercorp.com
"
 
But. I still unsinged this PR.
Would i sign CLA?

mzgol:
@mixed It looks good to me (although I'm not sure if names shouldn't start from large letters?), no idea why it fails. @scottgonzalez?

jzaefferer:
The CLA check is fine now.

mzgol:
LGTM

mzgol:
Landed, thanks!

PR #2444, opened 2015-07-02
Use `matchesSelector` for serializeArray.

sirlancelot:
This is slight performance enhancement.

Old method will create a jQuery object with a single element and pass it to `.is()` which loops through a single item array and winds up at `matchesSelector` after some more needless logic.

This method just uses `matchesSelector` directly which bypasses the looping and other logic.

```
   raw     gz Sizes
255090  75474 dist/jquery.js
 84466  29360 dist/jquery.min.js

   raw     gz Compared to master @ 8e4aac8cb03ffb88373ea99629165d82ff5eccdd
   +34    +12 dist/jquery.js
   +17     +7 dist/jquery.min.js
```

### DIFF
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -2,7 +2,7 @@ define([
 	"./core",
 	"./manipulation/var/rcheckableType",
 	"./core/init",
-	"./traversing", // filter
+	"./traversing", // filter, selector
 	"./attributes/prop"
 ], function( jQuery, rcheckableType ) {
 
@@ -93,8 +93,8 @@ jQuery.fn.extend({
 		.filter(function() {
 			var type = this.type;
 
-			// Use .is( ":disabled" ) so that fieldset[disabled] works
-			return this.name && !jQuery( this ).is( ":disabled" ) &&
+			// Test by selector because the disabled property ignores ancestor fieldsets
+			return this.name && !jQuery.find.matchesSelector( this, ":disabled" ) &&
 				rsubmittable.test( this.nodeName ) && !rsubmitterTypes.test( type ) &&
 				( this.checked || !rcheckableType.test( type ) );
 		})


sirlancelot:
This can go in to v2.x and legacy v1.x. I couldn't easily find how to that. Should I make a separate pull request to the `1.11-stable` branch?

gibson042:
> Should I make a separate pull request to the `1.11-stable` branch?

A separate pull request would need to target `compat`, but is not necessary here—this can be cleanly cherry-picked.

src/serialize.js
```diff
@@ -94,7 +94,7 @@ jQuery.fn.extend({
 			var type = this.type;
 
 			// Use .is( ":disabled" ) so that fieldset[disabled] works
```
gibson042: This comment needs updating to something like "Test by selector because the disabled _property_ ignores ancestor fieldsets".

src/serialize.js
```diff
@@ -94,7 +94,7 @@ jQuery.fn.extend({
...
-			return this.name && !jQuery( this ).is( ":disabled" ) &&
+			return this.name && !jQuery.find.matchesSelector( this, ":disabled" ) &&
```
gibson042: Input value serialization is not a performance-sensitive area of the codebase... what is the size impact of this change?

sirlancelot: Thanks for the feedback. JSPerf appears to be down right now, I'll put together a test when it's back.

gibson042: I don't care about performance, just the size difference on gzipped jquery.min.js (visible as the last output from running `grunt`).

sirlancelot: raw     gz Sizes
    255072  75469 dist/jquery.js
     84466  29360 dist/jquery.min.js
    
       raw     gz Compared to master @ 8e4aac8cb03ffb88373ea99629165d82ff5eccdd
       +16     +8 dist/jquery.js
       +17     +7 dist/jquery.min.js

gibson042:
Thanks very much. I think size trumps performance here, so we'll pass on this pull request. But your CLA signature will remain valid, and we welcome any future contributions.

Issue #2056, opened 2015-02-03
Update jscs to 1.11


markelog:
1.11 should be released later this week with a lot of jquery related changes.

On the last meeting, it was decided to update it after beta release so it wouldn't clutter other PRs.

hzoo:
Well now that 1.12 is released you can update with autofix as well!

markelog:
That's right :-)

PR #2446, opened 2015-07-03
Capitalizes the first letter of first word in comment

msarnicki:


markelog:
It seems there is a lot of places like this, it would be beneficial for us to do it in one commit, so we wouldn't clutter commit history.

<a href="http://jscs.info/">JSCS</a> has a <a href="http://jscs.info/rule/requireCapitalizedComments.html">rule</a> for this and we have a ticket #2056 to update it after 3.0 release.

I added this rule to "jquery" preset https://github.com/jscs-dev/node-jscs/commit/e024327898833ebba0bb776b4cc208f760769bbe so when will be dealing with all code style related issues we will catch all those "bad" comments (and possible autofix it) in one move.

Nevertheless, if this pr wouldn't happen, jquery preset wouldn't be updated, so thank you!

Issue #2102, opened 2015-02-19
Add catch to promise objects


tracker1:
This relates to #1722 but a far simpler request...  given that the `.then` method matches the Promise spec (with the extra progress parameter), it would be nice that there be a compatible `.catch` method... (per @dmethvin) this could call and return `.then(null,fn)` as a passthrough.

dmethvin:
I can see us creating `.catch(fn)` as an alias for `.then(null, fn)`. That is different than `.done(fn)` though because the latter adds a new handler to the existing Deferred.

tracker1:
of course, this would need to be done with `['catch']` so as not to trigger reserved word treatment in older browsers.

dcherman:
Giant +1 on this.  Since these Promises aren't based on a prototype, it's much easier for this to be supported internally rather than having to do something like wrap the constructor and decorate the returned promise.

gibson042:
I updated the title for accuracy. Suggested implementation:

```js
"catch": function( fn ) {
	return promise.then( null, fn );
}
```

markelog:
Sounds like a blocker to me

PR #2451, opened 2015-07-06
Deferred: add .catch handler

timmywil:
Fixes gh-2102

### DIFF
--- a/test/unit/deferred.js
+++ b/test/unit/deferred.js
@@ -167,6 +167,42 @@ test( "jQuery.Deferred.then - filtering (fail)", function( assert ) {
 	});
 });
 
+test( "jQuery.Deferred.catch", function( assert ) {
+	assert.expect( 4 );
+
+	var value1, value2, value3,
+		defer = jQuery.Deferred(),
+		piped = defer.catch(function( a, b ) {
+			return a * b;
+		}),
+		done = jQuery.map( new Array( 3 ), function() { return assert.async(); } );
+
+	piped.done(function( result ) {
+		value3 = result;
+	});
+
+	defer.fail(function( a, b ) {
+		value1 = a;
+		value2 = b;
+	});
+
+	defer.reject( 2, 3 ).catch(function() {
+		assert.strictEqual( value1, 2, "first reject value ok" );
+		assert.strictEqual( value2, 3, "second reject value ok" );
+		assert.strictEqual( value3, 6, "result of filter ok" );
+		done.pop().call();
+	});
+
+	jQuery.Deferred().resolve().catch(function() {
+		assert.ok( false, "then should not be called on resolve" );
+	}).then( done.pop() );
+
+	jQuery.Deferred().reject().catch( jQuery.noop ).done(function( value ) {
+		assert.strictEqual( value, undefined, "then fail callback can return undefined/null" );
+		done.pop().call();
+	});
+});
+
 test( "[PIPE ONLY] jQuery.Deferred.pipe - filtering (fail)", function( assert ) {
 
 	assert.expect( 4 );
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -33,6 +33,9 @@ jQuery.extend({
 					deferred.done( arguments ).fail( arguments );
 					return this;
 				},
+				"catch": function( fn ) {
+					return promise.then( null, fn );
+				},
 				// Keep pipe for back-compat
 				pipe: function( /* fnDone, fnFail, fnProgress */ ) {
 					var fns = arguments;


test/unit/deferred.js
```diff
@@ -167,6 +167,42 @@ test( "jQuery.Deferred.then - filtering (fail)", function( assert ) {
...
+	});
+
+	defer.reject( 2, 3 ).catch(function() {
+		assert.strictEqual( value1, 2, "first reject value ok" );
+		assert.strictEqual( value2, 3, "second reject value ok" );
+		assert.strictEqual( value3, 6, "result of filter ok" );
+		done.pop().call();
+	});
+
+	jQuery.Deferred().resolve().catch(function() {
```
mzgol: Spaces before `function` everywhere and between `}` & `)`?

test/unit/deferred.js
```diff
@@ -167,6 +167,42 @@ test( "jQuery.Deferred.then - filtering (fail)", function( assert ) {
...
+		assert.ok( false, "then should not be called on resolve" );
+	}).then( done.pop() );
+
+	jQuery.Deferred().reject().catch(jQuery.noop ).done(function( value ) {
```
mzgol: Space before `jQuery.noo[`

test/unit/deferred.js
timmywil: Yea, can we just go ahead and change that rule?

test/unit/deferred.js
```diff
@@ -167,6 +167,42 @@ test( "jQuery.Deferred.then - filtering (fail)", function( assert ) {
...
+
+	defer.fail(function( a, b ) {
+		value1 = a;
+		value2 = b;
+	});
+
+	defer.reject( 2, 3 ).catch(function() {
+		assert.strictEqual( value1, 2, "first reject value ok" );
+		assert.strictEqual( value2, 3, "second reject value ok" );
+		assert.strictEqual( value3, 6, "result of filter ok" );
```
gibson042: I'd move this into a `piped` callback, to accommodate async behavior changes (we make no guarantee that `piped.done` will have fired by this point).

test/unit/deferred.js
gibson042: Missing space.

gibson042:
Other than the above, LGTM.

test/unit/deferred.js
mzgol: I don't like exceptions too much, but it's already been discussed... https://github.com/jquery/contribute.jquery.org/issues/21

test/unit/deferred.js
timmywil: These tests are an exact copy of the `.then( null, fn )` tests above with `.then` replaced with `.catch`.

gibson042: Allright then, I'm in favor of the consistency and we'll change both places if/when they break.

test/unit/deferred.js
timmywil: Yea, I'm familiar with that discussion, but it seems like the core team is generally disgruntled when we see `} )`.

timmywil: There's a jscs rule that allows for character exceptions when checking for spaces. We could use this rule: http://jscs.info/rule/disallowSpacesInsideParentheses.html.

markelog: > but it seems like the core team is generally disgruntled when we see } ).

I think no one likes this, but it is chosen code-style, i think if don't want to use it, we need bring back that discussion with everyone else.

dmethvin: How big a group is "everyone else"? It seems like most of the core team is bothered by the space in `} )`

markelog: In my understanding, it is all jquery people who choose to participate in the possible ticket of the contribute repo, so we would document that style and show it to contributors when needed.

From the beginning, i said lots of times that this style is no good, showed new styled code, propose compromises, no one seems to notice. And funny thing is, many of us knew that style is horrible, but choose it anyway. My hypotheses is, that some of us just didn't want to argue about such a thing or wanted to "avoid exceptions".

And now, it seems, we are considering to diverge from the official guide lines and create new  `core` set of rules. And going with new debate further, changing and creating new code styles, rewriting perfectly working code... again and again.

Code style discussions are always time consuming and really pointless, we can do something more radical and choose to deprecate jquery code style altogether and choose one of the most popular code style out there, like airbnb or google.

If that would be the case, we would never spend another second debating space there, tabs here, or newline before something, because one of the other subjective feeling. 

Following the other people subjective feeling, while doing real work.

jaubourg: My 2 cents.

I've been using the "space-everywhere-no-exception" approach in my projects for 2 years now. Found it painful at first but got used to it and I find it much easier to comply with and explain to coworkers.

Like @markelog said, it's all just conventions and endless debates about them. Still, for the sake of simplicity in applying the code style and clarity in explaining it to contributors, the no-exception approach has always been a big win in anything I worked on.

Oh, and hi!

timmywil: @jaubourg HI! So you come out of the woodwork when it's a style
discussion!? ;)

markelog: LOL :)

dmethvin: Well for a long time (at least while @rwaldron was keeping an eye on things) we were pretty close to idiomatic.js which I still think is pretty good for ES5 JS. I'm not opposed to using another style guide, it increases the chances that a contributor will have a compatible style. There are a bunch of things in the Airbnb guide that apply to ES6+ and it will probably be a while before we need those, but when we do it would be nice to adopt a lot of theirs.

markelog: I remember some recent tweets from @rwaldron "advocating" to use airbnb - https://twitter.com/search?q=idiomatic%20airbnb%20from%3Arwaldron&src=typd&lang=en instead of idiomatic.

> There are a bunch of things in the Airbnb guide that apply to ES6+ and it will probably be a while before we need those

Well, i remember there was some discussion to use ES6 `import` instead of `amd` and who knows, maybe jQuery written on ES6 would be actually smaller after transpiling and minifying :-), we didn't yet try that :0)

jaubourg: > @jaubourg HI! So you come out of the woodwork when it's a style discussion!? ;)

I'm useful like that :P

Issue #1786, opened 2014-10-21
Android 2.3 erroneous handling of script exceptions


mzgol:
Originally reported by m_gol at: http://bugs.jquery.com/ticket/15203
<p>
One test in Ajax and one in Manipulation that concern throwing errors in async scripts are failing in Android 2.3, see e.g.: <a class="ext-link" href="http://swarm.jquery.org/job/3366" rel="nofollow"><span class="icon">&nbsp;</span>http://swarm.jquery.org/job/3366</a>
</p>
<p>
It may be needed to blacklist those tests if there's no way to prevent that behavior (and probably there isn't one).
</p>
<p>
See also: <a class="ext-link" href="https://github.com/jquery/jquery/pull/1573" rel="nofollow"><span class="icon">&nbsp;</span>https://github.com/jquery/jquery/pull/1573</a>
</p><p>Issue reported for jQuery 2.1.1</p>

mzgol:
Per the yesterday's meeting, we're cantfixing this issue on both branches.

PR #2458, opened 2015-07-08
Ajax: Account for Android 2.3 not firing window.onerror on script errors

mzgol:
Android 2.3 doesn't fire the window.onerror handler, just accept the reality
there and skip the test.

Refs gh-1573
Refs gh-1786

TBH, this seems pretty bad. My initial attempt was to add the code block at the end of the test but then, once the following:
```js
jQuery.ajax({
	url: "data/badjson.js",
	dataType: "script",
	throws: true
});
```
is invoked, `jQuery.active` stays `>=1` forever. This may affect all subsequent AJAX requests, right? If I understand it correctly, one error in a script loaded via `jQuery.ajax` with `dataType: "script"` breaks all subsequent `jQuery.ajax` uses, no `ajaxStart` & `ajaxStop` events will fire.

cc @jaubourg

### DIFF
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -1450,6 +1450,15 @@ module( "ajax", {
 	});
 
 	asyncTest( "#11743 - jQuery.ajax() - script, throws exception", 1, function() {
+		// Support: Android 2.3 only
+		// Android 2.3 doesn't fire the window.onerror handler, just accept the reality there.
+		if ( /android 2\.3/i.test( navigator.userAgent ) ) {
+			ok( true, "Test skipped, Android 2.3 doesn't fire window.onerror for " +
+				"errors in dynamically included scripts" );
+			start();
+			return;
+		}
+
 		var onerror = window.onerror;
 		window.onerror = function() {
 			ok( true, "Exception thrown" );
--- a/test/unit/manipulation.js
+++ b/test/unit/manipulation.js
@@ -2241,6 +2241,17 @@ test( "Ensure oldIE creates a new set on appendTo (#8894)", function() {
 });
 
 asyncTest( "html() - script exceptions bubble (#11743)", 2, function() {
+	// Support: Android 2.3 only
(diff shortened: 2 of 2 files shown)


jaubourg:
The ajax code (and thus the maintaining of the active count) is exception-proof as far as conversions are concerned.

Every conversion (including script to execution) is [`try`/`catch` protected](https://github.com/jquery/jquery/blob/master/src/ajax.js#L263). Even if we didn't have this, given we use a script tag in globalEval (used for same-domain scripts), the error is not actually thrown from within the ajax code anyway.

(as a side note, this `throws` option is a remnant from the past and is not used in the `ajaxTest` code anymore due to said change in globalEval, so it can be safely removed)

For cross-domain scripts, we only had issues with browsers that did not fire `error` events on script tags, but I'm pretty sure it's not a thing anymore with our current support grid.

mzgol:
@jaubourg I don't fully understand. This test indicates that Android 2.3 doesn't fire `window.onerror` for script errors in dynamically inserted scripts (even non-cross-domain) i.e. what `jQuery.ajax({..., dataType: 'script'})` does. When I just ignore that `window.onerror` doesn't fire in Android 2.3 in such a case, I'm left with `jQuery.ajax === 1` which then breaks subsequent tests. What's your take on this?

jaubourg:
So `ajax`doesn't throw (thanks to the `try`/`catch` I mentioned) and doesn't notify an error that can be inspected with a fail handler.

I'm beginning to suspect this wonderful browser simply halts the execution in case of an error but you'd need to test `globalEval` with a script that throws an exception in Android to be sure. Something we have no test for in [unit/core.js](https://github.com/jquery/jquery/blob/master/test/unit/core.js#L155) btw, unless I'm missing something.

mzgol:
Per yesterday's meeting, we're cantfixing #1786 on both branches so this PR is good to go (minus the `Fixes` reference that needs to be changed to `Refs`).

Issue #2323, opened 2015-05-18
Delete generated jsonp-callbacks


markelog:
Now we assign it to undefined when `undefined` when function is executed, we did this because we can't properly delete properties from the window objects on oldIE.

This issue still exist for supported IE8, but i think we should *discuss* using `removeProp` method in both branches instead.

secrettriangle:
While `delete`ing the function makes more semantic sense, V8 will deoptimize any objects that have had properties `delete`d. Just something to keep in mind.

markelog:
This is pure guessing, first of all, `window` object should be already deoptimized, since there is a lot of bounded properties, second, `window` is special host object so there might be another type of optimizations in play.

I wouldn't relay on those assumptions, until they proved by the performance benchmarking.

markelog:
@timmywil, @dmethvin, @jaubourg need your opinions here.

Also, would like to know if our perception of the https://github.com/jquery/jquery/issues/2317 didn't change.

dmethvin:
I agree that `window` is pretty magic and global so I doubt optimization comes into play here.

Would it make sense to create a single object on `window` at initialization and put all our JSONP callbacks there? That would work on IE8 too since we'd be `delete`ing the properties out of that regular object rather than on the global.

markelog:
Which would mean we would expose three globals and as soon as IE8 dead we could eliminate it, cool.
Although, it seems we would need to do it only because of the one browser.

timmywil:
I like the idea of using `removeProp`, which means we can delete in every browser except IE8 and IE8 receives no change.

PR #2464, opened 2015-07-10
Ajax: Remove jsonp callbacks through "jQuery#removeProp" method

markelog:
Fixes gh-2323

@timmywil that is that one i was talking about, but i was intended to go even further.

Registering callbacks not on window, but on special jQuery object, like `jQuery.jsonpCallbacks`, since according to the specification - http://json-p.org/ dots are allowed to be in get parameter so server scripts should allow them, but even their logic violate the spec, user could apply their own jsonp callback name generation. 

But it seems more radical and in need in discussion, thoughts?

### DIFF
--- a/src/ajax/jsonp.js
+++ b/src/ajax/jsonp.js
@@ -64,8 +64,14 @@ jQuery.ajaxPrefilter( "json jsonp", function( s, originalSettings, jqXHR ) {
 
 		// Clean-up function (fires after converters)
 		jqXHR.always(function() {
-			// Restore preexisting value
-			window[ callbackName ] = overwritten;
+			// If previous value didn't exist - remove it
+			if ( overwritten === undefined ) {
+				jQuery( window ).removeProp( callbackName );
+
+			// Otherwise restore preexisting value
+			} else {
+				window[ callbackName ] = overwritten;
+			}
 
 			// Save back as free
 			if ( s[ callbackName ] ) {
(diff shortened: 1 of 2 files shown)


markelog:
Because this change is a "trivial" one, decided upon and because alpha is upon us, i would like to merge this in next couple hours.

Please ping me up if someone disagrees.

timmywil:
LGTM

timmywil:
I'm not sure about supporting the use of dots in the name. Let's open a separate issue after this is merged and discuss it as a new feature. We need to gauge it's complexity and appeal.

markelog:
> I'm not sure about supporting the use of dots in the name.

Yeah, that is why i didn't proceed. Anyway, a bit busy right now, can merge this in 8 hours or so

src/ajax/jsonp.js
```diff
@@ -64,8 +64,14 @@ jQuery.ajaxPrefilter( "json jsonp", function( s, originalSettings, jqXHR ) {
...
 		jqXHR.always(function() {
-			// Restore preexisting value
-			window[ callbackName ] = overwritten;
+			// If previous value didn't exist - remove it
+			if ( overwritten === undefined ) {
+				jQuery( window ).removeProp( callbackName );
+
+			// Otherwise restore preexisting value
+			} else {
+				window[ callbackName ] = overwritten;
```
gibson042: This might be smaller if we dropped the `else`, always setting to `overwritten` but _then_ calling `removeProp` for the `undefined` case.

markelog: It doesn't

PR #2465, opened 2015-07-11
Attributes: .removeClass() -> .attr( "class", "" )

Mr21:
Hi :)

After asking a question about a specific part of `.removeClass` on IRC and after @Krinkle told me that the call of `.removeClass()` without any arguments cleaning all the `"class"` attribute. I watch the code and I've found that this particularity made the code different compared to `.addClass`.
And if you call `this.attr( "class", "" )` directly in a `if` we can save **17** bytes after gzipped.

I add another change in another commit to save other **2** bytes on `.hasClass`.

### DIFF
--- a/src/attributes/classes.js
+++ b/src/attributes/classes.js
@@ -14,9 +14,7 @@ function getClass( elem ) {
 jQuery.fn.extend( {
 	addClass: function( value ) {
 		var classes, elem, cur, curValue, clazz, j, finalValue,
-			proceed = typeof value === "string" && value,
-			i = 0,
-			len = this.length;
+			i = 0;
 
 		if ( jQuery.isFunction( value ) ) {
 			return this.each( function( j ) {
@@ -24,13 +22,10 @@ jQuery.fn.extend( {
 			} );
 		}
 
-		if ( proceed ) {
+		if ( typeof value === "string" && value ) {
+			classes = value.match( rnotwhite ) || [];
 
-			// The disjunction here is for better compressibility (see removeClass)
-			classes = ( value || "" ).match( rnotwhite ) || [];
-
-			for ( ; i < len; i++ ) {
-				elem = this[ i ];
+			while ( ( elem = this[ i++ ] ) ) {
 				curValue = getClass( elem );
 				cur = elem.nodeType === 1 &&
 					( " " + curValue + " " ).replace( rclass, " " );
@@ -57,20 +52,22 @@ jQuery.fn.extend( {
 
 	removeClass: function( value ) {
 		var classes, elem, cur, curValue, clazz, j, finalValue,
-			proceed = arguments.length === 0 || typeof value === "string" && value,
-			i = 0,
-			len = this.length;
+			i = 0;
 
 		if ( jQuery.isFunction( value ) ) {
 			return this.each( function( j ) {
 				jQuery( this ).removeClass( value.call( this, j, getClass( this ) ) );
 			} );
 		}
-		if ( proceed ) {
-			classes = ( value || "" ).match( rnotwhite ) || [];
 
-			for ( ; i < len; i++ ) {
-				elem = this[ i ];
+		if ( !arguments.length ) {
+			return this.attr( "class", "" );
+		}
+
+		if ( typeof value === "string" && value ) {
+			classes = value.match( rnotwhite ) || [];
+
+			while ( ( elem = this[ i++ ] ) ) {
 				curValue = getClass( elem );
 
 				// This expression is here for better compressibility (see addClass)
@@ -88,7 +85,7 @@ jQuery.fn.extend( {
 					}
 
 					// Only assign if different to avoid unneeded rendering.
-					finalValue = value ? jQuery.trim( cur ) : "";
+					finalValue = jQuery.trim( cur );
 					if ( curValue !== finalValue ) {
 						elem.setAttribute( "class", finalValue );
 					}
@@ -160,12 +157,13 @@ jQuery.fn.extend( {
 	},
 
 	hasClass: function( selector ) {
-		var className = " " + selector + " ",
-			i = 0,
-			l = this.length;
-		for ( ; i < l; i++ ) {
-			if ( this[ i ].nodeType === 1 &&
-				( " " + getClass( this[ i ] ) + " " ).replace( rclass, " " )
+		var className, elem,
+			i = 0;
+
+		className = " " + selector + " ";
+		while ( ( elem = this[ i++ ] ) ) {
+			if ( elem.nodeType === 1 &&
+				( " " + getClass( elem ) + " " ).replace( rclass, " " )
 					.indexOf( className ) > -1
 			) {
 				return true;


timmywil:
I like it.

Mr21:
After reviewing the code, I commit again to save 7 bytes on `.addClass` / `.removeClass` by removing the `len` variables.

src/attributes/classes.js
```diff
@@ -86,7 +84,7 @@ jQuery.fn.extend({
 					}
 
 					// Only assign if different to avoid unneeded rendering.
-					finalValue = value ? jQuery.trim( cur ) : "";
```
markelog: I wonder why it was done in the first place

src/attributes/classes.js
```diff
@@ -158,12 +156,13 @@ jQuery.fn.extend({
...
-		var className = " " + selector + " ",
-			i = 0,
-			l = this.length;
-		for ( ; i < l; i++ ) {
-			if ( this[i].nodeType === 1 &&
-				( " " + getClass( this[i] ) + " " ).replace( rclass, " " )
+		var className, elem,
+			i = 0;
+
+		className = " " + selector + " ";
```
markelog: Since we don't require to use single var per scope, we could do here `var className = " " + selector + " ";`, i bet you wouldn't became smaller, but i guess more readable

src/attributes/classes.js
```diff
@@ -158,12 +156,13 @@ jQuery.fn.extend({
...
+		while ( ( elem = this[ i++ ] ) ) {
+			if ( elem.nodeType === 1 &&
+				( " " + getClass( elem ) + " " ).replace( rclass, " " )
```
markelog: How much it would cost if we would introduce a variable for this line? Four lines of `if` is kinda messy

src/attributes/classes.js
```diff
@@ -78,7 +76,7 @@ jQuery.fn.extend({
 
 				if ( cur ) {
 					j = 0;
-					while ( (clazz = classes[j++]) ) {
+					while ( ( clazz = classes[ j++ ] ) ) {
```
markelog: I wonder if need those extra parentheses, is that really more readable even if it is the only expression?

@timmywil ?

src/attributes/classes.js
Mr21: Here, it was for removing all the classes when we call `.removeClass()` without any argument, but I write a special `if` (with the `.attr`) only for this case.

src/attributes/classes.js
Mr21: If I do that `jscs` (after doing `$> grunt`) tell me:
```
Var declarations should be joined at src/attributes/classes.js :
   160 |   i = 0;
   161 |
   162 |  var className = " " + selector + " ";
----------^
   163 |  while ( ( elem = this[ i++ ] ) ) {
   164 |   if ( elem.nodeType === 1 &&
>> 1 code style errors found!
Warning: Task "jscs:src" failed. Use --force to continue.
```

But if I do this:
``` javascript
var elem,
	className = " " + selector + " ",
	i = 0;
```
it cost +4 bytes (total -22)

and if I do this (more simple):
``` javascript
var elem,
	i = 0;
selector = " " + selector + " ";
```
it cost +5 bytes (total -21)

markelog: > If I do that jscs (after doing $> grunt) tell me:

Old config, we will update jscs soon enough. Ignore that for a minute. So what is the size diff without `className` initialization?

src/attributes/classes.js
Mr21: This version: add +5 (total -21) (clearly more readable)
``` javascript
hasClass: function( selector ) {
	var cur, elem,
		i = 0;

	selector = " " + selector + " ";
	while ( ( elem = this[ i++ ] ) ) {

		cur = elem.nodeType === 1 &&
			( " " + getClass( elem ) + " " ).replace( rclass, " " );

		if ( cur && cur.indexOf( selector ) > -1 ) {
			return true;
		}
	}

	return false;
}
```

src/attributes/classes.js
Mr21: I search about this and it's in all the jquery.js every where, but it doesnt appear in the minified version.

dmethvin: The idea of the extra parens is to give the reader another visual clue that the single `=` isn't a mistaken `==` or `===`. [JSHint complains](http://jshint.com/docs/options/#boss) if you don't add them. Since the extra parens are not required, the minifier removes them.

src/attributes/classes.js
```diff
@@ -24,19 +22,17 @@ jQuery.fn.extend({
...
 
-		if ( proceed ) {
-			// The disjunction here is for better compressibility (see removeClass)
-			classes = ( value || "" ).match( rnotwhite ) || [];
+		if ( typeof value === "string" && value ) {
+			classes = value.match( rnotwhite ) || [];
 
-			for ( ; i < len; i++ ) {
-				elem = this[ i ];
+			while ( ( elem = this[ i++ ] ) ) {
```
gibson042: This is technically a behavior change, albeit one I'm in favor of. Still, we should probably document that collections including falsy elements are invalid context for (all?) jQuery methods. Would anyone in @jquery/core oppose?

gibson042: Of course, scrolling down just two lines reveals that **this** change actually _doesn't_ affect behavior. Still, there are analogous places in the codebase where that doesn't hold, and my question stands.

src/attributes/classes.js
dmethvin: Probably to save the function call? Probably not worth the extra code. The trimming was added in e76ba32cebc26108e15b2e0e03c5152624c1bad9 to avoid leading/trailing space.

src/attributes/classes.js
```diff
@@ -56,20 +52,22 @@ jQuery.fn.extend({
...
 				jQuery( this ).removeClass( value.call( this, j, getClass( this ) ) );
 			});
 		}
-		if ( proceed ) {
-			classes = ( value || "" ).match( rnotwhite ) || [];
 
-			for ( ; i < len; i++ ) {
-				elem = this[ i ];
+		if ( !arguments.length ) {
+			return this.attr( "class", "" );
```
dmethvin: Good, we definitely want `.attr()` here since `.prop()` won't work on SVG elements. We should ensure we have a test with SVG if we don't already.

src/attributes/classes.js
gibson042: What @dmethvin said. It _was_ more readable (in my opinion), but then we started adding padding spaces to expression parentheses. I'm hoping jscs-dev/node-jscs/pull/1254 will gain some traction so we can change back.

src/attributes/classes.js
dmethvin: I don't think falsy values cause issues, there are a lot of places where Sizzle iterates node lists and uses this construct. I like the `while` better than the headless-`for` as a reader, so there's that. :smile:

gibson042: I like the `while` better too, but there are places where changing to it from `for` will introduce failures on collections including falsy members ([`.find`](https://github.com/Mr21/jquery/blob/classes-better-compressibility/src/traversing/findFilter.js#L55-L76), [`.closest`](https://github.com/Mr21/jquery/blob/classes-better-compressibility/src/traversing.js#L72), etc.). I'm suggesting that we do so anyway (as is being done here) and document that such collections are invalid.

src/attributes/classes.js
Mr21: No, no it's not for save a function call, it was a specific behaviour, if the `value` of the user is `undefined` we remove set an empty class attribute. It is really more simple to read the specific case of `.removeClass()` now.

src/attributes/classes.js
markelog: Yeah, that is all true, i was asking if we should still do that? Since this is a single expression and if something goes wrong in the loop, we definitely will notice that.

src/attributes/classes.js
markelog: Judgment call i suppose what is worth more size or readability here? @timmywil?

src/attributes/classes.js
markelog: So we would consistently use `while` everywhere? The reason to do this is to make code smaller?

>  I'm suggesting that we do so anyway (as is being done here) and document that such collections are invalid.

I'm a bit scare of that the less user complains the better, but that is not a strong position.

src/attributes/classes.js
Krinkle: The combined boolean and string assignment and repeated check of `cur` is a bit odd though. Perhaps like this:

```js
hasClass: function( selector ) {
    var cur, elem,
        i = 0;

    selector = " " + selector + " ";
    while ( ( elem = this[ i++ ] ) ) {
        if ( elem.nodeType === 1 ) {
            cur = ( " " + getClass( elem ) + " " ).replace( rclass, " " );
            if ( cur.indexOf( selector ) > -1 ) {
                return true;
            }
        }
    }

    return false;
}
```

It seems UglifyJS optimises this quite well. It produces a `for` statement for the setup, and a single `if` statement as its body.

src/attributes/classes.js
gibson042: Understood. I am in favor of continuing to wrap statement-head assignment expressions with parentheses.

dmethvin: Yeah me too.

markelog: Although it seems we in agreement but just in case - @timmywil, @mzgol, @jaubourg?

timmywil: That's fine.

mzgol:
@Mr21 Could you rebase it over latest `master` (we have a new JSCS in there and jsdom shouldn't cause problems anymore), invoke `npm install` (the `grunt` should work without problems you had afterwards) and then try what @markelog suggested in his comments?

markelog:
@Mr21 friendly ping

Mr21:
Oh sorry, I was extremely focused on something else this time.
I'll do it ASAP!

Issue #2466, opened 2015-07-13
Can't chain multiple find calls using Sizzle-less build


mlewand:
Looks like `uniqueSort` function is not resolved correctly when jQuery is builded without Sizzle dependency.

When I try to call chained `find` methods I'm getting following exception: `Uncaught TypeError: n.uniqueSort is not a function`

## TC

* Build jQuery with `grunt custom:-sizzle`
* Try to chain multiple `find` calls, e.g.

	```html
	<!doctype html>
	<html lang="en">
		<head>
			<title>title</title>
		</head>
		<body>
		<div>
		  foo
		</div>
			<div>
		  <span>bar</span>
		</div>
	
		<script src="dist/jquery.min.js"></script>
		<script>
		  console.log( $('div').find('span').length );
		</script>
		</body>
	</html>
	```

* An exception is thrown: `Uncaught TypeError: n.uniqueSort is not a function`.

It happens also with other calls, e.g. `$( 'div' ).has( 'span' )`.

Experienced with 84ccf26 and a2ae215.

gibson042:
Confirmed; caused by e1090c3d2b2a988a5b41f1f1ed9f8d6dcae02200. We need selector-sizzle to define both `uniqueSort` and `unique`. @mlewand, would you like to open a pull request?

gibson042:
Whoops, I see that you have without referencing this issue. I'll move discussion to gh-2467.

mlewand:
Yes, you're right I've already made a PR for that. I believe it was already referenced as I've referenced this issue in PR description.

> mlewand referenced this issue an hour ago

PR #2467, opened 2015-07-13
Added $.uniqueSort for selector-native.

mlewand:
I can see that some time ago in e1090c3d2b2a988a5b41f1f1ed9f8d6dcae02200 there was a change where you started using jQuery.uniqueSort rather than jQuery.unique.

This change was not reflected in `selector-native.js` file, thus exception reported in #2466.

### DIFF
--- a/src/selector-native.js
+++ b/src/selector-native.js
@@ -67,6 +67,28 @@ var hasDuplicate,
 
 		// Not directly comparable, sort on existence of method
 		return a.compareDocumentPosition ? -1 : 1;
+	},
+	uniqueSort = function( results ) {
+		var elem,
+			duplicates = [],
+			i = 0,
+			j = 0;
+
+		hasDuplicate = false;
+		results.sort( sortOrder );
+
+		if ( hasDuplicate ) {
+			while ( (elem = results[i++]) ) {
+				if ( elem === results[ i ] ) {
+					j = duplicates.push( i );
+				}
+			}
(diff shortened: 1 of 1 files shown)


mlewand:
I've sent CLA after sending this PR. Will bot update the label correctly?

dmethvin:
As long as you signed with the [same name and email as the commit](http://contribute.jquery.org/CLA/status/?repo=jquery&sha=20409ac8f9ef248993b0da852c7a5cac690e667d) it should be fine. If new commits are pushed to this PR it will do the check again, so hold on and let's see if any are needed.

Great catch by the way!

src/selector-native.js
```diff
@@ -121,6 +121,9 @@ jQuery.extend({
 
 		return results;
 	},
+	unique: function( results ) {
+		return this.uniqueSort( results );
```
gibson042: Since the "method" is actually context-free, I'd rather not introduce a reliance upon `this`. `jQuery.unique = jQuery.uniqueSort` or `var uniqueSort = function …; jQuery.extend({ …, unique: uniqueSort, uniqueSort: uniqueSort, … })` would be better. Other than that, though, LGTM!

mlewand: Sounds good, fix applied with daec351.

mlewand:
Is there anything more I can do for this PR?

gibson042:
Nope; we're all set.

PR #2477, opened 2015-07-16
Offset: Do not attempt to call getClientRects on window object

booleanbetrayal:
Noticed that a few libraries are attempting to call `offset` on `$(window)` (for better or worse). This method does not exist in that context and therefore produces a no method error. Here's a simple fix for that expected case.

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -86,7 +86,7 @@ jQuery.fn.extend({
 		var docElem, win, rect, doc,
 			elem = this[ 0 ];
 
-		if ( !elem ) {
+		if ( !elem || elem === window ) {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


mzgol:
Partial duplicate of #2310. This change is intended, `.offset()` is not supported on `$(window)`.

dmethvin:
@booleanbetrayal Do you have a list of the libraries? It would be worth filing a bug with them when you see it.

booleanbetrayal:
@dmethvin - I've already submitted a PR against a popular scrollspy library - waypoints: imakewebthings/waypoints#430

I'll try to reference #2310 as I encounter more.

felicienfrancois:
@mzgol 
> This change is intended, .offset() is not supported on $(window)
So maybe it should thow `.offset() is not supported on $(window)` or `.offset() is no longer supported on $(window)` instead of the browser error `elem.getClientRects is undefined`

mzgol:
@felicienfrancois We don't throw for invalid input, jQuery would have to be a lot larger than it is if we wanted to handle all such situations. If you pass invalid input, you're basically on your own, anything may happen.

PR #2480, opened 2015-07-19
Data: remove user data in cleanData

jbedard:
Shouldn't user data be removed like [v1](https://github.com/jquery/jquery/blob/22120cea66ad88f83b96535416f5baab605fe530/src/manipulation.js#L442) and [v2](https://github.com/jquery/jquery/blob/d837f119c3729565103005d5d7fa89e1dd8110cb/src/manipulation.js#L304)? Or was that change intentional?

Note that this needs to change if/when #2479 is merged.

### DIFF
--- a/test/unit/manipulation.js
+++ b/test/unit/manipulation.js
@@ -2064,7 +2064,7 @@ test( "jQuery.cleanData", function() {
 });
 
 test( "jQuery.cleanData eliminates all private data (gh-2127)", function() {
-	expect( 2 );
+	expect( 3 );
 
 	var div = jQuery( "<div/>" ).appendTo( "#qunit-fixture" );
 
@@ -2074,10 +2074,32 @@ test( "jQuery.cleanData eliminates all private data (gh-2127)", function() {
 
 	div.remove();
 
+	ok( !jQuery.hasData( div[ 0 ] ), "Removed element hasData should return false" );
+
 	ok( jQuery.isEmptyObject( jQuery._data( div[ 0 ] ) ),
 		"Private data is empty after node is removed" );
 
 	div.remove();
+});
+
+test( "jQuery.cleanData eliminates all public data", function() {
+	expect( 2 );
+
+	var key,
+		div = jQuery( "<div/>" );
+	div.data( "some", "data" );
+	ok( !jQuery.isEmptyObject( jQuery.data( div[ 0 ] ) ),  "Ensure some public data exists" );
+
+	div.remove();
+
+	ok( !jQuery.hasData( div[ 0 ] ), "Removed element hasData should return false" );
+
+	// Make sure the expando is gone
+	for ( key in div[ 0 ] ) {
+		if ( /^jQuery/.test( key ) ) {
+			ok( false, "Expando was not removed when there was no more data" );
+		}
+	}
 });
 
 test( "domManip plain-text caching (trac-6779)", function() {
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -270,19 +270,24 @@ jQuery.extend({
 			i = 0;
 
 		for ( ; (elem = elems[ i ]) !== undefined; i++ ) {
-			if ( jQuery.acceptData( elem ) && (data = elem[ dataPriv.expando ])) {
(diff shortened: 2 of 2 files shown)


mzgol:
`jQuery.cleanData` is used for removed elements to clean up after them. Since in jQuery 3 we attach the data directly to elements, it doesn't seem necessary to delete user data associated with them, does it? Once you lose reference to the node, the data is garbage collected as well.

This is an observable difference, though, so we might want to document it anyway.

jbedard:
Was there a reason to remove private data but not user data?

It isn't necessary for any memory/GC issues (unless someone is keeping removed elements around) but I think if `$.cleanData([e])` is called then `$.hasData(e)` should be false. Currently it is also different in master vs compat/v2.

(I just updated the tests to also do a `hasData` check)

dmethvin:
> Was there a reason to remove private data but not user data?

I can't think of one. It does seem odd that we're doing one but not the other.

mzgol:
>> Was there a reason to remove private data but not user data?
>
> I can't think of one. It does seem odd that we're doing one but not the other.

One potential reason is that user data never leaks and event handlers may register event data on other elements (like in the focusin case) so it has to be cleaned to not leak.

That said, we might want to clean user data as well for consistency with older versions. I don't have a strong opinion here.

dmethvin:
We definitely need to call the teardowns for special events, which we are doing, but I think that should still work even if we don't `delete` the data afterwards.

mzgol:
> that should still work even if we don't delete the data afterwards.

That's true but if you then re-attached the element to the DOM, effects could be... interesting. ;)

dmethvin:
We're always in the process of letting elements go when they get to this point, but yeah if you pull them out of the trash who knows if you can still eat them or not? :smile_cat:

jbedard:
Originally d702b7637a61e1973e08c27b8d8de2ed24a543e2 stopped clearing both user and private data and only cleared events. Not clearing private data caused #2127 so clearing private data was added back but user data was not. What if something/someone depends on user data similar to the tests depending on private data?

IMO just the consistency with 1.x/2.x/compat and avoiding doc changes is worth the extra code though.

mzgol:
Thanks @jbedard for finding all that info with links! The initial idea was that `jQuery.cleanData` is not really needed once we attach data directly to elements. Unfortunately, it turned out to be false for private data and I feel convinced by your argumentation that leaving user data there might have similar issues so I'd be fine for clearing both as it was done previously.

cc @rwaldron

mzgol:
@jbedard Could you create an issue and then reference it in the commit message as "Fixes gh-ISSUE_NUMBER"? We try to separate issues & PRs.

Otherwise, this LGTM.

timmywil:
:+1:

test/unit/manipulation.js
```diff
@@ -2074,12 +2074,34 @@ test( "jQuery.cleanData eliminates all private data (gh-2127)", function() {
 
 	div.remove();
 
+	ok( !jQuery.hasData( div[ 0 ] ), "Removed element hasData should reutrn false" );
```
timmywil: "return" :)

test/unit/manipulation.js
```diff
@@ -2074,12 +2074,34 @@ test( "jQuery.cleanData eliminates all private data (gh-2127)", function() {
...
+	expect( 2 );
+
+	var key,
+		div = jQuery( "<div/>" );
+	div.data( "some", "data" );
+	ok( !jQuery.isEmptyObject( jQuery.data( div[ 0 ] ) ),  "Ensure some public data exists" );
+
+	div.remove();
+
+	ok( !jQuery.hasData( div[ 0 ] ), "Removed element hasData should reutrn false" );
```
timmywil: "return"

jbedard:
Created #2503 and updated the commit message, and fixed the typos.

mzgol:
@jbedard Landed, thanks! I backported the tests to `compat` as well but with a slight change as on `compat` the value for the expando property is set to `undefined`, not deleted. This discrepancy should be removed when #2479 lands.

Issue #2478, opened 2015-07-18
catch property needs to be quoted in Android 2.3


mzgol:
Apparently Android 2.3 doesn't like a `catch` property and so it fails deferred tests completely since 84ccf2606c6b97d5875774bf774f9f2aae950ae7. We need to quote it as on the `compat` branch.

Do we have any way to ensure we don't change it back without turning on the JSHint `es3` option? :/ Or maybe via JSCS? cc @rwaldron @markelog.

mzgol:
Or maybe we should just turn on the `es3` option, preferably with a comment that Android 2.3 needs it.

markelog:
Probably easier to left a comment in the source code? As inline rule or just as a warning?

mzgol:
I'm concerned that since Android 2.3 chokes on it, there may be other
reserved words that don't work as property names. It would be safer to
switch es3 on (with a support comment that it's just for Android 2.3) and
be sure everything is quoted.

markelog:
>  there may be other reserved words that don't work as property names

Is there? Every reserved word should trigger such error, if that the case, then we should enable es3.

mzgol:
> Every reserved word should trigger such error

I didn't check all but at least `catch`, `class` & `else` trigger this error. They all work on Android 4.0.

mzgol:
PR: #2481.

PR #2481, opened 2015-07-19
Build: Acknowledge Android 2.3 is not ES5-compatible

mzgol:
Android 2.3 chokes on unquoted reserved words being used as property names
which was making Deferred tests not run.

Acknowledge the sad fact that Android 2.3 is not ES5-compliant browser
and enable the "es3" option in JSHint config.

Fixes gh-2478

### DIFF
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -12,8 +12,11 @@
 
 	"sub": true,
 
+	"es3": true,
+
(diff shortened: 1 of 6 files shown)


mzgol:
Landed at 1c59b308d201d6dd0f0aed2ad0256d01b9f68047.

PR #2487, opened 2015-07-21
Fix various spelling mistakes

bperel:


arthurvr:
You shouldn't create a new PR to update it, you can just update the existing one ;)

The changes here look good, but you'll need to [sign the CLA](http://contribute.jquery.org/CLA/status/?repo=jquery&sha=0947943116b92ce179ceef45fd3bed8b8adeee2f).

bperel:
I have signed the CLA but although my email address and name match my Git config the "CLA Verification Results" page says "The following authors don't have appropriate CLA signatures", followed with my info. Is there a delay after someone  sends the CLA form ?

mzgol:
The signature looks OK to me. @scottgonzalez, any idea why the check fails?

scottgonzalez:
> The signature looks OK to me. @scottgonzalez, any idea why the check fails?

The first signature for the CLA is "Bruno Pérel" which doesn't match because the accent is not present in the commit name.

bperel:
Thanks @scottgonzalez , should be good now :)

arthurvr:
Thanks @bperel :) Looks good to me.

mzgol:
@bperel Thx, landed.

PR #2490, opened 2015-07-22
Ajax: remove extra "?" in the URL

### LINKED ISSUE #1791: $.getJSON treats request as JSONP when URL contains ?? (two question marks)
Originally reported by Stijn at: http://bugs.jquery.com/ticket/15216
<p>
Similar to <a href="http://bugs.jquery.com/ticket/8417">http://bugs.jquery.com/ticket/8417</a> and its duplicates, but they're all about the data parameter, not the URL parameter.
</p>
<p>
Happens with current stable and with edge. Fiddle: <a class="ext-link" href="http://jsfiddle.net/r5zcw6wx/" rel="nofollow"><span class="icon">&nbsp;</span>http://jsfiddle.net/r5zcw6wx/</a>
</p>
<p>
Code:
</p>
<pre class="wiki">$.getJSON("http://date.jsontest.com/")
$.getJSON("http://date.jsontest.com/?foo")
$.getJSON("http://date.jsontest.com/??foo")
</pre><p>
Requests:
</p>
<pre class="wiki">GET http://date.jsontest.com/
GET http://date.jsontest.com/?foo
GET http://date.jsontest.com/jQuery2129957077760988448071399739372034175163514043816_1408610547781foo?_=1408610547782
</pre><p>
Of course we shouldn't be making requests with two question marks (naive string concat on our side), and this could be a dupe of the mentioned ticket, but I figured this was worth reporting anyway.
</p><p>Issue reported for jQuery 2.1.1</p>

Mr21:
Hi :)

Fixes #1791

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -13,6 +13,7 @@ define([
 
 var
 	rhash = /#.*$/,
+	rextraQuestions = /\?+/,
 	rts = /([?&])_=[^&]*/,
 	rheaders = /^(.*?):[ \t]*([^\r\n]*)$/mg,
 	// #7653, #8125, #8152: local protocol detection
@@ -489,11 +490,17 @@ jQuery.extend({
 		// Attach deferreds
 		deferred.promise( jqXHR );
 
-		// Remove hash character (#7531: and string promotion)
-		// Add protocol if not provided (prefilters might expect it)
 		// Handle falsy url in the settings object (#10093: consistency with old signature)
 		// We also use the url parameter if available
-		s.url = ( ( url || s.url || location.href ) + "" ).replace( rhash, "" )
+		s.url = ( ( url || s.url || location.href ) + "" )
+
+			// Remove hash character (#7531: and string promotion)
+			.replace( rhash, "" )
+
+			// Remove the extra "?" (gh-1791)
+			.replace( rextraQuestions, "?" )
+
+			// Add protocol if not provided (prefilters might expect it)
 			.replace( rprotocol, location.protocol + "//" );
 
 		// Alias method option to type as per ticket #12004
@@ -787,6 +794,7 @@ jQuery.extend({
 
 jQuery.each( [ "get", "post" ], function( i, method ) {
 	jQuery[ method ] = function( url, data, callback, type ) {
+
 		// Shift arguments if data argument was omitted
 		if ( jQuery.isFunction( data ) ) {
 			type = type || callback;
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -271,6 +271,33 @@ module( "ajax", {
 		error: true
 	});
 
+	ajaxTest( "jQuery.ajax() - remove extra question marks in the URL", 3, [
+		{
+			url: "data/??foo",
+			beforeSend: function( xhr, settings ) {
+				equal( settings.url, "data/?foo", "Make sure there is no extra '?' in the URL." );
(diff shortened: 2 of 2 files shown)


mzgol:
@Mr21 What's the size difference?

src/ajax.js
```diff
@@ -13,6 +13,7 @@ define([
 
 var
 	rhash = /#.*$/,
+	rextraQuestions = /\?+/,
```
mzgol: Perhaps `/\?\?+/` would be faster as in this PR you're currently changing single question marks to themselves.

Mr21: Yeah right, but it's not a problem if this url change?
`"/file.html?q=??"`

mzgol: Ah, I see, you're changing the first question mark only and that's why it needs to be one. Seems fine as it is then.

jaubourg: Actually it isn't. The `??` replacer was introduced so that you c

Mr21: @jaubourg: i never do some `jsonp` but it's not only these cases:
`index.html?callback=?`
`index.html?var=1&callback=?`
`http://ex.com/?callback=?` and dont find where we have to write two `?`.

Mr21:
@mzgol 
  +109    +40 dist/jquery.js
   +25     +8 dist/jquery.min.js

src/ajax.js
jaubourg: http://bugs.jquery.com/ticket/4897

Mr21: @jaubourg oh.. ok.
I read these tests: https://github.com/jquery/jquery/blob/master/test/unit/ajax.js#L704
Finally, I don't understand exactly what is it necessary for doing some `jsonp`...
The doc http://api.jquery.com/jquery.ajax/ doesn't speak about `??` anywhere :(

jaubourg: Not the first time we forgot to document something ;)

Not sure I understand your question but the fix (ie introduction of `??`) was for APIs that would have the callback as part of the URL, not as a parameter (see [the original ticket](http://bugs.jquery.com/ticket/4897)).

Mr21: Ahhhh I understood :)

It's really different with that behaviour :/
Because if we want to be friendly on `.html??var=` we have to be also friendly on `/callback=???var=` and `/callback=?/??var=` the regex is more complex and maybe it's not so necessary.

I think we have to use this regex: `.replace( /([^=]|=\?)\?+/, "$1?" )`
(some tests here: http://jsfiddle.net/Mr21/0sofoaay/)

jaubourg: Well, to be honest, I think we shouldn't try and fix anything because it is working as intended. The original bug report is bogus in that it is asking to cater to an illformed url.

mzgol:
@timmywil why did you remove the "Needs review" label? Shouldn't we discuss it?

dmethvin:
My take is [similar to @jaubourg](https://github.com/jquery/jquery/pull/2490/files#r37265083), this is a malformed URL because the second `?` should be URL encoded. For better or worse we are using `??` to have special meaning so it seems strange to subvert that in order to avoid mangling an already-invalid URL.

markelog:
@Mr21 we decided that this is notabug, nevertheless, thank you for the effort!

Mr21:
Right, no problem :) what's jaubourg told me increased a lot the complexity of the regex anyway...

Issue #2493, opened 2015-07-23
Table manipulation regression (XHTML)


anthonyryan1:
jQuery 3.0 introduced an XHTML regression in the appending and updating of table rows, it appears the outermost element in every `.append()`, `.html()`, `.load()`, etc gets removed and only the contents of that outermost element are actually appended or set as the innerHTML.

Simple test case (requires `Content-type: application/xhtml+xml; charset=utf-8;` to properly demonstrate the regression)

```html 
<!-- fill in doctype, html, head, body, etc. -->
<table>
<thead>
    <tr>
        <th>One</th>
        <th>Two</th>
        <th>Three</th>
    </tr>
</thead>
<tbody id="append"></tbody>
</table>
```

```javascript
$(document).ready(function(){
    $('#append').append('<tr><td>Four</td><td>Five</td><td>Six</td></tr>');
    $('#append').append('<tr><td>Seven</td><td>Eight</td><td>Nine</td></tr>');
});
```

In XHTML mode, the resulting DOM is the equivalent of:

```html
<table>
<thead>
    <tr>
        <th>One</th>
        <th>Two</th>
        <th>Three</th>
    </tr>
</thead>
<tbody id="append">
    <td>Four</td>
    <td>Five</td>
    <td>Six</td>
    <td>Seven</td>
    <td>Eight</td>
    <td>Nine</td>
</tbody>
</table>
```

I've also noticed the XHTML test harness is once again broken and likely how this regression occurred, but that's outside the scope of this issue.

kishu1:
NAVER - http://www.naver.com/
--------------------------------------------
김태훈(kishu@navercorp.com) 님은 현재 부재중입니다.</br>
보내신 메일 <[jquery] Table manipulation regression (XHTML) (#2493)> 은 저장되어 있으므로 다시 보내실 필요는 없습니다.</br>
김태훈(kishu@navercorp.com) 님이 남기신 메시지 입니다.
--------------------------------------------
[Absence period : 2015-07-23 10:00 ~ 2015-07-23 19:00 GMT +09:00]<br>개인 사유로 연차 중입니다. 급한 용무는 모바일(010-2940-0359)로 연락 바랍니다.
--------------------------------------------

Mr21:
If I test this in chrome:
``` html
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
[...]

PR #2499, opened 2015-07-25
Manipulation: Don't provide the parser with sloppy table markup

anthonyryan1:
While we can reply on parsers that were designed to cope with malformed syntax to understand what we mean, we shouldn't intentionally provide bad markup, not all parsers will accept it.

Reverts 0ea342a6a6dce793c1b0f14f051c2573f40f4e44
See also gh-2031, gh-2002
Fixes gh-2493

### DIFF
--- a/src/manipulation/wrapMap.js
+++ b/src/manipulation/wrapMap.js
@@ -6,21 +6,13 @@ var wrapMap = {
 	// Support: IE9
 	option: [ 1, "<select multiple='multiple'>", "</select>" ],
 
+	// XHTML parsers do not magically insert elements in the
+	// same way that tag soup parsers do. So we cannot shorten
+	// this by omitting <tbody> or other required elements.
 	thead: [ 1, "<table>", "</table>" ],
-
-	// Some of the following wrappers are not fully defined, because
-	// their parent elements (except for "table" element) could be omitted
-	// since browser parsers are smart enough to auto-insert them
-
-	// Support: Android 2.3
-	// Android browser doesn't auto-insert colgroup
 	col: [ 2, "<table><colgroup>", "</colgroup></table>" ],
-
-	// Auto-insert "tbody" element
-	tr: [ 2, "<table>", "</table>" ],
-
-	// Auto-insert "tbody" and "tr" elements
-	td: [ 3, "<table>", "</table>" ],
+	tr: [ 2, "<table><tbody>", "</tbody></table>" ],
+	td: [ 3, "<table><tbody><tr>", "</tr></tbody></table>" ],
 
 	_default: [ 0, "", "" ]
 };


timmywil:
It would be great if we had a test case for this. Maybe all we need is to finish https://github.com/jquery/jquery/issues/1731

mzgol:
> Maybe all we need is to finish #1731

Most likely. This will be my priority after dealing with this Android 2.3-related things.

markelog:
Hm, weird, i remember checking this in xhtml document, there should be more bugs like that, like auto-insertion of `tbody` could be a candidate

anthonyryan1:
Finishing #1731 would be sufficient to have caught this I believe, but as I outlined in that issue a while back, it's a rather tricky problem (in the same way unit testing under CSP proved very difficult to do in practice). The jQuery test suite was designed almost exclusively for HTML mode so far, and only a subset of the tests would be valid.

If we converted the entire test suite a a polyglot that can be run under XHTML and HTML, we introduce the possibility of regressions for people in the HTML mode, which is a very bad idea given that's likely how 99% of jQuery code is executed.

I do believe the unit testing of this belongs more in the scope of #1731 than here, since I'm fairly certain this functionality is already has tests, and would be happy to continue the discussion of possible solutions there.

mzgol:
@anthonyryan1 Is this the only XHTML-related issue you found in current code?

@jquery/core Do we want to land it as is or wait for a proper XHTML testing? I still have it on my plate but it may require some time so I can't commit to it before October.

anthonyryan1:
I've been running it in production with sizzle, deprecated, jsonp and a few other parts removed so I can't fully comment on those parts at the moment.

I'll schedule some testing that will include those.

dmethvin:
I think we should land this ASAP and let the XML test arrive when it does.

mzgol:
This will influence #2581 a little as there I'm currently removing one more such thing. It would be good to land this one and then I'll remove the removal from my PR.

src/manipulation/wrapMap.js
```diff
@@ -8,19 +8,15 @@ var wrapMap = {
 
 	thead: [ 1, "<table>", "</table>" ],
 
-	// Some of the following wrappers are not fully defined, because
-	// their parent elements (except for "table" element) could be omitted
-	// since browser parsers are smart enough to auto-insert them
-
 	// Support: Android 2.3
```
mzgol: This comment should be removed as it suggests we're only putting `<colgroup>` here because of Android 2.3 and it's XML as well. The XHTML-related comment from below should be moved above this line.

anthonyryan1:
Fixed the comment including removing the support comment for Android 2.3 anticipating #2581.

mzgol:
Landed, thanks!

Issue #2502, opened 2015-07-27
Switch support tests to use document.documentElement instead of document.body


mzgol:
See https://github.com/jquery/jquery/issues/2133#issuecomment-125375758. I used `document.body` for support tests in `1.11` because of test failures but it seems no longer to be a problem, I'm able to run support tests in IE8 before document ready and they give correct result. We can synchronize these code paths.

Refs gh-2133.

mzgol:
PR: #2504.

mzgol:
Landed at 9b04201ff21b6e5932dc21679774f398d6940cfa

mzgol:
This has been problematic in IE<8 AFAIK so I'd rather not backport this to 1.12.

Issue #2501, opened 2015-07-27
Support non-browser environments in jQuery Compat


mzgol:
Per today's meeting, we'll try to support non-browser environments in jQuery Compat.

mzgol:
PR: #2504.

markelog:
Landed via https://github.com/jquery/jquery/commit/04ec688e806660107a0b5823fabe2fe213f63b92

mzgol:
I'd like to *not* backport it to 1.12.

timmywil:
@mzgol I agree

markelog:
Since 1.12 will be released, i will remove `wontfix` label

mzgol:
I'd like it to stay wontfix. Supporting non-browser wnvironments requires some effort and IE<8-specific code can be problematic. I don't want to guarantee that 1.x is now good for non-browser, it better stay unsupported, it won't live much longer anyway.

Issue #2133, opened 2015-03-09
Synchronize use of document globals


gibson042:
Changes from 76df9e4e389d80bff410a9e5f08b848de1d21a2f (in particular, definition and use of the `document` and `documentElement` variables) never made it to the compat branch, introducing what I feel to be a needless divergence and source of merge conflicts.

@mzgol: agree?

mzgol:
I mostly didn't backport it because this is only required for non-browser environments which we don't support in the compat line. But I have nothing against porting them if that makes cherry-picking easier.

Note, though, that on `compat` we have a different jsHint config that assumes the browser environment; the one on `master` only declares `window` & timer functions.

mzgol:
I'll create a `document` variable as part of #2501 but `documentElement` doesn't seem worth it - on `master` it's mostly used in `selector-native` which is not supported on `compat` & support tests which use `document.body`, not `document.documentElement` on `compat`.

mzgol:
> support tests which use `document.body`, not `document.documentElement` on `compat`.

BTW, it's worth checking if https://github.com/jquery/jquery/pull/1342#issuecomment-23911836 still applies or maybe we could switch from `body` to `documentElement` for support tests even on `compat`.

mzgol:
It seems we can switch from `body` to `documentElement` on `compat`, support tests work fine in IE8 before doc ready. :) PR coming.

mzgol:
PR: #2504.

markelog:
Landed via https://github.com/jquery/jquery/commit/04ec688e806660107a0b5823fabe2fe213f63b92

mzgol:
I'd rather not backport it to 1.12 as there is way more compat code there and it may pose problems for Node.js.

mzgol:
This is a wontfix now since it was done in connection with https://github.com/jquery/jquery/issues/2501.

markelog:
Since 1.12 will be released, i will remove `wontfix` label

mzgol:
Last time I tried it it caused problems for IE<8. I'd rather it stayed wontfix unless we can verify IE 6-7 don't have problems with it.

PR #2504, opened 2015-07-28
Core: Support non-browser environments

mzgol:
Fixes gh-2133
Fixes gh-2501
Closes gh-2504
Refs gh-1950
Refs gh-1949
Refs gh-2397
Refs gh-1537
Refs gh-2504
Refs 842958e7aecd0d75a7ee9e2aaec83457701aa2f3

Second commit adds support for non-browser environments. Message of the first commit:

Core:CSS: Attach test nodes to documentElement, not body

Attaching test divs to document.documentElement instead of document.body
used to cause issues in jQuery 1.x; jQuery Compat doesn't execute any tests
on document ready, though so it could be aligned with master.

This makes jQuery Compat support tests work correctly even if jQuery is
included & used in head before body even exists - making it similar to
the master behavior.

Fixes gh-2502

EDIT: BTW, this PR decreases the gzipped size by 5 bytes. ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@
 
 /dist
 /node_modules
+
+/test/node_smoke_tests/lib/ensure_iterability.js
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,5 +1,6 @@
 {
 	"preset": "jquery",
 
-	"excludeFiles": [ "external", "src/intro.js", "src/outro.js" ]
+	"excludeFiles": [ "external", "src/intro.js", "src/outro.js",
+		"test/node_smoke_tests/lib/ensure_iterability.js" ]
 }
--- a/.jshintignore
+++ b/.jshintignore
@@ -9,3 +9,4 @@ test/data/readywaitasset.js
 test/data/readywaitloader.js
 test/data/support/csp.js
 test/data/support/getComputedSupport.js
+test/node_smoke_tests/lib/ensure_iterability.js
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -34,6 +34,18 @@ module.exports = function( grunt ) {
 				cache: "build/.sizecache.json"
 			}
 		},
+		babel: {
+			options: {
+				sourceMap: "inline",
+				retainLines: true
+			},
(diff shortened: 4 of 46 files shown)


mzgol:
Once this lands (I'd appreciate a quick review of this PR, @jquery/core), I'll work on a basic test suite for Android 2.3.

@timmywil @gibson042 Am I right that you prefer to not remove [Android 2.3 workarounds like this one](https://github.com/jquery/jquery/blob/6044fb6a7384aec85906949835ef9a58114896ce/test/unit/ajax.js#L1455-L1460) from `master` tests even when Android 2.3 won't even be reaching that code to reduce divergence from `master`? I guess we won't backport new Android 2.3-specific workarounds from `compat` to `master` so the divergence will happen anyway but whatever you prefer.

gibson042:
> @timmywil @gibson042 Am I right that you prefer to not remove [Android 2.3 workarounds like this one](https://github.com/jquery/jquery/blob/6044fb6a7384aec85906949835ef9a58114896ce/test/unit/ajax.js#L1455-L1460) from `master` tests even when Android 2.3 won't even be reaching that code to reduce divergence from `master`? I guess we won't backport new Android 2.3-specific workarounds from `compat` to `master` so the divergence will happen anyway (and the "workarounds" are mainly skipping the test which is easy to isolate) but whatever you prefer.

My preference is for everything under `/test` to be identical between branches, so we don't run into unnecessary merge conflicts. We may not get there, but I definitely don't want to invest effort in explicit divergence.

gibson042:
:clap:

This looks great, @mzgol.

timmywil:
> We may not get there, but I definitely don't want to invest effort in explicit divergence.

Agreed.

dmethvin:
Yes, merge conflicts are a pain so isolating those would be very helpful. That will become really clear if we ever do a style sweep on the unit tests!

mzgol:
> We may not get there, but I definitely don't want to invest effort in explicit divergence.

@gibson042 @timmywil @dmethvin works for me. You should all like this PR, then, since it makes `src/**` more similar between `master` & `compat`. ;)

PR #2507, opened 2015-07-29
Tests: don't use deprecated argument in test declaration

markelog:
@FagnerMartinsBrack hope you will be happy about this :-).

Did it with CST script manipulation - https://github.com/markelog/qunit-migration/blob/master/src/migration.js

@scottgonzalez, @jzaefferer, @Krinkle
I suppose it would be useful if qunit would provide scripts like that, here it only replaces deprecated argument, but it could do a lot more.

@nzakas i guess this could be another example of manipulation with CST.

### DIFF
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -6,7 +6,9 @@ module( "ajax", {
 });
 
 (function() {
-	test("Unit Testing Environment", 2, function () {
+	test("Unit Testing Environment", function () {
+		expect( 2 );
+
 		ok( hasPHP, "Running in an environment with PHP support. The AJAX tests only run if the environment supports PHP!" );
 		ok( !isLocal, "Unit tests are not ran from file:// (especially in Chrome. If you must test from file:// with Chrome, run it with the --allow-file-access-from-files flag!)" );
 	});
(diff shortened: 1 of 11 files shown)


FagnerMartinsBrack:
> @FagnerMartinsBrack hope you will be happy about this :-).

I won't be "happy", just thought it would be useful reminding about it ¬¬

jzaefferer:
There's a similar tool for migrating to the QUnit 2.0 API, though using regex replace: https://github.com/apsdehal/qunit-migrate

Surely would be good to have one reliable tool that handles the full migration.

markelog:
@FagnerMartinsBrack
> just thought it would be useful reminding about it ¬¬

It totally was! Your remainder prompt me to do this

> There's a similar tool for migrating to the QUnit 2.0 API

That sounds cool, although regexs aren't, it seems though, it missing this kind of manipulation.

dmethvin:
This LGTM. It would be convenient if `QUnit.config.requireExpect` let you require only the `expect()` format instead of the function arg, but I doubt this will be a problem in future tests since people tend to copy whatever style is nearby.

mzgol:
LGTM.

markelog:
In weekend i will try to apply manipulation from the module that @jzaefferer mentioned

Issue #2505, opened 2015-07-28
Add basic tests for Android 2.3 on master, drop full support


mzgol:
jQuery 3.0.0 will include only basic support for Android 2.3, bigger non-critical workarounds like `reliableMarginRight` will be `compat`-only.

Those tests need to be written.

To not create further divergence between `master` & `compat` test files, I propose to create a separate `android` "module" with those basic tests. Thoughts?

gibson042:
> To not create further divergence between `master` & `compat` test files, I propose to create a separate `android` "module" with those basic tests. Thoughts?

I like the idea, but would rather see a name like "basic" or "minimal" so we're not tied to a given client—the same suite might apply to e.g. jsdom or PhantomJS.

mzgol:
> I (...) would rather see a name like "basic" or "minimal" so we're not tied to a given client

Works for me. Good idea about jsdom & PhantomJS, btw, this might not be that hard to add once we have this.

mzgol:
PR (in progress but I need feedback before I continue): #2509.

mzgol:
This is not fixed until we reduce support as in the title. Only the first part (basic tests) have been finished.

mzgol:
PR: #2581

Issue #2483, opened 2015-07-20
Reduce Android 2.3 support on master (was: Investigate Android 2.3 support)


mzgol:
### We're already not fully supporting Android 2.3

The hard truth is we've never fully supported Android 2.3. First of all, our unit tests don't pass there in a couple of modules: http://swarm.jquery.org/job/942. There are serious AJAX issues with error handling: #1786, #2458 (#2458 seems pretty serious, see https://github.com/jquery/jquery/pull/2458#issuecomment-119572119), there are manipulation issues with checked radios: #1785.

Even though most of the tests pass, we cheat a little by putting Android 2.3 in special paths in some of them:

1. https://github.com/jquery/jquery/blob/1c59b308d201d6dd0f0aed2ad0256d01b9f68047/test/unit/ajax.js#L1507-L1512
2. https://github.com/jquery/jquery/blob/1c59b308d201d6dd0f0aed2ad0256d01b9f68047/test/unit/event.js#L408-L413
3. https://github.com/jquery/jquery/blob/1c59b308d201d6dd0f0aed2ad0256d01b9f68047/test/unit/manipulation.js#L419-L426 & https://github.com/jquery/jquery/blob/1c59b308d201d6dd0f0aed2ad0256d01b9f68047/test/unit/manipulation.js#L440-L444
4. https://github.com/jquery/jquery/blob/1c59b308d201d6dd0f0aed2ad0256d01b9f68047/test/unit/support.js#L236-L254

### Maintenance cost

There are various problems that happen only on Android 2.3, e.g. now it turns out it doesn't allow unquoted reserved words as property names, exactly as our favorite IE8: #2478. It happens quite often that something just doesn't work, e.g. currently for some reason AJAX tests hang very near the beginning: http://swarm.jquery.org/result/479018. Surprisingly, it works on `compat`: http://swarm.jquery.org/result/479031. Lots of hours are being constantly put into making this browser work with our tests.

### Market share

When we released the last stable version of jQuery, 2.1 (I'm omitting patches) [a year and a half ago](http://blog.jquery.com/2014/01/24/jquery-1-11-and-2-1-released/), Android 2.3 still had a very large market share. Currently [Google stats](http://developer.android.com/about/dashboards/index.html) show it at 5.6% of all Android versions (this is not a mobile market share but Android market share). Real percentage might be a little higher as it's counted via Google Play Store but it's still not that large as it was back then.

### Size

[...]

PR #2509, opened 2015-07-29
Tests: Add simple tests for Android 2.3

mzgol:
Fixes gh-2505
Refs gh-2483

@jquery/core This is an initial stab at simple tests for Android 2.3 but please review what I have before I write tests for other modules.

I've created a separate "module" called `basic` to contain those selected tests; I also added it to the list run by regular browsers so that we immediately know if those tests are ok.

Note that tests are *very* basic but testing the whole API even without edge cases (however you define them) would be way too much; if we run more than 10% of assertions in Android 2.3 it will be unstable (older browsers fare better with many modules with few tests in them rather than one module with all). I'd rather remove some CSS tests from this list than add more.

I've temporarily modified our Jenkins jobs to see how that would look:

1. the regular periodic run without Android 2.3: http://swarm.jquery.org/job/1009
2. the basic run just for Android 2.3: http://swarm.jquery.org/job/1011

Even these 18 CSS tests seem to *sometimes* run very slow on Android 2.3: they were running for 11 minutes & timed out, after restarting they passed almost immediately...

### DIFF
--- a/test/unit/basic.js
+++ b/test/unit/basic.js
@@ -1,0 +1,258 @@
+module( "basic", {
+	teardown: moduleTeardown
+} );
+
+if ( jQuery.ajax ) {
+test( "ajax", function( assert ) {
+	expect( 4 );
+
+	var done = jQuery.map( new Array( 3 ), function() { return assert.async(); } );
+
+	jQuery.ajax({
+		type: "GET",
+		url: url( "data/name.php?name=foo" ),
+		success: function( msg ) {
+			assert.strictEqual( msg, "bar", "Check for GET" );
+			done.pop()();
+		}
+	});
+
+	jQuery.ajax({
+		type: "POST",
+		url: url( "data/name.php" ),
+		data: "name=peter",
+		success: function( msg ) {
+			strictEqual( msg, "pan", "Check for POST" );
+			done.pop()();
+		}
+	});
+
+	jQuery( "#first" ).load( url( "data/name.html" ), function() {
+		assert.ok( /^ERROR/.test( jQuery( "#first" ).text() ),
+			"Check if content was injected into the DOM" );
+		done.pop()();
+	});
+} );
+}
+
+test( "attributes", function ( assert ) {
+	expect( 6 );
+
+	var a = jQuery( "<a/>" ).appendTo( "#qunit-fixture" ),
+		input = jQuery( "<input/>" ).appendTo( "#qunit-fixture" );
+
+	assert.strictEqual( a.attr( "foo", "bar" ).attr( "foo" ), "bar", ".attr getter/setter" );
+	assert.strictEqual( a.removeAttr( "foo" ).attr( "foo" ), undefined, ".removeAttr" );
+	assert.strictEqual( a.prop( "href", "#5" ).prop( "href" ),
+		location.href.replace( /\#.*$/, "" ) + "#5",
+		".prop getter/setter" );
+
+	a.addClass( "abc def ghj" ).removeClass( "def ghj" );
+	assert.strictEqual( a.hasClass( "abc" ), true, ".(add|remove|has)Class, class present" );
+	assert.strictEqual( a.hasClass( "def" ), false, ".(add|remove|has)Class, class missing" );
+
+	assert.strictEqual( input.val( "xyz" ).val(), "xyz", ".val getter/setter" );
+} );
+
+if ( jQuery.css ) {
+test( "css", function( assert ) {
+	expect( 3 );
+
+	var div = jQuery( "<div/>" ).appendTo( "#qunit-fixture" );
+
+	assert.strictEqual( div.css( "width", "50px" ).css( "width" ), "50px", ".css getter/setter" );
+
+	div.hide();
+	assert.strictEqual( div.css( "display" ), "none", "div hidden" );
+	div.show();
+	assert.strictEqual( div.css( "display" ), "block", "div shown" );
+} );
+}
+
+test( "core", function ( assert ) {
+	expect( 28 );
+
+	var elem = jQuery( "<div></div><span></span>" );
+
+	assert.strictEqual( elem.length, 2, "Correct number of elements" );
+	assert.strictEqual( jQuery.trim("  hello   "), "hello", "jQuery.trim" );
+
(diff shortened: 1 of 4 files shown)


dmethvin:
Anything that goes through only `basic` support doesn't test many code paths at all, yet most of them work with Android 2.3 don't they? It seems like we'd want to instead specifically exclude (and document) the cases we know don't work, like `reliableMarginRight`. That list of tests could be pulled out into its own module perhaps?

mzgol:
> Anything that goes through only `basic` support doesn't test many code paths at all, yet most of them work with Android 2.3 don't they?

Most currently do. The fact that `basic` won't test many "basic" things is one of the reasons why I'd prefer to just drop Android 2.3 on `master` and leave it just on `compat` to avoid such weird compatibility state but only @markelog supported me in this view and we decided we'll be just running basic tests, i.e. what this PR tries to do.

But yeah, I'm still not terribly convinced.

> It seems like we'd want to instead specifically exclude (and document) the cases we know don't work, like reliableMarginRight. That list of tests could be pulled out into its own module perhaps?

Pulling such tests to a separate module would be catering to Android 2.3 too much, IMO; we'd break module separation just because of one browser. IMO it's easier (& better) to just skip problematic tests as we started to do now.

The problem is not only failing tests, though. Those tests run extremely slow, usually ~1/3 of the modules time out and I have to restart those runs a few times before they stabilize and (mostly) pass to be sure we didn't regress. That's quite tedious. Running just very basic tests should at least help with this problem.

test/unit/basic.js
```diff
@@ -0,0 +1,84 @@
+module( "basic", {
+	teardown: moduleTeardown
+} );
+
+if ( jQuery.css ) {
+	test( "css", function() {
+		expect( 18 );
```
gibson042: This already looks like too much... I'd expect to see one `test` per module (as you have), but no more than 10 assertions per test, and hopefully even fewer. For the CSS module, maybe one assertion like `assert.equal( div.css( "width", "50px" ).css( "width" ), "50px", ".css getter/setter" )` and one for `.show`/`.hide` and call it done.

mzgol:
PR updated (traversing & wrap tests to be written).

mzgol:
Ready for final review.

dmethvin:
Now I'm really wanting to get some code coverage stats. :)

test/unit/basic.js
```diff
@@ -0,0 +1,271 @@
...
+	teardown: moduleTeardown
+} );
+
+if ( jQuery.ajax ) {
+test( "ajax", function( assert ) {
+	expect( 4 );
+
+	var requestsLeft = 3;
+
+	stop();
```
gibson042: Could you update these to use `assert.async` instead of `stop`/`start`?

markelog: We can do this across the all unit files, so it could be done later.

test/unit/basic.js
```diff
@@ -0,0 +1,271 @@
...
+
+	assert.ok( jQuery.isWindow( window ), "jQuery.isWindow(window)" );
+	assert.ok( !jQuery.isWindow( 2 ), "jQuery.isWindow(Number)" );
+
+	assert.strictEqual( jQuery.inArray( 3, [ "a", 6, false, 3, {} ] ), 3, "jQuery.inArray - true" );
+	assert.strictEqual( jQuery.inArray( 3, [ "a", 6, false, "3", {} ] ), -1, "jQuery.inArray - false" );
+
+	assert.strictEqual( elem.get( 1 ), elem[ 1 ], ".get" );
+	assert.strictEqual( elem.first()[ 0 ], elem.get( 0 ), ".first" );
+	assert.strictEqual( elem.last()[ 0 ], elem.get( 1 ), ".last" );
```
gibson042: Let's define expected values without `.get`.

mzgol:
> Now I'm really wanting to get some code coverage stats. :)

I had significant problems with getting coverage to be generated in our current setup (most likely I did everything wrong :D), in the end I hacked it around in PhantomJS but excluding jQuery.ajax tests as I had problems with getting it running and I got:

* Statements: 42.14% (1369 / 3249)
* Branches: 28.39% (926 / 3262)
* Functions: 42.58% (244 / 573)
* Lines: 42.14% (1369 / 3249)

test/unit/basic.js
mzgol: Sure, since I use the new syntax in these files I may as well get it right.

test/unit/basic.js
mzgol: Done.

markelog:
These stats for all modules? Could run just `basic` ones?

mzgol:
No, just for `basic`.

mzgol:
Travis fails because of problems with installing old jsdom... Can we get #2526 landed ASAP?

gibson042:
This looks nice. :+1:

timmywil:
:+1:

test/unit/basic.js
```diff
@@ -0,0 +1,268 @@
...
+		requestsLeft = 3;
+
+	jQuery.ajax({
+		type: "GET",
+		url: url( "data/name.php?name=foo" ),
+		success: function( msg ) {
+			assert.strictEqual( msg, "bar", "Check for GET" );
+			requestsLeft--;
+			if ( !requestsLeft ) {
+				done();
```
gibson042: You don't need to do the tracking yourself; just use `assert.async` multiple times (cf. [test/unit/deferred.js](https://github.com/jquery/jquery/blob/1c59b308d201d6dd0f0aed2ad0256d01b9f68047/test/unit/deferred.js#L411-L421)).

mzgol: Updated, although that doesn't save that much. Ideally `assert.async()` would accept a parameter indicating how many async operations are expected and that would require calling `done()` this many times.

BTW, `done.pop().call()` is not needed, it's just `done.pop()()` after all.

gibson042: Of course you're right about `.call()`, and I agree on the former point. In fact, I meant to create a ticket for that earlier, but it fell off my radar. Anyway, corrected now: jquery/qunit/issues/843

mzgol: Thanks for creating a ticket, I should've done it myself. :)

PR #2518, opened 2015-08-01
Skip empty values in jQuery.param

cawa-93:


mzgol:
Thanks for the PR, @cawa-93. Could you create an issue describing the problem first? Also, we'll need you to sign our [CLA](https://contribute.jquery.org/CLA/) to accept any contribution.

In addition, any PR that changes behavior needs to have a unit test added that confirms the fix works. Please make sure any changes you make don't break existing tests.

dmethvin:
Looks like this was an attempt to solve the problem in #2522? It would mean a breaking change in current behavior since at present `{ mydata: undefined }` translates to `mydata=` and now it would be missing.

PR #2520, opened 2015-08-01
Build: Add a comment explaining why the es3 option is needed

mzgol:
It might not be obvious to everyone that IE 9 & Android 4.0 are not
ES5-compliant browsers (by a large margin) so it's better to add a support
comment. This requires slight changes in parsing the config file
as it's not a pure JSON anymore. JSHint understands such files without
problems.

### DIFF
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -12,6 +12,9 @@
 
 	"sub": true,
 
+	// Support: IE < 10, Android < 4.1
+	// The above browsers are failing a lot of tests in the ES5
+	// test suite at http://test262.ecmascript.org.
 	"es3": true,
 
 	"globals": {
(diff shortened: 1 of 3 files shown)


src/.jshintrc
```diff
@@ -12,6 +12,9 @@
 
 	"sub": true,
 
+	// Support: IE < 10, Android < 4.1
+	// The above browsers are failing a lot of tests in the ES5
+	// test suite at http://test262.ecmascript.org.
```
gibson042: Damn, that's ugly. It also appears to be both undocumented and untested in JSHint (albeit very [longstanding](https://github.com/jshint/jshint/commit/b0afb93ca05493e5994bd908f54bbdf21f4e1f0e)), and just feels like bad form. On the other hand, dummy keys aren't really any better. Color me ambivalent about the change, and—per usual—frustrated with JSON.

mzgol: I didn't realize it's undocumented (I assume it should be, @rwaldron?). I agree it feels bad, it just felt more bad to me to have this undocumented and have someone remove the option prematurely, as we did on `master` before.

Issue #2513, opened 2015-07-31
Don't expose jQuery.access


mzgol:
`jQuery.access` is public & undocumented, is there a reason we should keep it that way? We removed `jQuery.swap` (#2058), `jQuery.buildFragment` (#2224) & `jQuery.domManip` (#2225).

See also #2512.

mzgol:
PR: #2524

milesj:
This was unfortunate. `$.access()` was great for building our own getter/setter combo methods.

timmywil:
Sorry for the inconvenience! I'd be willing to consider exposing this method (and documenting it) for plugins, but we don't like having public methods that aren't documented.

PR #2524, opened 2015-08-03
Core: Don't expose jQuery.access

mzgol:
jQuery.access was never documented, there is no need to keep it exposed.

Fixes gh-2513

### DIFF
--- a/src/core/access.js
+++ b/src/core/access.js
@@ -4,7 +4,7 @@ define([
 
 // Multifunctional method to get and set values of a collection
 // The value/s can optionally be executed if it's a function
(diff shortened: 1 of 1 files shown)


mzgol:
Because of how trivial it is I'd like to land it tomorrow.

dmethvin:
Yeah, pretty trivial so no problem assuming nobody is using it externally. It's not used in Migrate, the only place I might fear we'd use it.

mzgol:
I wanted to land it now but swarm is down. :(

mzgol:
Landed (swarm is OK again).

Issue #2512, opened 2015-07-31
Don't expose jQuery.dir & jQuery.sibling


mzgol:
We have a few more undocumented APIs shipped under the jQuery object; two of them are `jQuery.dir` & `jQuery.sibling`.

Is there any reason why they should be exposed & undocumented? We should minimize the number of such cases as we did with `jQuery.swap` (#2058), `jQuery.buildFragment` (#2224) & `jQuery.domManip` (#2225).

See also #2513.

mzgol:
PR: #2525

PR #2525, opened 2015-08-03
Traversing: Don't expose jQuery.dir & jQuery.sibling

mzgol:
jQuery.dir & jQuery.sibling are undocumented internal APIs; they shouldn't
be exposed.

Fixes gh-2512

-1 byte :)

### DIFF
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -1,11 +1,13 @@
 define([
 	"./core",
 	"./var/indexOf",
+	"./traversing/var/dir",
+	"./traversing/var/siblings",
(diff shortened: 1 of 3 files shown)


src/traversing.js
```diff
@@ -1,11 +1,13 @@
 define([
 	"./core",
 	"./var/indexOf",
+	"./traversing/var/dir",
+	"./traversing/var/siblings",
```
mzgol: We already have a local `sibling` function so I named it `siblings`.

dmethvin:
LGTM

Issue #2519, opened 2015-08-01
Don't install jsdom 3.x - it requires Python & Visual Studio on Windows


anthomaxcool:
Hi, I'm having some problems like the title says it.

In the terminal, it says:

Can't find Python executable ...
....
Tell the author that this fails on your system:
        node-gyp rebuild
....
Tell the author that this fails on your system:
       npm install && grunt

It also says that I need to include the file "npm-debug.log" with my support request... but I can't find it, this file doesn't exist.

So if needed, I copy pasted everything that was written in my terminal here ->

Running "jsdom" task
-

> contextify@0.1.14 install c:\Users\Anthony\website\jquery\node_modules\jsdom\n
ode_modules\contextify
> node-gyp rebuild

-
c:\Users\Anthony\website\jquery\node_modules\jsdom\node_modules\contextify>if no
t defined npm_config_node_gyp (node "c:\Program Files\nodejs\node_modules\npm\bi
n\node-gyp-bin\\..\..\node_modules\node-gyp\bin\node-gyp.js" rebuild )  else (re
build)
gyp ERR! configure error
gyp ERR! stack Error: Can't find Python executable "python", you can set the PYT
HON env variable.
gyp ERR! stack     at failNoPython (c:\Program Files\nodejs\node_modules\npm\nod
e_modules\node-gyp\lib\configure.js:103:14)
gyp ERR! stack     at c:\Program Files\nodejs\node_modules\npm\node_modules\node
-gyp\lib\configure.js:64:11
gyp ERR! stack     at FSReqWrap.oncomplete (evalmachine.<anonymous>:95:15)
gyp ERR! System Windows_NT 6.3.9600
gyp ERR! command "node" "c:\\Program Files\\nodejs\\node_modules\\npm\\node_modu
les\\node-gyp\\bin\\node-gyp.js" "rebuild"
gyp ERR! cwd c:\Users\Anthony\website\jquery\node_modules\jsdom\node_modules\con
textify
gyp ERR! node -v v0.12.4
gyp ERR! node-gyp -v v1.0.3
gyp ERR! not ok
npm ERR! Windows_NT 6.3.9600
npm ERR! argv "c:\\Program Files\\nodejs\\\\node.exe" "c:\\Program Files\\nodejs
\\node_modules\\npm\\bin\\npm-cli.js" "install" "jsdom@3.1.2"
npm ERR! node v0.12.4
npm ERR! npm  v2.10.1
npm ERR! code ELIFECYCLE

[...]

PR #2526, opened 2015-08-03
Build: Don't install jsdom 3 on Node.js 0.10 & 0.12 by default

mzgol:
jsdom 3 requires Python & Visual Studio on Windows which is a significant
barrier to contributors. Newer jsdom versions don't require pre-compiling
but work only on io.js. This commit installs the new jsdom everywhere (it
does install in old Node.js, it just won't work) and executes Node-related
tests only on newer Nodes or if a working jsdom version is installed. The
latter can be achieved by running the `old_jsdom` task.

Node.js is merging with io.js soon so this will become a smaller problem over
time.

One drawback is our Jenkins setup runs on Node 0.10 so it won't be running
Node tests anymore. We have Travis set up on io.js, though so all PRs
have those tests run. When the new LTS Node.js arrives (as it soon merges
with io.js) we should update our Jenkins infrastructure so that it runs on this
new version.

Fixes gh-2519


dmethvin:
Totally my bad on this, I noticed it and changed my normal grunt-and-test alias to skip it a while back, but forgot to open an issue because I was on a plane. It's definitely not worth extra work to run it for Windows contributors.

Issue #2532, opened 2015-08-10
Modify ajaxConvert to return a promise/thenable.


nathanhammond:
The internal `ajaxConvert` function always returns synchronously. This makes it impossible to support new techniques such as [async `JSON.parse`](http://azimi.me/2015/07/30/non-blocking-async-json-parse.html) by defining custom converters.

Considering:
- Ajax is already asynchronous.
- `ajaxConvert` is in an unreachable private scope.

I propose that we modify `ajaxConvert` to return a promise/thenable. If this is considered beneficial, I'll have a patch up with tests this week.

dcherman:
Ajax is not always asynchronous; the `async: false` option still exists.

nathanhammond:
@dcherman Yes, but the interface into it via `$.ajax` is already async (callbacks or thenable). I suppose that you could do something which created side effects in a synchronous manner without using callbacks, but then I believe you've explicitly opted not to use the `response`.

My goal is to be able to write async converters, and am able to make `ajaxConvert` run both synchronously and asynchronously if necessary.

dmethvin:
That's not quite true, the plumbing calls inside are implicitly async up to this point so for example a uesr could have a `success` callback (not a Promise) but that could still use a converter.

I'm not sure whether we can support async filters and converters, they may break assumptions. It certainly would break assumptions of sync ajax but those are exceptions so maybe there is a way around it.

markelog:
I guess you wouldn't use `ajax` if you have fetch right? What is your use-case, what use-cases are there?

nathanhammond:
As it stands `fetch` is still a relatively immature API–we can still do much more with the `XHR` API by comparison. I suspect that they will be used in parallel for some time until achieving some sort of feature parity, plus the latent cost of migration efforts. I suspect that jQuery and `$.ajax` will be an important part of migrating people *to* `fetch`, and adding support for async converters creates a bit of a technical bridge.

The use case I have is an app with dependence upon Ember Data, which in turn delegates to `$.ajax` to make requests. We have server responses that push our parse times over the 16ms threshold and we want to handle that parsing asynchronously using `return (new Response(string)).json();`.

Goals:
- We don't want to replace the battle-tested `$.ajax` with a hand-rolled version.
- We want an intermediate step in the `XHR` => `fetch` transition.
- We want to enable every application using Ember Data to immediately benefit from this technique. (Improving an entire ecosystem.)
[...]

PR #2534, opened 2015-08-11
Ajax: Allow for async converters.

nathanhammond:
This PR is a proof-of-concept for #2532. It's far from perfect, but all existing tests in the Ajax module pass. Assuming that something like this would be a welcome inclusion to jQuery, there are a number of things still to do:

- [x] Tests.
- [x] Performance impact review.
- [ ] Code reformat–current format is designed to minimize the diff and make it easier to understand.

At this point I'd appreciate a review of concept to make sure continued experimentation in this vein is something worth entertaining. If so, I'll sign the CLA. :)

/cc @dmethvin @dcherman

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -201,12 +201,24 @@ function ajaxConvert( s, response, jqXHR, isSuccess ) {
 	while ( current ) {
 
 		if ( s.responseFields[ current ] ) {
-			jqXHR[ s.responseFields[ current ] ] = response;
+			if ( response && jQuery.isFunction( response.then ) ) {
+				/* jshint ignore:start */
+				response.then( (function(current) { return function( response ) { jqXHR[ s.responseFields[ current ] ] = response; }; })( current ) );
+				/* jshint ignore:end */
+			} else {
+				jqXHR[ s.responseFields[ current ] ] = response;
+			}
 		}
 
 		// Apply the dataFilter if provided
 		if ( !prev && isSuccess && s.dataFilter ) {
-			response = s.dataFilter( response, s.dataType );
+			if ( response && jQuery.isFunction( response.then ) ) {
+				/* jshint ignore:start */
+				response = response.then( (function() { return function( response ) { response = s.dataFilter( response, s.dataType ); }; })() );
+				/* jshint ignore:end */
+			} else {
+				response = s.dataFilter( response, s.dataType );
+			}
 		}
 
 		prev = current;
@@ -256,7 +268,9 @@ function ajaxConvert( s, response, jqXHR, isSuccess ) {
 				if ( conv !== true ) {
 
 					// Unless errors are allowed to bubble, catch and return them
-					if ( conv && s[ "throws" ] ) {
+					if ( response && jQuery.isFunction( response.then ) ) {
+						response = response.then( conv );
+					} else if ( conv && s[ "throws" ] ) {
 						response = conv( response );
 					} else {
 						try {
@@ -270,7 +284,18 @@ function ajaxConvert( s, response, jqXHR, isSuccess ) {
 		}
 	}
 
-	return { state: "success", data: response };
+	if ( response && jQuery.isFunction( response.then ) ) {
+		return response.then(
+			function( response ) {
+				return { state: "success", data: response };
+			},
+			function( e ) {
+				return { state: "parsererror", error: conv ? e : "No conversion from " + prev + " to " + current };
+			}
+		);
+	} else {
+		return { state: "success", data: response };
+	}
 }
 
 jQuery.extend({
@@ -678,6 +703,13 @@ jQuery.extend({
 			// Convert no matter what (that way responseXXX fields are always set)
 			response = ajaxConvert( s, response, jqXHR, isSuccess );
 
+			if ( response && jQuery.isFunction( response.then ) ) {
+				response.then( processResponse, processResponse );
+			} else {
+				processResponse( response );
+			}
+
+			function processResponse( response ) {
 			// If successful, handle type chaining
 			if ( isSuccess ) {
 
@@ -748,6 +780,7 @@ jQuery.extend({
 				if ( !( --jQuery.active ) ) {
 					jQuery.event.trigger("ajaxStop");
 				}
+			}
 			}
 		}
 
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -14,6 +14,7 @@ module( "ajax", {
 });
 
 (function() {
+	var originalConverters = jQuery.ajaxSettings.converters;
 
 	if ( !jQuery.ajax || ( isLocal && !hasPHP ) ) {
 		return;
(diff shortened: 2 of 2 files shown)


nathanhammond:
One additional note, I'm using native `Promise`s right now as well as `Function#bind` which will blow up the tests in some targets. Again, this is a POC to identify if further exploration is worth pursuing, not a final implementation.

Please review for concept, I can follow up with production-worthy code. :)

src/ajax.js
```diff
@@ -333,7 +349,15 @@ jQuery.extend({
 			"text html": true,
 
 			// Evaluate text as a json expression
-			"text json": jQuery.parseJSON,
+			"text json": function(json) {
```
dmethvin: Instead of modifying the standard `json` converter, you could create a `jsonAsync`. That way we don't change behavior for existing code.

nathanhammond: Totally agree, this was to see if I could get the tests passing and nothing more. The actual implementation will be this: `return (new Response(string)).json();`

nathanhammond: And I'm not sure that I would actually include an async JSON converter in the default build–you probably have to opt in by defining your own converter for "text json" in your `$.ajax` options.

dmethvin: Right, I was thinking that as well. We'd just add the ability for the converter to be async as you've done here.

dcherman:
So looking at this, one thing that would end up being weird is Promise chaining in the converters.  Since it's supported to do things like go from text -> json -> custom types, it seems non-intuitive to go from the json -> custom type stage of things here.

Under normal promise usage, I'd expect to write this code since handler expect to be passed the resulting value of the promise, not the promise itself:

```js
function myConverter(response) {
  return new MyType(response);
}
```

With this implementation, it looks like I'd instead need to do this:

```js
function myConverter(response) {
   return response.then(function(json) {
     return new MyType(json);
   });
}
```

If the while loop was refactored to unroll the Promise before passing the result into subsequent converters ( or just calling subsequent converters using .then if it exists ), then userland code would end up looking like my first example which feels like a more normal usage of Promises.

src/ajax.js
```diff
@@ -270,7 +271,22 @@ function ajaxConvert( s, response, jqXHR, isSuccess ) {
 		}
 	}
 
-	return { state: "success", data: response };
+	if (response && response.then) {
```
gibson042: This fails to address converter chains (note the `while ( current )` above).

nathanhammond: Yep, completely aware, and already on my checklist of things to do above. :) In making this PR I intend to close the gaps in test coverage that allowed my async converter to be present and still have all tests pass. I appreciate the detailed review!

nathanhammond:
@dcherman I'm imagining the API being such that the converter public API doesn't change except in allowing for returning `then`ables from a converter. I don't want any consumer of the API to have to be aware of the implementation details, excepting the conflict between `async: false` and `then`able converters. So, identical to your first example, plus the ability to return a promise.

nathanhammond:
Updated to allow for chainable converters, still with the minimally invasive approach. In this API converters may simply return `then`ables, but this has consequences since the object itself may have a property named `then` which this approach branches on incorrectly. There's no way to do `instanceof` on `jQuery.Deferred`s so I'm unsure that there is a good path forward using jQuery-only constructs.

Doing this right requires an `instanceof` check, and I'm not sure if we want to tie jQuery to browser-supplied constructs without broad adoption like Promises. http://caniuse.com/#search=promises

Unless there is an appetite for changing the constructor for `jQuery.Deferred`, I'd say we've reached impasse.

dcherman:
Another point to consider is that in synchronous parsing, exceptions would be thrown if an error occurred depending on whether or not the `throws` option was provided.  I assume we're going to want to provide some type of similar handling for any Promise based conversions since no kind of Unhandled Rejection handler exists in jQuery Deferred/Promises?

nathanhammond:
@dcherman I almost got the use case you describe, I missed accounting for `throws` in this `reject` handler: https://github.com/jquery/jquery/pull/2534/files#diff-c3749d3acba09ca9ec16bb56e496408bR289

I still feel like we've reached impasse here based upon needing duck-typing to identify thenables (explicitly, `jQuery.Deferred`) and having valid converters that could create something that would be duck-typed as a thenable but not be one.

staabm:
Would it make sense to trigger `.ajaxError`[1] for async errors, when no error handlers are defined for the returned promise?

[1] http://api.jquery.com/ajaxerror/

jaubourg:
The `throws` option only makes sense for synchronous ajax so I think we can safely always reject in case of an exception for asynchronous conversion.

Unhandled errors are the primary reasons why the original `Deferred.then` did not catch exceptions. It's still _the_ major flaw of standard Promises but the ship has sailed I've been told over and over again. Quite sure we'll see it back at the port in the years to come but heh.

Anyway, it is not the responsibility of the ajax module to detect unhandled rejections since it has no way of knowing about it, only the Deferred module would. And even so, how do you detect an error is not handled ? It's a common pattern to store Promises to be used later (in a cache for instance) and that's just the tip of the iceberg.

nathanhammond:
@dcherman @staabm @jaubourg I believe I've addressed the `throws` option with my latest push. If anybody can think of tests that they want to see please let me know.

Of the four things on my list to address I've handled chainability and testing with my latest push. I'd appreciate (possibly final) reviews in this state before I reformat and make the diff much larger.

src/ajax.js
```diff
@@ -254,9 +254,9 @@ function ajaxConvert( s, response, jqXHR, isSuccess ) {
 
 				// Apply converter (if not an equivalence)
 				if ( conv !== true ) {
-
-					// Unless errors are allowed to bubble, catch and return them
-					if ( conv && s[ "throws" ] ) {
+					if ( jQuery.isThenable(response) ) {
```
gibson042: I'm not a fan of introducing this abstraction. `response && jQuery.isFunction( response.then )` or `jQuery.isFunction( response && response.then )` would be better.

src/ajax.js
```diff
@@ -270,7 +270,23 @@ function ajaxConvert( s, response, jqXHR, isSuccess ) {
 		}
 	}
 
-	return { state: "success", data: response };
+	if ( jQuery.isThenable(response) ) {
+		return response.then(
+			function(response) {
+				this.responseJSON = response;
```
gibson042: Why assume that Promise-based conversion provides JSON?

src/ajax.js
```diff
@@ -270,7 +270,23 @@ function ajaxConvert( s, response, jqXHR, isSuccess ) {
...
+				return { state: "success", data: response };
+			}.bind(jqXHR),
+			function(e) {
+				if ( this[ "throws" ] ) {
+					throw e;
```
gibson042: This doesn't smell right... assuming a valid `.then` implementation, rethrowing is a no-op and very likely invisible anyway.

test/unit/ajax.js
```diff
@@ -1197,6 +1197,25 @@ module( "ajax", {
 		{
 			url: url("data/json.php"),
 			converters: {
+				"text json": function( data ) {
+					ok( true, "promise converter called" );
+					return jQuery.Deferred().resolveWith( data );
```
gibson042: This is not sufficient coverage, because `jQuery.Deferred().resolveWith( data )` is synchronous.

src/ajax.js
dcherman: Throwing here would be entirely invisible since the end user has no way to attach an error handler at the end of the Promise chain.  The only valid thing to do is have the ajax promise be rejected

src/ajax.js
nathanhammond: I did it to make it testable and because I was doing the same comparison in multiple places. Don't care too much, will modify.

src/ajax.js
nathanhammond: Good catch, I was stuck in my own little silo.

src/ajax.js
nathanhammond: Addressed. :) It was working as I expected it to because of my unfamiliarity with sync vs. async things inside of `jQuery.Deferred`.

test/unit/ajax.js
nathanhammond: I want coverage to be "run all existing tests with a sync `"text json"` converter, and with an async `"text json"` converter." Any recommendation as to how to do that?

nathanhammond:
I've simplified my approach, swapped in an async "text json" converter, made all existing tests pass, and identified one gap in test coverage (converter + datafilter). I want coverage to be "run all existing tests with a sync "text json" converter, and with an async "text json" converter." Any recommendation as to how to do that?

I believe that this solution (much simpler than my last, and I believe actually correct) fully addresses previous comments and doesn't have the logic gaps I introduced by not understanding `jQuery.Deferred`.

gibson042:
> I want coverage to be "run all existing tests with a sync "text json" converter, and with an async "text json" converter." Any recommendation as to how to do that?

One converter returning `jQuery.Deferred().resolve(…)`, a different one returning `jQuery.Deferred().resolve(…).then()` would do it. You can write a function that accepts a converter and its description, and runs corresponding tests ([example](https://github.com/jquery/jquery/blob/master/test/unit/deferred.js#L5-L11)).

nathanhammond:
How's this for coverage? I believe that I made all of the converters async for one pass, and sync for the other. Assuming this looks good I'll reformat.

dmethvin:
See #2532

Issue #2332, opened 2015-05-19
Input's form attribute and document submit handler produce an IE8 error


Igor10k:
When you use an input with `form` attribute while also handling the document's `submit` event you get an error in IE8.

Here's an easy way to reproduce it.

```html
<html>
<body>
  <input type="text" form="test">
  <form action="" id="test"></form>
  <script src="//code.jquery.com/jquery-1.11.3.min.js"></script>
  <script>
    $(document).on('submit', 'form', function () {});
  </script>
</body>
</html>
```

When you click the input in IE8 you get this error.

![screen shot 2015-05-19 at 16 10 42](https://cloud.githubusercontent.com/assets/1244342/7703590/a24b82d2-fe41-11e4-891d-ee385b0faa23.png)

Here's a jsfiddle http://jsfiddle.net/1g19xt3p/1/
And since jsfiddle doesn't work in IE8 here's a direct link to the result https://jsfiddle.net/1g19xt3p/1/embedded/result/

A small research showed that here's the code responsible
```js
jQuery.event.special.submit = {
  setup: function() {
    // skipped
    jQuery.event.add( this, "click._submit keypress._submit", function( e ) {
      var elem = e.target,
        form = jQuery.nodeName( elem, "input" ) || jQuery.nodeName( elem, "button" ) ? elem.form : undefined;
      if ( form && !jQuery._data( form, "submitBubbles" ) ) {
        // skipped
      }
    });
  },
  // skipped
```

The thing is `elem.form` returns the `form` attribute value in IE8 that is a `String` and jQuery can't get the data out of a `String`.

dmethvin:
See [trac-12717](http://bugs.jquery.com/ticket/12717) and [trac-14126](http://bugs.jquery.com/ticket/14126). Also note that Microsoft will [stop supporting IE8 in January 2016](http://blogs.msdn.com/b/ie/archive/2014/08/07/stay-up-to-date-with-internet-explorer.aspx) so it's really not worth a lot of additional IE8 heroics at this point.

Igor10k:
@dmethvin 

I see your point, but I still think that it wouldn't hurt to change
```js
if ( form && !jQuery._data( form, "submitBubbles" ) ) {
```
to
```js
if ( form && typeof form !== "string" && !jQuery._data( form, "submitBubbles" ) ) {
```
or maybe even
```js
if ( typeof form == "object" && !jQuery._data( form, "submitBubbles" ) ) {
```

[...]

PR #2536, opened 2015-08-12
Event: Don't try to read data on a non-element on IE8 form attribute

adrianolek:
This one is related to the error described in #2332
which is caused by jQuery trying to set data on a string, eg having:
```html
<button form="foo-form">Submit</button>
```
jQuery would try to do:
```js
jQuery._data( "foo-form", "submitBubbles", true );
```

which will throw an error here:
https://github.com/jquery/jquery/blob/062a7d63e4d1a55d72d64152381d93d60b0231e4/src/data.js#L125

This PR won't fix the behavior of form attribute on IE8, but will allow to apply your own polyfill for submit events when you need it (right now jQuery will just crash, and after debugging the only option available turns out to be overloading the whole jQuery.event.special.submit.setup function).


markelog:
This pull has wrong target, use `compat` branch, sorry for the confusion, also, no tests?

Issue #1755, opened 2014-10-21
Add Vagrant Support or other lightweight server


mgol:
Originally reported by eric@… at: http://bugs.jquery.com/ticket/14488

<p>
The current unit test instructions advise developers to install some variant of *AMP, which is a heavy install to set up their development environment.  Adding a Vagrantfile and related configuration would allow the project to ship a portable server configuration along with the code, lowering the bar new developers have to clear in order to run and contribute to unit testing.
</p>

<p>
An example Vagrantfile is available in this pull request: <a class="ext-link" href="https://github.com/jquery/jquery/pull/1410" rel="nofollow"><span class="icon">&nbsp;</span>https://github.com/jquery/jquery/pull/1410</a>
</p><p>Issue reported for jQuery 1.10.2</p>

Related: https://github.com/jquery/jquery/issues/1999

dmethvin:
@jaubourg how far did you get on the node server in New York. It would be another way of solving this problem and probably a better one than Vagrant.

jaubourg:
I didn't work on this as much as I intended/wanted but all that's needed is some express routes and rewiring to those routes in the unit tests. The hard part is to actually determine what the unit tests are actually testing so as to simplify them and not have the node server code become a mess.

Should come soon(ish).

Once I have the unit tests rewired, I'll pass the ball to @gnarf (I guess) to see how to integrate the node web server in the testing infra.

mgol:
I think this is handled by our new Karma setup. While it can be improved (AMD support, jQuery modules support, etc.) we don't need a meta-issue for that anymore.

PR #2537, opened 2015-08-12
Improved guide to testing/contributing

riichard:
- Contributing guide missed the `npm build` step when setting up the repository. 
- Added the suggestion to use PHP's built in webserver for testing, next to setting up with the LAMP. PHP is by default installed on OSX.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,6 +95,12 @@ Get in the habit of pulling in the "upstream" master to stay up to date as jQuer
 $ git pull upstream master
 ```
 
+Run the build script
+
+```bash
+$ npm run build
+```
+
 Run the Grunt tools:
 
 ```bash


nathanhammond:
Heads up: `php -S` didn't work for me when I was setting it up just a few days ago. (Mac OS X, 10.10)

riichard:
Which port did you use?

README.md
```diff
@@ -173,11 +173,15 @@ grunt watch
 Run the unit tests with a local server that supports PHP. Ensure that you run the site from the root directory, not the "test" directory. No database is required. Pre-configured php local servers are available for Windows and Mac. Here are some options:
 
 - Windows: [WAMP download](http://www.wampserver.com/en/)
-- Mac: [MAMP download](http://www.mamp.info/en/index.html)
 - Linux: [Setting up LAMP](https://www.linux.com/learn/tutorials/288158-easy-lamp-server-installation)
 - [Mongoose (most platforms)](http://code.google.com/p/mongoose/)
+- Mac: [MAMP download](http://www.mamp.info/en/index.html)
```
mzgol: Why this change?

riichard: I shouldn't have committed this change. There is nothing different compared
to before.

I initially thought about altering the Mac installation description for
Lamp, because PHP was already installed on OS X.

Then I thought it would be better just to insert a quick tip for using PHPs
build server if you're on OS X or already have OS X installed.
Because maybe some people still want to use LAMP on OS X.

markelog: I'm not sure we still recommend to use those, for Windows, sure, but for the Linux or Mac? If it is a Linux user, their definitely knows who to install php and http-server. If it is a Mac user, their already have apache and php in the system.

mzgol: They still need to be configured, even on a Mac and it's not that easy.

The best solution would be to fix #1755 but it needs some effort.

README.md
```diff
@@ -173,11 +173,15 @@ grunt watch
...
 
+If you already have PHP installed, you can use it's built in webserver to complete the tests:
 
+```
+php -S localhost:8080
```
mzgol: I've checked locally but such a server doesn't seem to work well. It failed the "core: document ready when jQuery loaded asynchronously (#13655)" test and times out/fails a lot of AJAX tests. I guess a full Apache server might be needed to make it work correctly.

gibson042: > It failed the "core: document ready when jQuery loaded asynchronously (#13655)" test and times out/fails a lot of AJAX tests.

Confirmed. Accordingly, I don't think we can accept a suggestion to use it.

mzgol:
@riichard Did you see our comments? The standalone PHP server won't work and the Mac info was just swapped so these changes should be reverted. This will get README.md back to the state it was in, the only change will be in CONTRIBUTING.md. Would you like to apply these changes? Or do you have any other suggestions?

riichard:
@mzgol I see your comments, the tests worked fine with php running on my machine but I guess there are exceptions and differences to apache. It's best to leave that README.md suggestion out then.

I would still like to apply the changes to CONTRIBUTING.md.

mzgol:
> @mzgol I see your comments, the tests worked fine with php running on my machine

Did you run the whole suite? I remember I used to have a lot of issues with the standalone server, although now I see only 4 of them (I tried on OS X 10.11 GM Candidate, PHP 5.5.27):
screen shot 2015-09-11 at 20 37 51

If my memory is fine, though, it might work worse on older versions and even then the test suite still doesn't pass completely. So I think this advice would be risky in general.

> I would still like to apply the changes to CONTRIBUTING.md.

Sure. :) Just revert the other changes and we'll land this one.

riichard:
It was a challenge for me to revert that commit from github, but I got it done. The PHP server suggestion commit is removed.

The only commit left fixes CONTRIBUTING.md's missing 'npm run build' step.

mzgol:
LGTM.

markelog:
LGTM

Issue #2437, opened 2015-06-30
The event module is listed twice in the modules dropdown on the QUnit test page


mzgol:
The event module is listed twice in the modules dropdown on the QUnit test page. This is probably because of https://github.com/jquery/jquery/blob/a44cfa00665d21c3197e25c2d63741dc15f6ffb9/test/unit/ready.js#L1.

@timmywil, any idea why it's written this way?

Screenshot: 
![screen shot 2015-06-30 at 16 18 16](https://cloud.githubusercontent.com/assets/1758366/8433049/d6dbd02e-1f43-11e5-89eb-efef456d152b.png)

timmywil:
We can just call that the "ready" module; it needed to be separated because the ready tests need to load synchronously in AMD mode.

mzgol:
> We can just call that the "ready" module

:+1:

mzgol:
No matter which one of these `event` entries I choose, I still get the same `event` tests executed; the double entry doesn't appear in AMD mode, though. How is it all wired up?

Issue #2541, opened 2015-08-16
Cache `node_modules` folder at travis build


markelog:
See http://docs.travis-ci.com/user/caching/

mzgol:
I don't think it's a good idea. This can lead to Travis succeeding when `npm install` on a fresh `master` clone will fail because it will install newer dependencies that got broken while Travis has old working ones. I'd guess most of us will also not notice it unless we have a habit of removing `node_modules` locally & reinstalling deps regularly.

OTH, `npm install` on Travis is **really** fast so this won't speed up the build significantly.

markelog:
I guess you forgetting that we have two CIs

mzgol:
> I guess you forgetting that we have two CIs

Good point. But considering that `npm install` is fast on Travis in the jQuery repository (before Promises/A+ tests were added, the whole build completed in 45 seconds), what are the reasons for introducing the cache here?

markelog:
I don't understand the question :-), there is only upside no downside

mzgol:
For me the downside is e.g. that we might miss a broken subdependency if we didn't have commits recently and someone submits a PR. Only when merged we'll notice that Jenkins crashes. Also, one day we might switch to Travis completely and then we won't notice the problem at all?

What are the (visible) upsides? Saving 10 seconds of the build isn't a real upside for me.

markelog:
> Only when merged we'll notice that Jenkins crashes.

This never happened before and chances are, it will never happen. Third party dep of the dep should influence original dep tremendously in order for us to fail. I'm not even sure that happen anywhere ever.

But even if it does, we couldn't do anything in that case, only wait, when dep of the dep would release another version, so that fail would be just noise. We couldn't even downgrade, since original dep could use that bad dep with tilde or caret since the first version.

With conjunction of this question
> Also, one day we might switch to Travis completely and then we won't notice the problem at all?

It seems you try to solve issue that doesn't exist, if that situation would be indeed dangerous, then travis caching feature should be considered dangerous, since if that would be the issue, it would be the issue not only for the node community, but for every community there is.

mzgol:
>> Only when merged we'll notice that Jenkins crashes.
>
> This never happened before and chances are, it will never happen. Third party dep of the dep should influence original dep tremendously in order for us to fail. I'm not even sure that happen anywhere ever.

[...]

PR #2542, opened 2015-08-16
Update code style, partially use new qunit interface, give test module a new name

markelog:
Fixes gh-2056, gh-2437, gh-2540, gh-2571

On gh-2540
See http://qunitjs.com/upgrade-guide-2.x/

For most of the boring work was used https://github.com/apsdehal/qunit-migrate package

However, it can't update local `qunit` helpers, plus in some places deprecated `QUnit.asyncTest` method is still used.

--

Some of these changes strongly depend on each other, others are also fixed, since i touched relevant files, so i decided to put them in one pull, i hope that is okay.


mzgol:
I'm against the "cache node_modules" part, see https://github.com/jquery/jquery/issues/2541#issuecomment-131523350.

src/css/var/rmargin.js
```diff
@@ -1,3 +1,3 @@
-define(function() {
-	return (/^margin/);
-});
+define( function() {
+	return ( /^margin/ );
```
mzgol: Are the parens still needed? I think it's no longer required by JSHint.

test/data/testinit.js
```diff
@@ -160,7 +165,9 @@ this.ajaxTest = function( title, expect, options ) {
 					if ( options.teardown ) {
 						options.teardown();
 					}
-					start();
+
+					// Make sure all events will be called before done()
+					setTimeout( done );
```
mzgol: This seems dangerous; could you explain why a single `setTimeout` is guaranteed to fix it?

mzgol:
Overall, a good job!

test/unit/ready.js
```diff
@@ -1,18 +1,18 @@
-module( "event" );
+QUnit.module( "ready" );
```
mzgol: How does the dropdown from the screenshot at #2437 look like now? Is the double `event` entry changed to `event` & `ready`? Does choosing `ready` work?

test/data/testinit.js
markelog: This is done in case test body listens to ajax events, those events that do fire after `complete` should fire right after it

src/css/var/rmargin.js
markelog: Don't know, but that is not a subject of this PR, too much already is squashed in there :-)

test/unit/ready.js
markelog: yes and yes

build/tasks/build.js
```diff
@@ -82,13 +89,15 @@ module.exports = function( grunt ) {
 			contents = contents
 				.replace( /define\(\[[^\]]*\]\)[\W\n]+$/, "" );
 		}
+
 		// AMD Name
-		if ( (amdName = grunt.option( "amd" )) != null && /^exports\/amd$/.test( name ) ) {
-			if (amdName) {
+		if ( ( amdName = grunt.option( "amd" ) ) != null && /^exports\/amd$/.test( name ) ) {
```
gibson042: Can changes like this wait for [*SpacesInsideParenthesizedExpression](https://github.com/jscs-dev/node-jscs/pull/1254)?

src/css/var/rmargin.js
mzgol: Fair enough. :)

build/tasks/build.js
markelog: The rule itself is not such a big thing, **when** (and again, thousands apologies i still didn't get around to that request) this rule will be landed, we would need to update our code style, meaning, we would need to argue and probably vote again, after that, we would need to update documentation on the site, after that we would need to add it to the `jquery` preset and after that, we would need to wait on next the minor release of jscs.

That might take some time, whereas we have our code-style in current form, on which we all agreed upon. Whereas, if we would decide on that, we could auto-format it in the same easy way i did in this pull.

timmywil: > Whereas, if we would decide on that, we could auto-format it in the same easy way i did in this pull.

That's a good point. I don't think we need to wait for that in order to land this PR.

gibson042: I wasn't advocating blocking the PR, just backing out the changes pertaining to spaces in parenthesized expressions (which I believe would shrink it considerably, but I suppose would require a temporary `requireSpacesInsideParentheses` override).

As for revoting/updating documentation/etc., a rule requiring such spaces doesn't appear in http://contribute.jquery.org/style-guide/js/#spacing or the old [.jscs.json](https://github.com/markelog/jquery/commit/b7283d738ec4c2d9d87cb16c32a3209cab567ed1#diff-ed6a67a86c4c5aa1dd5fcb0acb6f5c67), and in fact was stealth-added (from the perspective of this codebase, not its contributors) by https://github.com/jscs-dev/node-jscs/commit/0f78baf08fba75aeba4575608aaa08c384875701 in response to https://github.com/jquery/contribute.jquery.org/pull/104 (which was about requiring spaces in function calls). Since you committed it, I can just directly ask how you would have treated _expression_ parentheses if JSCS were capable of differentiating them. But I will point out that I haven't encountered one argument in favor of requiring them that didn't hinge upon style guide automation, which falls apart as soon as that process becomes sufficiently sophisticated.

markelog: > which was about requiring spaces in function calls

Well, in that discussion Mike suggested to remove all exceptions, so we wouldn't create sophisticated logic to the parentheses rules in JSCS needed to the one preset, that was somewhat the point of that proposal.

> I can just directly ask how you would have treated expression parentheses if JSCS were capable of differentiating them

I would follow the style guide, no matter of it is, i.e. `jquery` preset should represent it.

Personally, i would add more exceptions to that rule as i suggested in jquery/contribute.jquery.org#104  including the "expression parentheses"

timmywil: I'm less bothered by the extra spaces in expression parenthesis as I am by the extra space at the end of our methods (i.e. `} );`), which I don't think would require changes to jscs if we were comfortable with things like `method( arg1, { object: value })`.

mzgol: > But I will point out that I haven't encountered one argument in favor of requiring them that didn't hinge upon style guide automation

I have one - it's just simpler to remember that an opening paren containing something is always followed by a space. Simpler for us, simpler for contributors.

But as long as the tooling enforces the agreed convention, I'm fine whatever we end up with.

mzgol: >
> I'm less bothered by the extra spaces in expression parenthesis as I am by
> the extra space at the end of our methods (i.e. } );), which I don't
> think would require changes to jscs if we were comfortable with things like method(
> arg1, { object: value }).
>

I like those spaces for consistency reasons. If the majority doesn't want
them, I'm fine with this exception (although I prefer not to have too many
exceptions as @mikesherov suggested) but it'd be good to have it checked or
we'll have half of the code in one style & half in the other.

timmywil: For now, I'm okay with leaving it as-is. If we agree to change it, we'll do another PR.

PR #2547, opened 2015-08-19
Event: Don't try to read data on a non-element on IE8 form attribute

adrianolek:
Rebased #2536 on `compat` branch & added a test.

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -896,8 +896,8 @@ if ( !support.submit ) {
 					form = jQuery.nodeName( elem, "input" ) || jQuery.nodeName( elem, "button" ) ?
 						elem.form :
 						undefined;
-
-				if ( form && !jQuery._data( form, "submit" ) ) {
+				// Don't try to read data on non-element (#14126)
+				if ( form && form.nodeType === 1 && !jQuery._data( form, "submit" ) ) {
 					jQuery.event.add( form, "submit._submit", function( event ) {
 						event._submitBubble = true;
 					});
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -1416,6 +1416,26 @@ test("Submit event can be stopped (#11049)", function() {
 	form.remove();
 });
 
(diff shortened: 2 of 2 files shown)


mzgol:
Please read https://github.com/jquery/jquery/issues/2332#issuecomment-103718392 and https://github.com/jquery/jquery/issues/2332#issuecomment-103911627. This approach will just add a silent failure and not fix the underlying issue.

adrianolek:
Yeah, that's what I said in the #2536 description. HTML5 form attributes don't work on IE8, and it's documented & expected. What's not expected is jQuery crashing (or is it?).
Do you really want to go the propHook way? If so, how much of the form attribute spec would you like to see implemented? Because right now replacing `elem.form` with `jQuery.prop( elem, "form" )` would only fix the IE8 delegated submit events, which would make it even more difficult to polyfill. IMO it should be all or nothing.
Or maybe you see the propHook working differently (eg. acting like the form attribute isn't there, so returning the form when element is inside it, and null when there's no parent form)?

gibson042:
> If so, how much of the form attribute spec would you like to see implemented?

None of it within jQuery proper. We'll just be respecting any consumer-defined hook controlling the output of `$elem.prop("form")`. Unless such a hook is defined, though, jQuery behavior will remain unchanged.

> Because right now replacing `elem.form` with `jQuery.prop( elem, "form" )` would only fix the IE8 delegated submit events

Which is literally the only place where we retrieve `elem.form`.

> which would make it even more difficult to polyfill.

How so?

adrianolek:
> How so?
If we fix the form attribute there (so it works according to HTML5 spec), the polyfill would have to take it into account and not try to polyfill IE8 delegated submit events.

But as I understand, literally the only change you want to do is replacing `elem.form` with `jQuery.prop( elem, "form" )`, right?

gibson042:
Correct. Then library consumers will be able to define `jQuery.propHooks.form` as part of their workarounds and/or polyfills.

PR #2551, opened 2015-08-22
Check before setting style property

morris:
According to [this jsperf](http://jsperf.com/check-style-attributes-before-setting-or-just-constant-), checking if a style property has changed before setting it gives a speedup of 40% (Chrome) to 80% (Firefox) if the value is the same. This check is also hard to to in userland because the user would have to replicate camel case, vendor prefixes, hooks etc. which are applied inside `jQuery.style`.

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -252,7 +252,9 @@ jQuery.extend({
 			if ( !hooks || !("set" in hooks) ||
 				(value = hooks.set( elem, value, extra )) !== undefined ) {
 
-				if ( style[ name ] !== value ) style[ name ] = value;
+				if ( style[ name ] !== value ) {
+					style[ name ] = value;
+				}
 			}
 
 		} else {
--- a/src/css.js
+++ b/src/css.js
@@ -252,7 +252,7 @@ jQuery.extend({
 			if ( !hooks || !("set" in hooks) ||
 				(value = hooks.set( elem, value, extra )) !== undefined ) {
 
-				style[ name ] = value;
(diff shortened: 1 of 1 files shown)


gibson042:
This seems unobtrusive, and I have no objection in principle, but can you provide a real-world scenario where this actually affects broader _application_ performance?

dmethvin:
Is the most common case that `.css()` is trying to set the same property value over and over again? It seems like you'd want to test and bias towards the case where the current value is not the same, because there the added `if` is performing extra unnecessary work. 

The current code runs about 2 million operations per second or 0.5 microseconds per operation. When jsperf starts reporting numbers like that on microbenchmarks you really need to be careful about interpreting them, and wonder whether it matters at all.

The worst case scenario would seem to be animations, and for animating a property you'd call this at most 60 times a second for a total cost of 30 microseconds, or 0.2 percent of your 16 millisecond frame. The style and layout work provoked in the browser by most CSS changes will often be one or two orders of magnitude bigger. I wouldn't think it's common in animations to set the same property over and over, but when it does happen we're not gaining much from avoiding it and we don't yet know the cost of checking for it when it's not common.

morris:
Valid points, I really overinterpreted the benchmark I posted. I ran another benchmark which sometimes randomly sets a different value (something like Math.random() < 0.01 ? 'red' : 'blue') and for most properties the check I wanted to introduce only pays off if the value is unchanged 99.9% of the time which is very unlikely; otherwise the regular css() is faster. It also seems browsers handle the case of setting the same value again pretty well. Sorry for taking your time!

PR #2553, opened 2015-08-23
Data: combine register and cache methods

jbedard:
This also avoids the `Data.accepts` call if the object already has data.

```
-115    -32 dist/jquery.js
 -74    -28 dist/jquery.min.js
```

### DIFF
--- a/src/data/Data.js
+++ b/src/data/Data.js
@@ -12,44 +12,38 @@ Data.uid = 1;
 
 Data.prototype = {
 
-	register: function( owner ) {
-		var value = {};
-
-		// If it is a node unlikely to be stringify-ed or looped over
-		// use plain assignment
-		if ( owner.nodeType ) {
-			owner[ this.expando ] = value;
-
-		// Otherwise secure it in a non-enumerable property
-		// configurable must be true to allow the property to be
-		// deleted when data is removed
-		} else {
-			Object.defineProperty( owner, this.expando, {
-				value: value,
-				configurable: true
-			} );
-		}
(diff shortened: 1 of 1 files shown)


mzgol:
cc @rwaldron

dover6114:
Yes

Dover

dover6114:
really help

Dover

src/data/Data.js
```diff
@@ -13,45 +13,38 @@ Data.accepts = jQuery.acceptData;
...
-	},
 	cache: function( owner ) {
+		// Check if the owner object already has a cache
+		var value = owner[ this.expando ];
+		if ( value ) {
+			return value;
+		}
+
+		// If not, create one
+		value = {};
```
mzgol: Is this shorter (after gzip) than the version with:
```js
		if ( !Data.accepts( owner ) ) {
			return {};
		}
```
?

jbedard: I believe it was but I'll have to double check. I think I tried a few different methods of doing this and found this to be smallest (after gzip).

jbedard:
@mzgol doing it this way seems to reduce the size the most. Having `return` statements seems to always add a few bytes.

I've updated the PR to save one more byte by changing the `if (value) return value` to `if (!value) { ..create.. }`.  This does nest everything into another `if` statement though, I guess it's debatable if that is better or worse then return statements in the middle of a method...

timmywil:
I usually prefer the early return stylistically. To me, it's worth the extra byte(s).

mzgol:
Landed, thanks!

PR #2554, opened 2015-08-23
Data: do not create data cache when fetching single property

jbedard:
This makes master like compat. The test can probably be put into compat as is.

### DIFF
--- a/test/unit/data.js
+++ b/test/unit/data.js
@@ -874,3 +874,19 @@ test( "Check that the expando is removed when there's no more data", function() 
 		}
 	}
 });
+
+test( ".data(prop) does not create expando", function() {
+	expect( 2 );
+
+	var key,
+		div = jQuery( "<div/>" );
+
+	div.data("foo");
+	equal( false, jQuery.hasData( div[0] ) );
+	// Make sure no expando has been added
+	for ( key in div[ 0 ] ) {
+		if ( /^jQuery/.test( key ) ) {
+			ok( false, "Expando was created on access" );
+		}
+	}
+});
--- a/src/data/Data.js
+++ b/src/data/Data.js
@@ -73,13 +73,11 @@ Data.prototype = {
 		return cache;
 	},
 	get: function( owner, key ) {
-		var cache = this.cache( owner );
-
 		return key === undefined ?
-			cache :
+			this.cache( owner ) :
 
 			// Always use camelCase key (gh-2257)
-			cache[ jQuery.camelCase( key ) ];
+			owner[ this.expando ] && owner[ this.expando ][ jQuery.camelCase( key ) ];
 	},
 	access: function( owner, key, value ) {
 


test/unit/data.js
```diff
@@ -874,3 +874,18 @@ test( "Check that the expando is removed when there's no more data", function()
...
+
+test( ".data(prop) does not create expando", function() {
+	expect( 2 );
+
+	var div = jQuery( "<div/>" );
+
+	div.data("foo");
+	equal( false, jQuery.hasData( div[0] ) );
+	// Make sure no expando has been added
+	for ( key in div[ 0 ] ) {
```
timmywil: `var key,` at the top.

timmywil:
I like this.

test/unit/data.js
markelog: > at the top.

We have an example of something like that in the guide https://contribute.jquery.org/style-guide/js/#spacing, but strictly that is not a requirement anymore

test/unit/data.js
```diff
@@ -874,3 +874,18 @@ test( "Check that the expando is removed when there's no more data", function()
...
+		if ( /^jQuery/.test( key ) ) {
+			ok( false, "Expando was created on access" );
+		}
+	}
+});
```
markelog: Additional newline is needed

test/unit/data.js
timmywil: I'm not sure what you mean. This is failing because `key` is missing a declaration.

jbedard:
Oops.  Should be fixed now...

test/unit/data.js
markelog: i mean you could do 
```js
for ( var key in div[ 0 ] )
```

mzgol:
```js
for ( var key in div[ 0 ] )
```

I really don't like this form. :) `var` is function scoped and we write it outside of any blocks inside a function to not suggest it behaves otherwise.

(i.e. this should be done exactly as this PR does it)

markelog: > I really don't like this form. :)

This is why we have a code-style, i don't like a lot of things in current code base, as i'm sure we all do, but we choose to follow it.

> (i.e. this should be done exactly as this PR does it)

No, test should trigger the error with minimum overhead, any other requirements are subjective

mzgol: > This is why we have a code-style

We can codify this, sure, if JSCS can do it. It's just how it is in the whole code base so it's a de-facto standard anyway.

markelog: > We can codify this, sure, if JSCS can do it. 

It totally can, it was like that for a while, but code-style changed and rule was removed from the preset. It is like that de-facto because code-style force it, but now it doesn't so we should ask contributor to do something that is not documented

mzgol:
Landed, thanks. I had to modify the assertion count as the test should invoke only one assertion.

Issue #2555, opened 2015-08-23
Don't expose jQuery.acceptData


jbedard:
Similar to the other "don't expose" cases. Is there any reason for this to be exposed and undocumented?

markelog:
Hm, i guess one might argue that this function could be useful to plugin writers, but it seems we just should stop exposing it instead

timmywil:
@arschmitz @scottgonzalez Do you use this in mobile/UI?

jbedard:
Another option is to still expose it but use the internal var instead of `jQuery.acceptData` to save some bytes and not support overriding it. Today it is also [copied into Data](https://github.com/jquery/jquery/blob/master/src/data/Data.js#L12), so overriding it would be buggy anyway.

arschmitz:
@timmywil this is not used by either ui or mobile

markelog:
> @timmywil this is not used by either ui or mobile

It seems we're good of stop exposing it, @jbedard would you like to send us a pr?

mzgol:
> Another option is to still expose it but use the internal var instead of `jQuery.acceptData` to save some bytes and not support overriding it.

By "exposing" we mean putting it on the `jQuery` object so that it's accessible in the compiled version. If you just use AMD modules you can get to a lot of internal stuff already, we just don't guarantee the file hierarchy doesn't change: it might do even in a patch release.

PR #2557, opened 2015-08-25
Data: Don't expose jQuery.acceptData

jbedard:
jQuery.acceptData is an undocumented internal API that shouldn't be exposed.

Fixes gh-2555

The tests are little ugly but the rest is nice...
```
   -67    -11 dist/jquery.js
   -65    -19 dist/jquery.min.js
```

### DIFF
--- a/src/data/Data.js
+++ b/src/data/Data.js
@@ -2,14 +2,13 @@ define([
 	"../core",
 	"../var/rnotwhite",
 	"./accepts"
-], function( jQuery, rnotwhite ) {
+], function( jQuery, rnotwhite, acceptData ) {
 
 function Data() {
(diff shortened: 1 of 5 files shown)


mzgol:
LGTM, thanks for the PR!

We need to wait for #2542 to land first (it should happen soon). And then rebase all ready PRs, ugh. This will not be nice. ;)

mzgol:
Landed in 224271982eb9cd351d7db1b38c740b4e927e6f97 (on `master`) and bec2ba231288ec4abba63c3a5ca8ad26715a318b (on `compat`), thanks!

I've modified the structure a little as the new `acceptData` was more in line with our `var`-modules.

PR #2568, opened 2015-09-02
Manipulation: Switch rnoInnerhtml to a version more performant in IE

SeanHenderson:
IE versions greater then 9 do not handle the old regular expression well
with large html content.

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -27,7 +27,10 @@ define([
 
 var
 	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([\w:-]+)[^>]*)\/>/gi,
-	rnoInnerhtml = /<(?:script|style|link)/i,
+	// Support: IE 10-11, Edge 10240+
+	// In IE/Edge using regex groups here causes severe slowdowns.
+	// See https://connect.microsoft.com/IE/feedback/details/1736512/
+	rnoInnerhtml = /<script|<style|<link/i,
 	// checked="checked" or checked
 	rchecked = /checked\s*(?:[^=]|=\s*.checked.)/i,
 	rscriptTypeMasked = /^true\/(.*)/,
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,12 +4,16 @@ module.exports = function( grunt ) {
 	function readOptionalJSON( filepath ) {
 		var data = {};
 		try {
-			data = grunt.file.readJSON( filepath );
+			data = JSON.parse( stripJSONComments(
+				fs.readFileSync( filepath, { encoding: "utf8" } )
+			) );
 		} catch ( e ) {}
 		return data;
 	}
 
-	var gzip = require( "gzip-js" ),
+	var fs = require( "fs" ),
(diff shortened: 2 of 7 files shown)


mzgol:
We need you to sign our [CLA](https://contribute.jquery.org/CLA/) before accepting a PR.

SeanHenderson:
done

mzgol:
Did you invoke `npm install`? The Git commit hook would tell you the commit message is incorrect, it's missing a component `Manipulation`. The format for marking the bug closed is also different, please read https://contribute.jquery.org/commits-and-pull-requests/#commit-guidelines.

Otherwise, the code looks good.

SeanHenderson:
I don't have node installed. I have updated the commit message as requested.

mzgol:
`Component:Manipulation` should be `Manipulation: COMMIT_MESSAGE`, for example:
```
Manipulation: Switch rnoInnerhtml to a version more performant in IE
```

Otherwise, it looks good.

src/manipulation.js
```diff
@@ -27,7 +27,7 @@ define([
 
 var
 	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([\w:-]+)[^>]*)\/>/gi,
-	rnoInnerhtml = /<(?:script|style|link)/i,
+	rnoInnerhtml = /<script|<style|<link/i,
```
markelog: Hm, i wonder, do we have similar regexps? Would be helpful to have a comment, which would point to the microsoft issue and perf suite

SeanHenderson:
I have added additional detail to the commit message.  Similar regular expressions exist but I didn't find one that will be used on large strings that might have the same kind of impact.

markelog:
> I have added additional detail to the commit message.

Could you add it beside the regexp itself? We might forget that it is slow in Edge and remove it sometime in the future.

mzgol:
> Could you add it beside the regexp itself? Otherwise we might forget that it is slow in Edge and remove it sometime in the future.

@SeanHenderson For reference, it should look something like that:
```js
	// Support: IE 10-11, Edge 10240+
	// In IE/Edge using regex groups here causes severe slowdowns.
	// See https://connect.microsoft.com/IE/feedback/details/1736512/
	rnoInnerhtml = /<script|<style|<link/i,
```

> Similar regular expressions exist but I didn't find one that will be used on large strings that might have the same kind of impact.

I'm not sure about that. The above `rxhtmlTag` regex is used in [jQuery.htmlPrefilter](https://github.com/jquery/jquery/blob/5fe76c663f8a4986af62edb434a1708c006d0b21/src/manipulation.js#L222) which, in turn is used on most inputs to [jQuery.fn.html](https://github.com/jquery/jquery/blob/5fe76c663f8a4986af62edb434a1708c006d0b21/src/manipulation.js#L392). It seems this may be causing exactly the same slowdowns (although it would be good to perf-test it to make sure).

Maybe there are others?

SeanHenderson:
Sigh, I work with TFS still getting used to GIT.  I have created a new branch and a new pull request so I am closing off this one.

Issue #2563, opened 2015-09-01
html() uses a regular expression that performs badly in IE > 9


SeanHenderson:
rnoInnerhtml = /<(?:script|style|link)/i

The above causes significant performance issues in IE 10 IE 11 and EDGE when inserting large amounts of HTML using the html function.

It can be re-written as the following:
 rnoInnerhtml = /<script|<style|<link/i,

The above performs at the same level in other browsers and exponentially faster in IE .
My tests show for two columns for 4096 rows the difference is 18 seconds versus 5 milliseconds (IE 11). 

http://jsfiddle.net/Lwa0t5rp/3/

Note when testing you the first test will always be slower than the second in chrome / firefox

Ticket logged at Microsoft however i think jQuery should be fixed anyways.
https://connect.microsoft.com/IE/feedback/details/1736512/regular-expressions-are-slower-in-edge-than-in-ie8

mzgol:
Thanks for the report! I confirmed this is very slow in Edge (build 10240). Would you like to submit a PR to take the credit for the perf fix?

SeanHenderson:
https://github.com/jquery/jquery/pull/2568  Not sure if this is done correctly / if this should also go in other branches as well

PR #2574, opened 2015-09-03
Manipulation: Switch rnoInnerhtml to a version more performant in IE

SeanHenderson:
IE versions greater than 9 do not handle the old regular expression well with large html content
This is due to the use of a non-capturing group after a very common html character (<).

Test suite: http://jsfiddle.net/Lwa0t5rp/3/
Microsoft bug: https://connect.microsoft.com/IE/feedback/details/1736512/regular-expressions-are-slower-in-edge-than-in-ie8

Fixes gh-2563
Closes gh-2568

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -27,7 +27,10 @@ define([
 
 var
 	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([\w:-]+)[^>]*)\/>/gi,
-	rnoInnerhtml = /<(?:script|style|link)/i,
+	// Support: IE 10-11, Edge 10240+
+	// In IE/Edge using regex groups here causes severe slowdowns.
+	// See https://connect.microsoft.com/IE/feedback/details/1736512/
+	rnoInnerhtml = /<script|<style|<link/i,
 	// checked="checked" or checked
 	rchecked = /checked\s*(?:[^=]|=\s*.checked.)/i,
 	rscriptTypeMasked = /^true\/(.*)/,


mzgol:
@SeanHenderson A tip: you should never commit anything to `master` yourself, this makes it very hard for you to pull upstream changes if this PR needs to be rebased over latest code; or if you wanted to create a separate, parallel PR. Always create a dedicated branch just for a PR and have `master` always mirror the official `master`. https://contribute.jquery.org/commits-and-pull-requests/ has more info.

This PR can stay based on `master` if you prefer this way, this is mainly an advise for the future.

mzgol:
(BTW, if you have JSHint errors on a freshly cloned repository, don't care about it too much - JSHint has broken our build so it's red temporarily...)

SeanHenderson:
I had a look created a branch tried to reset my master but failed. 
(warning: refname 'jquery/master' is ambiguous.)  

If it will not effect me in future I rather have this PR go in as is.  
The next changes will definitely be made as you have suggested.

mzgol:
@SeanHenderson Did you read my comment at https://github.com/jquery/jquery/pull/2568#issuecomment-137452926? I think some other regexes might be affected as well.

SeanHenderson:
Hi I did read your comment and had a look at the performance of that regular expression.  I saw it does have an issue but I do not have a suitable replacement at the moment.  I traced the changes that introduced the optimised code path and can see it has changed quite bit over time.  At one point regular expression was written similar to the way I have it now.  I would prefer to create a separate PR for any changes to other regular expressions.

mzgol:
@SeanHenderson OK, I landed this PR on both branches, thanks! If you had time to look at these other regexes, that'd be great.

PR #2575, opened 2015-09-06
Event: Use form prop so that a propHook can be used

adrianolek:
As @gibson042 suggested, this enables 3rd party polyfills/workarounds to fix the IE8 delegated submit issues when HTML5 form attribute is used (see [#2332 (comment)](https://github.com/jquery/jquery/pull/2332#issuecomment-103911627), [#2547 (comment)](https://github.com/jquery/jquery/pull/2547#issuecomment-136599348))

Fixes gh-2332

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -894,7 +894,12 @@ if ( !support.submit ) {
 				// Node name check avoids a VML-related crash in IE (#9807)
 				var elem = e.target,
 					form = jQuery.nodeName( elem, "input" ) || jQuery.nodeName( elem, "button" ) ?
-						elem.form :
+
+						// Support: IE <=8
+						// We use jQuery.prop instead of elem.form
+						// to allow fixing the IE8 delegated submit issue (gh-2332)
+						// by 3rd party polyfills/workarounds.
+						jQuery.prop( elem, "form" ) :
 						undefined;
 
 				if ( form && !jQuery._data( form, "submit" ) ) {


mzgol:
LGTM.

timmywil:
Me too!

src/event.js
```diff
@@ -894,7 +894,7 @@ if ( !support.submit ) {
 				// Node name check avoids a VML-related crash in IE (#9807)
 				var elem = e.target,
 					form = jQuery.nodeName( elem, "input" ) || jQuery.nodeName( elem, "button" ) ?
-						elem.form :
+						jQuery.prop( elem, "form" ) :
```
gibson042: To prevent premature reversion, I'd like to see a comment explaining why this line uses `jQuery.prop` instead of `elem.form` (e.g., like [this](https://github.com/jquery/jquery/commit/ae215fdcf843862dbddbdfd68117ac2926e35a42#diff-89c55bb1a12d31cb4eb5d007633dc33bR14)). It should probably reference gh-2332, and be preceded by `Support: IE <=8`.

adrianolek:
Question about cs; should I add the comment before [`var elem`](https://github.com/adrianolek/jquery/blob/ie8-form-attr/src/event.js#L895) or leave it where it is now, but add blank line (for [jscs](https://travis-ci.org/jquery/jquery/jobs/79728588#L249))

mzgol:
@adrianolek 
> Question about cs; should I add the comment before [`var elem`](https://github.com/adrianolek/jquery/blob/ie8-form-attr/src/event.js#L895) or leave it where it is now, but add blank line (for [jscs](https://travis-ci.org/jquery/jquery/jobs/79728588#L249))?

The latter. It should be as close to where it applies as possible.

mzgol:
Landed at ead83b9c8a6b751f31f0197ff0364e95a1ca24cd, thanks!

PR #2581, opened 2015-09-08
Ajax:Attributes:CSS:Manipulation: Reduce Android 2.3 support

mzgol:
Drop non-critical workarounds for Android 2.3.

Fixes gh-2483
Fixes gh-2505

### DIFF
--- a/src/ajax/parseJSON.js
+++ b/src/ajax/parseJSON.js
@@ -2,11 +2,7 @@ define( [
 	"../core"
 ], function( jQuery ) {
 
-// Support: Android 2.3
-// Workaround failure to string-cast null input
-jQuery.parseJSON = function( data ) {
-	return JSON.parse( data + "" );
-};
+jQuery.parseJSON = JSON.parse;
 
 return jQuery.parseJSON;
 
--- a/src/attributes/support.js
+++ b/src/attributes/support.js
@@ -18,11 +18,6 @@ define( [
 	// Must access selectedIndex to make default options select
 	support.optSelected = opt.selected;
 
-	// Support: Android<=2.3
-	// Options inside disabled selects are incorrectly marked as disabled
-	select.disabled = true;
-	support.optDisabled = !opt.disabled;
-
 	// Support: IE<=11+
 	// An input loses its value after becoming a radio
 	input = document.createElement( "input" );
--- a/src/attributes/val.js
+++ b/src/attributes/val.js
@@ -105,8 +105,7 @@ jQuery.extend( {
 					if ( ( option.selected || i === index ) &&
 
 							// Don't return options that are disabled or in a disabled optgroup
-							( support.optDisabled ?
-								!option.disabled : option.getAttribute( "disabled" ) === null ) &&
+							!option.disabled &&
 							( !option.parentNode.disabled ||
 								!jQuery.nodeName( option.parentNode, "optgroup" ) ) ) {
 
--- a/src/css.js
+++ b/src/css.js
@@ -351,16 +351,6 @@ jQuery.each( [ "height", "width" ], function( i, name ) {
 	};
 } );
 
-// Support: Android 2.3
-jQuery.cssHooks.marginRight = addGetHookIf( support.reliableMarginRight,
-	function( elem, computed ) {
-		if ( computed ) {
-			return swap( elem, { "display": "inline-block" },
-				curCSS, [ elem, "marginRight" ] );
-		}
-	}
-);
-
 // These hooks are used by animate to expand properties
 jQuery.each( {
 	margin: "",
--- a/src/css/support.js
+++ b/src/css/support.js
@@ -29,10 +29,7 @@ define( [
 	// so they're executed at the same time to save the second computation.
 	function computeStyleTests() {
 		div.style.cssText =
-
-			// Support: Android 2.3
-			// Vendor-prefix box-sizing
-			"-webkit-box-sizing:border-box;box-sizing:border-box;" +
+			"box-sizing:border-box;" +
 			"display:block;position:absolute;" +
 			"margin:0;margin-top:1%;margin-right:50%;" +
 			"border:1px;padding:1px;" +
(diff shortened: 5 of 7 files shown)


markelog:
LGTM

Although i'm pretty said i couldn't come up with some @jquerybot joke :-(

src/ajax/parseJSON.js
```diff
@@ -2,11 +2,7 @@ define( [
 	"../core"
 ], function( jQuery ) {
 
-// Support: Android 2.3
-// Workaround failure to string-cast null input
-jQuery.parseJSON = function( data ) {
-	return JSON.parse( data + "" );
-};
+jQuery.parseJSON = JSON.parse;
```
markelog: Yey!

markelog:
What sup with - 
https://github.com/jquery/jquery/blob/0e982433eb94391b3e9f6838d9b8fbf9bb31abf9/src/effects.js#L294
https://github.com/jquery/jquery/blob/0e982433eb94391b3e9f6838d9b8fbf9bb31abf9/src/event.js#L728
https://github.com/jquery/jquery/blob/0e982433eb94391b3e9f6838d9b8fbf9bb31abf9/src/css/support.js#L33
https://github.com/jquery/jquery/blob/0e982433eb94391b3e9f6838d9b8fbf9bb31abf9/src/core.js#L253

This [comment](https://github.com/jquery/jquery/blob/0e982433eb94391b3e9f6838d9b8fbf9bb31abf9/Gruntfile.js#L141) should be changed.

I assume you didn't touch tests to lessens conflifcs with compat.

Also, you indicated in the commit that this is `Core` module, whereas you changing more then that.

mzgol:
> https://github.com/jquery/jquery/blob/0e982433eb94391b3e9f6838d9b8fbf9bb31abf9/src/effects.js#L294

A critical crash so I left it in.

> https://github.com/jquery/jquery/blob/0e982433eb94391b3e9f6838d9b8fbf9bb31abf9/src/event.js#L728

This is required so that preventing default works on Android 2.3 - a pretty basic thing.

> https://github.com/jquery/jquery/blob/0e982433eb94391b3e9f6838d9b8fbf9bb31abf9/src/css/support.js#L33

Dropping this would mean nothing box-sizing related works in Android 2.3. That would be like dropping support completely IMO.

> https://github.com/jquery/jquery/blob/0e982433eb94391b3e9f6838d9b8fbf9bb31abf9/src/core.js#L253

Without this workaround even the `basic` tests don't pass.

> This [comment](https://github.com/jquery/jquery/blob/0e982433eb94391b3e9f6838d9b8fbf9bb31abf9/Gruntfile.js#L141) should be changed.

Why? And to what?

> I assume you didn't touch tests to lessen conflicts with compat.

Yep.

> Also, you indicated in the commit that this is Core module, whereas you changing more then that.

What do you suggest to put as a component then?

markelog:
> A critical crash so I left it in.

Can we update the comments to indicate partial support?

> This is required so that preventing default works on Android 2.3 - a pretty basic thing.
...
Dropping this would mean nothing box-sizing related works in Android 2.3. That would be like dropping support completely IMO.

I'd say if is not in `basic` tests it should be removed or putted in `basic` tests

> Why? And to what?

nvm

> What do you suggest to put as a component then?

List them all

mzgol:
> Can we update all support comments to indicate partial support?

Propose something. :) For me what's currently there sounds enough but if you have a better suggestion...

>> What do you suggest to put as a component then?
>
> List them all

`Ajax:Attributes:CSS:Manipulation: Reduce Android 2.3 support`?

>> This is required so that preventing default works on Android 2.3 - a pretty basic thing.
...
Dropping this would mean nothing box-sizing related works in Android 2.3. That would be like dropping support completely IMO.
>
> I'd say if is not in `basic` tests it should be removed or putted in `basic` tests

Good idea. I'll check if it's already covered and if it's not, I'll add one test.

markelog:
> Propose something. :) For me what's currently there sounds enough but if you have a better suggestion...

`Support:` to `Partial support:`? I think if ppl would check out code like that might get an impression we support it all the way

> Ajax:Attributes:CSS:Manipulation: Reduce Android 2.3 support?

Yeah

mzgol:
>> Propose something. :) For me what's currently there sounds enough but if you have a better suggestion...
>
> `Support:` to `Partial support:`? I think if ppl would check out code like that might get an impression we support it all the way

People could already get that impression as we have a lot of code supporting IE6-8 in jQuery (main line); they mostly come from Sizzle. Also, we keep things like `Support: IE 8-9 only` even when we don't support IE 8; just to not lose information. So it already doesn't mean that we support a particular browser if it appears under `Support:`.

Also, we have support pragmas for PhantomJS, we've had one for jsdom - we don't fully support any of them.

This would IMO needlessly complicate https://github.com/jquery/contribute.jquery.org/issues/95. The only ultimate truth about our browser support is https://jquery.com/browser-support/, you can't get that from comments in the code.

markelog:
> People could already get that impression as we have a lot of code supporting IE6-8 in jQuery (main line); they mostly come from Sizzle.

I would think devs would check out the github

> Also, we keep things like Support: IE 8-9 only

Where?

> Also, we have support pragmas for PhantomJS, we've had one for jsdom - we don't fully support any of them.

I think we should.

If something is one way it doesn't mean we can't do better.

> This would IMO needlessly complicate jquery/contribute.jquery.org#95.

How so?

> The only ultimate truth about our browser support is https://jquery.com/browser-support/

Thats true, not sure how is this related though

mzgol:
>> Also, we keep things like Support: IE 8-9 only
>
> Where?

I don't have a link but it was decided a while ago that we don't drop info in which browsers a particular bug existed when we drop the browser. @gibson042 remembers for sure as he advocated for it fiercelessly.

But search for `IE8-` on https://code.jquery.com/jquery-git.js, you'll find something (and not all from Sizzle).

mzgol:
To sum up, the following:
```js
// Support: ABrowser 9-23 only, BBrowser 11-32+
```
means that the following code supports the mentioned browsers/environments, not that jQuery itself does. Which means comments like `// Support: IE8-10 only` are fine.

markelog:
> fiercelessly

i guess you mean "ferociously"? I'm not sure if it was "decided", although i don't have strong opinion about it.

I guess there is no arguing that comments like `// Support: IE8-10 only` might be confusing, yeah, we can left it as is, we probably will, it would be good though, to make clear our intention without distortion, i was proposing to build solution together not to ferociously challenge it.

> But search for IE8- on https://code.jquery.com/jquery-git.js, you'll find something (and not all from Sizzle).

Not sure how is this related.

mzgol:
>> But search for IE8- on https://code.jquery.com/jquery-git.js, you'll find something (and not all from Sizzle).
>
> Not sure how is this related.

You asked where we keep comments like `Support: IE 8-9 only` in code that doesn't support IE 8 and I showed you where.

markelog:
Yeah, i asked the github link

mzgol:
If it's on https://code.jquery.com/jquery-git.js and it's not in Sizzle then it's on GitHub... Just search for it.

mzgol:
@markelog I was wrong about `box-sizing`: it's only needed for one support test and I've already removed another support test meant for Android 2.3. I don't think we need to keep it (I've only left it in tests to reduce `master`/`compat` differences).

Issue #2432, opened 2015-06-26
Inadequate/dangerous jQuery behavior for 3rd party text/javascript responses 


homakov:
Because of this https://github.com/jquery/jquery/blob/250a1990baa571de60325ab2c52eabb399c4cf9e/src/ajax/script.js#L18 every text/javascript response gets executed. Even if we made a request to another service. CORS was created exactly to prevent this kind of behavior in JSONP (arbitrary code execution).

So when we do $.get('http://weather.com/sf-weather') or like in Rails' jquery_ujs a form is being sent automatically, the attacker can respond us with text/javascript and execute arbitrary code on our origin.  Demo `$.get('http://sakurity.com/jqueryxss')`

The fix is to not execute responses from 3rd party origins by default and make it an option. Don't know who to cc to discuss it.

P.S. I would switch it off for same origin either, because using subtle redirect_to saving tricks we can redirect user to local JSONP endpoint and still get an XSS but those are much more sophisticated vectors.

dmethvin:
There would be some pretty significant backcompat implications here.  $.get is just a thon wrapper around $.ajax. If the "intelligent guess" logic is a problem the caller can use $.ajax with an explicit "text" data type and do the conversion manually/explicitly.

homakov:
@dmethvin  I generally agree that it's developer's responsibility to use $.ajax properly with all options to control the response processing. However if we google "jquery cors request" good half of the results doesn't use dataType setting. 

Anyway if it seems hard to monkey patch it for cross domain requests, I'm fine with it.

dmethvin:
I should have made it clear that I *really* dislike the "intelligent guess" logic too, for this very reason. If it was taken out and the caller had to declare their data type (or explicitly tell us to guess?) the hole would close. But I suspect it would break a ton of the world's sites. What do others think?

homakov:
> But I suspect it would break a ton of the world's sites

I don't see how it can break something because the fix is only for cross domain requests and doesn't use dataType: 'javascript' and expecting javascript (not jsonp). Why would a website load JS from other domains via $.ajax? This is really weird and I feel it's safe to patch.

cuidingfeng:
https://github.com/cuidingfeng/jsplus

timmywil:
@homakov Thanks for opening an issue. Would you be comfortable opening a PR with the proposed changes? I think we could evaluate this change more thoroughly if we could see some code.

homakov:
[...]

PR #2588, opened 2015-09-10
Ajax: Mitigate possible XSS vulnerability

markelog:
Fixes gh-2432

### DIFF
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -71,6 +71,54 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - do not execute js (crossOrigin)", 2, function( assert ) {
+		return {
+			create: function( options ) {
+				options.crossDomain = true;
+				return jQuery.ajax( url( "data/script.php?header=ecma" ), options );
+			},
+			success: function() {
+				assert.ok( true, "success" );
+			},
+			complete: function() {
+				assert.ok( true, "complete" );
+			}
+		};
+	} );
+
+	ajaxTest( "jQuery.ajax() - execute js for crossOrigin when dataType option is provided", 3,
+		function( assert ) {
+			return {
+				create: function( options ) {
+					options.crossDomain = true;
+					options.dataType = "script";
+					return jQuery.ajax( url( "data/script.php?header=ecma" ), options );
+				},
+				success: function() {
+					assert.ok( true, "success" );
+				},
+				complete: function() {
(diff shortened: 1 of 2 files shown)


markelog:
@timmywil remind me please, did we decided to do it in this or next version? In other words, should we close or land it?

mzgol:
I think we [decided to land it](http://irc.jquery.org/%23jquery-meeting/default_%23jquery-meeting_20150921.log.html#t09:14:49). :)

jaubourg:
Still not a fan of this hack. It looks like old ajax code with tests and special cases hidden deep into the code.

mzgol:
@jaubourg You're thinking from the code perspective, I (others mostly agreed IIRC) just think it's really bad that `$.get(SOME_URL)` can execute code, I think most people doesn't expect it. Methods should be as secure by default as they can be.

jaubourg:
I'm thinking about code perspective because there are better, more consistent ways to handle this than hack into the conversion logic like that. Like providing the options object to converters so that they can handle special cases properly and report sensible errors.

Now, from a behavior point of view, this will effectively prevent valid cross-domain script execution because the xhr transport will provide a text response and will need to use the converter whether the dataType was provided explicitely or not.

mzgol:
OK, I had an impression that you don't want to prevent `$.get(CROSS_DOMAIN_URL)` from executing script and not that you're just against this specific way of achieving that.

jaubourg:
OK, so I've been thinking about this and the best way to handle this is to inhibit auto detection of the script dataType in the context of a cross-domain request which can be achieved by adding the following prefilter before the existing one in src/ajax/script:

```js
// Prevent auto-execution of scripts when no explicit dataType was provided
jQuery.ajaxPrefilter( function( s ) {
	if ( s.crossDomain ) {
		s.contents.script = false;
	}
} );
```

I still think this should be done userland but, anyway, that's the way to achieve this properly.

timmywil:
Thanks @jaubourg. I think that's a good suggestion.

reedloden:
Please see https://github.com/jquery/jquery/issues/2432#issuecomment-199098662

test/unit/ajax.js
```diff
@@ -71,6 +71,54 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - do not execute js (crossOrigin)", 2, function( assert ) {
```
FagnerMartinsBrack: Ain't you using `assert.expect` anymore? xD

PR #2589, opened 2015-09-11
Attributes: Allow .add/remove/toggle/hasClass(".class-name")

ryanwalters:
Many times it is useful to declare a class once and reuse it throughout a piece of code (e.g. plugins).

    var CUSTOM_CLASS = '.class-name',
        $el = $(CUSTOM_CLASS);

    ...

    if ($someEl.hasClass(CUSTOM_CLASS)) {
        $someEl.removeClass(CUSTOM_CLASS);
    }

    ...

    $someOtherEl.find(CUSTOM_CLASS);

    // etc

However, this is not possible currently since `.addClass`, `.removeClass`, `.toggleClass`, and `.hasClass` do not accept class names which include the dot, while `$()` and most tree traversing methods either accept or require it.

I think it would be beneficial to allow these other methods to accept class names which include the dot.

### DIFF
--- a/src/attributes/classes.js
+++ b/src/attributes/classes.js
@@ -5,7 +5,8 @@ define( [
 	"../core/init"
 ], function( jQuery, rnotwhite, dataPriv ) {
 
-var rclass = /[\t\r\n\f]/g;
+var rclass = /[\t\r\n\f]/g,
+	rdot = /\./g;
 
 function getClass( elem ) {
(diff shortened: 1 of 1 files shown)


mzgol:
You're confusing CSS selectors with class names. `$('.class-name')` will find elements with a class `class-name`, not `.class-name`.

It's not a good idea to start class names with a dot because of issues like that.

PR #2599, opened 2015-09-15
Update attr.js

Brooooooklyn:
fix issues that attr handler would affect to DocumentFragment.


Brooooooklyn:
fix issues that attr handler would affect to DocumentFragment.

timmywil:
Thank you for your willingness to contribute. However, we first need a ticket that demonstrates the problem before accepting any pull requests. Please open an issue and provide a small test case from http://jsbin.com or http://jsfiddle.net to help us assess the issue.

PR #2601, opened 2015-09-18
Docs: Change "baar" to "bar"

garysye:
Changes "baar" to "bar" when used with "foo" in readme and comments of js files

Hi, I wasn't sure if this was a typo or intentional, but I saw "baar" being used with "foo" in the readme and some comments.  It's just a minor change, but this is a fantastic library so I thought that it would look better if it was consistent across the repo, so I changed it to "bar".

Keep up the great work!

image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,7 +302,7 @@ t( testName, selector, [ "array", "of", "ids" ] );
 Example:
 
 ```js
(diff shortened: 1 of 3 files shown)


markelog:
Thank you!

Issue #2498, opened 2015-07-24
XHR throws an exception accessing responseText, if responseType === "arraybuffer"


metadings:
If you try using `xhr.responseType = "arraybuffer"` (`jqXHR.xhrFields = { responseType: "arraybuffer" }`), the code issues an exception `InvalidStateError` (in Firefox, when accessing `responseText`, instead of `response`):

	// Support: IE9
	// Accessing binary-data responseText throws an exception
	// (#11426)
	typeof xhr.responseText === "string" ? {
		text: xhr.responseText
	} : undefined

I changed jquery-1.6.1; now I'm going to work with a current version to make a PR or does anyone have something laying around...?

metadings:
Probably I need to go back to a `try...catch` block around `responseText`, `responseXml` and `response`, so I'm just getting the whole response and then sorting out...

dmethvin:
It seems like this is subverting the normal `$.ajax` processing model and confusing the code, as far as it's concerned this is still a request where it should look at `responseText`. I am wondering if [defining a converter for an "arraybuffer" dataType](http://api.jquery.com/jquery.ajaxtransport/) would be the appropriate way to do this. Can you provide a complete example?

metadings:
I defined it using 

	// Install plain dataType
	jQuery.ajaxSetup({
		accepts: {
			plain: "text/plain"
		},
		responseFields: {
			plain: "response"
		}
		/* , converters: {
			"* plain": function (arrayBuffer) {
				if (arrayBuffer) {
					return new Uint8Array(arrayBuffer);
				}
			}
		} /**/
	});

	// Handle cache's special case and global
	jQuery.ajaxPrefilter("plain", function ( s, originalSettings, jqXHR ) {
		s.xhrFields = {
			responseType: "arraybuffer"
		};
		// s.dataTypes[0] = "plain";
		// return "plain";
	});

However I didn't get the converter right, do you know how to do it?

dmethvin:
Although #2521 didn't have quite the right approach, I think the main issue is that the normal path that the code takes generates an error when the `.responseType` is a blob. So it seems like we should fix only that specific problem inside jQuery, we [already have a fix for IE9](https://github.com/jquery/jquery/blob/9adfad19865837f5dffedb1eb41e407f196ca515/src/ajax/xhr.js#L87-L92) in that area. Once that is done the other issues can be resolved via the approach you took in #2521, outside jQuery itself, using the extensions that `$.ajax()` provides.

[...]

PR #2607, opened 2015-09-21
Update xhr.js

bugdanov:
xhr.responseText is undefined when xhr responseType is 'blob'

### DIFF
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -91,7 +91,7 @@ jQuery.ajaxTransport( function( options ) {
 									// Support: IE9
 									// Accessing binary-data responseText throws an exception
 									// (#11426)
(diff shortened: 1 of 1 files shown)


gibson042:
The preexisting `typeof` check already handles `undefined`... are you trying to address an exception? If so, in which clients? We'd need a unit test covering the case, and a fix that doesn't [break in IE9](http://bugs.jquery.com/ticket/11426) as this one does.

markelog:
It would be great if you would sign CLA, provided isolated example failure through http://jsfiddle.net/ or similar resource and added test to support your change

dmethvin:
Since the reporter is not responding and we already have the essence of this bug covered by the issue in #2498, I'm going to close it.

PR #2608, opened 2015-09-21
Build: remove needless instruction from .gitattributes

markelog:
"JS files must always use LF for tools to work" -
assuming that was relevant when we were young and trees were tall

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,2 @@
 # Auto detect text files and perform LF normalization
 * text=auto
-
-# JS files must always use LF for tools to work
-*.js eol=lf


mzgol:
Does it work with jquery-release? I don't remember which tools we were talking about here. cc @scottgonzalez

scottgonzalez:
I don't recall either, but I *think* it had to do with scripts that split on `\n`.

@jzaefferer Do you remember where the big discussion happened? I recall it being a big discussion that lasted a while and we tried a few different settings before standardizing on settings for all projects.

jzaefferer:
See https://github.com/jquery/jquery-ui/pull/1198 

The only information I can gather from that is that its important to keep this for some tools to work properly on Windows. Should be a good enough reason to keep it.

dmethvin:
Yes it's a Windows thing. If that isn't there the repo starts to get polluted with files that use CRLF line endings.

mzgol:
> If that isn't there the repo starts to get polluted with files that use CRLF line endings.

The first instruction changes it back to LF before committing, doesn't it? So this shouldn't be a problem, the repo will be LF-only. The problem may be some tools may assume UNIX line endings and break if you have local files with Windows ones.

dmethvin:
Yep, you're right. In particular some of our build steps take local files and combine them, and if those files have CRLF the result will be different depending on whether Windows or Unix was used to build. We have some [last-ditch checks](https://github.com/jquery/jquery/blob/master/build/tasks/dist.js#L35-L39) to detect that but they just cause the build to error out.

markelog:
Is that hard, to make that logic newline type independent?

markelog:
Friendly ping

timmywil:
I think we'll leave this as-is. It doesn't hurt to have it and it works right now.

PR #2610, opened 2015-09-22
Give the module "selector" name

trupanka:
Fixes the following error with custom build and require.js AMD loader:

`Error: Script error for: jquery-src/selector`
`Mismatched anonymous define() module: undefined`

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -1,1 +1,1 @@
-define( [ "./selector-sizzle" ], function() {} );
+define( "selector", [ "./selector-sizzle" ], function() {} );


mzgol:
I don't think this is a good idea. Please see https://github.com/jquery/jquery/blob/99e8ff1baa7ae341e94bb89c3e84570c7c3ad9ea/src/exports/amd.js#L13-L16.

mzgol:
What exactly are you trying to achieve? Where did you stumble into a problem? Since recommendations for AMD say modules shouldn't declare their names, I doubt going into that direction would solve more problems than it'd create.

trupanka:
I'm trying to build custom bundle with r.js optimizer with the following configuration:
```
{
    "dir": "builds/dev/js",
    "baseUrl": "lib",
    "paths": {
        "requirejs": "../node_modules/requirejs/require",
        "jquery-src": "../node_modules/jquery/src",
        "sizzle": "../node_modules/sizzle/dist/sizzle",
        "jquery-bundle-2.1.4": "../builds/dev/js/vendor/jquery-2.1.4/main"
    },
    "packages": [
        { "name": "jquery-2.1.4", "location": "vendor/jquery-2.1.4" }
    ],
    "bundles": {
        "jquery-bundle-2.1.4": [ "jquery" ]
    },
    "modules": [
        {
            "name": "require",
            "include": ["requirejs", "jquery"],
            "create": true
        }
    ],
    "optimize": "none"
}
```
In load-time I get error with `Mismatched anonymous define` message.

mzgol:
You're using bith the `jquery` module (which I assume is the built file) and `jquery-src` which contains individual modules, you also declare a `jquery-2.1.4` package as one that needs to be loaded as a CommonJS module; it seems pretty convoluted. Does anything from http://requirejs.org/docs/errors.html#mismatch ring a bell?

In any case, I doubt that naming the module here would be a good idea, certainly against RequireJS own recommendations so please look for help on Stack Overflow or other forums as this is most likely a configuration conflict. Our packages are quite normal so there shouldn't be any problem here.

If you get more concrete information and there's indeed sth wrong in jQuery itself, please open an issue.

PR #2612, opened 2015-09-24
Compatible with nonce of CSP2.0

zzbo:
Compatible with nonce of CSP2.0

### DIFF
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -55,7 +55,7 @@ jQuery.ajaxTransport( "script", function( s ) {
 
 				//Compatible with nonce of CSP2.0
 				if ( s.nonce ) {
-					script.attr( "nonce", s.nonce ); 
+					script.attr( "nonce", s.nonce );
 				}
 
 				// Use native DOM manipulation to avoid our domManip AJAX trickery
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -52,7 +52,7 @@ jQuery.ajaxTransport( "script", function( s ) {
 						}
 					}
 				);
-				
+
 				//Compatible with nonce of CSP2.0
 				if ( s.nonce ) {
 					script.attr( "nonce", s.nonce );
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -55,7 +55,7 @@ jQuery.ajaxTransport( "script", function( s ) {
 
 				//Compatible with nonce of CSP2.0
 				if ( s.nonce ) {
-					script.attr( "nonce", s.nonce );
+					script.attr( "nonce", s.nonce ); 
 				}
 
 				// Use native DOM manipulation to avoid our domManip AJAX trickery
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -55,7 +55,7 @@ jQuery.ajaxTransport( "script", function( s ) {
 				
 				//Compatible with nonce of CSP2.0
 				if (s.nonce) {
-					script.attr('nonce', s.nonce);
+					script.attr( "nonce", s.nonce );
 				}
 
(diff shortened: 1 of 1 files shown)


dmethvin:
This seems like a reasonable change, although it's too bad that it's yet another `$.ajax` option. If you'd like to get credit please sign the CLA.

zzbo:
@dmethvin  I applied for a CLA, but the CLA Verification Results: zhenbo.zheng <bleon1#126.com>: Suspicious name: zhenbo.zheng. Actually, 'zhenbo.zheng' is my full name, counld you give me some tips to finish it?

mzgol:
@zzbo Is your full name "zhenbo.zheng" with a dot inside? I guess the algorithm checks for a space.

markelog:
What is the use case? How do you get nonce id? 

We need tests too.

zzbo:
@markelog nonce id is provided by the user, if user not set 'nonce' on CSP2.0, jQuery can not use 'jsonp' and 'getScript'.

markelog:
Nonce has to be generated for every new request, so where you getting it from? Do you want generate your scripts on the fly, do you get it from HTTP-API?

Not to mention that nonce is applied only to inline script, so i'm not sure what exactly happening here.

What is your use case?

zzbo:
@markelog Nonce id generated by back-end, back-end can set CSP2.0 response header, back-end also can print the nonce id to the page, so you can get it very easy. I think how to get the nonce id is not a question, the question is how can I use JQuery's jsonp and getScript functions on a CSP2.0 page.

markelog:
Okay, i still do not get how your app supposed to work, but lets past that, what anything nonce has to do with jsonp, if it could be applied only to inline scripts?

zzbo:
@markelog Nonce is not only applied to inline scripts, but also applied to external script. The question still is how can I use JQuery's jsonp and getScript functions on a CSP2.0 page.

markelog:
> Nonce is not only applied to inline scripts, but also applied to external script.

Would you mind sharing a link with relevant parts of specification?

zzbo:
@markelog No matter what content of the external script, if you inject a script tag(JSONP, getScript) to page without nonce attribute, you will encounter the same problem. You can go to learn more about CSP2.0 policy before we discuss the issue.

markelog:
We would need someone to provide a resource that we can reference to, justifying inclusion of new option. Specification, draft anything.

I only encounter resources specifying inline script interaction with nonce attribute, so it would be great to see a more complete one.

zzbo:
@markelog How can I use JQuery's jsonp and getScript functions on a CSP2.0 page.

zzbo:
@markelog 
```javascript
<?php 
	header("content-security-policy: script-src 'nonce-aaaa'");
?>
<!DOCTYPE html>
<html>
<head>
	<title>Hello CSP2</title>
	<script nonce="aaaa" src="./jquery-1.10.2.js" type="text/javascript"></script>
</head>
<body>
	<script nonce="aaaa">document.write("12345");</script>
	<script nonce="aaaa">
		$.ajax({
			url : 'http://www.abc.com',
			dataType : 'jsonp',
			success : function (data) {}
		});

	</script>
</body>
</html>
```

markelog:
@zzbo is reproducible, thank you, still it would be cool to have reference on specification, so we would be sure that is not vendor initiative.

Also, we would need tests. And i'm thinking, we might not need to introduce `nonce` attribute but look at first script attribute which already added to the page, that would be a half solution though :-(.

dmethvin:
Here's the spec ref: http://www.w3.org/TR/CSP2/#script-src-nonce-usage

dmethvin:
Closing due to inactivity. We can reopen if the OP or other interested parties reply.

PR #2615, opened 2015-09-25
Core: Add .extend test for defined accessor properties

ConnorAtherton:
I've added two tests to ensure that `extend` correctly handles defined getter properties.


ConnorAtherton:
I've signed the CLA but the check is still failing. Would it be possible for someone to verify my details are being added correctly?

markelog:
Would you mind moving it to different test?

markelog:
Okay, moved it for you, thank you though! PRs with tests only are awesome :-)

PR #2616, opened 2015-09-26
Alias ajax as http

coolaj86:
You've already aliased `type` to `method`, and fixed other semantic "bugs". Why not go all the way? Kick the decade old misnomer in the face and embrace a strong and clear message to the rising generation of developers.

It's just so difficult to explain how x stands for "JSON" and "Binary Data", y'know?

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -837,5 +837,7 @@ jQuery.each( [ "get", "post" ], function( i, method ) {
 	};
 } );
 
+jQuery.http = jQuery.ajax;
+
 return jQuery;
 } );


jhicken:
+1

nekcih:
+1

dmethvin:
To prevent spamming of +1s I've locked the PR.

dmethvin:
I don't think this is a very good idea.

1. It doesn't meet the criteria in our [feature guidelines](https://github.com/jquery/jquery/wiki/Adding-new-features). You can alias the name to anything you want in your own code, it doesn't require changes inside jQuery. It would just create an additional documentation burden and confuse people who knew this as `$.ajax` for the past decade.

2. The renamings we've done in the past like `type` to `method` or `andSelf` to `addBack`  were driven by confusion about functionality. Nobody is has expressed confusion about `$.ajax` and if they are a name change to `$.http` certainly won't clarify things.

3. If you were to pick a bad name for Google to digest, I don't think you could pick one worse than "http". But you're in luck, if you search for "jquery http" you will still get relevant jQuery results in the first page. So if people really do search for "http" thinking that is the name, the docs problem is solved without having to rewrite any docs or change code.

I've reopened the thread, but `+1` is not a convincing reply. Again, if you really like this simply add the line in your own code. There is no reason for jQuery itself to add this.

Issue #2343, opened 2015-05-22
IE 11 delegate vs direct radio click handlers behave differently


baronworks:
With IE 11 found a case where click handler on a radio input behaves differently when delegated vs directly-bound.  
```javascript
// delegated click handler, works when clicked but not when using arrows.
$('#testform').on('click', '.radios', function( e ) {
     console.log('delegated click handler', $(this).val() );
});
//directly-bound click handler, always works even when using arrows
$('.radios').on('click', function( e ) {
    console.log('direct click handler', $(this).val() );
});
```
jquery version: 1.11.* and  2.1.3, most testing done with 1.11.3
works in browsers: Firefox: 38.0.1, IE 9 & 10, Opera 27.0.1689.66, Chrome 42.0.2311.15
does not work in: IE 11 Version: 11.0.9600.17801
OS: Windows 7 64-bit 

Created a jsfiddle using jquery 1.11.0 at http://jsfiddle.net/damnmaudite/kt99moLf/24/
Note when actually clicking the radio option everything works, but when using arrows to select a value in IE-11 the delegated event is not triggered..

dmethvin:
This arises from a combination of browser bugs and lies. We have some [code](https://github.com/jquery/jquery/blob/a644101ed04d0beacea864ce805e0c4f86ba1cd1/src/event.js#L430) to guard against a problem in Firefox regarding delegated *right*-click events that bubble as if they are `click`; TBH I am not sure Firefox still has this problem. 

But to make it worse, when you use the arrow keys in this example  in Firefox and Chrome the event says the left mouse button was clicked which is a lie. In IE it says no button was clicked (`-1`) which is true, but that triggers our Firefox bug check. I think we can fix this by changing `!event.button` to `event.button <= 1`. Or perhaps this guard can be removed completely if Firefox fixed the bug long ago.

BTW, I can't find where any standard defines what happens with the arrow keys on a radio button group.

baronworks:
Thanks for the quick response.  Played with the code using ```event.button <= 1```, and everything appears to work as expected.  Will create a local patch for our purposes that overrides ```jQuery.event.handlers``` helper method until this is added to the 3.0 milestone.

dmethvin:
It appears that Firefox hasn't fixed this bug, which is [14  years old](https://bugzilla.mozilla.org/show_bug.cgi?id=184051). So we should take the conservative approach and leave our guard in place, just modifying the `event.button` check.

mzgol:
[...]

Issue #2410, opened 2015-06-18
middle click delegation


zba:
As you probably know, event delegation is working only for 1st button click, this is because of line:

https://github.com/jquery/jquery/blob/master/src/event.js#L481

> // Avoid non-left-click bubbling in Firefox (#3861)

I read the original bug request and tried to remove this "avoid" here:  http://jsfiddle.net/Lwxdhk9b/

For me I not found anything "bad" if  click delegation is allowed, but I found something very bad if they disallowed, especially in case I can't switch this behavior  without fork of almost whole jQuery. please take a look for searching better solution for this old issue.

zba:
I have workaround for this here:

    $(document)
        .on('clickReal click', 'a', function (e) {
        console.log('click', e);
    })
        .on('click', function (e) {
        if (e.button) {
            e.type = 'clickReal';
            $(this).trigger(e);
        }
    });

dmethvin:
See also #2343 which I think has the same root cause.

dmethvin:
This should have been fixed by #2617, can you verify it works for you @zba? The latest version is in http://code.jquery.com/jquery-git.js .

dmethvin:
I am pretty sure this was fixed, please let us know if you find otherwise.

PR #2617, opened 2015-09-28
Event: Fix delegated radio events when arrow keys are used

dmethvin:
Fixes gh-2343, gh-2410

This turned out to be trickier than I hoped. Click events that are fired by `$.fn.trigger()` don't have an `event.button` set at all. Also I couldn't figure out a way to test this, for similar reasons. I can set up a unit test that has a fake event with similar incorrect settings as Firefox gives us, but that seems like a bad idea. Ideally this would be a feature detect but that is also impossible since it requires a real user-initiated click and not a fake one.

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -488,9 +488,10 @@ jQuery.event = {
 		// Find delegate handlers
 		// Black-hole SVG <use> instance trees (#13180)
 		//
-		// Support: Firefox
-		// Avoid non-left-click bubbling in Firefox (#3861)
-		if ( delegateCount && cur.nodeType && ( !event.button || event.type !== "click" ) ) {
+		// Support: Firefox<=42+
+		// Avoid non-left-click in FF but don't block IE radio events (#3861, gh-2343)
+		if ( delegateCount && cur.nodeType &&
+			( event.type !== "click" || isNaN( event.button ) || event.button < 1 ) ) {
 
 			for ( ; cur !== this; cur = cur.parentNode || this ) {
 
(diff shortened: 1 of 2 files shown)


markelog:
How about we create a test in https://github.com/jquery/jquery/tree/master/test/integration, right now, @timmywil checks this before the release and later we will automate it - https://github.com/jquery/jquery/issues/2329

src/event.js
```diff
@@ -488,9 +488,10 @@ jQuery.event = {
 		// Find delegate handlers
 		// Black-hole SVG <use> instance trees (#13180)
 		//
-		// Support: Firefox
-		// Avoid non-left-click bubbling in Firefox (#3861)
-		if ( delegateCount && cur.nodeType && ( !event.button || event.type !== "click" ) ) {
+		// Support: Firefox<=42+
+		// Avoid non-left-click in FF but don't block IE radio events (#3861, gh-2343)
+		if ( delegateCount && cur.nodeType &&
+			( event.type !== "click" || isNaN( event.button ) || event.button < 1 ) ) {
```
gibson042: What about `… && !(event.type === "click" && event.button >= 1)`?

dmethvin: I thought about that, the original `!event.button` handles both `0` and `undefined` but it's not obvious. The three conditions make it a bit clearer I think and the second two checks aren't reached except on click events, which are low frequency.

gibson042: That's true both ways, and there's no practical performance difference. It's just smaller without the `isNaN`, and the explanation is already handled pretty well by our support comment (which could become "Suppress clicks of non-primary mouse buttons" if you want to increase clarity a little more).

PR #2618, opened 2015-09-28
Attributes: Remove undocumented .toggleClass( boolean ) signature

### LINKED ISSUE #2491: Remove the unspecified-class .toggleClass signature
cf. jquery/api.jquery.com/issues/781

`.toggleClass( [boolean] )` is [inaccurately documented](http://api.jquery.com/toggleClass/#toggleClass-state), (probably) not working correctly, easy to manually implement by library consumers, and I can't even come up with a reasonable use case calling for removal/restoration of all classes.

dmethvin:
Fixes #2491 

I may have gone a little crazy with the refactoring here, so feel free to yell if something looks off. It passes unit tests, and I removed the tests that dealt with the className storage. 

The commit messages needs a little work, I guess I'd describe this as "removing the incorrectly documented `.toggleClass( [ boolean ] )` signature. Here is what we currently say which did not match the old implementation:

> As of jQuery 1.4, if no arguments are passed to .toggleClass(), all class names on the element the first time .toggleClass() is called will be toggled.

Ref https://github.com/jquery/api.jquery.com/issues/781

### DIFF
--- a/src/attributes/classes.js
+++ b/src/attributes/classes.js
@@ -1,9 +1,8 @@
 define( [
 	"../core",
 	"../var/rnotwhite",
-	"../data/var/dataPriv",
 	"../core/init"
-], function( jQuery, rnotwhite, dataPriv ) {
+], function( jQuery, rnotwhite ) {
 
 var rclass = /[\t\r\n\f]/g;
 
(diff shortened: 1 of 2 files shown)


src/attributes/classes.js
```diff
@@ -100,60 +99,29 @@ jQuery.fn.extend( {
...
-				// then remove the whole classname (if there was one, the above saved it).
-				// Otherwise bring back whatever was previously saved (if anything),
-				// falling back to the empty string if nothing was stored.
-				if ( this.setAttribute ) {
-					this.setAttribute( "class",
-						className || value === false ?
-						"" :
-						dataPriv.get( this, "__className__" ) || ""
-					);
+				if ( checker.call( self, className ) ) {
```
gibson042: I'd get rid of `checker`, leaving this test something like `stateVal === false || self.hasClass( className )`.

src/attributes/classes.js
```diff
@@ -100,60 +99,29 @@ jQuery.fn.extend( {
...
-						self.addClass( className );
-					}
-				}
-
-			// Toggle whole class name
-			} else if ( value === undefined || type === "boolean" ) {
-				className = getClass( this );
-				if ( className ) {
+		var type = typeof value,
+			classNames = type === "string" ? value.match( rnotwhite ) : "",
```
gibson042: For easier cherry-picking (since oldIE doesn't allow numeric-property access to strings) and compressibility, this should probably be `classNames = type === "string" && value.match( rnotwhite ) || []`.

test/unit/attributes.js
```diff
@@ -1245,29 +1245,6 @@ var testToggleClass = function( valueObj, assert ) {
...
-	e.toggleClass( true );
-	assert.ok( e.is( ".testD.testE" ), "Assert class present (restored from data)" );
-	e.toggleClass();
-	e.toggleClass( false );
-	e.toggleClass();
-	assert.ok( e.is( ".testD.testE" ), "Assert class present (restored from data)" );
-
-	// Cleanup
-	e.removeClass( "testD" );
-	assert.expectJqData( this, e[ 0 ], "__className__" );
```
gibson042: :D

src/attributes/classes.js
mzgol: IE8 supports numeric string property access so this should no longer be an
issue.

-- 
Michał Gołębiowski

dmethvin: @gibson042 `.toggleClass("cls", true)` needs to add the class if it's there and avoid removing it. If `stateVal` is `true` and the element currently had the class, it would be removed rather than kept. It could be done with a few extra conditionals though, instead of the call.

src/attributes/classes.js
mzgol: IE8 supports numeric string property access so this should no longer be an issue.

dmethvin: Oh, this was an out-and-out mistake from an earlier iteration. I definitely wanted a `[]` there!

src/attributes/classes.js
gibson042: Ah, right, we'd need two checks against `stateVal`. It might still be preferable, but I'm now less enthusiastic.
```js
if ( stateVal !== true && self.hasClass( className ) ) {
	self.removeClass( className );
} else if ( stateVal !== false ) {
	self.addClass( className )
}
```

src/attributes/classes.js
markelog: So you would need to add a fallback here and we are good to go?

dmethvin: yeah, just  need to change `""` to `[]` I think. I was also going to review whether using direct logic rather than a function was about the same size. I'll get to it t his week.

src/attributes/classes.js
```diff
@@ -100,60 +99,26 @@ jQuery.fn.extend( {
...
-				// then remove the whole classname (if there was one, the above saved it).
-				// Otherwise bring back whatever was previously saved (if anything),
-				// falling back to the empty string if nothing was stored.
-				if ( this.setAttribute ) {
-					this.setAttribute( "class",
-						className || value === false ?
-						"" :
-						dataPriv.get( this, "__className__" ) || ""
-					);
+				if ( stateVal === false || stateVal !== true && self.hasClass( className ) ) {
```
Mr21: What is the use of `stateVal === false ||` ?

dmethvin: In `.toggleClass()` the `stateVal` can be either `true`, `false` or `undefined`. The boolean cases indicate the caller wants to force the class to be added or removed. For the `undefined` case, the class should be added if it does not have the class and removed if it does have the class. So if you walk through the three states you can double-check that expression works for all the cases. I added a few extra tests below to catch some of the possible mistakes someone could make in editing this line. :smile_cat:

Mr21: Oh yes, my bad... I realize now the 6th case here, doesn't work without your `===false||`, you're right.
``` javascript
$( ".class" ).toggleClass( "class" )        -> removeClass
$( ".class" ).toggleClass( "class", false ) -> removeClass
$( ".class" ).toggleClass( "class", true  ) -> addClass

$( ":not(.class)" ).toggleClass( "class" )        -> addClass
$( ":not(.class)" ).toggleClass( "class", true  ) -> addClass
$( ":not(.class)" ).toggleClass( "class", false ) -> removeClass // without stateVal === false || is only for this case
```

PR #2621, opened 2015-09-30
jQuery.param() was not compatible to JavaScripts encodeURIComponent and decodeURIComponent

sgreiner:


dmethvin:
Any time you change unit tests it implies a change in documented and/or supported behavior. We're following the standard, is there a reason to change this? See [trac-3400](http://bugs.jquery.com/ticket/3400).

markelog:
@sgreiner could you help us out and answer @dmethvin ?

dmethvin:
I'm going to close this since @sgreiner hasn't responded and also I don't think it can be landed as it is. However, see #2658 which I just opened after reviewing this. If you would like to submit a pull request for that ticket please let us know.

Issue #2622, opened 2015-09-30
css(): getting opacity value just after setting results in invalid value for iframe in Firefox


pmstss:
**Test**: https://jsfiddle.net/t7xsbe6o/

**Expected result**: alert with message "PASSED" (e.g. in Chrome)
Firefox 40.0.3 result: "Expected opacity value is 0, but returned value is: 1"

**jQuery versions**: 2.1.4, 2.1.3 and all other available at jsfiddle - result is the same

**OS**: Windows 7 x64

**Test details:**
Firstly opacity of the `#testDiv` element is set to 0 (`$testDiv.css('opacity', 0)`). Then value is received as `$testDiv.css('opacity')` - and it's not 0, which was set.

Didn't  try with another properties; tried only with opacity.

**Workarounds**:
it works fine in the following tests:
1) setTimeout before get call: https://jsfiddle.net/t7xsbe6o/1/
2) jQuery free test: https://jsfiddle.net/vgpavw9g/
3) iframe free test: https://jsfiddle.net/67oos5aj/

pmstss:
have just updated Firefox to version 41.0, result is the same

dmethvin:
Same failed result in Firefox 43.0a2 as well. Historically, manipulating an iframe with an instance of jQuery not in that iframe has been dicey and we don't have any unit tests to ensure that it works in the only case where it can, e.g. same-origin. Can you report this to Mozilla and link to it here?

pmstss:
Thanks for feedback! But what exactly to report to Mozilla? Mentioned jQuery free test (https://jsfiddle.net/vgpavw9g/) works. What are the differences between that test and what happened inside jQuery (was going to investigate it later, but may be you can clarify this at once)?

dmethvin:
Okay, sorry, missed that. What we're doing is [this code](https://jsfiddle.net/vgpavw9g/1/) with `window` because of the multiple-browser bugs described in [this file](https://github.com/jquery/jquery/blob/99e8ff1baa7ae341e94bb89c3e84570c7c3ad9ea/src/css/var/getStyles.js). Maybe Firefox has fixed their original bug and caused this in the process?

pmstss:
It works with `elem.ownerDocument.defaultView`, mentioned in getStyles.js - see https://jsfiddle.net/vgpavw9g/3/ - but not with `window` (parent), like in your https://jsfiddle.net/vgpavw9g/1/ test. Though in Chrome it works in both cases, which is strange for me.

[...]

Issue #2259, opened 2015-05-04
Problem calling css in parent (from within a frame)


MartinKosicky:
Hi well I am trying to do a $(someElement).css("display"); someElement was returned from parent window and was passed to iframe via a function. This call in firefox returns undefined due to the following code in jQuery:

    getStyles = function( elem ) {
    // Support: IE<=11+, Firefox<=30+ (#15098, #14150)
    // IE throws on elements created in popups
    // FF meanwhile throws on frame elements through "defaultView.getComputedStyle"
    if ( elem.ownerDocument.defaultView.opener ) {
        return elem.ownerDocument.defaultView.getComputedStyle( elem, null );
    }

    return window.getComputedStyle( elem, null );
    };

When in an iframe in firefox elem.ownerDocument.defaultView.opener is not set but elem.ownerDocument.defaultView.parent is. So I suggest replace it with

    if ( elem.ownerDocument.defaultView.opener || elem.ownerDocument.defaultView.parent ) {

timmywil:
Thank you for opening an issue. This may very well be a bug, but it would be helpful if you could create a test case on http://jsbin.com or similar.

timmywil:
Closing due to inactivity.

PR #2623, opened 2015-10-01
CSS: getting proper window object for getComputedStyle call in getStyles

pmstss:
Problem is described in #2622. Origin seems to be that `getComputedStyle` is called on main `window` object, and not on iframe `window` object. 
In this fix `window` object is _always_ received from an element (`elem.ownerDocument.defaultView`) with additional check for `window`/`document` elements - and thus check for `opener` is removed as useless.
Similar issue is described in #2259, but in proposed there solution check for `parent` in additional to `opener` is made. I think getting `window` object from element every time is better.

### DIFF
--- a/src/css/var/getStyles.js
+++ b/src/css/var/getStyles.js
@@ -1,15 +1,9 @@
-define( function() {
-	return function( elem ) {
+define( [
+	"../../core"
+], function( jQuery ) {
 
-		// Support: IE<=11+, Firefox<=30+ (#15098, #14150)
-		// IE throws on elements created in popups
-		// FF meanwhile throws on frame elements through "defaultView.getComputedStyle"
-		var view = elem.ownerDocument.defaultView;
-
-		if ( !view.opener ) {
-			view = window;
-		}
-
-		return view.getComputedStyle( elem );
-	};
+return function( elem ) {
+    return ( jQuery.isWindow( elem ) ? elem : elem.nodeType === 9 ?
+        elem.defaultView : elem.ownerDocument.defaultView ).getComputedStyle( elem );
+};
 } );


pmstss:
Btw, could anybody clarify why more verbose version https://github.com/pmstss/jquery/blob/windowObjectFixVerbose/src/css/var/getStyles.js is not concatenated properly (jshint fails cause `getWindow` strangely disappears from result jquery.js)? Did I miss something obvious, or is it build bug?

dmethvin:
I'm still not sure why this used to fail in Firefox and doesn't now--or am I being dense about something? It seems like `elem.ownerDocument.defaultView` should have gotten us to the right window, but the comment says it fails in older FF. Were you able to test it on all browsers including IE11? If the problems are all in the past I'm okay with removing the clarifying comment but otherwise it would be helpful to keep.

I'm not sure about the issue with the longer code, although I'm wondering if we need that elsewhere.

timmywil:
Hi @pmstss! Thanks for the PR. Could you add a test that ensure we fix the issue described in #2622?

pmstss:
Hi @timmywil! There are some still unresolved oddities with my fix, which i want to clarify first, but i'm short in time now. 
As @dmethvin already mentioned, we must be sure that this fix doesn't not break previous e488d98, and check via test/index.html shows that this is true, but: if I extract test html structure and corresponding scripting into separate html file - it is not passed. 
So, at the current time I think that: 
a) something wrong with test in e488d98 (not with test itself, but with the environment in which test is running)
b) i assume my fix breaks e488d98 (though test/index.html doesn't show it)

timmywil:
@pmstss Perhaps FF has fixed the issue from https://github.com/jquery/jquery/commit/e488d985cfb10ab8c684bbc6a9b8ff3eae23bf83. Nevertheless, I think we need a test similar to the test in the [fiddle from #2622](https://jsfiddle.net/timmywil/t7xsbe6o/4/), since this wasn't failing before in FF. As long as FF doesn't throw an error in versions current and current - 1, we can merge the change.

pmstss:
Test from e488d98 is not passed anymore with my changes and Firefox 42. 
Additional debugging revealed one more oddity: test is passed when "display:none" is set in html (https://jsfiddle.net/femxL9du/3/) but is not passed, when the same is set via script (https://jsfiddle.net/femxL9du/4/).

#2622 should be fixed in some other way.

PR #2624, opened 2015-10-01
Event: Move .bind() and .delegate() to deprecated

### LINKED ISSUE #2288: Deprecated `bind`, `unbind`, `delegate`, `undelegate` methods
I guess it is time to do this.

1. Move them to `deprecated` module
2. Add warnings into `jquery-migrate`
3. Update the docs

dmethvin:
Fixes #2288 

At the moment we don't have any tests at all for .bind() and .delegate(), it seems like we would at least want a sanity check for them. At the very least they should be tested in Migrate.

While I'm in here I'm wondering if we should deprecate anything else in the aliases.

### DIFF
--- a/src/event/alias.js
+++ b/src/event/alias.js
@@ -19,24 +19,6 @@ jQuery.each( ( "blur focus focusin focusout resize scroll click dblclick " +
 jQuery.fn.extend( {
 	hover: function( fnOver, fnOut ) {
 		return this.mouseenter( fnOver ).mouseleave( fnOut || fnOver );
-	},
-
-	bind: function( types, data, fn ) {
-		return this.on( types, null, data, fn );
-	},
-	unbind: function( types, fn ) {
-		return this.off( types, null, fn );
-	},
-
-	delegate: function( selector, types, data, fn ) {
-		return this.on( types, selector, data, fn );
-	},
-	undelegate: function( selector, types, fn ) {
-
-		// ( namespace ) or ( selector, types [, fn] )
-		return arguments.length === 1 ?
-			this.off( selector, "**" ) :
-			this.off( types, selector || "**", fn );
 	}
 } );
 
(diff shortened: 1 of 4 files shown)


src/event/alias.js
```diff
@@ -19,24 +19,6 @@ jQuery.each( ( "blur focus focusin focusout resize scroll click dblclick " +
 jQuery.fn.extend( {
 	hover: function( fnOver, fnOut ) {
```
dmethvin: In case you haven't heard, [I'm not a fan of this method](http://blog.methvin.com/2011/10/this-hover-cruft-is-full-of-ills.html).

mzgol: I agree, it's confusing. We could deprecate it in 3.0 & remove in 4.0 perhaps?

mzgol:
LGTM. Tests would be nice but since you're just moving code around it's not a blocker here.

BTW, we've had `.bind()`/`.delegate()` tests until I switched them to `.on()` in 8ca9f931ec311b6f73990eac7665451a28bceac3 & 84a94acae1ed7d65d91df235985e433d34486dc3 (one of my first contributions to jQuery!). But it had to be done as we had too low coverage of `.on()`/`.off()` testing before that. :)

Should we duplicate a few basic `.on()` tests, creating `.bind()`/`.delegate()` versions?

dmethvin:
I don't mind adding tests to this PR, just some basic sanity tests that are copies of what we already have but calling the deprecated methods. And we already have an empty `test/unit/deprecated.js`, yay!

timmywil:
In my opinion, anything more than

```
equal( jQuery.on, jQuery.bind, ... );
```

is just redundant.

mzgol:
@timmywil They're not equal, `.bind()` invokes `.on()`. That's why real tests are needed.

timmywil:
Of course. It's been a long time since I've used them. :P

markelog:
Tests are added, are we good to go?

dmethvin:
Should be good to go. Just added the Migrate issue

markelog:
Okay, merging

PR #2625, opened 2015-10-02
Misspelled comment

GreenCoder64:


GreenCoder64:
Hi @jquerybot, I have already did the prescribed, however I am still deemed to have not signed the CLA, please verify.

FagnerMartinsBrack:
[That's probably intentional](https://en.wikipedia.org/wiki/If_and_only_if):

> In logic and related fields such as mathematics and philosophy, if and only if (shortened iff) is a biconditional logical connective between statements.

dmethvin:
Yes, it's intentional, see https://github.com/jquery/jquery/pull/749#discussion_r718127

@GreenCoder64 As the message above says, if you signed the CLA with the same name and email as your commit there should be no problem. Also you should do work on a branch for your own sanity, otherwise your master branch gets out of sync and makes it harder for you to do other pull requests. See the [Contributing](http://learn.jquery.com/contributing/) section for some pointers.

GreenCoder64:
How about;
```javascript
// The broken getElementById methods don't pick up programatically-set names,
```

```javascript
// In-page `selector#id sibing-combinator selector` fails
```

```javascript
// Defend against cloned attroperties (jQuery gh-1709)
```

```javascript
// Defend against cloned attroperties (jQuery gh-1709)
```

```javascript
// For mousenter/leave call the handler if related is outside the target.
```

PR #2627, opened 2015-10-02
Core: Support Symbol wrapper objects in jQuery.type

ChristianGrete:
In ECMAScript 2015 (ES6), the native `typeof` operator returns `"symbol"`
for `Symbol` primitives. As it is possible to wrap symbols using the
`Object` constructor, symbols can be objects as well as any other
primitive type in JavaScript and should be determined by `jQuery.type`.

### DIFF
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -270,6 +270,12 @@ QUnit.test( "type", function( assert ) {
 	assert.equal( jQuery.type( new MyNumber( 1 ) ), "number", "Number" );
 	assert.equal( jQuery.type( new MyString( "a" ) ), "string", "String" );
 	assert.equal( jQuery.type( new MyObject() ), "object", "Object" );
+
+	// Prevent reference errors
+	if( typeof Symbol === "function" ) {
+		assert.equal( jQuery.type( Symbol() ), "symbol", "Symbol" );
+		assert.equal( jQuery.type( Object( Symbol() ) ), "symbol", "Symbol" );
+	}
 } );
 
 QUnit.asyncTest( "isPlainObject", function( assert ) {
--- a/src/core.js
+++ b/src/core.js
@@ -442,7 +442,7 @@ if ( typeof Symbol === "function" ) {
 /* jshint ignore: end */
 
 // Populate the class2type map
-jQuery.each( "Boolean Number String Function Array Date RegExp Object Error".split( " " ),
+jQuery.each( "Boolean Number String Function Array Date RegExp Object Error Symbol".split( " " ),
 function( i, name ) {
 	class2type[ "[object " + name + "]" ] = name.toLowerCase();
 } );


dmethvin:
I think this is a good time for some review. Is there a reason for `jQuery.type` to return `"symbol"`? On the one hand yes, it's a new type, but on the other why is a `Symbol` being passed to anything that jQuery is handling since we never do anything with it? If we are going to make this change I think it would be good to also take a look at all our other type-sniffing methods like `isPlainObject` and `isNumeric` to see what they do with `Symbol` as well.

mzgol:
We never do anything with it but `jQuery.type` is a public API so it should be accurate. I think this is a good thing to fix, especially that the patch is small.

mzgol:
@dmethvin 
```js
var s1 = Symbol('id');
var s2 = new Object(s1);

$.isPlainObject(s1); // false
$.isPlainObject(s2); // false

$.isEmptyObject(s1); // true
$.isEmptyObject(s2); // true
// But also:
$.isEmptyObject(42); // true

$.isNumeric(s1); // throws on parseFloat(s1)
$.isNumeric(s2); // throws on parseFloat(s2)
```

Throwing in `$.isNumeric` can be prevented by changing `parseFloat( obj )` to `parseFloat( String( obj ) )`.

markelog:
@mzgol in what browsers do you have this result?

mzgol:
I tested in Chrome but those methods are generic enough it should work in
the same way in every browser supporting symbols. Do you have different
results somewhere?

-- 
Michał Gołębiowski

markelog:
Since `symbol` is the new entity, vendors might implement it with some subtle differences, violating the spec or exploiting some undefined behaviours. Whereas right now, most code that released with ES6 probably uses transpiler, whereas transpiler might or not might use a shim (babel, for example, does for `Symbol`).

So in order to be sure in results you shown above, we need to properly research it, checking it with any possible environment

mzgol:
@markelog Good point. Let's see what exactly happens there then, according to the spec:

1. `$.isPlainObject` returns `false` because for both boxed & unboxed symbols `s.constructor` is the global `Symbol`; they behave the same because primitives are auto-boxed on property access. For the latter, as http://www.ecma-international.org/ecma-262/6.0/#sec-objects specifies, every object has a `constructor` field pointing to the constructor; as http://www.ecma-international.org/ecma-262/6.0/#sec-symbol-constructor specifies, the `Symbol` function creates an instance of the `Symbol` value so it matches.
2. `$.isEmptyObject` returns `true` because it returns `false` only inside the loop `for ( name in obj ) { return false }` and symbols don't have any string-keyed enumerable properties and every primitive is converted to an object when iterated via `for..in` (http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-forin-div-ofheadevaluation-tdznames-expr-iterationkind, point 7.b.).
3. `$.isNumeric` throws because it invokes `parseFloat` on a symbol because `parseFloat` converts its parameter to string via the abstract `ToString` operation (http://www.ecma-international.org/ecma-262/6.0/#sec-parsefloat-string, point 1) and `ToString` on a symbol throws a `TypeError` according to the table at http://www.ecma-international.org/ecma-262/6.0/#sec-tostring.

So my previous comment was correct anyway.

markelog:
Mostly, i meant that bit - 

> violating the spec

and 

> babel, for example, does use it for Symbol

That happend before, like with v8 initial implementation of promises

mzgol:
Yeah, but considering that Symbols are not in IE, other browsers are mostly evergreen so they are patched regularly and jQuery doesn't use Symbols internally I don't think we should fix any possible Symbol-related bugs in Core, we should just make sure we behave correctly for standards-based implementations. We *might* make an exception for Babel/Traceur but I wouldn't care too much about real environments here.

mzgol:
FWIW, Node 0.12 & 4.0 works as Chrome 45 with these functions and Node 0.10 doesn't support Symbols.

markelog:
@mzgol if everything that browsers did would be according to the specification, we could have remove all those evergreen hacks

mzgol:
@markelog Yes, but see the part:
> jQuery doesn't use Symbols internally

We don't fix IndexedDB bugs inside jQuery, for the same reason I don't think we should fix any potential Symbol-related bugs.

markelog:
> We don't fix IndexedDB bugs inside jQuery

I don't understand the reference

>  for the same reason I don't think we should fix any potential Symbol-related bugs.

That means we don't support `Symbol`

mzgol:
Oh well, these tests are simple so I just run them:
```js
var s1 = Symbol('id');
var s2 = new Object(s1);

console.assert($.isPlainObject(s1) === false);
console.assert($.isPlainObject(s2) === false);

console.assert($.isEmptyObject(s1) === true);
console.assert($.isEmptyObject(s2) === true);
// But also:
console.assert($.isEmptyObject(42) === true);

try { $.isNumeric(s1); console.assert(false); } catch (e) {}
try { $.isNumeric(s2); console.assert(false); } catch (e) {}
```
It passes in Chrome 45, Firefox 41, Edge 12 & Safari 9. Safari < 9 doesn't support `Symbol`. We can add proper tests for those things, I guess.

markelog:
> We can add proper tests for those things, I guess.

That would be cool to have, yeah, if they would fail, we can open discussion of whenever or not we  should fixing potential bugs. I think we should also try to run them in popular shims for it too.

ChristianGrete:
Well... The discussion is whether or not to implement this feature and the arguments against it is that it is not used by jQuery internally (or that the average jQuery user doesn’t do anything with symbols), right?

markelog:
@ChristianGrete nuh, we just being careful here, discussing specifics of the implementation and possible risks. We all agree that this should be implemented

markelog:
@mzgol would you mind adding your tests with follow-up commit?

mzgol:
> @mzgol would you mind adding your tests with follow-up commit?

Sure. Should I land this PR then?

markelog:
Well, @timmywil added `Needs review` and didn't remove it, also, did you try run them with shim?

mzgol:
Not yet, I won't have time for that before the Summit.

timmywil:
I'm fine with landing this before we add more tests, but we need an issue to track that.

markelog:
It would be cool to test in the shim, just to be sure

timmywil:
Fair enough.

mzgol:
@markelog I assume you mean to test Babel with the core-js shim? In that case I'd like to first fix the test at [test/node_smoke_tests/iterable_with_symbol_polyfill.js](https://github.com/jquery/jquery/blob/03eaadb131df925d1072afd2496ee3b41d2f1fc6/test/node_smoke_tests/iterable_with_symbol_polyfill.js) to work in browsers (this will allow us to update jsdom) and then adding a similar test for this issue will be simple. The initial work is not a one-liner, though, so I won't have time to do it before the summit.

markelog:
@mzgol i'd say your conformation that this is works is enough, after that i can create ticket like @timmywil suggest and track everything else from there

mzgol:
I won't be 100% sure if it works before I test it. ;) That said, this reminded me that unless you enable the non-default [es6.spec-symbols transformer](http://babeljs.io/docs/advanced/transformers/es6/spec-symbols/) typeof isn't shimmed by Babel so for the majority of users `typeof symbol` will return `object`; try e.g. [this code](http://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&playground=false&code=console.log(typeof%20Symbol())%3B%0A) in IE11.

markelog:
> I won't be 100% sure if it works before I test it

I meant you could just test it locally, without adding anything to the dist, conform that it is works and we could deal with everything else later - that would be enough for me.

mzgol:
It's not that simple; I'd have to be careful in checking if I'm testing the right thing, there's also the issue about `typeof` not being shimmed by default which I described above. I don't have time now and I don't want to rush it.

markelog:
> about typeof not being shimmed

Transformed, not shimmed, otherwise we will be lost in terminology :-), when you use babel shim collection, you would need to use such transforms, since shims are not included by default, i assume that is reason why such transforms are not included too.

So it should be pretty easy, but i guess we can wait :-)

mzgol:
The Babel's idea is that some transformers fix edge cases and are enough non-performant that the majority of users won't need them. They are marked with the `es6.spec.` prefix & disabled by default. I don't have data about it but I assume most people will run Babel with default settings, adding only the `browser-polyfill.js` file (which contains compiled `core-js`) or enabling Babel runtime which adds it dynamically (at least that's what I see in most projects I look at). Therefore, most Babel users won't have `typeof` transpiled to be ES6-compliant so this patch won't work for them in non-modern browsers.

I'm not sure if we should care about it but that's how it is.

markelog:
In any case, that is not our problem, if some shim requires some deps, it is their issue, we can do only so much, so i would suggest to check correctness with full version

mzgol:
OK, I was wrong, since `jQuery.type` in no way uses `typeof`, the non-compliant default typeof Symbol treating is not a problem.

The patch makes it work in IE11 as seen in [this simple test case](http://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&playground=false&code=const%20class2type%20%3D%20%7B%7D%3B%0A%0Aconst%20toString%20%3D%20class2type.toString%3B%0A%0A%2F%2F%20Populate%20the%20class2type%20map%0A%24.each(%20%22Boolean%20Number%20String%20Function%20Array%20Date%20RegExp%20Object%20Error%20Symbol%22.split(%20%22%20%22%20)%2C%0Afunction(%20i%2C%20name%20)%20%7B%0A%09class2type%5B%20%22%5Bobject%20%22%20%2B%20name%20%2B%20%22%5D%22%20%5D%20%3D%20name.toLowerCase()%3B%0A%7D%20)%3B%0A%0Afunction%20type(%20obj%20)%20%7B%0A%09if%20(%20obj%20%3D%3D%20null%20)%20%7B%0A%09%09return%20obj%20%2B%20%22%22%3B%0A%09%7D%0A%0A%09%2F%2F%20Support%3A%20Android%3C4.0%20(functionish%20RegExp)%0A%09return%20typeof%20obj%20%3D%3D%3D%20%22object%22%20%7C%7C%20typeof%20obj%20%3D%3D%3D%20%22function%22%20%3F%0A%09%09class2type%5B%20toString.call(%20obj%20)%20%5D%20%7C%7C%20%22object%22%20%3A%0A%09%09typeof%20obj%3B%0A%7D%3B%0A%0A%0Aconsole.log(type(Symbol('foo'))%2C%20type(new%20Object(Symbol('bar'))))%3B%0A).

test/unit/core.js
```diff
@@ -270,6 +270,12 @@ QUnit.test( "type", function( assert ) {
 	assert.equal( jQuery.type( new MyNumber( 1 ) ), "number", "Number" );
 	assert.equal( jQuery.type( new MyString( "a" ) ), "string", "String" );
 	assert.equal( jQuery.type( new MyObject() ), "object", "Object" );
+
+	// Prevent reference errors
+	if( typeof Symbol === "function" ) {
```
markelog: Could you move that bit in separate test?

markelog:
@mzgol sounds like you conformed it :-)
@ChristianGrete would you mind taking in the consideration my [suggestion](https://github.com/jquery/jquery/pull/2627#discussion_r41774895)? After that, i think we all set!

markelog:
Okay, i moved that test for you, thank you!

ChristianGrete:
@markelog Of course not, I’m totally fine with that. Just wasn’t sure where to properly put the test, so I added it there. 😉

Issue #2005, opened 2015-01-13
Improve tag name related regexes


markelog:
<a href="https://github.com/jquery/jquery/pull/1988">See</a> discussion in related pull, specifically @gibson042 <a href="https://github.com/jquery/jquery/pull/1988#issuecomment-69500885">comment</a>.

markelog:
@LeonardoBraga interested in providing fix for it?

LeonardoBraga:
@marklog - Yes, definitely. Will work on the PR ASAP.

mzgol:
@LeonardoBraga Still interested?

LeonardoBraga:
@mzgol Sure. I'll work on it over the weekend and send the PR. Thanks.

dmethvin:
@LeonardoBraga just pinging to see if you're going to send a PR. We'd love to have your contribution!

LeonardoBraga:
@dmethvin I'm sorry for the delay on sending the PR. Should I also apply the same changes to other branches and send additional PRs?

gibson042:
No need; this is easy to cherry pick.

PR #2634, opened 2015-10-07
Manipulation: Bring tagname regexes up to spec

LeonardoBraga:
Closes gh-2005

Implements the recommendations from @gibson042 in [this comment](https://github.com/jquery/jquery/pull/1988#issuecomment-69500885).

### DIFF
--- a/src/core/var/rsingleTag.js
+++ b/src/core/var/rsingleTag.js
@@ -1,5 +1,5 @@
 define( function() {
 
 	// Match a standalone tag
-	return ( /^<([\w-]+)\s*\/?>(?:<\/\1>|)$/ );
+	return ( /^<([a-z][^\/\0>:\x20\t\r\n\f]*)[\x20\t\r\n\f]*\/?>(?:<\/\1>|)$/i );
 } );
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -26,7 +26,7 @@ define( [
 	dataPriv, dataUser, acceptData ) {
 
 var
-	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([\w:-]+)[^>]*)\/>/gi,
+	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([a-z][^\/\0>\x20\t\r\n\f]*)[^>]*)\/>/gi,
 
 	// Support: IE 10-11, Edge 10240+
 	// In IE/Edge using regex groups here causes severe slowdowns.
--- a/test/unit/manipulation.js
+++ b/test/unit/manipulation.js
@@ -500,6 +500,59 @@ QUnit.test( "html(String) tag-hyphenated elements (Bug #1987)", function( assert
 	assert.equal( j.children().text(), "text", "Tags with multiple hypens behave normally" );
 } );
 
+QUnit.test( "html(String) respect the spec on how to handle special and name-terminating chars for tag names (gh-2005)", function( assert ) {
+
+	assert.expect( 240 );
+
+	var wrapper, child, nodeName,
+		nameTerminatingChars = "\x20-\t-\r-\n-\f".split( "-" ),
+		specialChars = "[ ] { } _ - = + \\ ( ) * & ^ % $ # @ ! ~ ` ' ; ? ¥ « µ λ ⊕ ≈ ξ ℜ ♣ €".split( " " );
+
+	specialChars.push( specialChars.join( "" ) );
+
+	function buildChild( method, tagName ) {
+		wrapper = jQuery( "<div></div>" );
+		wrapper[method]( tagName );
+		return wrapper.children()[0];
+	}
+
+	function assertSpecialCharsSupport( method, nodeName ) {
+		child = buildChild( method, "<" + nodeName + "></" + nodeName + ">" );
+		assert.ok( jQuery.nodeName( child, nodeName.toUpperCase() ), "Paired tagnames with special characters" );
+
+		child = buildChild( method, "<" + nodeName + ">" );
+		assert.ok( jQuery.nodeName( child, nodeName.toUpperCase() ), "Unpaired tagnames with special characters" );
+
+		child = buildChild( method, "<" + nodeName + "/>" );
+		assert.ok( jQuery.nodeName( child, nodeName.toUpperCase() ), "Self-closing tagnames with special characters" );
+	}
+
+	function assertNameTerminatingCharsHandling( method, character ) {
+		nodeName = "div" + character + "-this-will-be-discarded";
+
+		child = buildChild( method, "<" + nodeName + "></" + nodeName + ">" );
+		assert.ok( jQuery.nodeName( child, "DIV" ), "Paired tagnames with name terminating characters" );
+
+		child = buildChild( method, "<" + nodeName + ">" );
+		assert.ok( jQuery.nodeName( child, "DIV" ), "Unpaired tagnames with name terminating characters" );
+
+		child = buildChild( method, "<" + nodeName + "/>" );
+		assert.ok( jQuery.nodeName( child, "DIV" ), "Self-closing tagnames with name terminating characters" );
+	}
+
+	jQuery.each( specialChars, function( i, character ) {
+		nodeName = "valid" + character + "tagname";
+
+		assertSpecialCharsSupport( "html", nodeName );
+		assertSpecialCharsSupport( "append", nodeName );
+	} );
+
+	jQuery.each( nameTerminatingChars, function( i, character ) {
+		assertNameTerminatingCharsHandling( "html", character );
+		assertNameTerminatingCharsHandling( "append", character );
+	} );
+} );
+
 QUnit.test( "IE8 serialization bug", function( assert ) {
 
 	assert.expect( 2 );
--- a/src/manipulation/var/rtagName.js
+++ b/src/manipulation/var/rtagName.js
@@ -1,3 +1,3 @@
 define( function() {
-	return ( /<([\w:-]+)/ );
+	return ( /<([a-z][^\/\0>\x20\t\r\n\f]+)/i );
 } );


mzgol:
You have to exclude NULL (`\0`) as well, @gibson042 mentioned that in his comment.

LeonardoBraga:
@mzgol Added the missing NULL (`\0`) character to the exclusion list.

src/core/var/rsingleTag.js
```diff
@@ -1,5 +1,5 @@
 define( function() {
 
 	// Match a standalone tag
-	return ( /^<([\w-]+)\s*\/?>(?:<\/\1>|)$/ );
+	return ( /^<([a-zA-Z][^\s\/\0>]*)\s*\/?>(?:<\/\1>|)$/ );
```
gibson042: Some comments on this regular expression:
* `\s` is too broad—we should instead use [`\x20\t\r\n\f`](http://www.w3.org/TR/css3-selectors/#whitespace) (assuming https://github.com/whatwg/html/issues/242 is accepted)
* It shouldn't start accepting `:`, which is special in XHTML—such text should go through the [namespace-aware `innerHTML` path](https://github.com/jquery/jquery/pull/1988#discussion_r22442173)
* It would be smaller if it were case-insensitive rather than explicitly matching `A-Z`. Altogether, I got good results with `/^<([a-z][^:\/\0>\x20\t\r\n\f]*)[\x20\t\r\n\f]*\/?>(?:<\/\1>|)$/i`.

src/manipulation.js
```diff
@@ -26,7 +26,7 @@ define( [
 	dataPriv, dataUser, acceptData ) {
 
 var
-	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([\w:-]+)[^>]*)\/>/gi,
+	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([a-z][^\s\/\0>]*)[^>]*)\/>/gi,
```
gibson042: See above regarding `\s` and sizing, but note that this one should accept `:`. Likewise on all counts for `rtagName` as well.

src/core/var/rsingleTag.js
LeonardoBraga: Sure. I'll apply the aforementioned changes.

Regarding making it case insensitive, I considered that and did a quick research about it, but all the pieces of information I found said that the performance would be considerably lower than adding both [a-zA-Z]. Just let me know if you'd rather taking the chance and making it case insensitive.

test/unit/manipulation.js
```diff
@@ -500,6 +500,29 @@ QUnit.test( "html(String) tag-hyphenated elements (Bug #1987)", function( assert
...
+		wrapper = jQuery( "<div></div>" );
+		wrapper[method]( "<" + nodeName + "></" + nodeName + ">" );
+		assert.ok( jQuery.nodeName( wrapper.children()[0], nodeName.toUpperCase() ), "Tagnames with special characters" );
+	}
+
+	jQuery.each( specialChars, function ( i, character ) {
+		nodeName = "valid" + character + "tagname";
+
+		assertSupport( "html", nodeName );
+		assertSupport( "append", nodeName );
```
gibson042: I'd like to see `jQuery( "<" + nodeName + ">" )` tests as well, and negative tests for [name-terminating whitespace characters](https://html.spec.whatwg.org/multipage/syntax.html#tag-name-state).

src/core/var/rsingleTag.js
gibson042: It's only used in association with `jQuery.parseHTML`; any gains would be unnoticeable in the cost of creating elements.

LeonardoBraga: Got it. I'll update the code and amend the PR.

markelog:
Friendly ping

test/unit/manipulation.js
LeonardoBraga: > (...) and negative tests for name-terminating whitespace characters.

Could you please elaborate on this request? Do you mean test support for elements with tags ending in one of the characters `\x20\t\r\n\f` i.e. `<div\t>`?

LeonardoBraga:
@gibson042 Please ignore my previous question. I think I got what you meant and updated the PR.

test/unit/manipulation.js
```diff
@@ -500,6 +500,53 @@ QUnit.test( "html(String) tag-hyphenated elements (Bug #1987)", function( assert
 	assert.equal( j.children().text(), "text", "Tags with multiple hypens behave normally" );
 } );
 
+QUnit.test( "html(String) respect the spec for tag names (gh-2005) ", function( assert ) {
```
timmywil: Thank you for this! I think my only note is to be more specific about how we're respecting the spec here.

LeonardoBraga: My pleasure to help!

Would ...
> html(String) respect the spec on how to handle special and name-terminating chars for tag names (gh-2005)

... better the describe the changes?

timmywil: Works for me!

LeonardoBraga: @timmywil Done.

timmywil: @LeonardoBraga Are you sure? It doesn't look like the change got committed.

LeonardoBraga: @timmywil My apologies. It's updated now.

LeonardoBraga:
Just pinging to know if there's anything else I can do to help the team land this PR. Thanks!

src/manipulation.js
```diff
@@ -26,7 +26,7 @@ define( [
 	dataPriv, dataUser, acceptData ) {
 
 var
-	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([\w:-]+)[^>]*)\/>/gi,
+	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([a-z][^\x20\t\r\n\f\/\0>]*)[^>]*)\/>/gi,
```
dmethvin: Does this regex need the two negated `>` in a row? There's one in the modified part and one that was there originally.

test/unit/manipulation.js
```diff
@@ -500,6 +500,53 @@ QUnit.test( "html(String) tag-hyphenated elements (Bug #1987)", function( assert
 	assert.equal( j.children().text(), "text", "Tags with multiple hypens behave normally" );
 } );
 
+QUnit.test( "html(String) respect the spec on how to handle special and name-terminating chars for tag names (gh-2005)", function( assert ) {
+
+	assert.expect( 120 );
+
+	var wrapper, child, nodeName,
+		nameTerminatingChars = "\x20-\t-\r-\n-\f".split( "-" ),
+		specialChars = "[ ] { } _ - = + \\ ( ) * & ^ % $ # @ ! ~ ` ' ; ?".split( " " );
```
dmethvin: Should this include any non-ASCII characters as well? If they're allowed we should be testing them.

src/manipulation.js
LeonardoBraga: Without the two negated `>`, some tests fail such as [this one](https://github.com/jquery/jquery/blob/master/test/unit/manipulation.js#L1619-L1621). It seems that before the `\w` was taking care of the job of the first negation, but now we're finding the name-terminating characters explicitly, we'll need to take care of this ourselves.

test/unit/manipulation.js
LeonardoBraga: Surely - added 10 non-ASCII characters to the list. Please let me know if adding more is needed.

src/manipulation.js
dmethvin: OK that makes sense, thanks!

test/unit/manipulation.js
```diff
@@ -500,6 +500,53 @@ QUnit.test( "html(String) tag-hyphenated elements (Bug #1987)", function( assert
 	assert.equal( j.children().text(), "text", "Tags with multiple hypens behave normally" );
 } );
 
+QUnit.test( "html(String) respect the spec on how to handle special and name-terminating chars for tag names (gh-2005)", function( assert ) {
+
+	assert.expect( 160 );
+
+	var wrapper, child, nodeName,
+		nameTerminatingChars = "\x20-\t-\r-\n-\f".split( "-" ),
```
gibson042: This should also include `/` and `>`.

gibson042: Or rather, those cases should also be tested. It looks like `>` is covered implicitly, so maybe just add `buildChild( method, "<" + nodeName + "/>" )` into the helpers.

test/unit/manipulation.js
```diff
@@ -500,6 +500,53 @@ QUnit.test( "html(String) tag-hyphenated elements (Bug #1987)", function( assert
...
+		specialChars = "[ ] { } _ - = + \\ ( ) * & ^ % $ # @ ! ~ ` ' ; ? ¥ « µ λ ⊕ ≈ ξ ℜ ♣ €".split( " " );
```
gibson042: @jquery/core, thoughts on embedding non-ASCII characters directly? I _think_ we use escape sequences everywhere else, but AFAIK every environment will handle UTF-8.

dmethvin: I thought about it too, we don't allow non-ASCII in the built source but I don't think it is a problem in our unit test files.

dmethvin: Also I don't think we want to be crazy enough to try surrogate pairs and the like. I just wanted a few basic tests and these are good enough for me at least.

gibson042:
Sorry for the delay, @LeonardoBraga. Did you check the size impact of moving all the `\/\0>` sequence to the front of their character sets?

Other than these minor comments, it looks good to me.

test/unit/manipulation.js
gibson042: Likewise.

gnarf: Yeah - build output should not have them, if the server doesn't send a utf8 header having `var π` will explode things --  d3 has been suspect to this issue for a while: https://github.com/mbostock/d3/issues/1195

tests probably don't matter so much.

LeonardoBraga:
@gibson042 Moving `\/\0>` to the front of the exclusion sequences produces the following differences:

```
  =     +2 dist/jquery.js                                                   
  =     -1 dist/jquery.min.js                                               
```

Based on my (limited) knowledge about the libraries usage in general, I lean towards benefiting the minified version (by having `\/\0>` in the beginning of their exclusions lists). The PR was updated to address the comments above and move `\/\0>` to the beginning. Please let me know if any additional changes are required.

LeonardoBraga:
Just touching base to see if there's anything else I can do to help you guys have this PR landed. Thanks!

mgol:
@LeonardoBraga Sorry for the delay, we're preparing for 1.12/2.2 simultaneous release, your PR is scheduled for 3.0.0 so we'll get to it once we're done with 1.12/2.2. :)

gibson042:
Thanks for your patience, @LeonardoBraga.

LeonardoBraga:
Thank **you** all! I'm just glad I was able to help. :-)

PR #2648, opened 2015-10-13
CSS: Make show/hide/toggle methods a module

### LINKED ISSUE #2193: Break out .show(), .hide(), and .toggle() to a new module
Once #2180 lands, cleave a new module out of showHide so that it could be excluded when not needed.

dmethvin:
Fixes #2193

Unit test changes some uses of .show() and .hide() to .css( "display", ... ),
there was already an implicit assumption in several of the existing tests.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -70,6 +70,7 @@ module.exports = function( grunt ) {
 					ajax: [ "manipulation/_evalUrl", "event/ajax" ],
 					callbacks: [ "deferred" ],
 					css: [ "effects", "dimensions", "offset" ],
+					"css/showHide": [ "effects" ],
 					sizzle: [ "css/hiddenVisibleSelectors", "effects/animatedSelector" ]
 				}
 			}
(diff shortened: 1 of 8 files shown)


mzgol:
LGTM

Gruntfile.js
```diff
@@ -70,6 +70,7 @@ module.exports = function( grunt ) {
 					ajax: [ "manipulation/_evalUrl", "event/ajax" ],
 					callbacks: [ "deferred" ],
 					css: [ "effects", "dimensions", "offset" ],
+					"css/showHide": [ "effects" ],
```
timmywil: While showHide does depend on effects, I don't think removing showHide should also remove effects.

timmywil: However, removing effects should remove showHide: `"effects": [ "css/showHide" ],`

dmethvin: If you don't remove effects, showHide will be [pulled back in](https://github.com/jquery/jquery/blob/67b76f5986a78d868553b9aec0d0192f5259a078/src/effects.js#L11) I think?

On the opposite side, you can remove effects but still want to use the non-animated `.show()` and `.hide()` methods so I don't think it should be force-removed.

timmywil: You're right. I was only thinking of animations pertaining to `.show()` and `.hide()`, not all animations that need to show and hide elements.

zqw:
good

Issue #2055, opened 2015-02-03
Sizzle.attr TypeError


rafaelchavesfreitas:
The code below throws an Uncaught TypeError Exception because Sizzle.attr tries to call getAttribute of a comment element but comments does not have this function.

```Javascript
$(function(){
    $('body').on('DOMNodeInserted', '*[data-test]', function () { 
		console.log(this); 
	}).append('<!-- new comment -->');
});
```

gibson042:
Related: https://github.com/jquery/sizzle/pull/256#issuecomment-40119966

We could punt on this, but it's so small that we might as well come up with a valid test and cherry-pick fc2ba2e1361126c39f955437ee025cfca3bffa65 from `compat` (where it somewhat indirectly prevents this issue from surfacing).

timmywil:
:+1: for nodeType check.

timmywil:
It looks like we already cherry-picked that commit.

timmywil:
9d820fbd

PR #2659, opened 2015-10-18
Event: Ensure delegation doesn't error on comment nodes

timmywil:
From my preliminary tests, the `DOMNodeInserted` event is still supported in our supported browsers.

Fixes gh-2055

### DIFF
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -1852,6 +1852,33 @@ QUnit.test( "delegated event with intermediate DOM manipulation (#13208)", funct
 	jQuery( "#anchor2" ).trigger( "click" );
 } );
 
+QUnit.test( "ignore comment nodes in event delegation (gh-2055)", function( assert ) {
+	assert.expect( 1 );
+
+	// Test if DOMNodeInserted is supported
+	// This is a back-up for when DOMNodeInserted support
+	// is eventually removed from browsers
+	function test() {
+		var ret = false;
+		var $fixture = jQuery( "#qunit-fixture" );
+		$fixture.on( "DOMNodeInserted", function() {
+			ret = true;
+			$fixture.off( "DOMNodeInserted" );
+		} ).append( "<div></div>" );
+		return ret;
+	}
+
+	var $foo = jQuery( "#foo" ).on( "DOMNodeInserted", "[id]", function() {
+		assert.ok( true, "No error thrown on comment node" );
+	} ),
+		$comment = jQuery( document.createComment( "comment" ) )
+			.appendTo( $foo.find( "#sap" ) );
+
+	if ( !test() ) {
+		fireNative( $comment[0], "DOMNodeInserted" );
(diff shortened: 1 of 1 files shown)


dmethvin:
This was fixed in Sizzle, right? 

Seems like the test is going to be brittle since these events are deprecated and will be removed eventually. If this is the only text-node event we need to guard against, how about if you add a check for DOMNodeInserted in this unit test before doing the actual test? That way we could do something like `assert.ok( true, "DOM Mutation events not supported")`.

gibson042:
Sizzle still throws on text node input, but 9d820fbde6d89bc7a06e2704be61cf6c0b4d6e3c in jQuery protects it from seeing them.

test/unit/event.js
```diff
@@ -1852,6 +1852,14 @@ QUnit.test( "delegated event with intermediate DOM manipulation (#13208)", funct
 	jQuery( "#anchor2" ).trigger( "click" );
 } );
 
+QUnit.test( "ignore comment nodes in event delegation (gh-2055)", function( assert ) {
+	assert.expect( 1 );
+
+	jQuery( "#foo" ).on( "DOMNodeInserted", "[id]", function() {
+		assert.ok( true, "No error thrown on comment node" );
+	} ).find( "#sap" ).append( "<!-- comment -->" );
```
gibson042: The deprecation of `DOMNodeInserted` event could be addressed by a plan B that uses [fireNative](https://github.com/jquery/jquery/blob/0fb84fa8ccefcd07febf282fd7b80262ad70add7/test/data/testinit.js#L114-L123) to synthesize an event on a comment node.

timmywil:
Added fallback for when DOMNodeInserted is no longer supported.

test/unit/event.js
```diff
@@ -1852,6 +1852,33 @@ QUnit.test( "delegated event with intermediate DOM manipulation (#13208)", funct
...
+			ret = true;
+			$fixture.off( "DOMNodeInserted" );
+		} ).append( "<div></div>" );
+		return ret;
+	}
+
+	var $foo = jQuery( "#foo" ).on( "DOMNodeInserted", "[id]", function() {
+		assert.ok( true, "No error thrown on comment node" );
+	} ),
+		$comment = jQuery( jQuery.parseHTML( "<!-- comment -->" ) )
```
gibson042: This should be `document.createComment`, since we don't assert that `jQuery.parseHTML` handles comments (and in fact [explicitly avoid doing so](https://github.com/timmywil/jquery/blob/gh-2055/test/unit/core.js#L609)).

timmywil: Well, I wouldn't call that explicitly avoiding it, just explicitly accounting for their possible absence, but I forgot that some browsers remove them.

timmywil: I see, we mean the same thing, I think.

Issue #2590, opened 2015-09-12
Offset: Fractions test fails in Chrome 45


Krinkle:
Two days ago, TestSwarm updated the jquery-core browser set from Chrome 43/44 to Chrome 44/45.

This revealed a test failure.

<http://swarm.jquery.org/result/619430>:
> offset: fractions (see #7730 and #7885) (1, 1, 2)
> Check top
> Expected: 1000
> Result: 999.984375

Looks like Chrome 45 is more precise than our tests expect.

mzgol:
Yeah, I've noticed it even before updating the browsers as I've seen it on Chrome locally but I wanted to be sure it's not just my machine. But thanks for the report anyway!

mzgol:
Actually, it seems Chrome 45 is *less* precise than older versions. I'll need to investigate further, I don't want to just decrease the test expectation without making sure this output is acceptable.

dmethvin:
@mzgol I'm going to take this one for a jqsummit project.

PR #2660, opened 2015-10-18
Offset: allow small differences in offset.top

Queeniebee:
Fixes gh-2590

### DIFF
--- a/test/unit/offset.js
+++ b/test/unit/offset.js
@@ -548,8 +548,10 @@ QUnit.test( "fractions (see #7730 and #7885)", function( assert ) {
 
 	result = div.offset();
 
-	assert.equal( result.top, expected.top, "Check top" );
-	assert.equal( result.left, expected.left, "Check left" );
+        // Support: Chrome 45-46+
+        // In recent Chrome these values differ a little.
+        assert.ok( Math.abs( result.top - expected.top ) < 0.25, "Check top within 0.25 of expected" );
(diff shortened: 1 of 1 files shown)


mzgol:
Hmm, why is @ajpiano marked as an author of a commit? :)

test/unit/offset.js
```diff
@@ -548,7 +548,7 @@ QUnit.test( "fractions (see #7730 and #7885)", function( assert ) {
 
 	result = div.offset();
 
-	assert.equal( result.top, expected.top, "Check top" );
+	assert.equal( Math.abs( result.top - expected.top ) < 0.25, true, "Check top within 0.25 of expected" );
```
mzgol: Since we only check within a range for top, not left, it'd be good to have a support comment, like:

    // Support: Chrome 45-46+
    // In recent Chrome those values differ a little.

dmethvin:
No wonder the CLA check didn't fail! We're using an old @ajpiano machine, had to upgrade everything. 

I'll just have her amend the commit to update the author info and we can force-push the branch.

test/unit/offset.js
timmywil: Also, maybe we can make this `assert.ok( expression )` instead of `.assert.equal( expression, true )`.

dmethvin:
Closed #2660 via 9f9e204bba41b7a9cde5ba7e065d817ef8b18c41

Issue #2662, opened 2015-10-18
Make jQuery.isNumeric consistent


timmywil:
Some people understand `isNumeric` as, "can I do math on this thing?" Others are asking, "Is this actually a number?" Now that we have no internal usage of `jQuery.isNumeric`, I think we can choose one or the other and update the documentation. Right now, it is inconsistent and a confusing confluence of the two.

timmywil:
@dmethvin, @gibson042 and I propose allowing all input and returning true for strings and numbers – whether primitive or object – that can be numeric.

timmywil:
```js
jQuery.isNumeric = function( input ) {
  var num,
    type = jQuery.type( input );
  return ( type === "number" || type === "string" ) && ( input - parseFloat( input ) + 1 ) >= 0;
};
```

dmethvin:
I added the behavior change tag since there's that one strange case with a custom `.toString()` that we will no longer handle. It can just be mentioned in the release notes.

PR #2663, opened 2015-10-18
Core: make isNumeric limited to strings and numbers

timmywil:
- This is a little more bulletproof in regards to future JS type additions.
- It is only different than what we currently have in one case.
- While the documentation has good examples, we need to be more explicit about how our isNumeric tries to coerce strings.

Fixes gh-2662

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -210,12 +210,12 @@ jQuery.extend( {
 
 	isNumeric: function( obj ) {
 
-		// parseFloat NaNs numeric-cast false positives (null|true|false|"")
-		// ...but misinterprets leading-number strings, particularly hex literals ("0x...")
-		// subtraction forces infinities to NaN
-		// adding 1 corrects loss of precision from parseFloat (#15100)
-		var realStringObj = obj && obj.toString();
-		return !jQuery.isArray( obj ) && ( realStringObj - parseFloat( realStringObj ) + 1 ) >= 0;
+		// As of jQuery 3.0, isNumeric is limited to
+		// strings and numbers (primitives or objects)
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -210,12 +210,11 @@ jQuery.extend( {
 
 	isNumeric: function( obj ) {
 
-		// parseFloat NaNs numeric-cast false positives (null|true|false|"")
-		// ...but misinterprets leading-number strings, particularly hex literals ("0x...")
-		// subtraction forces infinities to NaN
-		// adding 1 corrects loss of precision from parseFloat (#15100)
-		var realStringObj = obj && obj.toString();
-		return !jQuery.isArray( obj ) && ( realStringObj - parseFloat( realStringObj ) + 1 ) >= 0;
+		// isNumeric is now limited to strings and numbers (primitives or objects)
```
dmethvin: The word "now" will be a bit confusing over time, I'd just remove it.

src/core.js
```diff
@@ -210,12 +210,12 @@ jQuery.extend( {
 
 	isNumeric: function( obj ) {
 
-		// parseFloat NaNs numeric-cast false positives (null|true|false|"")
-		// ...but misinterprets leading-number strings, particularly hex literals ("0x...")
-		// subtraction forces infinities to NaN
-		// adding 1 corrects loss of precision from parseFloat (#15100)
```
stevemao: I would like to keep this comment here :)
@timmywil can we bring it back?

timmywil: Yea, that shouldn't have been removed.

Issue #2647, opened 2015-10-13
$().prop("tabIndex") returns wrong value (JSBin example included)


aschmalzhaf:
Hello, 

$(...).prop("tabIndex") behaves differently in JQuery 1.11.3 and 2.1.4 when Internet Explorer 11 is used. 
This is true at least for images (dom nodes of type <img></img>) that doesn't have an attribute "tabindex" explicitly given. 

Please see also the playground example here: 
http://jsbin.com/herehipoke/1/edit?html,output

When changing the URL of the loaded jQuery version, you can see that the tabIndex is different ("0" vs. "-1"). 

This is because of changes in attributes/porp.js file...
The propHook for tabIndex is different: 

The current code line (master and 2.1 stable) show the code of propHooks - tabIndex as follows:
```javascript
get: function( elem ) {
return elem.hasAttribute( "tabindex" ) || rfocusable.test( elem.nodeName ) || elem.href ?
elem.tabIndex : -1;
}
```

1.11.1 looks like that:
```javascript
get: function( elem ) {
var tabindex = jQuery.find.attr( elem, "tabindex" );
return tabindex ?
parseInt( tabindex, 10 ) :
rfocusable.test( elem.nodeName ) || rclickable.test( elem.nodeName ) && elem.href ? 0 : -1;
}
```

I've seen that jQuery Compat 3.0 Alpha uses the 1.11 logic again. Maybe it would make sense to move that kind of logic also into jQuery 2/3 as Internet Explorers are still supprted (at least the newer ones starting from 9 upwards). 

Best regards, 
Alexander

mzgol:
Thanks for the report!

jQuery Compat 3.0 is the continuation of the jQuery 1.x line so nothing changed here; jQuery 3.0 is a continuation of the jQuery 2.x line and the code there is the same as in 2.x.

It does seem the `compat` code is still needed on the main branch as IE 11 is still broken here. Fortunately it seems fixed in Edge 12.

We'll most likely not fix it in jQuery 2.1.x (there are many bug fixes people would like to have fixed there, we can't backport all of them) but it'd be good to fix it for 3.0.0.

mzgol:
[...]

PR #2664, opened 2015-10-18
Attributes: fix tabIndex on <img> in IE11

Queeniebee:
Fixes gh-2647

### DIFF
--- a/src/attributes/prop.js
+++ b/src/attributes/prop.js
@@ -5,7 +5,8 @@ define( [
 	"../selector"
 ], function( jQuery, access, support ) {
 
-var rfocusable = /^(?:input|select|textarea|button)$/i;
+var rfocusable = /^(?:input|select|textarea|button)$/i,
+    rclickable = /^(?:a|area)$/i;
 
 jQuery.fn.extend( {
 	prop: function( name, value ) {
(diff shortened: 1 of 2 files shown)


src/attributes/prop.js
```diff
@@ -55,10 +56,19 @@ jQuery.extend( {
 	propHooks: {
 		tabIndex: {
 			get: function( elem ) {
-				return elem.hasAttribute( "tabindex" ) ||
-					rfocusable.test( elem.nodeName ) || elem.href ?
-						elem.tabIndex :
-						-1;
+
+			    // elem.tabIndex doesn't always return the
```
dmethvin: Do we need a `// Support` comment?

src/attributes/prop.js
```diff
@@ -55,10 +56,19 @@ jQuery.extend( {
...
+			    // correct value when it hasn't been explicitly set
+			    // http://fluidproject.org/blog/2008/01/09/getting-setting-and-removing-tabindex-values-with-javascript/
```
dmethvin: This link seems broken for me?

mzgol: I'll switch it to the Web Archive one in a followup. :) We have the same link on compat so we might need to fix both.

src/attributes/prop.js
mzgol: Good remark, I'll add it later.

mgol: Fixed in 9cb89bf91d034ec7166d9215e2f80fa765292975.

src/attributes/prop.js
mgol: Fixed in 9cb89bf91d034ec7166d9215e2f80fa765292975.

src/attributes/prop.js
markelog: I wonder how long your todo list, if "later" is four months :)

mgol: Too long. :-) I'm working on that!

Issue #1759, opened 2014-10-21
removeAttr("selected") should not set property to false


mzgol:
Originally reported by wchen@… at: http://bugs.jquery.com/ticket/14633
<p>
<a class="ext-link" href="http://jsfiddle.net/b85D4/5/" rel="nofollow"><span class="icon">&nbsp;</span>http://jsfiddle.net/b85D4/5/</a>
</p>
<p>
In Firefox the selected value is 1. In Chrome the selected value is 2.
</p>
<p>
removeAttr has special handing for boolean attributes to set the corresponding DOM property to false: <a class="ext-link" href="https://github.com/jquery/jquery/blob/master/src/attributes/attr.js#L86" rel="nofollow"><span class="icon">&nbsp;</span>https://github.com/jquery/jquery/blob/master/src/attributes/attr.js#L86</a>
</p>
<p>
Removing an attribute and setting the DOM property to false are very different. In this case, as specified (<a class="ext-link" href="http://www.whatwg.org/specs/web-apps/current-work/multipage/the-button-element.html#dom-option-selected" rel="nofollow"><span class="icon">&nbsp;</span>http://www.whatwg.org/specs/web-apps/current-work/multipage/the-button-element.html#dom-option-selected</a>), setting the property to false means setting the dirtiness to true. This changes how the option should react to "selected" attribute changes. In the test case, it means that adding the "selected" attribute at the end should not select the option.
</p>
<p>
Firefox respects the specified behaviour and does not select the second option.
</p><p>Issue reported for jQuery 2.0.3</p>

AurelioDeRosa:
Will this change be landed in 3.0?

bzbarsky:
@timmywil Is there an ETA on when this will actually appear in a jQuery release?  Right now doing this:

                $('#foo').removeAttr('selected');
                $('#foo').attr('selected', 'selected');

on an `<option id="foo">` will leave it unselected.  And people keep reporting this as a browser bug every so often...

timmywil:
@bzbarsky That is a different issue. To set the dynamic selected state, use the property instead of the attribute. The selected attribute is used to set the default state for when the page loads.

```js
$("#foo").prop("selected", true);
```

dmethvin:
I don't see a difference in behavior though: http://jsbin.com/lipunucamu/edit?html,js,output

timmywil:
@dmethvin You can set the default state like that if the element hasn't been interacted with yet.

bzbarsky:
[...]

PR #2665, opened 2015-10-18
Attributes: do not set properties to false when removing booleans

timmywil:
- This is just legacy behavior that somehow stuck around. No docs changes needed.

Fixes gh-1759


dmethvin:
LGTM.

Looking at this code, I wonder if we should just check for the presence of `elem.removeAttribute` rather than checking `elem.nodeType === 1`. That's a separate issue though.

PR #2671, opened 2015-10-22
Ajax: Only form-encode requests with a body

### LINKED ISSUE #2658: Only change `%20` to `+` for application/x-www-form-urlencoded
This resurrects [trac-3400](http://bugs.jquery.com/ticket/3400) which has the argument for the change. Currently we *always* change encoded spaces (`%20`) to the plus sign (`+`) in `jQuery.param()`. That is only correct behavior for posted form data, but not other cases. The conversion should be moved to `jQuery.ajax()` and only be done when the method is POST.

dmethvin:
Fixes #2658 

This is a (minor) behavior change for `jQuery.param()` and `jQuery.serialize()` since they no longer convert `%20` to `+`, and for `jQuery.ajax()` when the `data` ends up on the URL.

I had some conflict trying to determine the right place to put this conversion. If we want prefilters to see it this would have to be moved higher. However, the prefilters have the opportunity to change all kinds of stuff that might affect whether we do this at all. So I decided to do it a bit further down once the `hasContent` variable is set and the content-type is known. Other suggestions welcome.

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -20,7 +20,6 @@ var
 	// #7653, #8125, #8152: local protocol detection
 	rlocalProtocol = /^(?:about|app|app-storage|.+-extension|file|res|widget):$/,
 	rnoContent = /^(?:GET|HEAD)$/,
-	rformContent = /x-www-form-urlencoded/,
 	rprotocol = /^\/\//,
 
 	/* Prefilters
@@ -604,8 +603,9 @@ jQuery.extend( {
 					cacheURL + ( rquery.test( cacheURL ) ? "&" : "?" ) + "_=" + nonce++;
 			}
 
-		// Change '%20' to '+' if this is encoded form content (gh-2658)
-		} else if ( s.data && s.processData && rformContent.test( s.contentType ) ) {
+		// Change '%20' to '+' if this is encoded form body content (gh-2658)
+		} else if ( s.data && s.processData &&
+			( s.contentType || "" ).indexOf( "application/x-www-form-urlencoded" ) === 0 ) {
 			s.data = s.data.replace( r20, "+" );
 		}
 
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -12,6 +12,7 @@ define( [
 ], function( jQuery, document, rnotwhite, location, nonce, rquery ) {
 
 var
+	r20 = /%20/g,
 	rhash = /#.*$/,
 	rts = /([?&])_=[^&]*/,
 	rheaders = /^(.*?):[ \t]*([^\r\n]*)$/mg,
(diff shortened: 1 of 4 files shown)


src/ajax.js
```diff
@@ -19,6 +20,7 @@ var
 	// #7653, #8125, #8152: local protocol detection
 	rlocalProtocol = /^(?:about|app|app-storage|.+-extension|file|res|widget):$/,
 	rnoContent = /^(?:GET|HEAD)$/,
+	rformContent = /x-www-form-urlencoded/,
```
gibson042: If we're testing by regex, this should be `/^application\/x-www-form-urlencoded[\x20\t]*(?:;.*|)$/` for maximum [spec](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) adherence. Equivalently, we could test by `(s.contentType || "").split( /[\x20\t;]+/ )[ 0 ] === "application/x-www-form-urlencoded"`. But in either case, both the below test and [jsonp](https://github.com/dmethvin/jquery/blob/2658-form-encoding/src/ajax/jsonp.js#L28-L29) should use the same logic.

test/unit/ajax.js
```diff
@@ -1129,6 +1129,32 @@ QUnit.module( "ajax", {
...
+		};
+	} );
+
+	ajaxTest( "jQuery.ajax() - data - URL data isn't form encoded (gh-2658)", 1, function( assert ) {
+		return {
+			url: "bogus.html",
+			data: { devo: "A Beautiful World" },
+			type: "get",
+			beforeSend: function( _, s ) {
+				assert.strictEqual( s.url, "bogus.html?devo=A%20Beautiful%20World", "data is NOT '+'-encoded" );
```
gibson042: This is wrong; the [GET logic](https://html.spec.whatwg.org/multipage/forms.html#submit-mutate-action) still calls for adding `+`s if using the (default) `application/x-www-form-urlencoded` encoding.

dmethvin: Here is the idea I thought of during the meeting today. Since the default content-type for `$.ajax` is `application/x-www-form-urlencoded`, this must  continue to be done for GET if we want to be spec compliant as @gibson042 points out above. However, now that we've fixed `$.param` the user would have the ability to set content-type to avoid the conversion. They can either do that through the ajax options on the call (good) or `$.ajaxSetup` (not a good idea due to side-effects to other callers). We probably need an example of a good content-type to use in the docs, although that may depend on the server's config and behavior.

gibson042: As pointed out in the meeting, it is strange—but not tremendously so—to let `contentType` affect query string generation. That said, this doesn't seem like the end of the world. I'd be ok with it, but also ok with just dropping the assertion (essentially leaving percent-encoding vs. plus-encoding unspecified for automatically generated query strings).

dmethvin: OK, since we've gotten several complaints about `+`-encoding of URL strings in the past I am ok with leaving it as well. I will remove this URL test for GET leaving that behavior unspecified, but add a few more unit tests for the POST case including one with a non-default content-type.

PR #2675, opened 2015-10-26
info

Enolak:


mzgol:
https://github.com/jquery/jquery/pull/1263#issuecomment-17630863

Issue #2100, opened 2015-02-17
Check document.readyState === "interactive" on $(document).ready


stackia:
https://github.com/jquery/jquery/blob/master/src/core/ready.js#L81

As the comment mentioned, IE9 will probably give an incorrect "interactive" readyState. But that only affects IE9. Should "interactive" be accepted on other browsers? I'm using require.js to load my scripts and exactly jQuery and many other scripts are loaded after DOMContentLoaded but before fully loaded. Scripts depending on $(document).ready have to wait until load event is triggered.

Currently I use <code>document.readyState === "interactive" ? init() : $(document).ready(init);</code> to get around this problem.

https://github.com/ded/domready/blob/master/src/ready.js#L16
This is another implementation of domReady which accept "interactive" readyState on browsers excluding IE.

Is there any reason why jQuery doesn't do that?

dmethvin:
To recast the question in the form of a bug/feature, I think it would be whether jQuery should support being lazy-loaded so that it would run `.ready()` handlers before `window.onload` fires. You can read through http://bugs.jquery.com/ticket/10067 for background. It seems like we could use the `.doScroll` hack that domready uses.

mzgol:
> It seems like we could use the `.doScroll` hack that domready uses.

What `doScroll` hack? ;) It was needed for IE<8, now the code looks [like that](https://github.com/jquery/jquery/blob/9368a8cd8ca81902b8359188668973e489fca6db/src/core/ready.js#L120-L126).

mzgol:
FWIW, this is the mentioned hack (last commit before dropping IE<8): https://github.com/jquery/jquery/blob/a467f8653a6fab2903148df80ab0ce9f5f4fd04f/src/core/ready.js#L125-L152

dmethvin:
I meant the check for `.doScroll()` they use in domready. It turns out that method went away in IE11, so by using that as an indirect feature check you only penalize IE9 and IE10. At  least that's what I think it would mean.

mzgol:
Ah, the library domready, not our core/ready. ;) Sorry!

timmywil:
@dmethvin To sum it up, we're thinking we could check if "interactive" as long as we use a scroll hack to determine interactive where not supported?

dmethvin:
[...]

PR #2678, opened 2015-10-28
Core: use interactive to evaluate dom ready

timmywil:
- Now that dom ready is always synchronous, we don't have to deal with ensuring sync execution order.
- I actually couldn't reproduce the IE9 issue where DOM below scripts is not actually interactive, but I ran into a different issue with IE10 (I think it was mostly a test issue, but it had to do with document.write behaving differently so I didn't mess with it). Either way, I just kept the proposed solution of checking doScroll, effectively excluding IE9-10.

Fixes gh-2100

### DIFF
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -70,10 +70,10 @@ jQuery.ready.promise = function( obj ) {
 
 		// Catch cases where $(document).ready() is called
 		// after the browser event has already occurred.
-		// We once tried to use readyState "interactive" here,
-		// but it caused issues like the one
-		// discovered by ChrisS here: http://bugs.jquery.com/ticket/12282#comment:15
-		if ( document.readyState === "complete" ) {
+		// Support: IE9-10 only
+		// Older IE sometimes signals "interactive" too soon
+		if ( document.readyState === "complete" ||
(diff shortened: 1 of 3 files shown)


timmywil:
@dmethvin @mikesherov Would like feedback from you two specifically given your familiarity with the old tickets.

Krinkle:
Hm.. so in the old code it seems that if the code runs after DOMContentLoaded (readyState interactive) but before window-load (readyState complete) then it wouldn't trigger immediately but fallback to window-load.

The commit mentions interactive but the code does not. I assume `!== loading` is used as proxy for  `== interactive` here. We may want to document in the code why that is.

On a related note: I tried to work on a polyfill for a readyState promise (whatwg/html#127) and ran into a bug on Android 2.3 where `interactive` is reported as `loaded` (and there is no `complete`). See <https://gist.github.com/Krinkle/fc0b4d4626f2409219ee>. I guess that is why we use `!== loading`?

timmywil:
@Krinkle Yes, I believe that was why originally. I used it here just because it's shorter.

dmethvin:
The logic LGTM. I'd put a support comment there, something to the effect of:

```js
// Support: IE9-10 only
// Older IE signals "interactive" before scripts are ready
```

PR #2680, opened 2015-10-29
Change comment structure

zachpanz88:
Maintain the same comment structure throughout the entire project

### DIFF
--- a/build/release/release-notes.js
+++ b/build/release/release-notes.js
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
-/*
- * jQuery Release Note Generator
- */
+
+// jQuery Release Note Generator
 
 var http = require( "http" ),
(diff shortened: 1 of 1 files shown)


dmethvin:
I don't think that file is even used anymore since we switched from Trac to Github Issues. Timmy, do you use it during a release? At this point I think it can be removed.

@zachpanz88 I'm not sure that comment deviates from our style, it's a banner comment. The actual components of the library use JSCS to check style automatically (see the Gruntfile) so they should be okay. We don't currently check all the unit test files, if you're interested in tackling that let us know. It should just be a case of adding them to the JSCS list in the Gruntfile and then editing the files.

timmywil:
@dmethvin Yup, I think it can be removed.

timmywil:
File removed.

PR #2682, opened 2015-10-31
Ajax: Don't throw exceptions on binary data response

### LINKED ISSUE #2498: XHR throws an exception accessing responseText, if responseType === "arraybuffer"

dmethvin:
Fixes #2498

The added unit test shows how this could be used to support an
ArrayBuffer return, but $.ajax does not support it natively.
The goal with this change was to avoid the exception.

### DIFF
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -88,12 +88,10 @@ jQuery.ajaxTransport( function( options ) {
 									xhrSuccessStatus[ xhr.status ] || xhr.status,
 									xhr.statusText,
 
-									// XHR2 responseText throws an exception on binary data,
-									// return it raw and let a converter or caller handle it
-									// https://xhr.spec.whatwg.org/#the-responsetext-attribute
-									// (trac-11426, gh-2498)
-									xhr.responseType === "arraybuffer" ||
-									xhr.responseType === "blob" ||
+									// Support: IE9 only
+									// IE9 has no XHR2 but throws on binary (trac-11426)
+									// For XHR2 non-text, let the caller handle it (gh-2498)
+									( xhr.responseType || "text" ) !== "text"  ||
 									typeof xhr.responseText !== "string" ?
 										{ binary: xhr.response } :
 										{ text: xhr.responseText },
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -88,12 +88,15 @@ jQuery.ajaxTransport( function( options ) {
 									xhrSuccessStatus[ xhr.status ] || xhr.status,
 									xhr.statusText,
 
-									// Support: IE9
-									// Accessing binary-data responseText throws an exception
-									// (#11426)
-									typeof xhr.responseText === "string" ? {
-										text: xhr.responseText
-									} : undefined,
+									// XHR2 responseText throws an exception on binary data,
+									// return it raw and let a converter or caller handle it
+									// https://xhr.spec.whatwg.org/#the-responsetext-attribute
+									// (trac-11426, gh-2498)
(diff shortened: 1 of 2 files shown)


src/ajax/xhr.js
```diff
@@ -88,12 +88,15 @@ jQuery.ajaxTransport( function( options ) {
 									xhrSuccessStatus[ xhr.status ] || xhr.status,
 									xhr.statusText,
 
-									// Support: IE9
```
dmethvin: I removed this comment because basically all XHR2 impls are supposed to throw on binary data in cases like ArrayBuffer.

src/ajax/xhr.js
```diff
@@ -88,12 +88,15 @@ jQuery.ajaxTransport( function( options ) {
...
-									// Accessing binary-data responseText throws an exception
-									// (#11426)
-									typeof xhr.responseText === "string" ? {
-										text: xhr.responseText
-									} : undefined,
+									// XHR2 responseText throws an exception on binary data,
+									// return it raw and let a converter or caller handle it
+									// https://xhr.spec.whatwg.org/#the-responsetext-attribute
+									// (trac-11426, gh-2498)
+									xhr.responseType === "arraybuffer" ||
```
dmethvin: Alternatively, we could look for the cases we wanted to force into the `text` path, but that's a little more code since XHR1 doesn't have `.responseType` and we actually want to handle JSON or XML conversions ourselves via `text`.

gibson042: > If `responseType` is not the empty string or "text", throw an InvalidStateError exception.

So maybe our logic should be
```js
(xhr.responseType || "text") !== "text" || typeof xhr.responseText !== "string" ?
	{ binary: xhr.response } :
	{ text: xhr.responseText }
```

dmethvin: That will force the `json` and `document` (XML) cases down the wrong branch, or at least wrong for the current logic in ajax.js that wants to convert the `text` result explicitly rather than letting XHR do it. This was due to various bugs and/or inability to control cases like [trac-13276](http://bugs.jquery.com/ticket/13276). @jaubourg says in that ticket he could only make it happen in IE9 so maybe we could hew closer to the spec once we drop IE9 support after 3.0? At that point we should be able to drop the `typeof` check as well and just look at `.responseType`.

mzgol: > once we drop IE9 support after 3.0

You're really forward-thinking! ]:->

FWIW, even Angular 2 will support even IE9 in the end (Amadeus that works on Angular with Google has to support IE9 because of its high market share in some countries like South Korea) so I don't envision us dropping support soon.

dmethvin: Per the discussion in the core team meeting I'm going to try the approach proposed by @gibson042 .

dmethvin: Yep, that works. I was mistakenly thinking that XHR *set* the `.responseType` property based on the returned headers, but it is only set by the caller.

dmethvin:
OK, this should take care of it, I tried to make the comments a bit less wordy and added back the IE9 comment because the `typeof` check is only needed there.

gibson042:
LGTM

Issue #2079, opened 2015-02-10
native abort does not abort jqXHR in 2.1.x


salomvary:
If the native XMLHttpRequest is aborted (eg. calling abort(), suspending the computer or unloading the page) jqXHR does not fire any callbacks. Expected: fire error and complete callbacks.

Test code to reproduce:
```js
var wrappedXhr;
$.ajax({
    xhr: function() {
      wrappedXhr = jQuery.ajaxSettings.xhr();  
      return wrappedXhr;
    },
    url: '/echo/json/?delay=100',
    complete: console.log.bind(console, 'jQuery complete'),
    error: console.log.bind(console, 'jQuery error'),
    success: console.log.bind(console, 'jQuery success')
})
.progress(console.log.bind(console, 'jQuery progress'))
.fail(console.log.bind(console, 'jQuery fail'))
.always(console.log.bind(console, 'jQuery always'));
wrappedXhr.abort();
```

[Test code on jsfiddle](http://jsfiddle.net/salomvary/agg3ovxz/6/).

Note: the bug is **not** present in 1.11.x and 3.0 (master) versions.

dmethvin:
Since 1.11.x polls it would seem to be okay, as you say. Not really sure why 3.0/master is okay though, seems like we would need an `onabort` hander to be notified of this. I'm not seeing any notifications if I change the version to master.

timmywil:
Should we add an onabort handler in master?

dmethvin:
Yes, it looks like we need to add one so that compat and master will have the same behavior.

timmywil:
So, after digging into the issue, this isn't the blocker I thought it was. While we don't try to hide the native XHR, we don't encourage direct manipulation either. If you use jQuery's abort method instead, everything works fine: http://jsfiddle.net/timmywil/agg3ovxz/9/.

@dmethvin Do you think this is still worth a fix? One minor speedbump is that the status text for the OP's case in compat is "error" when it should be "abort", so fixing this requires changes on both branches and I'm not sure it's something we necessarily support.

salomvary:
@timmywil using jQuery's abort method is not an option when the *browser aborts* the request for some reason (computer suspend for example).

dmethvin:
So just to refresh myself without having to re-investigate, does this affect our current 3.0 master or 3.0 compat? The initial comment said it did not.

mzgol:
[...]

PR #2684, opened 2015-11-02
Ajax: trigger error callback on native abort

timmywil:
Fixes gh-2079

### DIFF
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -38,7 +38,7 @@ QUnit.module( "ajax", {
 	);
 
 	ajaxTest( "jQuery.ajax() - success callbacks", 8, function( assert ) {
-	    return {
+		return {
 			setup: addGlobalEvents( "ajaxStart ajaxStop ajaxSend ajaxComplete ajaxSuccess", assert ),
 			url: url( "data/name.html" ),
 			beforeSend: function() {
@@ -431,6 +431,25 @@ QUnit.module( "ajax", {
 				assert.strictEqual( xhr.readyState, 0, "XHR readyState indicates successful abortion" );
 			},
 			error: true,
+			complete: function() {
+				assert.ok( true, "complete" );
+			}
+		};
+	} );
+
+	ajaxTest( "jQuery.ajax() - native abort", 2, function( assert ) {
+		return {
+			url: url( "data/name.php?wait=1" ),
+			xhr: function() {
+				var xhr = new window.XMLHttpRequest();
+				setTimeout( function() {
+					xhr.abort();
+				}, 100 );
+				return xhr;
+			},
+			error: function( xhr, msg ) {
+				assert.strictEqual( msg, "error", "Native abort triggers error callback" );
+			},
 			complete: function() {
 				assert.ok( true, "complete" );
 			}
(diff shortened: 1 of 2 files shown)


test/unit/ajax.js
```diff
@@ -38,7 +38,7 @@ QUnit.module( "ajax", {
 	);
 
 	ajaxTest( "jQuery.ajax() - success callbacks", 8, function( assert ) {
-	    return {
+			return {
```
scottgonzalez: Bad indentation.

timmywil: Actually, this converts spaces to tabs.

mzgol: Yeah, but ut;s still indented one tab too many, isn't it?

test/unit/ajax.js
```diff
@@ -437,6 +437,26 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - native abort", 2, function( assert ) {
+		return {
+			url: url( "data/name.php?wait=1" ),
+			xhr: function() {
+				var xhr = new window.XMLHttpRequest();
+				setTimeout( function() {
+					xhr.statusText = "abort";
```
gibson042: Why?

test/unit/ajax.js
timmywil: Good call.

test/unit/ajax.js
timmywil: Debugging, woops.

timmywil:
Bah, IE9 doesn't have onabort. The only other solution I see is to use .onreadystatechange, but that would change other stuff as well.

dmethvin:
Could we just document that IE9 doesn't support this? It would be a shame to let IE9 prevent us from doing the right thing for everyone else.

timmywil:
That's probably better than switching everything back to `onreadystatechange`. Anyone against this?

gibson042:
Looking at the compat code, I don't think use of `onreadystatechange` would be that cumbersome.

timmywil:
Afaict, it does require determining success/error manually again, which has historically been cumbersome.

timmywil:
Yea, we end up losing a lot of advantages.

gibson042:
No, we can keep the current structure and only use `onreadystatechange` for the abort case.

timmywil:
Ah, well I tried that first, but `onreadystatechange` is called before other handlers, and brings up some IE9 quirks, such as not being to access any properties on a manual, native abort – which makes it difficult to test – and not having a solid way to tell the difference between abort and other errors.

gibson042:
For example, verify `readyState === 4` and then delay its body (inside of which is an `if ( callback )` guard) by a tick so that it only executes if none of `onload`/`onerror`/`onabort` have.

timmywil:
Ah, a `setTimeout` could do the trick to allow other callbacks to run first.

timmywil:
@dmethvin @gibson042 This works. Let me know if you have ideas to get it smaller.

jaubourg:
> Ah, a `setTimeout` could do the trick to allow other callbacks to run first.

Except for synchronous requests, right?

timmywil:
> Except for synchronous requests, right?

Right

PR #2687, opened 2015-11-04
Event: Add basic unit tests for event aliases

dmethvin:
I don't think we need anything fancy here. A few questions inline.

### DIFF
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -2816,25 +2816,23 @@ QUnit.test( "originalEvent property for Chrome, Safari, Fx & Edge of simulated e
 	jQuery( "#donor-input" ).trigger( "focus" );
 } );
 
-if ( jQuery.fn.click ) {
-	QUnit.test( "Event aliases", function( assert ) {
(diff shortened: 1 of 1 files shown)


test/unit/event.js
```diff
@@ -2816,6 +2816,28 @@ QUnit.test( "originalEvent property for Chrome, Safari, Fx & Edge of simulated e
...
+
+	jQuery.each( aliases, function( i, name ) {
+
+		if ( !jQuery.fn[ name ] ) {
+			assert.ok( true, "no method present for jQuery.fn." + name );
+			return;
+		}
+
+		// e.g. $(elem).click(...).click();
+		jQuery( document.body )[ name ]( function( event ) {
```
gibson042: An element under `#qunit-fixture` would probably be better than `document.body`, to make leaks even less likely.

dmethvin:
All feedback incorporated. Unfortunately it's not possible at the moment to run all the event tests because they make scattered use of event aliases themselves. I opened #2691 for that.

gibson042:
LGTM.

mzgol:
LGTM

Issue #1864, opened 2014-11-13
Make event triggering a separate module


dmethvin:
If a custom build doesn't need manually triggered events, it could save a bunch of bytes by removing `jQuery.event.trigger()`, `jQuery.event.simulate()`, `jQuery.fn.trigger()`, and `jQuery.fn.triggerHandler()`. This would mostly be useful for the 2.x (now 3.x-master) branch because the 1.x (3.x-compat) branch relies pretty heavily on manually triggered events to fix up IE8. On master, manual triggering is used to create bubbling focusin/out events and fire ajax events, but there are other approaches devs can use for both situations.

mzgol:
@timmywil @dmethvin Moving to 3.1.0 as it won't be a breaking change and so doesn't have to go in 3.0.0. Revert the change if you disagree.

dmethvin:
@timmywil I can grab this if you want.

timmywil:
@dmethvin Feel free!

PR #2692, opened 2015-11-05
Event: Separate trigger/simulate into its own module

dmethvin:
Fixes gh-1864

This also pulls the focusin/out special event into its own module, since that
depends on simulate(). NB: The ajax module triggers events pretty heavily.

I *think* I got all module plumbing right but it could definitely use a good look.

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -5,9 +5,11 @@ define( [
 	"./ajax/var/location",
 	"./ajax/var/nonce",
 	"./ajax/var/rquery",
+
 	"./core/init",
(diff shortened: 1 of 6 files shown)


timmywil:
I'd like to leave the support test in its own module. It doesn't affect the build size, so it's just a matter of consistently putting any support tests in their own modules.

timmywil:
But LGTM

dmethvin:
I realized this was missing the README docs for the new module, so I added that in the commit to master.

PR #2693, opened 2015-11-06
Event: Make event aliases optional in unit tests

dmethvin:
Relatively easy fixes, although I did have a question about one strange VML-related test.

### DIFF
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -900,49 +900,6 @@ QUnit.test( "mouseenter, mouseleave don't catch exceptions", function( assert ) 
 	}
 } );
 
-if ( jQuery.fn.click ) {
-
-	QUnit.test( "trigger() shortcuts", function( assert ) {
-		assert.expect( 5 );
-
-		var counter, clickCounter,
-			elem = jQuery( "<li><a href='#'>Change location</a></li>" ).prependTo( "#firstUL" );
-		elem.find( "a" ).on( "click", function() {
-			var close = jQuery( "spanx", this ); // same with jQuery(this).find("span");
-			assert.equal( close.length, 0, "Context element does not exist, length must be zero" );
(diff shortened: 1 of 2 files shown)


test/unit/event.js
```diff
@@ -900,49 +900,6 @@ QUnit.test( "mouseenter, mouseleave don't catch exceptions", function( assert )
 	}
 } );
 
-if ( jQuery.fn.click ) {
```
dmethvin: Just moved this down and made it a `QUnit.skip` test.

timmywil:
LGTM

test/unit/event.js
```diff
@@ -2843,6 +2800,46 @@ QUnit.test( "originalEvent property for Chrome, Safari, Fx & Edge of simulated e
...
+	assert.equal( counter, 1, "Check that click, triggers onclick event handler also" );
+
+	clickCounter = 0;
+	jQuery( "#simon1" )[ 0 ].onclick = function() {
+		clickCounter++;
+	};
+	jQuery( "#simon1" ).click();
+	assert.equal( clickCounter, 1, "Check that click, triggers onclick event handler on an a tag also" );
+
+	// test that special handlers do not blow up with VML elements (#7071)
```
dmethvin: This doesn't look like it belongs on an event alias unit test. It also is appending to head and not removing the markup. Do we need this test at all now? @timmywil do you remember anything about https://github.com/jquery/jquery/commit/69866fd2e5ba2456c2ae84a5ba711a1db2a25091 ?

timmywil: I don't recall the specifics, but it looks like we had that xml in the index.html to start. That said, I think it's safe to remove at the end of the test. Regardless, I agree that it should be moved to non-alias tests.

mzgol:
LGTM.

Issue #1729, opened 2014-10-20
Have $(window).outerWidth() include the scrollbar width


mzgol:
Originally reported by antti@… at: http://bugs.jquery.com/ticket/14989
<p>
It is sometimes annoying that $(window).width() or $(window).outerWidth() do not include the scrollbar width in their value. Especially when trying to match the window width to the CSS media queries.
</p>
<p>
Currently e.g. if you have a media query in the CSS such as
</p>
<pre class="wiki">@media screen and (max-width: 1600px) {
  body {background: red;}
}
</pre><p>
When the red background is actually applied to the document by the browser, jQuery tells the window width is less than 1600px (1600 - scrollbar width).
</p>
<p>
This is annoying e.g. in cases where there are some mobile-specific stuff that needs to be done in both, CSS and JS. And they would need to be done at the exactly same width of the document/window.
</p>
<p>
However, plain JavaScript returns 1600 as the window width when asked at the point of the breakpoint as follows:
</p>
<pre class="wiki">console.log(window.innerWidth);
</pre><p>
Is it possible to get either $(window).width() or $(windor).outerWidth() working consistently with the browser's actual width? $(document).width() and $('body').width() will not return the same as "window.innerWidth" either, they return the same as $(window).width() and $(window).outerWidth().
</p>
<p>
Fixing this would greatly help making consistent code with the CSS media queries.
</p>
<p>
Possible workarounds currently are:
</p>
<ul><li>Using plain javascript, i.e. window.innerWidth
</li><li>Checking $('...').is(':visible') for some element that is visible only in the target width
</li></ul><p>
Here's an example of what I mean:
<a class="ext-link" href="http://jsfiddle.net/aq5vb/" rel="nofollow"><span class="icon">&nbsp;</span>http://jsfiddle.net/aq5vb/</a>
</p><p>Issue reported for jQuery 2.1.0</p>

timmywil:
@dmethvin Just to sum up, is the idea here to leave `.width(window)` as-is, but add scroll bar width to `.outerWidth(window)`? Although, I think @scottgonzalez still makes a good case against it.

dmethvin:
Actually, I read the comment as, "it looks like doing this is okay." @scottgonzalez can you clarify? I agree it's possible this will break code *somewhere* . On the other hand, if we can't think of any and this is a 3.0 it would be nice to expose some way to get this dimension.

scottgonzalez:
[...]

PR #2694, opened 2015-11-06
Dimensions: outerWidth/Height include scrollbar

dmethvin:
Fixes gh-1729

This is a behavior change so shouldn't be picked back into 1.12/2.2.

### DIFF
--- a/src/dimensions.js
+++ b/src/dimensions.js
@@ -19,6 +19,11 @@ jQuery.each( { Height: "height", Width: "width" }, function( name, type ) {
 
 				if ( jQuery.isWindow( elem ) ) {
 
+					// $( window ).outerWidth/Height return w/h including scrollbars (gh-1729)
+					if ( funcName.indexOf( "outer" ) === 0 ) {
+						return elem[ "inner" + name ];
+					}
+
 					// As of 5/8/2012 this will yield incorrect results for Mobile Safari, but there
 					// isn't a whole lot we can do. See pull request at this URL for discussion:
 					// https://github.com/jquery/jquery/pull/764
--- a/src/dimensions.js
+++ b/src/dimensions.js
@@ -20,14 +20,9 @@ jQuery.each( { Height: "height", Width: "width" }, function( name, type ) {
 				if ( jQuery.isWindow( elem ) ) {
 
 					// $( window ).outerWidth/Height return w/h including scrollbars (gh-1729)
-					if ( funcName.indexOf( "outer" ) === 0 ) {
-						return elem[ "inner" + name ];
-					}
-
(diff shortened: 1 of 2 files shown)


src/dimensions.js
```diff
@@ -19,6 +19,11 @@ jQuery.each( { Height: "height", Width: "width" }, function( name, type ) {
 
 				if ( jQuery.isWindow( elem ) ) {
 
+					// $( window ).outerWidth/Height return w/h including scrollbars (gh-1729)
```
dmethvin: This can probably be combined into a single return with the value below but I wanted to resolve the status of that comment first.

src/dimensions.js
```diff
@@ -19,6 +19,11 @@ jQuery.each( { Height: "height", Width: "width" }, function( name, type ) {
...
+					if ( funcName.indexOf( "outer" ) === 0 ) {
+						return elem[ "inner" + name ];
+					}
+
 					// As of 5/8/2012 this will yield incorrect results for Mobile Safari, but there
```
dmethvin: Does this bug still exist in Mobile Safari? If not let's remove the comment, if so I can add a `// Supports` for it. I don't have an iPhone to check and I'm not sure the emulators are reliable enough.

test/unit/dimensions.js
```diff
@@ -235,6 +236,45 @@ QUnit.test( "outerWidth()", function( assert ) {
 	div.remove();
 } );
 
+QUnit.test( "outerHeight()", function( assert ) {
```
dmethvin: Hard to believe, but we didn't have any unit test coverage for this API.

dmethvin: Oh, we *did* but it was further down the file and I missed it. I can remove this and/or beef up the tests there.

src/dimensions.js
mzgol: For such use cases emulators should be fine. Also, BrowserStack now has real iOS devices for iOS 7 & 8.

I've just checked and the issue exists in iOS 6 but doesn't in 7/8/9. We still support iOS 6 in 1.12/2.2 but not in 3.0 and this change is breaking so this is remark is moot now.

This makes sense, btw, as iOS 7 experienced a major redesign, the Safari scrolling behavior was also changed in a number of ways so it's not surprising behavior changed here as well.

mzgol: Interestingly enough, I see the bug in iOS 6 in the iPhone 4S simulator but not in the iPhone 5 one...

In any case, I'd assume we don't need to care about it anymore.

dmethvin: Thanks for checking!

dmethvin:
Updated to remove comment and fix unit tests.

Issue #1712, opened 2014-10-16
.width(value) sets incorrect width value if block has border and value specified in ems and box-sizing:border-box


mzgol:
Originally reported by tde: http://bugs.jquery.com/ticket/14038

API documentation specifies: "Note that .width("value") sets the content width of the box regardless of the value of the CSS box-sizing property."

There are two key factors leading to this bug

"box-sizing:border-box"
"border:1px solid black"
width specified in ems
Code that reproduces this bug

```html
<html>
	<head>
		<style>
			#container {width:100%;}
			#box {border:1px solid black; margin:0 auto; box-sizing:border-box;}
		</style>
		<script type="text/javascript" src="http://code.jquery.com/jquery-2.0.1.js"></script>
		<script>
			var $ = jQuery.noConflict();
			$(document).ready(function() {
				$('#box').width('10em'); // width should be 10em, but jquery set it to 12em
			});
		</script>
	</head>
	<body>
		<div id="container">
			<div id="box">
				<p>some text</p>
			</div>
		</div>
	</body>
</html>
```

Gerst20051:
:+1: just ran into this bug (is it a bug or expected behavior?) but as pointed out above setting the width using ```.css('width', width);``` works as expected.

mzgol:
@Gerst20051 We've accepted it as a bug and (currently) plan to fix it for 3.0.0. We still [have a lot to do](https://github.com/jquery/jquery/milestones/3.0.0) for 3.0.0 though!

PR #2695, opened 2015-11-06
Dimensions: properly manipulate non-px values

timmywil:
Fixes gh-1712

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -27,8 +27,6 @@ var
 	// except "table", "table-cell", or "table-caption"
 	// See here for display values: https://developer.mozilla.org/en-US/docs/CSS/display
 	rdisplayswap = /^(none|table(?!-c[ea]).+)/,
-	rnumsplit = new RegExp( "^(" + pnum + ")(.*)$", "i" ),
-
 	cssShow = { position: "absolute", visibility: "hidden", display: "block" },
 	cssNormalTransform = {
 		letterSpacing: "0",
@@ -59,11 +57,14 @@ function vendorPropName( name ) {
 }
 
 function setPositiveNumber( elem, value, subtract ) {
-	var matches = rnumsplit.exec( value );
+
+	// Any relative (+/-) values have already been
+	// normalized at this point
+	var matches = rcssNum.exec( value );
 	return matches ?
 
 		// Guard against undefined "subtract", e.g., when used as in cssHooks
-		Math.max( 0, matches[ 1 ] - ( subtract || 0 ) ) + ( matches[ 2 ] || "px" ) :
+		Math.max( 0, matches[ 2 ] - ( subtract || 0 ) ) + ( matches[ 3 ] || "px" ) :
 		value;
 }
 
@@ -336,16 +337,25 @@ jQuery.each( [ "height", "width" ], function( i, name ) {
 		},
 
 		set: function( elem, value, extra ) {
-			var styles = extra && getStyles( elem );
-			return setPositiveNumber( elem, value, extra ?
-				augmentWidthOrHeight(
+			var matches,
+				styles = extra && getStyles( elem ),
+				subtract = extra && augmentWidthOrHeight(
 					elem,
 					name,
 					extra,
 					jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
 					styles
-				) : 0
-			);
+				);
+
+			// Convert to pixels if value adjustment is needed
+			if ( subtract && ( matches = rcssNum.exec( value ) ) &&
+				( matches[ 3 ] || "px" ) !== "px" ) {
+
+				elem.style[ name ] = value;
+				value = jQuery.css( elem, name );
+			}
+
+			return setPositiveNumber( elem, value, subtract );
 		}
 	};
 } );
--- a/test/unit/dimensions.js
+++ b/test/unit/dimensions.js
@@ -418,36 +418,41 @@ QUnit.test( "getters on non elements should return null", function( assert ) {
 } );
 
 QUnit.test( "setters with and without box-sizing:border-box", function( assert ) {
-	assert.expect( 20 );
+	assert.expect( 60 );
 
-	// Support: Android 2.3 (-webkit-box-sizing).
-	var el_bb = jQuery( "<div style='width:114px;height:114px;margin:5px;padding:3px;border:4px solid white;-webkit-box-sizing:border-box;box-sizing:border-box;'>test</div>" ).appendTo( "#qunit-fixture" ),
-		el = jQuery( "<div style='width:100px;height:100px;margin:5px;padding:3px;border:4px solid white;'>test</div>" ).appendTo( "#qunit-fixture" ),
-		expected = 100;
+	var parent = jQuery( "#foo" ).css({ width: "200px", height: "200px", "font-size": "16px" }),
+		el_bb = jQuery( "<div style='margin:5px;padding:1px;border:2px solid black;box-sizing:border-box;'></div>" ).appendTo( parent ),
+		el = jQuery( "<div style='margin:5px;padding:1px;border:2px solid black;'></div>" ).appendTo( parent );
 
-	assert.equal( el_bb.width( 101 ).width(), expected + 1, "test border-box width(int) by roundtripping" );
-	assert.equal( el_bb.innerWidth( 108 ).width(), expected + 2, "test border-box innerWidth(int) by roundtripping" );
-	assert.equal( el_bb.outerWidth( 117 ).width(), expected + 3, "test border-box outerWidth(int) by roundtripping" );
-	assert.equal( el_bb.outerWidth( 118, false ).width(), expected + 4, "test border-box outerWidth(int, false) by roundtripping" );
-	assert.equal( el_bb.outerWidth( 129, true ).width(), expected + 5, "test border-box innerWidth(int, true) by roundtripping" );
+	jQuery.each( {
+		"number": { set: 100, expected: 100 },
+		"em": { set: "10em", expected: 160 },
+		"percentage": { set: "50%", expected: 100 }
+	}, function( units, values ) {
+		assert.equal( el_bb.width( values.set ).width(), values.expected, "test border-box width(" + units + ") by roundtripping" );
+		assert.equal( el_bb.innerWidth( values.set ).width(), values.expected - 2, "test border-box innerWidth(" + units + ") by roundtripping" );
+		assert.equal( el_bb.outerWidth( values.set ).width(), values.expected - 6, "test border-box outerWidth(" + units + ") by roundtripping" );
+		assert.equal( el_bb.outerWidth( values.set, false ).width(), values.expected - 6, "test border-box outerWidth(" + units + ", false) by roundtripping" );
+		assert.equal( el_bb.outerWidth( values.set, true ).width(), values.expected - 16, "test border-box innerWidth(" + units + ", true) by roundtripping" );
 
-	assert.equal( el_bb.height( 101 ).height(), expected + 1, "test border-box height(int) by roundtripping" );
-	assert.equal( el_bb.innerHeight( 108 ).height(), expected + 2, "test border-box innerHeight(int) by roundtripping" );
-	assert.equal( el_bb.outerHeight( 117 ).height(), expected + 3, "test border-box outerHeight(int) by roundtripping" );
-	assert.equal( el_bb.outerHeight( 118, false ).height(), expected + 4, "test border-box outerHeight(int, false) by roundtripping" );
-	assert.equal( el_bb.outerHeight( 129, true ).height(), expected + 5, "test border-box innerHeight(int, true) by roundtripping" );
+		assert.equal( el_bb.height( values.set ).height(), values.expected, "test border-box height(" + units + ") by roundtripping" );
+		assert.equal( el_bb.innerHeight( values.set ).height(), values.expected - 2, "test border-box innerHeight(" + units + ") by roundtripping" );
+		assert.equal( el_bb.outerHeight( values.set ).height(), values.expected - 6, "test border-box outerHeight(" + units + ") by roundtripping" );
+		assert.equal( el_bb.outerHeight( values.set, false ).height(), values.expected - 6, "test border-box outerHeight(" + units + ", false) by roundtripping" );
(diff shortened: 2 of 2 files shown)


src/css.js
```diff
@@ -58,13 +58,31 @@ function vendorPropName( name ) {
 	}
 }
 
+// Use font-size to convert to px
+function convertToPx( elem, value ) {
+	var fontSize = jQuery.css( elem, "font-size" );
+	elem.style.fontSize = value;
+	value = parseFloat( jQuery.css( elem, "font-size" ) );
```
dmethvin: This will force a synchronous layout, right? I suppose there's no way to avoid it but I wonder if there is a less expensive way to do it, for example with a different element attached to the body.

test/unit/dimensions.js
```diff
@@ -447,6 +447,39 @@ QUnit.test( "setters with and without box-sizing:border-box", function( assert )
 	assert.equal( el.outerHeight( 129, true ).height(), expected + 5, "test border-box innerHeight(int, true) by roundtripping" );
 } );
 
+QUnit.test( "setting ems on an element with box-sizing: border-box (gh-1712)", function( assert ) {
+	assert.expect( 20 );
+
+	var el_bb = jQuery( "<div style='margin:5px;padding:1px;border:2px solid black;-webkit-box-sizing:border-box;box-sizing:border-box;'></div>" ).appendTo( "#nothiddendiv" ),
```
gibson042: I think this should use `.css( "box-sizing", "border-box" )` to avoid the vendor-prefixing. Also, why append to nothiddendiv instead of qunit-fixture?

test/unit/dimensions.js
```diff
@@ -447,6 +447,39 @@ QUnit.test( "setters with and without box-sizing:border-box", function( assert )
...
+		el = jQuery( "<div style='margin:5px;padding:1px;border:2px solid black;'></div>" ).appendTo( "#nothiddendiv" ),
+		// font-size of nothiddendiv is 16px
```
gibson042: Mmm, that would be better if it were explicit in this test.

src/css.js
```diff
@@ -58,13 +58,31 @@ function vendorPropName( name ) {
...
+	elem.style.fontSize = fontSize;
```
gibson042: This function has some problems:
* It doesn't work when `value` is expressed with `%`, because `height`/`width` and `font-size` have different referents
* It leaves an undesirable inline `font-size`
* (nit) It invokes `parseFloat` directly rather than relying on `jQuery.css( elem, prop, "" )` for coercion

src/css.js
```diff
@@ -58,13 +58,31 @@ function vendorPropName( name ) {
...
-	var matches = rnumsplit.exec( value );
-	return matches ?
+	var unit,
+		matches = rnumsplit.exec( value );
+	if ( matches ) {
+		unit = matches[ 2 ] || "px";
+
+		// Only need to convert if we're manipulating the value
+		if ( unit !== "px" && subtract ) {
+			matches[ 1 ] = convertToPx( elem, value );
```
gibson042: I _love_ this force-to-px strategy, but get the feeling that `setPositiveNumber` is the wrong place for it. What if we moved everything out into the height/width cssHooks (the only place providing `subtract` anyway), where we know which property is being set?
```js
set: function( elem, value, extra ) {
	var valueParts,
		styles = extra && getStyles( elem ),
		subtract = extra && augmentWidthOrHeight(
			elem,
			name,
			extra,
			jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
			styles
		);

	// Match units if we have margin/border/padding pixels to subtract
	if ( subtract && (valueParts = rcssNum.exec( value )) && (valueParts[ 3 ] || "px") !== "px" ) {
		elem.style[ name ] = value;
		value = jQuery.css( elem, name );
	}

	return setPositiveNumber( elem, value, subtract );
}
```

test/unit/dimensions.js
mzgol: This doesn't need the prefix since we dropped Android 2.3 testing of this file.

src/css.js
timmywil: Yea, the only issue I have with this is that we'd be doing the regex check twice. I wonder if there's some way to share the match with `setPositiveNumber`.

test/unit/dimensions.js
timmywil: Yea, some of this was copy/paste from the tests above these. The vendor-prefixed `box-sizing` was from there. It could be removed in both tests, tho.

test/unit/dimensions.js
timmywil: I suppose. Other tests would fail if the font-size on nothiddendiv was changed, but I'm fine with setting it in the test.

src/css.js
timmywil: > It doesn't work when value is expressed with %, because height/width and font-size have different referents.

Good point.

> It leaves an undesirable inline font-size.

We *could* avoid this one.

> It invokes parseFloat directly rather than relying on jQuery.css( elem, prop, "" ) for coercion.

I forgot about that.

src/css.js
timmywil: I was thinking that using an element added to the DOM would trigger layout even more.

test/unit/dimensions.js
timmywil: vendor-prefixed box-sizing removed.

src/css.js
timmywil: This is been removed.

test/unit/dimensions.js
timmywil: Tests now explicit.

src/css.js
timmywil: Change to use a modified version of your solution. This function is gone.

src/css.js
timmywil: I've made it so we reuse matches when possible. Also, removed `rnumsplit`. `rcssNum` seems like enough.

timmywil:
Added tests for percentages as well.

src/css.js
timmywil: @gibson042 Thanks for this, btw. I remember thinking of using `font-size` to get pixels long ago. Not sure why I was thinking that now.

mzgol:
I like that the source diff is quite small. :)

test/unit/dimensions.js
```diff
@@ -418,36 +418,44 @@ QUnit.test( "getters on non elements should return null", function( assert ) {
...
-	assert.equal( el.outerWidth( 129, true ).width(), expected + 5, "test border-box innerWidth(int, true) by roundtripping" );
-
-	assert.equal( el.height( 101 ).height(), expected + 1, "test border-box height(int) by roundtripping" );
-	assert.equal( el.innerHeight( 108 ).height(), expected + 2, "test border-box innerHeight(int) by roundtripping" );
-	assert.equal( el.outerHeight( 117 ).height(), expected + 3, "test border-box outerHeight(int) by roundtripping" );
-	assert.equal( el.outerHeight( 118, false ).height(), expected + 4, "test border-box outerHeight(int, false) by roundtripping" );
-	assert.equal( el.outerHeight( 129, true ).height(), expected + 5, "test border-box innerHeight(int, true) by roundtripping" );
+	assert.expect( 60 );
+
+	var parent = jQuery( "#qunit-fixture" ).css({ width: "200px", height: "200px", "font-size": "16px" }),
```
gibson042: Eeek! Pick a different `parent` so we're not screwing with the fixture container.

src/css.js
```diff
@@ -336,16 +333,23 @@ jQuery.each( [ "height", "width" ], function( i, name ) {
 		},
 
 		set: function( elem, value, extra ) {
-			var styles = extra && getStyles( elem );
-			return setPositiveNumber( elem, value, extra ?
-				augmentWidthOrHeight(
+			var matches = rcssNum.exec( value ),
```
gibson042: `matches` would be more clear (and [consistent](https://github.com/timmywil/jquery/blob/width-1712/src/css/adjustCSS.js#L6)) as `valueParts`.

src/css.js
```diff
@@ -58,12 +56,11 @@ function vendorPropName( name ) {
 	}
 }
 
-function setPositiveNumber( elem, value, subtract ) {
-	var matches = rnumsplit.exec( value );
```
gibson042: Stripping this `exec` will break our [other application](https://github.com/timmywil/jquery/blob/width-1712/src/css.js#L394) of `setPositiveNumber`.

src/css.js
gibson042: rcssNum allows leading `+=` and `-=`, which is invalid here. We should either keep `rnumsplit`, or verify `&& !matches[ 1 ]` in the condition below.

src/css.js
```diff
@@ -336,16 +333,23 @@ jQuery.each( [ "height", "width" ], function( i, name ) {
...
 					elem,
 					name,
 					extra,
 					jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
 					styles
-				) : 0
-			);
+				);
+
+			if ( subtract && ( matches[ 3 ] || "px" ) !== "px" ) {
```
gibson042: This should have an explanatory comment.

src/css.js
timmywil: Huh, interesting it didn't break the tests.

src/css.js
timmywil: I'd rather add the verification. As far as the variable name, I think the use of `matches` for regex checks is older than `valueParts`, and is actually what was used in `setPositiveNumber`.

test/unit/dimensions.js
timmywil: Heh, make up your mind.

src/css.js
gibson042: Yeah, it looks like we're missing some for setting `border` and `padding` to negative values.

src/css.js
timmywil: Yup, meant to add one.

src/css.js
gibson042: Verification works; if it were me you know I'd go with whichever yields a smaller gzip. And `matches` is fine here... I meant for that to be a comment about the `setPositiveNumber` parameter name, but I guess it's made moot by the other reasons to remove it.

test/unit/dimensions.js
gibson042: I said the `font-size` should be explicit (which I stand behind), but we should never be manipulating `#qunit-fixture` directly—only its _contents_ are fair game.

src/css.js
timmywil: Actually, I'm not sure the verification is necessary since all hooks are called *after* the value has been adjusted in `adjustCSS`.

src/css.js
timmywil: Match returned here.

test/unit/dimensions.js
timmywil: Yea, thought you were suggesting to use `#qunit-fixture` *and* be explicit about `font-size` when you asked whey I was appending to `#nothiddendiv`, but I think I misread that.

src/css.js
```diff
@@ -59,11 +57,14 @@ function vendorPropName( name ) {
 }
 
 function setPositiveNumber( elem, value, subtract ) {
-	var matches = rnumsplit.exec( value );
+
+	// Any relative (+/-) values have already been
+	// normalized at this point
+	var matches = rcssNum.exec( value );
```
gibson042: That's not true for e.g. `.css( "borderWidth", "+=2px" )`.

timmywil: I think it is, actually.

timmywil: screenshot 2015-11-09 16 00 18

src/css.js
```diff
@@ -336,16 +337,25 @@ jQuery.each( [ "height", "width" ], function( i, name ) {
...
 					extra,
 					jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
 					styles
-				) : 0
-			);
+				);
+
+			// Convert to pixels if value adjustment is needed
+			if ( subtract && ( matches = rcssNum.exec( value ) ) &&
+				( matches[ 3 ] || "px" ) !== "px" ) {
```
gibson042: Does this need a `!matches[ 1 ]` guard? Without it, `.width( "+=10%" )` will:
* `elem.style[ "width" ] = "+=10%"` (noop, or are there still supported environments where this throws?)
* `value = jQuery.css( elem, "width" )` (current width of {X}px)
* `return setPositiveNumber( elem, "{X}px", subtract )`
  * `var matches = rcssNum.exec( value )` (`[ "{X}px", undefined, X, "px" ]`)
  * `return Math.max( 0, X - ( subtract || 0 ) ) + "px" )`
* Set inline `width` to that value

Whereas with a guard, we'd just set an inline width of the obviously invalid `+=10%`.

timmywil: From what I can tell, you're forgetting that the value has already been adjusted to something settable by the time the hook is executed.

timmywil: I've come to this conclusion from looking at https://github.com/timmywil/jquery/blob/width-1712/src/css.js#L237. `adjustCSS` works out the real value to set. Then, later in the function, `hook.set` is called.

src/css.js
gibson042: /me slaps forehead

I guess that's kind of the point, huh?

src/css.js
gibson042: Yeah, you're right. Inbound relative values are made absolute before the set hook sees them.

test/unit/dimensions.js
```diff
@@ -418,36 +418,44 @@ QUnit.test( "getters on non elements should return null", function( assert ) {
...
+
+		assert.equal( el.height( values.set ).height(), values.expected, "test non-border-box height(" + units + ") by roundtripping" );
+		assert.equal( el.innerHeight( values.set ).height(), values.expected - 2, "test non-border-box innerHeight(" + units + ") by roundtripping" );
+		assert.equal( el.outerHeight( values.set ).height(), values.expected - 6, "test non-border-box outerHeight(" + units + ") by roundtripping" );
+		assert.equal( el.outerHeight( values.set, false ).height(), values.expected - 6, "test non-border-box outerHeight(" + units + ", false) by roundtripping" );
+		assert.equal( el.outerHeight( values.set, true ).height(), values.expected - 16, "test non-border-box innerHeight(" + units + ", true) by roundtripping" );
+	} );
+
+	// Reset styles
+	parent.css({ width: "", height: "", "font-size": "" });
```
gibson042: This reset can be removed.

timmywil: Removed.

Issue #1757, opened 2014-10-21
Problem with the context when loading a script in an iframe


mzgol:
Originally reported by jjgcastela@… at: http://bugs.jquery.com/ticket/14568
<p>
When you create a script node and append it to a head or a body of an iframe, the code is evaluated in the main window (where the jQuery was loaded) instead of in the window of the iframe.
I have been able to solve it adding the desired 'document' as the second argument of the 'globalEval' function this way:
</p>
<pre class="wiki">globalEval: function( code , doc) {
	var script,
	indirect = eval;
	code = jQuery.trim( code );
	if ( code ) {
		// If the code includes a valid, prologue position
		// strict mode pragma, execute code by injecting a
		// script tag into the document.
		if ((doc &amp;&amp; doc != document) || code.indexOf("use strict") === 1 ) {
			if(!doc) doc = document;
			script = doc.createElement("script");
			script.text = code;
			doc.head.appendChild( script ).parentNode.removeChild( script );
		} else {
			// Otherwise, avoid the DOM node creation, insertion
			// and removal by using an indirect global eval
			indirect( code );
		}
	}
}

</pre><p>
And then in the 'domManip' function, adding this parameter in the call to 'globalEval':
</p>
<pre class="wiki">jQuery.globalEval( node.textContent.replace( rcleanScript, "" ), doc );

</pre><p>Issue reported for jQuery 2.0.3</p>

peyerluk:
Here's an updated version of the current `globalEval` with a second param `context` analogouous to [jQuery.parseHTML()](http://api.jquery.com/jquery.parsehtml/):

```js
globalEval: function( code, context ) {
  context = context || document;
  var script = context.createElement( "script" );

  script.text = code;
  context.head.appendChild( script ).parentNode.removeChild( script );
}
```

PR #2696, opened 2015-11-06
Manipulation: execute scripts from iframe in the iframe's context

timmywil:
Fixes gh-1757

### DIFF
--- a/test/unit/manipulation.js
+++ b/test/unit/manipulation.js
@@ -2162,6 +2162,17 @@ testIframeWithCallback(
 	}
 );
 
+testIframeWithCallback(
+	"domManip executes scripts in iframes in the iframes' context",
+	"manipulation/scripts-context.html",
+	function( frameWindow, bodyElement, html, assert ) {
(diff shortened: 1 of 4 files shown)


test/unit/manipulation.js
```diff
@@ -2162,6 +2162,16 @@ testIframeWithCallback(
 	}
 );
 
+testIframeWithCallback(
+	"domManip executes scripts in iframes in the iframes' context",
+	"manipulation/scripts-context.html",
+	function( bodyElement, html, assert ) {
+		assert.expect( 1 );
+		jQuery( bodyElement ).append( html );
+		assert.ok( !window.scriptTest, "script executed in iframe context" );
```
mzgol: It would be good if we had a way of checking if it was set in the iframe context (and not just not set in the current one) but otherwise LGTM.

timmywil: Good call. Added the iframe window to arguments to make sure the variable is on that window.

mzgol:
LGTM

PR #2699, opened 2015-11-09
Ajax: Golf away 21 bytes

gibson042:


mzgol:
Nice!

src/ajax.js
```diff
@@ -462,14 +462,14 @@ jQuery.extend( {
...
-					return state === 2 ? responseHeadersString : null;
+					return completed ? responseHeadersString : null;
 				},
 
 				// Caches the header
 				setRequestHeader: function( name, value ) {
-					var lname = name.toLowerCase();
-					if ( !state ) {
-						name = requestHeadersNames[ lname ] = requestHeadersNames[ lname ] || name;
+					if ( completed == null ) {
```
dmethvin: Why not a strict comparison to `undefined` here?

gibson042: I reserve those for cases where we explicitly want to differentiate `null` and `undefined` in order to keep the minified output more compressible. Each `=== undefined` costs about a byte.

dmethvin: I'm okay with that as long as we document it somewhere. The first thing I did when I saw this was look for a place where it might be assigned `null` thinking the check for both was important. The style guide example [implies](http://contribute.jquery.org/style-guide/js/#equality) we do it to check for both.

gibson042: Any thoughts on how to document it? I see what you mean about the style guide, and even though this _technically_ fits because I'd allow `null` to pass, the fact is that it will never come up so it's arguably against the spirit of our style guide. I could also just suck it up and use `=== undefined` if you think the current approach is confusing.

dmethvin: What does everyone else think? To me the `=== undefined` is clearer and worth a byte.

timmywil: I'm ambivalent. I'd be fine with whichever.

Issue #2319, opened 2015-05-15
Single-member read methods should always return undefined for empty context


gibson042:
Derived from gh-2134

Known offenders (`$()[ method ]()` currently returns `null`):
* scrollTop
* scrollLeft
* width
* height
* innerWidth
* innerHeight
* outerWidth
* outerHeight

timmywil:
Following up on what I was saying in gh-2134, the (outer|inner)width/height ones will be probably be in one PR and the scroll ones in another. I think those can be two tickets. Atomized issues will be more helpful in tracking our progress (and more helpful to me when triaging). Any broad issues that are more guidelines to what specific actions need to be taken can be moved to the Roadmap. Make sense?

gibson042:
I think that runs the risk of incompleteness. I know that these eight methods need to change, but there might be more. Without a tracking ticket, we could land a fix for the scroll methods and a fix for the dimension methods, then erroneously consider the work done. And I really don't want to look in multiple places to understand what's going on.

timmywil:
There's only one place: the Roadmap. You're just tracking it in that instead of a ticket. The item on the Roadmap does not need to be removed just because tickets were closed. We can leave it there until we're sure the it's all done. That's its purpose. Another advantage to this is that we will no longer leave tickets lying open just because we're unsure if it's all done. No more ambiguity in issues. And combined with the Roadmap, there is no incompleteness.

timmywil:
Also, having a Roadmap and keeping it updated with the broader issues gives us all one place to see what's going on without sifting through the issues. I know I've found myself many times looking for that one ticket that explains the other ones. I'm proposing we put all that in one place.

gibson042:
The roadmap plus GitHub issues makes two places. And there's no [mention](https://github.com/jquery/jquery/search?q=roadmap) of the former in this repository, which makes the project even less accessible to would-be contributors. That's correctable, though, and I'm willing to go with it, but honestly don't see the advantage over adding a new issue tag.

markelog:
I didn't participate in the related ticket, so sorry for this question - why returned value should be `undefined`?

Sometimes it is known for getters to return `null` if there is nothing to return. I believe this is why it was done in jQuery in the first place and it is better corresponds with DOM methods, like `document.createElement("div").getAttribute("non-existent") // null` and it is common in many languages, although such methods usually deal with objects. 

On the other hand `Object.getOwnPropertyDescriptor({}, "foo"); // undefined`. 

[...]

PR #2701, opened 2015-11-09
Dimensions: Empty sets should return undefined

dmethvin:
Ref gh-2319

Split into two PRs since it affects two modules (dimensions and offset).

This reverts the patch done for http://bugs.jquery.com/ticket/12283 so it's either a wontfix or a worksforme.

### DIFF
--- a/src/dimensions.js
+++ b/src/dimensions.js
@@ -45,7 +45,7 @@ jQuery.each( { Height: "height", Width: "width" }, function( name, type ) {
 
 					// Set width or height on the element
 					jQuery.style( elem, type, value, extra );
-			}, type, chainable ? margin : undefined, chainable, undefined );
+			}, type, chainable ? margin : undefined, chainable );
 		};
 	} );
 } );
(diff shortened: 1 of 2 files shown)


src/dimensions.js
```diff
@@ -45,7 +45,7 @@ jQuery.each( { Height: "height", Width: "width" }, function( name, type ) {
 
 					// Set width or height on the element
 					jQuery.style( elem, type, value, extra );
-			}, type, chainable ? margin : undefined, chainable, null );
+			}, type, chainable ? margin : undefined, chainable, undefined );
```
gibson042: Again, I believe the last argument is unnecessary.

test/unit/dimensions.js
```diff
@@ -53,7 +53,7 @@ function testWidth( val, assert ) {
 
 	blah = jQuery( "blah" );
```
gibson042: Maybe we could fix this to `jQuery()` and give the variable a better name?

dmethvin: I'll do that too. These tests are pretty crusty.

PR #2702, opened 2015-11-09
Offset: offsetLeft/Top on empty set returns undefined

dmethvin:
~~Closes~~ Fixes gh-2319

Ref gh-2701

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -202,7 +202,7 @@ jQuery.each( { scrollLeft: "pageXOffset", scrollTop: "pageYOffset" }, function( 
 			} else {
 				elem[ method ] = val;
 			}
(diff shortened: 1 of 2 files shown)


src/offset.js
```diff
@@ -202,7 +202,7 @@ jQuery.each( { scrollLeft: "pageXOffset", scrollTop: "pageYOffset" }, function(
 			} else {
 				elem[ method ] = val;
 			}
-		}, method, val, arguments.length, null );
+		}, method, val, arguments.length, undefined );
```
gibson042: I'm pretty sure you can just omit the `undefined`.

dmethvin: Yep!

Issue #2048, opened 2015-01-30
Fix event delegation with selector-native


timmywil:
When event delegation started using Sizzle's seed handling for scoped matches, it broke event delegation when selector-native replaced Sizzle. Worse comes to worse, we can just defer to matchesSelector and eat the scope issue.

timmywil:
By the way, I was wrong about this being broken. I was thinking we weren't dealing with the `seed` argument in selector-native, but it turns out we were. There's still a behavior difference due to the nature of QSA's scoping problem, but that pretty much requires Sizzle to fix.

mzgol:
> There's still a behavior difference due to the nature of QSA's scoping problem, but that pretty much requires Sizzle to fix.

...or the new @gibson042's in-jQuery implementation that borrows some ideas from Sizzle but only as much as needed (see https://github.com/jquery/sizzle/issues/293#issuecomment-154581860 for more info). But that's a plan for the next major.

dmethvin:
At some point the [DOM `matches` and `findAll`](http://dev.w3.org/2006/webapi/selectors-api2/#interface-definitions) would help here, but nobody seems to be in a hurry to implement them.

mzgol:
Until we drop IE 11 (which will happen once we conquer Mars) we need all this parsing logic anyway.

timmywil:
> ...or the new @gibson042's in-jQuery implementation that borrows some ideas from Sizzle but only as much as needed (see jquery/sizzle#293 (comment) for more info). But that's a plan for the next major.

Or possibly later. Something I like about selector-native is that it doesn't even tokenize the selector, which would be required even in a streamlined engine. Rather than an in-jQuery implementation, I'd prefer to see a new version of Sizzle that drops all polyfills (for native selectors) and POS handling, and just manipulates the selector into something that can first go through qsa or matchesSelector, then filtered. Regardless, it won't be for 3.0, and it's something we can discuss later when the time comes.

Issue #1742, opened 2014-10-21
Make unit tests work with selector-native.js


mzgol:
Originally reported by dmethvin at: http://bugs.jquery.com/ticket/13537
<p>
It would be good to have the native selector engine run a clean subset of unit tests as a sanity check. It wouldn't hurt to wait on this until we get some user feedback though.
</p><p>Issue reported for jQuery 2.0b1</p>

timmywil:
I can resolve this when I get to a bug with selector-native.

mzgol:
Since we only have `master` now, I'm removing the `master-only` label, it's no longer useful.

PR #2703, opened 2015-11-10
Selector: pass jQuery unit tests with selector-native

timmywil:
- Ignore certain tests that are obviously not supported
- Beefed up the sortOrder, uniqueSort, isXMLDoc, and attr functions to pass our tests.

Fixes gh-1742
Fixes gh-2048

### DIFF
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -622,7 +622,7 @@ QUnit.test( "jQuery('html')", function( assert ) {
 		assert.equal( jQuery( "element:not(<div></div>)" ).length, 0,
 			"When html is within parens, do not recognize as html." );
 	} else {
-		assert.ok( true, "Complex :not not supported in selector-native" );
+		assert.ok( "skip", "Complex :not not supported in selector-native" );
 	}
 	assert.equal( jQuery( "\\<div\\>" ).length, 0, "Ignore escaped html characters" );
 } );
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -145,7 +145,7 @@ QUnit.test( "jQuery()", function( assert ) {
 		"Empty attributes object is not interpreted as a document (trac-8950)" );
 } );
 
-QUnit.test( "jQuery(selector, context)", function( assert ) {
+QUnit[ jQuery.find.compile ? "test" : "skip" ]( "jQuery(selector, context)", function( assert ) {
 	assert.expect( 3 );
 	assert.deepEqual( jQuery( "div p", "#qunit-fixture" ).get(), q( "sndp", "en", "sap" ), "Basic selector with string as context" );
 	assert.deepEqual( jQuery( "div p", q( "qunit-fixture" )[ 0 ] ).get(), q( "sndp", "en", "sap" ), "Basic selector with element as context" );
@@ -618,8 +618,12 @@ QUnit.test( "jQuery('html')", function( assert ) {
 
 	//equal( jQuery( "element[attribute=<div></div>]" ).length, 0,
 	//	"When html is within brackets, do not recognize as html." );
-	assert.equal( jQuery( "element:not(<div></div>)" ).length, 0,
-		"When html is within parens, do not recognize as html." );
+	if ( jQuery.find.compile ) {
+		assert.equal( jQuery( "element:not(<div></div>)" ).length, 0,
+			"When html is within parens, do not recognize as html." );
+	} else {
+		assert.ok( true, "Complex :not not supported in selector-native" );
+	}
 	assert.equal( jQuery( "\\<div\\>" ).length, 0, "Ignore escaped html characters" );
 } );
 
--- a/test/unit/effects.js
+++ b/test/unit/effects.js
@@ -1527,16 +1527,16 @@ QUnit.test( "Animate callbacks have correct context", function( assert ) {
 	this.clock.tick( 10 );
 } );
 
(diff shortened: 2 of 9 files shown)


mzgol:
This looks awesome at first glance!

test/unit/core.js
```diff
@@ -618,8 +618,12 @@ QUnit.test( "jQuery('html')", function( assert ) {
 
 	//equal( jQuery( "element[attribute=<div></div>]" ).length, 0,
 	//	"When html is within brackets, do not recognize as html." );
-	assert.equal( jQuery( "element:not(<div></div>)" ).length, 0,
-		"When html is within parens, do not recognize as html." );
+	if ( jQuery.find.compile ) {
+		assert.equal( jQuery( "element:not(<div></div>)" ).length, 0,
+			"When html is within parens, do not recognize as html." );
+	} else {
+		assert.ok( true, "Complex :not not supported in selector-native" );
```
gibson042: Is there a way to mark such assertions as skipped (e.g., `ok( true, … )` → `ok( "skip", … )`)? It won't bubble up to QUnit, but it might help us in the future.

dmethvin: I like the idea of marking individual tests somehow, and "skip" seems as good as any. Seems like it would be a useful QUnit feature as well?

test/unit/serialize.js
```diff
@@ -142,8 +142,12 @@ QUnit.test( "serialize()", function( assert ) {
 		"action=Test&radio2=on&check=on&hidden=&foo%5Bbar%5D=&name=name&search=search&email=dave%40jquery.com&number=43&select1=&select2=3&select3=1&select3=2&select5=3&T3=%3F%0D%0AZ&H1=x&H2=&PWD=&T1=&T2=YES&My%20Name=me&S1=abc&S3=YES&S4=",
 		"Multiple form serialization as query string" );
 
-	assert.equal( jQuery( "#form, #testForm :input" ).serialize(),
-		"action=Test&radio2=on&check=on&hidden=&foo%5Bbar%5D=&name=name&search=search&email=dave%40jquery.com&number=43&select1=&select2=3&select3=1&select3=2&select5=3&T3=%3F%0D%0AZ&H1=x&H2=&PWD=&T1=&T2=YES&My%20Name=me&S1=abc&S3=YES&S4=",
-		"Mixed form/input serialization as query string" );
+	if ( jQuery.find.compile ) {
+		assert.equal( jQuery( "#form, #testForm :input" ).serialize(),
```
gibson042: I think this test should execute regardless of selector engine: `#form, #testForm :input` → `#form, #testForm input, #testForm select, #testForm textarea, #testForm button`.

test/unit/effects.js
```diff
@@ -1527,7 +1527,7 @@ QUnit.test( "Animate callbacks have correct context", function( assert ) {
 	this.clock.tick( 10 );
 } );
 
-QUnit.test( "User supplied callback called after show when fx off (#8892)", function( assert ) {
+QUnit[ jQuery.find.compile ? "test" : "skip" ]( "User supplied callback called after show when fx off (#8892)", function( assert ) {
```
gibson042: Can we keep this test? The right way to do it would be testing visibility with `supportjQuery`, but it would also be acceptable to replace use of `:hidden`/`:visible`.

test/unit/event.js
```diff
@@ -1750,7 +1750,8 @@ QUnit.test( "jQuery.off using dispatched jQuery.Event", function( assert ) {
 		.remove();
 } );
 
-QUnit.test( "delegated event with delegateTarget-relative selector", function( assert ) {
+// selector-native does not support scope-fixing in delegation
+QUnit[ jQuery.find.compile ? "test" : "skip" ]( "delegated event with delegateTarget-relative selector", function( assert ) {
```
gibson042: This one is half-preservable.

test/unit/selector.js
```diff
@@ -261,7 +279,7 @@ QUnit.test( "disconnected nodes", function( assert ) {
 	assert.equal( $div.is( "div" ), true, "Make sure .is('nodeName') works on disconnected nodes." );
 } );
 
-QUnit.test( "disconnected nodes - jQuery only", function( assert ) {
+QUnit[ jQuery.find.compile ? "test" : "skip" ]( "disconnected nodes - jQuery only", function( assert ) {
```
gibson042: I think all the `jQuery only` suffixes can be dropped, since they were only here as a poor man's `QUnit.skip`.

gibson042:
Awesome work!

test/unit/core.js
gibson042: It's come up before (jquery/qunit/pull/434, jquery/qunit/issues/326), but never really solidified enough. I'll give it some thought.

timmywil: Sounds good.

test/unit/effects.js
timmywil: Will switch to supportjQuery.

test/unit/event.js
timmywil: True, but I think the preservable parts aren't the important parts.

test/unit/selector.js
timmywil: Removed

test/unit/serialize.js
timmywil: Done.

dmethvin:
Honestly I wasn't ever sure we'd get this one done. Great work!

PR #2706, opened 2015-11-11
Add bower.json for better dependency injection support

evanshortiss:
Since the project doesn't contain a bower.json, using tools like wiredep to inject dependencies is not possible.

Adding the bower.json with a "main" field resolves this.

Let me know if any fields aren't set to appropriate values

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,24 @@
+{
+  "name": "jquery",
+  "version": "3.0.0-alpha1+compat",
+  "homepage": "http://jquery.com",
+  "authors": [
+    "jQuery Foundation and other contributors"
+  ],
+  "description": "JavaScript library for DOM operations",
+  "main": "dist/jquery.js",
+  "keywords": [
+    "jquery",
+    "browser",
+    "javascript",
+    "library"
+  ],
+  "license": "MIT",
+  "ignore": [
+    "**/.*",
+    "node_modules",
+    "bower_components",
+    "test",
+    "tests"
+  ]
+}


mzgol:
Thanks for the PR but tagged commits [like 2.1.4](https://github.com/jquery/jquery/tree/2.1.4) do contain bower.json, it's generated during a release process. The main repo doesn't contain bower.json on purpose.

dmethvin:
Hey @evanshortiss, the main jquery repo doesn't have a `/dist` folder so even if we added this I don't think it would do what you expected. We don't want to require committing all the built files to our *development* repo on every change that we make, which this PR would imply we do. 

Is there anywhere we could put this information that would help people to understand why there is no `bower.json` file? Obviously that would require having the person look at that documentation before making a PR, so it would be good to know where you looked before creating your PR. Possible locations would be the CONTRIBUTING.md or README.md.

evanshortiss:
Thanks @mzgol I'll look into it. In the latest release the _bower.json_ is missing, but I can see in the build script there's a function to generate it.

@dmethvin I would have thought the same based on the repo, but your releases _do_ contain the _/dist_ folder so that's when it can be added. I'm assuming it's checked in, tagged, then deleted? 

Check out this [commit/tag](https://github.com/jquery/jquery/tree/f9846aec1dadc4bc80ca2df47d3ecc1b9f7a4f84), it has the _/dist_ files included.

timmywil:
@evanshortiss Future releases will be published to https://github.com/jquery/jquery-dist. The bower registry will be changed to point to that location and the `bower.json` will only be created there.

evanshortiss:
Thanks @timmywil, that's great news.

mzgol:
@timmywil Just to make sure - do we still need to go through jquery-dist now that we don't plan to publish compat? One of the motivations was that we couldn't have one tag point to 2 different places.

timmywil:
@mzgol Yes, we'll still use jquery-dist. There were other motivations. It makes the release process simpler and keeps our source and distribution files separate. It plays nicely with bower, and it makes it easy to see what was published.

Issue #2546, opened 2015-08-19
jQuery.when doesn't recognize solitary thenables


gibson042:
`jQuery.when( a )` does not look for a `then` method on its argument, which differs from its behavior in longer `jQuery.when( a, … )` invocations. It should treat non-`Deferred` thenables identically in both cases.

http://jsfiddle.net/0r62we79/9/

marcandre:
And I thought I was going crazy...
```
var failed = new Promise(function(resolve, reject) { reject('fail!'); });
$.when(failed).then(function(){console.log('Should not be here')});  // Not ok
$.when(failed, failed).then(function(){console.log('Should not be here either')}); // Ok
```
 Currently outputs "Should not be here" to the console.

Issue #2018, opened 2015-01-19
Remove single-promiseable exception in jQuery.when


gibson042:
From [http://api.jquery.com/jQuery.when](http://api.jquery.com/jQuery.when/):
> If a single Deferred is passed to `jQuery.when()`, its Promise object (a subset of the Deferred methods) is returned

For consistency, we should always create a new master Deferred.

~~We [follow through](http://jsfiddle.net/noyw6Lzd/), but the code—both ours and library consumers—is more complicated than it needs to be as a result. The upcoming major version bump is a chance to unify single- and multiple-argument calls to `jQuery.when`, if we agree that doing so is valuable.~~ (we did not)

dmethvin:
I agree it makes sense to create a new Deferred rather than borrow the first one, so this is good with me as well.

jaubourg:
If I read correctly, this is not just about not reusing the original object. It's about ensuring `$.when` always resolves as an array.

Currently, `$.when` is idempotent. Meaning that `$.when( $.when( "string" ) );` will resolve as "string" no matter the number of calls. Making it "less complicated than it needs to be" would mean that this would become `[ [ "string" ] ]` (and on and on). You can then forget about using `$.when` for handling inputs that may or may not be a Deferred/Promise because you would effectively change (and by change I mean obfuscate/complicate) the resolve value in the process.

The actual solution is to have another utility that would be used for arrays __explicitly__:

```js
jQuery.whenArray = function( array ) {
    return jQuery.when.call( jQuery, array );
};
```

dmethvin:
If we want `$.when` to accept any kind of thenable or standard Promise, I think we'd want to handle the `$.when( thenable )` case and create a new `Deferred` there. So always creating a new one seems like the best way to go for consistency and predictability.

I think that changing the array case would be a lot more visible and breaking, and a separate API might be the best way to "document" the confusion people have in trying to call `$.when` with an array.

jaubourg:
> If we want $.when to accept any kind of thenable or standard Promise, I think we'd want to handle the $.when( thenable ) case and create a new Deferred there. So always creating a new one seems like the best way to go for consistency and predictability.

You're perfectly right.

gibson042:
This is in fact a double request; shame on me. Sounds like we're in agreement in always creating a new Deferred, so :+1: there.

The other bit, correctly identified by @jaubourg, is me going off half-cocked about yet _another_ inconsistency in our API.

> Currently, `$.when` is idempotent

[...]

PR #2707, opened 2015-11-11
Deferred: synchronize single and multiple target handling in $.when

timmywil:
- This probably needs revision, but in the process of fixing #2546, I took a stab at #2018, but I'm not as familiar with this code as @jaubourg or @gibson042.
- -6 bytes

Fixes gh-2546
Fixes gh-2018

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -294,19 +294,17 @@ jQuery.extend( {
 	},
 
 	// Deferred helper
-	when: function( subordinate /* , ..., subordinateN */ ) {
+	when: function() {
 		var method,
 			i = 0,
 			resolveValues = slice.call( arguments ),
 			length = resolveValues.length,
 
 			// the count of uncompleted subordinates
-			remaining = length !== 1 ||
-				( subordinate && jQuery.isFunction( subordinate.promise ) ) ? length : 0,
+			remaining = length,
 
 			// the master Deferred.
-			// If resolveValues consist of only a single Deferred, just use that.
-			master = remaining === 1 ? subordinate : jQuery.Deferred(),
+			master = jQuery.Deferred(),
 
 			// Update function for both resolve and progress values
 			updateFunc = function( i, contexts, values ) {
@@ -316,14 +314,17 @@ jQuery.extend( {
 					if ( values === progressValues ) {
 						master.notifyWith( contexts, values );
 					} else if ( !( --remaining ) ) {
-						master.resolveWith( contexts, values );
+						master.resolveWith(
+							contexts.length === 1 ? contexts[ 0 ] : contexts,
+							values
+						);
 					}
 				};
 			},
 			progressValues, progressContexts, resolveContexts;
 
 		// Add listeners to Deferred subordinates; treat others as resolved
-		if ( length > 1 ) {
+		if ( length > 0 ) {
 			progressValues = new Array( length );
 			progressContexts = new Array( length );
 			resolveContexts = new Array( length );
@@ -345,14 +346,13 @@ jQuery.extend( {
 						updateFunc( i, progressContexts, progressValues )
 					);
 				} else {
-					--remaining;
+					updateFunc( i, resolveContexts, resolveValues )( resolveValues[ i ] );
 				}
 			}
-		}
 
 		// If we're not waiting on anything, resolve the master
-		if ( !remaining ) {
-			master.resolveWith( resolveContexts, resolveValues );
+		} else {
+			master.resolveWith();
 		}
 
 		return master.promise();
--- a/test/unit/deferred.js
+++ b/test/unit/deferred.js
@@ -667,7 +667,6 @@ QUnit.test( "jQuery.when", function( assert ) {
 		"undefined": undefined,
 		"a plain object": {},
 		"an array": [ 1, 2, 3 ]
-
 	}, function( message, value ) {
 		assert.ok(
 			jQuery.isFunction(
@@ -698,12 +697,10 @@ QUnit.test( "jQuery.when", function( assert ) {
 	} );
 
 	jQuery.each( [ 1, 2, 3 ], function( k, i ) {
-
 		jQuery.when( cache || jQuery.Deferred( function() {
 				this.resolve( i );
 			} )
 		).done( function( value ) {
-
 			assert.strictEqual( value, 1, "Function executed" + ( i > 1 ? " only once" : "" ) );
 			cache = value;
 		} );
@@ -759,10 +756,8 @@ QUnit.test( "jQuery.when - joined", function( assert ) {
 				expected = shouldResolve ? [ 1, 1 ] : [ 0, undefined ],
 				expectedNotify = shouldNotify && [ willNotify[ id1 ], willNotify[ id2 ] ],
 				code = "jQuery.when( " + id1 + ", " + id2 + " )",
-				context1 = defer1 && jQuery.isFunction( defer1.promise ) ? defer1.promise() :
-					( defer1.then ? window : undefined ),
-				context2 = defer2 && jQuery.isFunction( defer2.promise ) ? defer2.promise() :
-					( defer2.then ? window : undefined );
+				context1 = defer1 && jQuery.isFunction( defer1.promise ) ? defer1.promise() : window,
+				context2 = defer2 && jQuery.isFunction( defer2.promise ) ? defer2.promise() : window;
 
 			jQuery.when( defer1, defer2 ).done( function( a, b ) {
 				if ( shouldResolve ) {
@@ -879,4 +874,37 @@ QUnit.test( "jQuery.when - chaining", function( assert ) {
 	} );
 
 	defer.resolve( "other deferred" );
+} );
+
+QUnit.test( "jQuery.when - solitary thenables", function( assert ) {
+
+	assert.expect( 1 );
+
+	var done = assert.async(),
+		rejected = new Promise( function( resolve, reject ) {
+			setTimeout( function() {
+				reject( "rejected" );
+			}, 100 );
+		} );
+
+	jQuery.when( rejected ).then(
+		function() {
+			assert.ok( false, "Rejected, solitary, non-Deferred thenable should not resolve" );
+			done();
+		},
+		function() {
+			assert.ok( true, "Rejected, solitary, non-Deferred thenable rejected properly" );
+			done();
+		}
+	);
+} );
+
+QUnit.test( "jQuery.when does not reuse a solitary jQuery Deferred (gh-2018)", function( assert ) {
+
+	assert.expect( 2 );
+	var defer = jQuery.Deferred().resolve(),
+		promise = jQuery.when( defer );
+
+	assert.equal( promise.state(), "resolved", "Master Deferred is immediately resolved" );
+	assert.notStrictEqual( defer.promise(), promise, "jQuery.when returns the master deferred's promise" );
 } );


test/unit/deferred.js
```diff
@@ -759,10 +756,8 @@ QUnit.test( "jQuery.when - joined", function( assert ) {
 				expected = shouldResolve ? [ 1, 1 ] : [ 0, undefined ],
 				expectedNotify = shouldNotify && [ willNotify[ id1 ], willNotify[ id2 ] ],
 				code = "jQuery.when( " + id1 + ", " + id2 + " )",
-				context1 = defer1 && jQuery.isFunction( defer1.promise ) ? defer1.promise() :
-					( defer1.then ? window : undefined ),
-				context2 = defer2 && jQuery.isFunction( defer2.promise ) ? defer2.promise() :
-					( defer2.then ? window : undefined );
+				context1 = defer1 && jQuery.isFunction( defer1.promise ) ? defer1.promise() : window,
+				context2 = defer2 && jQuery.isFunction( defer2.promise ) ? defer2.promise() : window;
```
timmywil: This is a behavior difference. For `$.when( 1 )`, `window` is the context instead of `undefined`.

src/deferred.js
```diff
@@ -345,14 +346,11 @@ jQuery.extend( {
 						updateFunc( i, progressContexts, progressValues )
 					);
 				} else {
-					--remaining;
+					updateFunc( i, resolveContexts, resolveValues )( resolveValues[ i ] );
```
dmethvin: I see what you did there!

dmethvin:
This logic could be simplified a lot further if we got rid of the special case handling for Deferred and just used the thenables path. It would mean we'd lose progress notification on Deferreds but I'm not even sure how aggregated and interspersed progress info from several Deferreds would be used effectively by a caller. In any case the behavior of progress notifications with `jQuery.when` is not documented.

timmywil:
> I'm not even sure how aggregated and interspersed progress info from several Deferreds would be used effectively by a caller.

Perhaps a user could have progress bars for each request along with a progress bar for all of them?

timmywil:
While I'm not opposed to doing this, isn't removing progress handling from `$.when` a separate issue? I didn't see any responses to your comment about removing progress. I'm wondering if we have consensus.

dmethvin:
We can make it a separate thing and land this PR first. It seems that `$.when` progress is broken for a common use case anyway, the natural thing you might want from progress on the master is whether the other promises have resolved. http://stackoverflow.com/questions/26066198/jquery-when-progress-for-array-of-deferred-and-or-promise

test/unit/deferred.js
```diff
@@ -880,3 +875,27 @@ QUnit.test( "jQuery.when - chaining", function( assert ) {
...
+
+	assert.expect( 1 );
+
+	var rejected = new Promise( function( resolve, reject ) {
+		setTimeout( function() {
+			reject( "rejected" );
+		}, 100 );
+	} );
+
+	QUnit.stop();
```
gibson042: `var done = assert.async()`. We don't need `QUnit.stop` anymore. :smiley_cat:

test/unit/deferred.js
```diff
@@ -880,3 +875,27 @@ QUnit.test( "jQuery.when - chaining", function( assert ) {
...
+
+	jQuery.when( rejected ).then(
+		function() {
+			assert.ok( false, "Rejected, solitary, non-Deferred thenable should not resolve" );
+		},
+		function() {
+			assert.ok( true, "Rejected, solitary, non-Deferred thenable rejected properly" );
+		}
+	).always( function() {
```
gibson042: I'd stay away from `.always` and just put `done()` inside the callbacks.

src/deferred.js
```diff
@@ -345,14 +346,11 @@ jQuery.extend( {
...
 				}
 			}
-		}
-
-		// If we're not waiting on anything, resolve the master
-		if ( !remaining ) {
-			master.resolveWith( resolveContexts, resolveValues );
+		} else {
+			master.resolveWith( resolveContexts );
```
gibson042: `resolveContexts` is guaranteed to be `undefined` here... was it smaller to keep it in?

src/deferred.js
```diff
@@ -345,14 +346,11 @@ jQuery.extend( {
 						updateFunc( i, progressContexts, progressValues )
 					);
 				} else {
-					--remaining;
+					updateFunc( i, resolveContexts, resolveValues )( resolveValues[ i ] );
 				}
 			}
-		}
-
-		// If we're not waiting on anything, resolve the master
```
gibson042: I'd like to see some form of this comment remain.

src/deferred.js
```diff
@@ -316,14 +314,17 @@ jQuery.extend( {
 					if ( values === progressValues ) {
 						master.notifyWith( contexts, values );
 					} else if ( !( --remaining ) ) {
-						master.resolveWith( contexts, values );
+						master.resolveWith(
+							contexts.length === 1 ? contexts[ 0 ] : contexts,
+							values
```
gibson042: I don't like the incongruity of popping out the context but not the value. And looking at the discussion in #2018 around casting and idempotence, I'm almost certain that both should be unwrapped when `contexts.length === 1`.

src/deferred.js
```diff
@@ -316,14 +314,17 @@ jQuery.extend( {
...
 				};
 			},
 			progressValues, progressContexts, resolveContexts;
 
 		// Add listeners to Deferred subordinates; treat others as resolved
-		if ( length > 1 ) {
+		if ( length > 0 ) {
 			progressValues = new Array( length );
 			progressContexts = new Array( length );
 			resolveContexts = new Array( length );
```
gibson042: If I remember correctly, it was also my intention to reverse the below `if` blocks, privileging `then` over `promise`. I think that fits in this `jQuery.when` overhaul.

test/unit/deferred.js
gibson042: Not since at least jQuery 1.8.3: http://jsfiddle.net/av4shtg8/ . I complained about the `cast`/`all` mixing in #2018, but lost.

This test code should not be altered, and the behavior change should be reverted. If anything, the problem is that the [Callbacks `apply`](https://github.com/timmywil/jquery/blob/when-2546/src/callbacks.js#L80) converts our `undefined` context into the global object, but there's nothing we can do about that language-level operation.

gibson042:
Where are the tests for #2018?

test/unit/deferred.js
timmywil: But, why? What's the point of returning `undefined` instead of `window`, or vice versa? I imagine neither is being used very often. And why are we continuing to do different things for `$.when( 1 )` and `$.when( 1, 1 )`. jQuery 1.8.3 looks just as bad, only swapped. Time to get them to do the same thing.

test/unit/deferred.js
timmywil: Something wrong with `always`?

src/deferred.js
timmywil: Yup, was thinking I at least wanted to pass `undefined` as the context, but `resolveWith` can be called with no arguments here.

src/deferred.js
timmywil: Got lost during revisions.

test/unit/deferred.js
timmywil: Yea, I started with that. I went with `QUnit.stop` given how prevalent it is in our code, but I'll switch it back.

test/unit/deferred.js
timmywil: I changed it, but I'm just curious why you preferred it.

src/deferred.js
dmethvin: When would the `promise` branch be used if that was done?

src/deferred.js
timmywil: An array of values passed to `resolveWith` will get spread by apply, unlike with contexts. It does the same thing whether you pass an array of length 1 or just the value.

test/unit/deferred.js
dmethvin: Either context seems sufficiently useless that it does seem unlikely anyone is using them.

timmywil:
I didn't realize there are tests needed for #2018. Is there a behavior difference that I'm not accounting for? It seemed like just a refactor that would be covered by existing tests.

src/deferred.js
timmywil: Ah, yes, that would be a change in behavior requiring tests. If we reverse them, will the `promise` case ever get used – since Deferred has `.then`?

mzgol:
> Is there a behavior difference that I'm not accounting for?

You're returning a new Deferred instead of reusing the old one; that's a behavior change; the new bahavior is one that many promise libs are careful about. This definitely needs to be tested.

timmywil:
@gibson042 Now that I think about it, I could add tests that explicitly address the fact that we're not returning the original Deferred anymore. Is that what you mean?

timmywil:
> You're returning a new Deferred instead of reusing the old one.

Yup, silly me.

test/unit/deferred.js
timmywil: I'm inclined to leave it. Returning `window` and `[ window, window ]` seems better than `undefined` and `[ window, window ]`. However, I agree that getting `window` out of passing `undefined` is an odd language quirk, but I *think* all the windows will be undefineds in global strict mode.

src/deferred.js
gibson042: jQuery collections.

test/unit/deferred.js
gibson042: > I'm inclined to leave it. Returning `window` and `[ window, window ]` seems better than `undefined` and `[ window, window ]`

Meh.

> I *think* all the windows will be undefineds in global strict mode.

Correct.

src/deferred.js
gibson042: :+1:

src/deferred.js
timmywil: Interesting. I'm happy to try that. I prefer the `then` branch as well.

dmethvin: Good point. Note that this also means that in most cases jQuery Deferreds won't send progress events so this is another reason to just remove that entirely, or replace it with progress events on the master that signal when the individual promises complete.

timmywil: Welllll, maybe I spoke too soon. Giving the `then` branch precedence changes more than I expected. Lots of failures having to do with unexpected contexts and resolve values when passing multiple Deferreds. Let's make that a separate issue if we still want to do it.

test/unit/deferred.js
gibson042: > And why are we continuing to do different things for `$.when( 1 )` and `$.when( 1, 1 )`… Time to get them to do the same thing.`

Look, I made the [**same** argument](https://github.com/jquery/jquery/issues/2018#issuecomment-70524248), but [lost](https://github.com/jquery/jquery/issues/2018#issuecomment-70531678). And the differences go beyond context—`jQuery.when( value )` and `jQuery.when( value1, …, valueN )` are actually _different_ operations, with different signatures.

src/deferred.js
dmethvin: Moving to a separate ticket is fine with me.

gibson042: Fine with me, too.

test/unit/deferred.js
gibson042: `.always` is a nonstandard method specific to jQuery Deferreds, and not under test in this block. Using `.then` means that it won't need to change if we alter the implementation of `jQuery.when` in the future (a minor benefit to be sure, but one that costs basically nothing now).

src/deferred.js
timmywil: Issue [here](https://github.com/jquery/jquery/issues/2709).

test/unit/deferred.js
```diff
@@ -880,3 +875,36 @@ QUnit.test( "jQuery.when - chaining", function( assert ) {
...
+} );
+
+QUnit.test( "jQuery.when does not reuse a solitary jQuery Deferred (gh-2018)", function( assert ) {
+
+	assert.expect( 2 );
+	var defer = jQuery.Deferred().resolve(),
+		promise = jQuery.when( defer );
+
+	assert.equal( promise.state(), "resolved", "Master Deferred is immediately resolved" );
+	assert.notEqual( defer.promise(), promise, "jQuery.when returns the master deferred's promise" );
```
gibson042: `notEqual` → `notStrictEqual`

timmywil: Thanks.

test/unit/deferred.js
timmywil: Got it.

timmywil:
To sum up for the log, this PR led to opening two more issues for Deferred. #2709, which we probably won't get to before 3.0, and #2710, which we will. The concerns about progress + when are not addressed here, but we will address them before 3.0 is released.

PR #2713, opened 2015-11-13
Effects: Don't expose jQuery.speed

sculove:
jQuery.speed is undocumented internal APIs
they shouldn't be exposed.

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -429,7 +429,7 @@ jQuery.Animation = jQuery.extend( Animation, {
 	}
 } );
 
-jQuery.speed = function( speed, easing, fn ) {
+function speedOption( speed, easing, fn ) {
 	var opt = speed && typeof speed === "object" ? jQuery.extend( {}, speed ) : {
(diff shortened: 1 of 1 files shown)


mzgol:
cc @gnarf @gibson042

mzgol:
We should either hide it as you did in this PR (thanks for it, @sculove!) or document it if it's needed externally.

scottgonzalez:
We use this in jQuery UI: https://github.com/jquery/jquery-ui/blob/0bfbd21d4fefa98d165b7d50277bd23be84e919a/ui/effect.js#L796

timmywil:
If UI uses it, we should probably document it.

mzgol:
> If UI uses it, we should probably document it.

Agreed. And test. There are currently no tests for `jQuery.speed` as far as I can tell.

timmywil:
Closing this in favor of documenting. Thank you for bringing this to our attention, @sculove.

PR #2718, opened 2015-11-13
Core: do not expose second argument of the `jQuery.globalEval`

markelog:
Ref jquery/api.jquery.com#831

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -8,8 +8,10 @@ define( [
 	"./var/class2type",
 	"./var/toString",
 	"./var/hasOwn",
-	"./var/support"
-], function( arr, document, slice, concat, push, indexOf, class2type, toString, hasOwn, support ) {
+	"./var/support",
+	"./core/DOMEval"
+], function( arr, document, slice, concat,
+	push, indexOf, class2type, toString, hasOwn, support, DOMEval ) {
 
 var
 	version = "@VERSION",
@@ -258,12 +260,8 @@ jQuery.extend( {
 	},
 
 	// Evaluates a script in a global context
-	globalEval: function( code, context ) {
-		context = context || document;
-		var script = context.createElement( "script" );
-
-		script.text = code;
-		context.head.appendChild( script ).parentNode.removeChild( script );
+	globalEval: function( code ) {
+		DOMEval( code );
 	},
 
 	// Convert dashed to camelCase; used by the css and data modules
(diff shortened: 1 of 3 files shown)


mzgol:
This should be a variable, i.e. it should lie in /core/var/.

markelog:
I don't think so @mzgol

mzgol:
Why not? A module that exports a single thing and doesn't have side effects should be a variable. Look at [`src/css`](https://github.com/jquery/jquery/tree/78b9eac1198627eb4dad0cc35334c7704449f310/src/css/var).

I see, though, that some other modules from core are not following the same principle, e.g. [`src/core/access`](https://github.com/jquery/jquery/blob/78b9eac1198627eb4dad0cc35334c7704449f310/src/core/access.js) should be a variable as well.

cc @timmywil

markelog:
> Why not?

Didn't you answer your own question with -

> I see, though, that some other modules from core are not following the same principle, e.g. src/core/access.js should be a variable as well.

There is a lot of modules like that. I don't think we should impose additional restrictions on the contributions process.

mzgol:
@markelog From what I remember this was not an arbitrary restriction, the concept of `var`-modules was created by @timmywil to save size in the built file. Refactoring your PR in this way saves 2 bytes.

I thought it was going to bring bigger gains, though... So, out of curiosity I applied it to other similar modules and... I got back to the size from this PR (i.e. I lost 2 bytes).

@timmywil, since those var-modules don't seem to save size as much as I thought (or not at all sometimes), could you clarify what's their purpose? Because, as @markelog pointed out, currently the decision whether to create a var-module or a regular one seems completely random looking at all the files.

timmywil:
The point of the var modules is to translate...

```js
// File var/arr.js
define( function() {
	return [];
} );
```

to 

```js
var arr = [];
```

This is necessary whether the size gains are minor or not. For functions, though, the use of this translation is more subjective. I've mainly used var modules for smaller, one- to two-line, functions. I don't think we need to do that here.

markelog:
I know it might be sound... ahhh, weird :-), but should we document that?

timmywil:
> should we document that?

Meh.

timmywil:
We could stick a note about it in CONTRIBUTING.md.

markelog:
Yeah, my thought exactly, not on contribute.jquery.com or anything like that

Issue #1732, opened 2014-10-20
Preserve URL hash when requesting via $.ajax


mzgol:
Originally reported by pferreir at: http://bugs.jquery.com/ticket/10495
<p>
While trying to inject the Facebook "like button" JS in a page in execution time, I got some funny behavior:
</p>
<div class="code"><pre><span class="kd">var</span> <span class="nx">fjs</span> <span class="o">=</span> <span class="nx">$</span><span class="p">(</span><span class="s1">'script:first'</span><span class="p">);</span>
<span class="nx">$</span><span class="p">(</span><span class="s1">'&lt;script/&gt;'</span><span class="p">).</span><span class="nx">attr</span><span class="p">({</span><span class="s1">'async'</span><span class="o">:</span> <span class="kc">true</span><span class="p">,</span> <span class="s1">'src'</span><span class="o">:</span> <span class="s1">'//connect.facebook.net/en_US/all.js#xfbml=1&amp;appId=xxx'</span><span class="p">}).</span><span class="nx">insertBefore</span><span class="p">(</span><span class="nx">fjs</span><span class="p">)</span>
</pre></div><p>
fails (same with the equivalent <em>before()</em> call).
</p>
<p>
While
</p>
<div class="code"><pre><span class="kd">var</span> <span class="nx">fjs</span> <span class="o">=</span> <span class="nx">$</span><span class="p">(</span><span class="s1">'script:first'</span><span class="p">).</span><span class="nx">get</span><span class="p">(</span><span class="mi">0</span><span class="p">);</span>
<span class="kd">var</span> <span class="nx">e</span> <span class="o">=</span> <span class="nx">$</span><span class="p">(</span><span class="s1">'&lt;script/&gt;'</span><span class="p">).</span><span class="nx">attr</span><span class="p">({</span><span class="s1">'async'</span><span class="o">:</span> <span class="kc">true</span><span class="p">,</span> <span class="s1">'src'</span><span class="o">:</span> <span class="s1">'//connect.facebook.net/en_US/all.js#xfbml=1&amp;appId=xxx'</span><span class="p">}).</span><span class="nx">get</span><span class="p">(</span><span class="mi">0</span><span class="p">);</span>
<span class="nx">fjs</span><span class="p">.</span><span class="nx">parentNode</span><span class="p">.</span><span class="nx">insertBefore</span><span class="p">(</span><span class="nx">e</span><span class="p">,</span> <span class="nx">fjs</span><span class="p">)</span>
[...]

PR #2721, opened 2015-11-16
Ajax: Preserve URL hash on requests

dmethvin:
Fixes gh-1732

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -412,8 +412,8 @@ jQuery.extend( {
 			// Loop variable
 			i,
 
-			// hash from url, removed during manipulation
-			hash,
+			// uncached part of the url
+			uncached,
 
 			// Create the final options object
 			s = jQuery.ajaxSetup( {}, options ),
@@ -590,7 +590,7 @@ jQuery.extend( {
 		if ( !s.hasContent ) {
 
 			// Remember the hash so we can put it back
-			hash = s.url.slice( cacheURL.length );
+			uncached = s.url.slice( cacheURL.length );
 
 			// If data is available, append data to url
 			if ( s.data ) {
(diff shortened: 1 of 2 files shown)


src/ajax.js
```diff
@@ -605,6 +611,12 @@ jQuery.extend( {
 					cacheURL + ( rquery.test( cacheURL ) ? "&" : "?" ) + "_=" + nonce++;
 			}
 
+			// Put hash back on the actual URL that will be requested (gh-1732)
+			s.url = cacheURL + hash;
+
+			// cacheURL doesn't have the anti-cache param or hash
+			cacheURL = cacheURL.replace( rts, "" );
```
gibson042: The final value of `cacheURL` only matters if `s.ifModified` is true, but I think it should **include** `_` parameters that come from the user rather than us. What if we updated like so?

```js
uncached = s.url.slice( cacheURL.length );

if ( s.data ) {
	cacheURL += ( rquery.test( cacheURL ) ? "&" : "?" ) + s.data;
	…
}

if ( s.cache === false ) {
	cacheURL = cacheURL.replace( rts, "" );
	uncached = ( rquery.test( cacheURL ) ? "&" : "?" ) + "_=" + (nonce++) + uncached;
}

s.url = cacheURL + uncached;
```

dmethvin:
@mzgol I can almost guarantee you that I will continue to forget. It seems that any PR containing "Fixes #xxxx" will put a reference in the issue so it can be detected that way. I guess we could have a bot auto-tag it?

mzgol:
@dmethvin That's the problem with automation that's not automated, isn't it? :smile: 

> I guess we could have a bot auto-tag it?

Sounds like a good idea.

Issue #2716, opened 2015-11-13
Add tests that use jQuery.speed explicitly


timmywil:
From https://github.com/jquery/jquery/pull/2713

A lot of things wouldn't work if `jQuery.speed` didn't work, but since we've decided to document `jQuery.speed`, we need tests that correspond to how it might be used externally.

PR #2724, opened 2015-11-16
Effects: add tests for using jQuery.speed directly

timmywil:
Fixes gh-2716

### DIFF
--- a/test/unit/effects.js
+++ b/test/unit/effects.js
@@ -2495,4 +2495,89 @@ QUnit.test( "Show/hide/toggle and display: inline", function( assert ) {
 	} );
 } );
 
+function testEasing( assert, speed, easing, complete ) {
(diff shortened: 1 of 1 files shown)


dmethvin:
Looks pretty thorough. Could we add a test for `jQuery.fx.off` returning `duration: 0` as well? Right now the only test we do for it is for an [edge case](https://github.com/jquery/jquery/blob/cf7102c3f1bcbd135f24947a3a3216cff272bdc2/test/unit/effects.js#L1530-l1544).

timmywil:
@dmethvin Actually, that's [here](https://github.com/jquery/jquery/pull/2724/files#diff-c11dc311e64f99c637c0ef4fc61054a6R2578)

dmethvin:
dunno how i missed that! LGTM

PR #2726, opened 2015-11-16
Update testsuite.css

mgracer48:
Codeacy changes
I fixed some performence issues reported by codeacy

[Review on Reviewable](https://reviewable.io/reviews/jquery/jquery/2726)

### DIFF
--- a/test/data/testsuite.css
+++ b/test/data/testsuite.css
@@ -1,12 +1,12 @@
-ol#empty {
+#empty {
 	opacity: 0;
 }
 
-div#fx-tests h4 {
+#fx-tests h4 {
 	background: red;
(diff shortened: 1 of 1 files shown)


timmywil:
Thank you for the PR, but CSS performance is sufficient in the test suite.

mgracer48:
Oh

dmethvin:
Just to clarify, the changes aren't *wrong*, they just aren't necessary. Also, there are often times in our unit tests where we do somewhat, um, *strange* things in order to test that jQuery works correctly. Best practices are definitely good at the application level when you're writing real-world code, but for testing we often go against those rules.

PR #2728, opened 2015-11-17
Build: Build dependencies "added" to published npm package.

### LINKED ISSUE #2727: Package Distribution Ignores Build Dependencies (.npmignore)
This may be an edge case issue, or a even non-issue, but I'd love to hear your thoughts!

Required build dependencies are being ignored based on the `.npmignore` file. The following commits contribute to this change: 35f8e15fd73256ec02397d7bf93893e38afc9bc9 and b04b3861802606a114c936e753c0f20da6e8b236.

Currently, I'm relying on the `npm` published package for a module. As it stands right now, the distributed package cannot be re-built (module exclusion) as the various required files are being ignored – I'm looking to make `grunt custom:-deprecated,-effects,...` a `postinstall` hook. 

It seems the current solution would be to include `jquery` as a `git submodule` and building, which _works_, it just feels a little dirty (...blame the package managers for this one. :wink:)

It's also worth pointing out that I'm bundling for the browser – so the exclusion of modules/file size has some effect.

Thanks!

jbottigliero:
Resolves #2727

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -7,11 +7,4 @@
 /.mailmap
 /.travis.yml
 
-/build
 /speed
-/test
-/Gruntfile.js
-
-/external/qunit
-/external/requirejs
-/external/sinon


timmywil:
See #2727.

PR #2729, opened 2015-11-18
1.12 stable

rm-f:


mzgol:
This PR looks like a mistake... Please be careful next time.

PR #2735, opened 2015-11-20
removed variable self-assignment

cboynton-cweil:
this is a small optimization that removes a variable assignment to itself

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -540,7 +540,6 @@ jQuery.extend( {
 
 				// Support: IE8-11+
 				// Anchor's host property isn't correctly set when s.url is relative
-				urlAnchor.href = urlAnchor.href;
 				s.crossDomain = originAnchor.protocol + "//" + originAnchor.host !==
 					urlAnchor.protocol + "//" + urlAnchor.host;
 			} catch ( e ) {


dmethvin:
The comment above that line explains exactly why it is needed.

cboynton-cweil:
Thank you for the quick feedback. In my opinion, that comment does not properly explain why the property should be assigned to itself. Are you suggesting by assigning this property to itself, there is a special side-effect operation that occurs in Internet Explorer to re-evaluate the href, which in turn modifies the host property for relative URLs?

dmethvin:
Yes. The source code doesn't stand alone, but I think the comment above it makes it clear enough that the assignment isn't a mistake. If you look at the blame view you can see the discussion on the ticket. https://github.com/jquery/jquery/pull/1880

cboynton-cweil:
Thank you for the confirmation and clarification. I originally did follow your suggestion of looking at the blame, but I failed to review at the attached "Closes" issue where the details were described. I appreciate that you pointed that out so I won't make that mistake again. I am patiently waiting for the day IE is universally deprecated everywhere so these hacks are not needed.

dmethvin:
That one was a pretty brutally complex PR, it went through several iterations. There's always a tradeoff between how much explanation goes in the code and how much goes elsewhere. In jQuery at least, the `// Supports:` comments usually have a longer story behind them that you can find in the PR or ticket.

Issue #2736, opened 2015-11-20
Add unhandled exception notices for Deferred


dmethvin:
Right now, if there is an exception in a `.then()` it is silently swallowed.  I don't think we should ship 3.0.0 with that behavior.

http://jsbin.com/bodevaqive/edit?js,console

Maybe a `console.warn` message? We may end up yelling about rejections that might have been dealt with by a handler attached later in time. 

One possible tradeoff would be to only generate the console warnings for errors that seem to be programming mistakes such as `TypeError` or `ReferenceError`. That prevents the really bad mistakes such as the example above with a typo in a function name.

For more complex cases we could recommend that people use native Promise if it's consistently available across their supported platforms, since it will be easier to debug.

mzgol:
Unhandled rejections may be handled later so going through `console` seems wrong to me. We could expose something like `jQuery.Deferred.unhandledRejections` and/or a special event people could hook to to see rejections.

If this went through a property and not events, we'd need to remove stuff from there once it's handled so we'd have to be careful about perf.

Although this:
> One possible tradeoff would be to only generate the console warnings for errors that seem to be programming mistakes such as `TypeError` or `ReferenceError`.

seems like a good tradeoff to me. But maybe we could do more.

mzgol:
Also, it's worth noting what Node.js did to make working with rejections easier: https://nodejs.org/api/process.html#process_event_unhandledrejection

tl;dr: a `unhandlerRejection` event. We could do both that *and* a console.warn for `TypeError`/`ReferenceError`.

timmywil:
I'm not a fan of console messages from a production library. What about removing the default unhandled rejection handler – which would just rethrow – if another handler is attached?

gibson042:
I don't have time to address this today, but we can talk about it at the next meeting. I'm opposed to the core library having an opinion, and still maintain that it's a plugin problem for now. I'm happy to write it... have we documented current best practices?

timmywil:
@gibson042 There's a note about it in the blog post, but I don't think we've added anything to api.jquery.com about it. Is the deferred.then page the best place for that?

timmywil:
Also, I agree. Supporting Promises/A+ is a mixed bag for this reason. I'm inclined to leave this up to the user, but warn strongly against allowing phantom errors.

dmethvin:
Having a default that swallows errors is really bad though, especially compared to the current behavior that gives immediate feedback about programming errors.

timmywil:
[...]

PR #2737, opened 2015-11-23
Deferred: Warn on exceptions that are likely programming errors

dmethvin:
Fixes gh-2736

Still a WIP with some details to be worked out. I'm not really happy with the stack trace that results, it's only showing the immediate prior call. The gist, however, is that nearly all exceptions of these types are programming mistakes and should be reported ASAP. If devs don't like being warned about them they can set `jQuery.Deferred.exceptionHook = null` and not be bothered, or define a custom hook that filters out the noisy false positives.

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -157,6 +157,10 @@ jQuery.extend( {
 											mightThrow();
 										} catch ( e ) {
 
+											if ( jQuery.Deferred.exceptionHook ) {
+												jQuery.Deferred.exceptionHook( e );
+											}
+
 											// Support: Promises/A+ section 2.3.3.3.4.1
 											// https://promisesaplus.com/#point-61
 											// Ignore post-resolution exceptions
@@ -358,6 +362,23 @@ jQuery.extend( {
 		return master.promise();
 	}
 } );
+
+if ( window.console && window.console.warn ) {
+
+	// These usually indicate a programmer mistake during development,
+	// warn about them ASAP rather than swallowing them by default.
+
+	var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
+
+	jQuery.Deferred.exceptionHook = function( error ) {
+		if ( error && rerrorNames.test( error.name ) ) {
+			window.console.warn( "jQuery.Deferred exception: " + error.message );
+			if ( window.console.trace ) {
+				window.console.trace();
+			}
+		}
+	};
+}
 
 return jQuery;
 } );
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -158,7 +158,8 @@ jQuery.extend( {
 										} catch ( e ) {
 
 											if ( jQuery.Deferred.exceptionHook ) {
-												jQuery.Deferred.exceptionHook( e );
+												jQuery.Deferred.exceptionHook( e,
+													process.stackTrace );
 											}
 
 											// Support: Promises/A+ section 2.3.3.3.4.1
@@ -186,6 +187,12 @@ jQuery.extend( {
 							if ( depth ) {
 								process();
 							} else {
+
+								// Call an optional hook to record the stack, in case of exception
+								// since it's otherwise lost when execution goes async
+								if ( jQuery.Deferred.getStackHook ) {
+									process.stackTrace = jQuery.Deferred.getStackHook();
+								}
 								window.setTimeout( process );
 							}
 						};
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -166,7 +166,7 @@ jQuery.extend( {
 											// Ignore post-resolution exceptions
 											if ( depth + 1 >= maxDepth ) {
 
-												// Only substitue handlers pass on context
+												// Only substitute handlers pass on context
 												// and multiple values (non-spec behavior)
 												if ( handler !== Thrower ) {
 													that = undefined;
@@ -362,23 +362,6 @@ jQuery.extend( {
 		return master.promise();
 	}
 } );
-
-if ( window.console && window.console.warn ) {
-
-	// These usually indicate a programmer mistake during development,
-	// warn about them ASAP rather than swallowing them by default.
-
-	var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
-
-	jQuery.Deferred.exceptionHook = function( error ) {
-		if ( error && rerrorNames.test( error.name ) ) {
-			window.console.warn( "jQuery.Deferred exception: " + error.message );
-			if ( window.console.trace ) {
-				window.console.trace();
-			}
-		}
-	};
-}
 
 return jQuery;
 } );
--- a/src/deferred/exceptionHook.js
+++ b/src/deferred/exceptionHook.js
@@ -11,10 +11,8 @@ jQuery.Deferred.exceptionHook = function( error, stack ) {
 
 	// Support: IE9
 	// Console exists when dev tools are open, which can happen at any time
-	var warn = window.console && window.console.warn;
-
-	if ( warn && error && rerrorNames.test( error.name ) ) {
-		warn.call( window.console, "jQuery.Deferred exception: " + error.message, stack );
+	if ( window.console && window.console.warn && error && rerrorNames.test( error.name ) ) {
+		window.console.warn( "jQuery.Deferred exception: " + error.message, stack );
 	}
 };
 
--- a/src/deferred/exceptionHook.js
+++ b/src/deferred/exceptionHook.js
@@ -1,0 +1,21 @@
+define( [
+	"../core",
+	"../deferred"
+], function( jQuery ) {
+
+// These usually indicate a programmer mistake during development,
+// warn about them ASAP rather than swallowing them by default.
+var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
+
+jQuery.Deferred.exceptionHook = function( error ) {
+
+	// Support: IE9
+	// Console exists when dev tools are open, which can happen at any time
+	var warn = window.console && window.console.warn;
+
+	if ( error && rerrorNames.test( error.name ) ) {
+		warn( "jQuery.Deferred exception: " + error.message );
+	}
+};
+
+} );
--- a/src/deferred/exceptionHook.js
+++ b/src/deferred/exceptionHook.js
@@ -7,17 +7,13 @@ define( [
 // warn about them ASAP rather than swallowing them by default.
 var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
 
-jQuery.Deferred.exceptionHook = function( error ) {
+jQuery.Deferred.exceptionHook = function( error, stack ) {
 
 	// Support: IE9
 	// Console exists when dev tools are open, which can happen at any time
-	var stack,
-		warn = window.console && window.console.warn;
+	var warn = window.console && window.console.warn;
 
 	if ( warn && error && rerrorNames.test( error.name ) ) {
-		if ( jQuery.Deferred.getStackHook ) {
-			stack = jQuery.Deferred.getStackHook();
-		}
 		warn.call( window.console, "jQuery.Deferred exception: " + error.message, stack );
 	}
 };
--- a/src/deferred/exceptionHook.js
+++ b/src/deferred/exceptionHook.js
@@ -11,10 +11,14 @@ jQuery.Deferred.exceptionHook = function( error ) {
 
 	// Support: IE9
 	// Console exists when dev tools are open, which can happen at any time
-	var warn = window.console && window.console.warn;
+	var stack,
+		warn = window.console && window.console.warn;
 
-	if ( error && rerrorNames.test( error.name ) ) {
-		warn( "jQuery.Deferred exception: " + error.message );
+	if ( warn && error && rerrorNames.test( error.name ) ) {
+		if ( jQuery.Deferred.getStackHook ) {
+			stack = jQuery.Deferred.getStackHook();
+		}
+		warn.call( window.console, "jQuery.Deferred exception: " + error.message, stack );
 	}
 };
 
--- a/test/unit/deferred.js
+++ b/test/unit/deferred.js
@@ -540,11 +540,11 @@ QUnit[ window.console ? "test" : "skip" ]( "jQuery.Deferred.exceptionHook", func
 		defer.then( function() {
 			// Should get an error
 			jQuery.barf();
-		} ).catch( jQuery.noop ),
+		} ).then( null, jQuery.noop ),
 		defer.then( function() {
 			// Should NOT get an error
 			throw new Error( "Make me a sandwich" );
-		} ).catch( jQuery.noop )
+		} ).then( null, jQuery.noop )
 	).then( function( ) {
 		window.console.warn = oldWarn;
 		done();
--- a/test/unit/deferred.js
+++ b/test/unit/deferred.js
@@ -525,23 +525,30 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
 	} catch ( _ ) {}
 } );
 
-QUnit.test( "jQuery.Deferred.exceptionHook", function( assert ) {
+QUnit[ window.console ? "test" : "skip" ]( "jQuery.Deferred.exceptionHook", function( assert ) {
 
 	assert.expect( 1 );
 
 	var done = assert.async(),
 		defer = jQuery.Deferred(),
-		oldWarn = console.warn;
+		oldWarn = window.console.warn;
 
-	console.warn = function( s ) {
-		assert.ok( true, "Warned -- " + s );
-		oldWarn.apply( console, arguments );
-		console.warn = oldWarn;
+	window.console.warn = function( msg ) {
+		assert.ok( true, "Message: " + msg );
 	};
-
-	defer.then( function() {
-		jQuery.barf();
-	} ).then( null, done );
+	jQuery.when(
+		defer.then( function() {
+			// Should get an error
+			jQuery.barf();
+		} ).catch( jQuery.noop ),
+		defer.then( function() {
+			// Should NOT get an error
+			throw new Error( "Make me a sandwich" );
+		} ).catch( jQuery.noop )
+	).then( function( ) {
+		window.console.warn = oldWarn;
+		done();
+	} );
 
 	defer.resolve();
 } );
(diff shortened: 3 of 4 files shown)


timmywil:
Yea, I wonder if `error.stack` is a more appropriate stack trace than `console.trace()`.

src/deferred.js
```diff
@@ -359,5 +363,22 @@ jQuery.extend( {
 	}
 } );
 
+if ( window.console && window.console.warn ) {
+
+	// These usually indicate a programmer mistake during development,
+	// warn about them ASAP rather than swallowing them by default.
+
+	var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
```
gibson042: I still don't see this as library code. The whole mess is more appropriate for userland.

src/deferred.js
```diff
@@ -157,6 +157,10 @@ jQuery.extend( {
 											mightThrow();
 										} catch ( e ) {
 
+											if ( jQuery.Deferred.exceptionHook ) {
+												jQuery.Deferred.exceptionHook( e );
```
gibson042: Calling it `exceptionHook` doesn't change the fact that it is invoked for every Promise-like rejection.

gibson042: If we're comfortable with the above, though, a hook here seems like an interesting approach. Since only the exception is passed, there's no access to a relevant thenable and therefore no value for anything other than debugging.

src/deferred.js
```diff
@@ -359,5 +363,22 @@ jQuery.extend( {
 	}
 } );
 
+if ( window.console && window.console.warn ) {
+
+	// These usually indicate a programmer mistake during development,
```
gibson042: How true is this? An obvious and common example is `JSON.parse` throwing SyntaxError, but there are also cases like `Date.prototype.toISOString` throwing RangeError. Promise examples often encourage developers to optimize for non-exceptional execution, and this could get noisy when they do something like:
```js
get( … ).then( assertSuccess ).then( JSON.parse ).then( handleResponse )
	.catch( handleError )
	.then( finish );
```

src/deferred.js
dmethvin: Agreed, perhaps `rejectionHook` instead?

src/deferred.js
dmethvin: You've probably caught me editorializing a bit there, I admit. :smile: I think exceptions (vs programmatic rejections) are much more likely to be programming errors but I don't have data to back up that assertion. For that case it would seem really unusual for a server to deliver malformed JSON, but if it did occasionally do so the worst we've done is said that on the console.

scottgonzalez:
Is my understanding correct that this:
* Warns on all exceptions, not just unhandled exceptions
* Only listens for thrown exceptions and not explicitly rejected promises?

src/deferred.js
dmethvin: I'd agree if the developer ergonomics weren't so horrific. If the default is to swallow but we just add the ability for a hook, people won't know enough to add a hook. Like I said in the meeting, devs need a solution really bad, so here is a really bad solution.

dmethvin:
> Warns on all exceptions, not just unhandled exceptions

No, it only warns on the derivative `Error` types listed there.

> Only listens for thrown exceptions and not explicitly rejected promises?

Yes, non-exception rejections (not an exception or explicitly `throw`n error) don't go through this path. jQuery Deferred hasn't ever warned on unhandled explicit rejections (such as an ajax 404) so I don't feel bad about excluding them.

scottgonzalez:
>> Warns on all exceptions, not just unhandled exceptions

> No, it only warns on the derivative Error types listed there.

What I mean is that this isn't a hook for unhandled exceptions. It's a hook for all exceptions.

dmethvin:
> What I mean is that this isn't a hook for unhandled exceptions. It's a hook for all exceptions.

Oh, sorry, yes. Exceptions like `SyntaxError` and explicitly `throw`n errors will pass through here.

src/deferred.js
```diff
@@ -359,5 +363,22 @@ jQuery.extend( {
...
+	// warn about them ASAP rather than swallowing them by default.
+
+	var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
+
+	jQuery.Deferred.exceptionHook = function( error ) {
+		if ( error && rerrorNames.test( error.name ) ) {
+			window.console.warn( "jQuery.Deferred exception: " + error.message );
+			if ( window.console.trace ) {
+				window.console.trace();
```
timmywil: Could we do `error.stack` here?

src/deferred.js
```diff
@@ -359,5 +363,22 @@ jQuery.extend( {
...
+if ( window.console && window.console.warn ) {
+
+	// These usually indicate a programmer mistake during development,
+	// warn about them ASAP rather than swallowing them by default.
+
+	var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
+
+	jQuery.Deferred.exceptionHook = function( error ) {
+		if ( error && rerrorNames.test( error.name ) ) {
+			window.console.warn( "jQuery.Deferred exception: " + error.message );
```
mzgol: What if `console` doesn't exist?

mzgol: Specifically, that's the case in IE 9 without open Dev Tools.

scottgonzalez: See L366.

dmethvin: There's a guard above for that case but you're right, it needs to be in the function itself to prevent a really nasty Heisenbug where it would crash, but only if you didn't have Dev Tools open. :smiling_imp: 

In environments without the console my intent was that you just wouldn't get any output, which is basically no worse than the current default. I think escalating to something that is still there like `window.alert` would be too intrusive, but someone could add that as a hook if they were trying to debug an IE9-specific problem.

src/deferred.js
staabm: What about pluging in this "error-handler" via a separate AMD Module...? That way one can build jQuery without it (for production) and with it for development etc

src/deferred.js
mzgol: Ah, OK. I *think* if you open the site with DevTools open and then close it the `console` object is still there but I'd have to check.

test/unit/deferred.js
```diff
@@ -525,6 +525,27 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
...
 
+QUnit.test( "jQuery.Deferred.exceptionHook", function( assert ) {
+
+	assert.expect( 1 );
+
+	var done = assert.async(),
+		defer = jQuery.Deferred(),
+		oldWarn = console.warn;
+
+	console.warn = function( s ) {
```
markelog: We have sinon lib btw

src/deferred.js
timmywil: @staabm not a bad idea

test/unit/deferred.js
mzgol: @markelog :+1:

dmethvin:
Pushed a few changes here. I tried `sinon.spy` but was getting an error from its guts and didn't really need it anyway.

src/deferred/exceptionHook.js
```diff
@@ -0,0 +1,21 @@
...
+var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
+
+jQuery.Deferred.exceptionHook = function( error ) {
+
+	// Support: IE9
+	// Console exists when dev tools are open, which can happen at any time
+	var warn = window.console && window.console.warn;
+
+	if ( error && rerrorNames.test( error.name ) ) {
+		warn( "jQuery.Deferred exception: " + error.message );
```
dmethvin: Because the Deferred is [resolved through a setTimeout](https://github.com/jquery/jquery/blob/e077ffb083743f4a4b990f586c9d25d787e7b417/src/deferred.js#L185) it loses the stack. We could use some chicanery to save a stack beforehand but that would impact size and performance for the comparatively rare case where an exception occurs. So I pulled out the stack trace entirely since it only refers to things inside jQuery itself.

src/deferred/exceptionHook.js
```diff
@@ -0,0 +1,21 @@
+define( [
+	"../core",
+	"../deferred"
+], function( jQuery ) {
+
+// These usually indicate a programmer mistake during development,
+// warn about them ASAP rather than swallowing them by default.
+var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
```
dmethvin: The `error.stack` property isn't standardized but it looks like Chrome and Safari do add one for `throw new Error()`. Using that would make this a much noisier hook so I stayed with identifying the ES-defined exception names.

test/unit/deferred.js
```diff
@@ -525,6 +525,34 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
...
 
+QUnit[ window.console ? "test" : "skip" ]( "jQuery.Deferred.exceptionHook", function( assert ) {
+
+	assert.expect( 1 );
+
+	var done = assert.async(),
+		defer = jQuery.Deferred(),
+		oldWarn = window.console.warn;
+
+	window.console.warn = function( msg ) {
```
markelog: Did you try sinon for this?

test/unit/deferred.js
```diff
@@ -525,6 +525,34 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
 	} catch ( _ ) {}
 } );
 
+QUnit[ window.console ? "test" : "skip" ]( "jQuery.Deferred.exceptionHook", function( assert ) {
```
markelog: I think we can extend QUnit API for this kind of pattern

dmethvin: Something like `QUnit.testIf(condition, function)` would be useful I agree.

test/unit/deferred.js
dmethvin: I tried sinon spy but they didn't offer that much value for such a simple case and I kept getting a strange error. Basically I gave up because I felt I was wasting time.

test/unit/deferred.js
gibson042: `testIf` has already been contemplated [multiple](https://github.com/jquery/qunit/issues/637#issuecomment-53498233) [times](https://github.com/jquery/qunit/pull/714#issuecomment-67778915), and is easy to add manually. Let's try it out in jQuery core. :+1:

markelog: /cc @leobalter

test/unit/deferred.js
markelog: Hm, weird, sinon should be useful in cases like this, especially that it provides integration with assertion libraries like standart `assert` module and `chai`. See - http://chaijs.com/plugins/sinon-chai

@leobalter is there integration way with qunit?

test/unit/deferred.js
leobalter: It's not the first time we have this API suggestion on QUnit, I'll discuss it on QUnit issues.

test/unit/deferred.js
leobalter: > sinon spy but they didn't offer that much value for such a simple case

Agreed. A few years ago I tried it and saw a bloat library polluting the global scope and QUnit's api. At that moment I've made [DexterJS](https://github.com/leobalter/dexterjs) with a very simple spy/stub API alternative but it's not maintained since I didn't have many contributions and stopped using mocks.

test/unit/deferred.js
```diff
@@ -525,6 +525,34 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
...
+		defer = jQuery.Deferred(),
+		oldWarn = window.console.warn;
+
+	window.console.warn = function( msg ) {
+		assert.ok( /barf/.test( msg ), "Message: " + msg );
+	};
+	jQuery.when(
+		defer.then( function() {
+			// Should get an error
+			jQuery.barf();
```
leobalter: Instead of calling `jQuery.barf`, we could make a direct call to the assertion here, the `assert.expect` is enough.

test/unit/deferred.js
markelog: > Agreed.

Yeah, but if we would use it across all the code base i.e. everywhere, it seems as matter of consistency and usability for other not that simple cases like this one.

>  A few years ago I tried it and saw a bloat library polluting the global scope and QUnit's api. 

Is it possible to provide an extension to qunit as separate package? Since our extensions already complicate our test code, i fear core testsuite might became more vague if we would extend it even more.

leobalter: The problem with Dexter was the ajax mock. Maybe re-releasing it with only the spy and stub API will serve most of what many libraries want. It does not need to rely on any test framework.

I'll also check if I should even split the spy and stub features in different modules.

Anyway, you can solve this current PR without any mock/spy function as I mentioned in https://github.com/jquery/jquery/pull/2737/files#r46825810

markelog: Well, i would like qunit api to be able to integrate with `sinon`, like it can be done with `chai`, `assert` or `jasmine` libs.

leobalter: AFAIK, sinon already connects to QUnit through sinon-qunit (See: http://sinonjs.org/qunit/).

You may also call sinon api directly, which seems better than having a module messing QUnit's own API.

markelog: Of course it is :-), thank you Leo!

src/deferred/exceptionHook.js
```diff
@@ -0,0 +1,21 @@
...
+
+jQuery.Deferred.exceptionHook = function( error ) {
+
+	// Support: IE9
+	// Console exists when dev tools are open, which can happen at any time
+	var warn = window.console && window.console.warn;
```
dcherman: This needs to have the context bound to `console` doesn't it?  Otherwise it throws `Uncaught TypeError: Illegal invocation` in Chrome (probably others too)

mzgol: That's true... @dmethvin did you check it in Chrome?

src/deferred/exceptionHook.js
dcherman: Does simply accessing the `stack` property of an Error have a perf impact?  You suggested this in the meeting, but you could also provide `jquery.longStackTraces.js` as a plugin or something to get functionality similar to Bluebird as long as the appropriate hooks exist in core

src/deferred/exceptionHook.js
dmethvin: Yep, this needs to be fixed, thanks!

dmethvin: So since I wasn't calling the old console API (just shimming it) the error didn't occur. :confounded:

src/deferred/exceptionHook.js
dmethvin: Adding a hook so that a plugin could provide better stack traces seems like a good way to go, I'm looking into it.

dmethvin:
Updated to add the stack hook. I'm not fully confident about what's here, would love a review, esp from @gibson042.

test/unit/deferred.js
```diff
@@ -525,6 +525,66 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
...
+	var done = assert.async(),
+		defer = jQuery.Deferred(),
+		oldWarn = window.console.warn;
+
+	jQuery.Deferred.getStackHook = function() {
+		// Default exceptionHook assumes the stack is in a form console.warn can log,
+		// but a custom getStackHook+exceptionHook pair could save a raw form and
+		// format it to a string only when an exception actually occurs.
+		// For the unit test we just ensure the plumbing works.
+console.trace();
```
dmethvin: This needs to be removed but I wanted to be sure we had the valid stack at this point of execution.

test/unit/deferred.js
dmethvin: If you want to come up for a spec for `QUnit.testIf()` we can polyfill it until we upgrade, or we could upgrade QUnit and make the change then. Let's do that change outside this ticket though.

src/deferred.js
```diff
@@ -182,6 +187,12 @@ jQuery.extend( {
 							if ( depth ) {
 								process();
 							} else {
+
+								// Call an optional hook to record the stack, in case of exception
+								// since it's otherwise lost when execution goes async
+								if ( jQuery.Deferred.getStackHook ) {
+									process.stackTrace = jQuery.Deferred.getStackHook();
```
dmethvin: I struggled with finding a good place to put the stack trace, it needs to be per-promise and not leak memory.

timmywil: Perhaps it doesn't need to be stored, just logged?

timmywil: I'm thinking of something akin to a verbose mode, which could be turned on and run for all deferreds during debugging.

dmethvin: The problem is that we have to get the stack before the `setTimeout` and at that point we don't know whether there will be a problem or not. If you run the tests with the console.trace() there right now you can see how noisy that would be. We only want to dump the stack if there's an exception.

src/deferred/exceptionHook.js
```diff
@@ -0,0 +1,21 @@
...
+var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
+
+jQuery.Deferred.exceptionHook = function( error, stack ) {
+
+	// Support: IE9
+	// Console exists when dev tools are open, which can happen at any time
+	var warn = window.console && window.console.warn;
+
+	if ( warn && error && rerrorNames.test( error.name ) ) {
+		warn.call( window.console, "jQuery.Deferred exception: " + error.message, stack );
```
dcherman: Not that this matters in the slightest, but seeing as how you're accessing the global immediately above this as feature detection, is there any reason to `.call` rather than just `window.console.warn(...)`?  Might even save some bytes ;)

dmethvin: Yeah, technically I could just let gzip do its thang and avoid the variable, which would also remove the need for the `.call`. I'm okay with that if people prefer it.

mzgol: Sounds like a good idea.

-- 
Michał Gołębiowski

dmethvin:
I've put together a [plugin](https://github.com/dmethvin/jquery-deferred-reporter) that uses the hook for collecting the stack and then shows it when an exception occurs. In the README I mention that  having to collect the stack in advance makes Deferreds slower [by roughly 2x](https://jsfiddle.net/h20r0e6z/) which is why it isn't on by default. However, since it's such a [small amount of code](https://github.com/dmethvin/jquery-deferred-reporter/blob/1901ae9b535730469cd2e4d8764801776b5df143/src/reporter.js#L13-L22) I wonder whether we shouldn't just have a flag or other simple one-liner to enable this rather than making people load another module/plugin to do it.

timmywil:
I like that it's a separate plugin, with its own README explaining why it's necessary. A flag can easily be kept on by accident.

PR #2744, opened 2015-11-27
Effects: Possible workaround for Android || assignment issue.

aleph1:
If the ternary operator works in this context we lose one var declaration and save bytes. Unfortunately I don't have an older Android device to test this on.


aleph1:
I tried following the instructions on the CLA Verification Results page, but I seem to still be getting a jQuery Foundation CLA error.

markelog:
In associated ticket discussion - http://bugs.jquery.com/ticket/12497 it is stated that this bug does reproduce only on certain devices and on certain version of 2.3. Plus we are on the verge of dropping android 2.3 completely. 

So it would be safer to just wait until that time will come.

Thank you for the effort though!

PR #2746, opened 2015-11-29
Updated links to https where they are supported.

jdufresne:


README.md
```diff
@@ -23,7 +23,7 @@ What you need to build your own jQuery
 
 In order to build jQuery, you need to have the latest Node.js/npm and git 1.7 or later. Earlier versions might work, but are not supported.
 
-For Windows, you have to download and install [git](http://git-scm.com/downloads) and [Node.js](http://nodejs.org/download/).
+For Windows, you have to download and install [git](http://git-scm.com/downloads) and [Node.js](https://nodejs.org/download/).
```
markelog: I guess we can change it to https://nodejs.org/en/download/?

README.md
```diff
@@ -176,9 +176,9 @@ grunt watch
 Run the unit tests with a local server that supports PHP. Ensure that you run the site from the root directory, not the "test" directory. No database is required. Pre-configured php local servers are available for Windows and Mac. Here are some options:
 
 - Windows: [WAMP download](http://www.wampserver.com/en/)
-- Mac: [MAMP download](http://www.mamp.info/en/index.html)
```
markelog: I guess we need to change it to https://www.mamp.info/en/downloads/

README.md
```diff
@@ -176,9 +176,9 @@ grunt watch
...
+- Mac: [MAMP download](https://www.mamp.info/en/index.html)
 - Linux: [Setting up LAMP](https://www.linux.com/learn/tutorials/288158-easy-lamp-server-installation)
-- [Mongoose (most platforms)](http://code.google.com/p/mongoose/)
+- [Mongoose (most platforms)](https://code.google.com/p/mongoose/)
```
markelog: Do we still need that? @timmywil ?

README.md
mzgol: @markelog :+1: The current URL seems to be non-human-friendly.

But it can be done in a separate commit as well.

markelog: I would keep them in one commit, so we wouldn't "trash" the commit history

mzgol:
LGTM. I think we can leave changing/dropping some URLs to separate commits, I like this one being focused just on changing http to https.

README.md
mzgol: I think it's fine to separate them: https://github.com/jquery/jquery/pull/2746#issuecomment-162600983.

Not very important, though.

Issue #2730, opened 2015-11-18
.attr() getter doesn't accept some characters in Firefox


mzgol:
With the following HTML body:
```html
<h1 Ω="omega">Hello!</h1>
```
and the following code:
```js
var h1 = $('h1');
console.log(h1.attr('Ω'));
console.log(h1[0].getAttribute('Ω'));
```
the `h1.attr('Ω')` part prints `"omega"` in Chrome but undefined in Firefox. The `getAttribute` part prints `"omega"` in both browsers.

This has hit AngularJS, see https://github.com/angular/angular.js/pull/13318#issuecomment-157812532 for more info.

EDIT: @dmethvin says it works in IE 11.

mzgol:
It worked in 1.6.4, hasn't been working since 1.7.0.

mzgol:
Test case: http://jsfiddle.net/cz57ey00/

mzgol:
@dmethvin discovered the culprit is `String#toLowerCase` which changes `Ω` to `ω`. `.getAttribute('ω')` works in Chrome but not in Firefox.

What does the spec say?

mzgol:
So, the spec says that:

1. `getAttribute(name)` for non-`null` values returns an attribute value for name `name`: https://dom.spec.whatwg.org/#dom-element-getattribute
2. When getting the element name the `name` parameter is converted to ASCII lowercase: https://dom.spec.whatwg.org/#concept-element-attributes-get-by-name
3. Converting to ASCII lowercase means changing only the values between U+0041 to U+005A which is `A-Z`.

This means Firefox is right and we're wrong but... every other browser is wrong as well - Chrome 46, IE 9-11, Safari 9 and even the old Opera Presto 12.16 returns the attribute value for `getAttribute('ω')`.

@bzbarsky, should the spec be changed considering Firefox is the exception here?

mzgol:
As for jQuery itself, @dmethvin suggested we could get rid of the [`.toLowerCase()` part](https://github.com/jquery/jquery/blob/47ccf3daadc4b312f850502300129952e70f9d9d/src/attributes/attr.js#L42) in 3.0.0 if it was only needed for old IE.

If it doesn't cause major problems for IE 6-8 we could try to use our own `toLowerCase` in 1.12 that would only transform `A-Z` to `a-z`.

mzgol:
I'll set a milestone to 3.0.1 for now but if we could get to it before that'd be great.

bzbarsky:
[...]

PR #2749, opened 2015-11-30
Attributes: exclusively lowercase A-Z in attribute names

timmywil:
Fixes gh-2730

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -7,7 +7,14 @@ define( [
 ], function( jQuery, access, support, rnotwhite ) {
 
 var boolHook,
-	attrHandle = jQuery.expr.attrHandle;
+	attrHandle = jQuery.expr.attrHandle,
+
+	// Exclusively lowercase A-Z in attribute names (gh-2730)
+	// https://dom.spec.whatwg.org/#converted-to-ascii-lowercase
+	raz = /[A-Z]/g,
+	lowercase = function( ch ) {
+		return ch.toLowerCase();
+	};
 
 jQuery.fn.extend( {
 	attr: function( name, value ) {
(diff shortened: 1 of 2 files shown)


test/unit/attributes.js
```diff
@@ -451,7 +451,7 @@ QUnit.test( "attr(String, Object)", function( assert ) {
 
 	$radio = jQuery( "<input>", {
 		"value": "sup",
-		"type": "radio"
+		"TYPE": "radio"
```
mzgol: What's the reason for this change?

timmywil: To make sure we use the type attrhook even if the name that comes in is not lowercase.

mzgol: Do we have one that tests the lowercase version, just in case?

src/attributes/attr.js
```diff
@@ -7,7 +7,14 @@ define( [
 ], function( jQuery, access, support, rnotwhite ) {
 
 var boolHook,
-	attrHandle = jQuery.expr.attrHandle;
+	attrHandle = jQuery.expr.attrHandle,
+
+	// Exclusively lowercase A-Z in attribute names (gh-2730)
+	// https://dom.spec.whatwg.org/#converted-to-ascii-lowercase
+	raz = /[A-Z]/g,
```
dmethvin: Not that it makes a lot of difference, but you could use `[A-Z]+` here to reduce the potential number of function calls?

test/unit/attributes.js
dmethvin: Do we need that since the `value` case above tests all-lowercase? Or is there some specific concern about `type`?

timmywil: Yea, lowercase is covered elsewhere. I wanted to have one place where we ensure we still use the `type` attrHook even when the name passed is not lowercase. I'll add a comment.

src/attributes/attr.js
timmywil: Yup

dmethvin:
LGTM

PR #2751, opened 2015-12-03
Remove duplicated word text on comment

netil:
Just removes duplicated `name` text comment.


dmethvin:
I love it when I get the easy ones. Thanks @netil!

PR #2772, opened 2015-12-14
Serialize: newfunction serializeObject

DarkScorpion:
Added new function, that converts contents in the form to object { name: value} which is convenient to operate and manipulate the data. Such as the creation of a complex calculator.
If you hang an event listener "onChange" , you can monitor changes entered data in one object and recalculated or show / hide additional fields . All in one object !

### DIFF
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -92,6 +92,16 @@ jQuery.fn.extend( {
 	serialize: function() {
 		return jQuery.param( this.serializeArray() );
 	},
+	serializeObject: function() {
+		var arr = this.serializeArray();
+
+		var temp = {};
+		for ( var i in arr ) {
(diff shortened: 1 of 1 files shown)


DarkScorpion:
Example of tracking changes in form.

dmethvin:
Thanks, but someone beat you to it! There are several good plugins available for serializing form data to an object in different ways, just search here on Github. See for example [serializeJSON](https://github.com/marioizquierdo/jquery.serializeJSON) and [serializeObject](https://github.com/hongymagic/jQuery.serializeObject) but there are others. The one you propose doesn't cover all the cases they do, so users of those plugins wouldn't benefit from switching to it and/or would be asking us to enhance it. Also see our guidelines for adding features at https://github.com/jquery/jquery/wiki/Adding-new-features .

Issue #2777, opened 2015-12-16
CSS: Enable support to get css path


techunits:
Current implementations doesn't have support to get CSS path from jquery core. It should be implemented.

mzgol:
What do you mean by "getting CSS path"?

mzgol:
See https://github.com/jquery/jquery/pull/2778#issuecomment-165072219

PR #2778, opened 2015-12-16
CSS: Enable support to get css path #2777

techunits:
I have implemented the reported features as well tested that as per the given standards. Please let me know if I am missing something.

The use of the implementation as follows:

$(selector).getCSSPath();

returns: html>body>ul.test-class>li.item:nth-child(1)

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -14,6 +14,7 @@ define( [
 	"./css/adjustCSS",
 	"./css/addGetHookIf",
 	"./css/support",
+	"./css/getCSSPath",
 
 	"./core/init",
 	"./core/ready",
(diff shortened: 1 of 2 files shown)


techunits:
@timmywil 
Hello Tim,
Can you please have a check on the pull request, and let me know the outcome, whether this is eligible.

Thanks
Sougata

mzgol:
Thanks for the PR but most people don't need this to be in Core and this can be implemented as a plugin. See https://github.com/jquery/jquery/wiki/Adding-new-features for more info.

dmethvin:
I agree with @mzgol. Also, even if it is released as a plugin it should have documentation and unit tests. I think this method as implemented is fragile and some good unit tests would find the cases where it fails.

Issue #2529, opened 2015-08-06
`getAttribute` call on `document-fragment`


pkaske:
**Hello jquery community,**

**First of all:**
Here's the [demo](http://pkaske.github.io/polymer-problems/jquery-event-delegation/index.html)

All relevant output show's up in the browser's console.
Here's a screenshot that shows the line in jQuery where the problem happens.
(You can see that `elem` is a `document-fragment`)
![selection_008](https://cloud.githubusercontent.com/assets/11567985/9084426/e9871192-3b73-11e5-8e19-7cd9f1b75dde.png)

**Source of the demo**
Can be found [here](https://github.com/pkaske/polymer-problems/tree/master/jquery-event-delegation)

**Situation**
You have a delegate click event handler that targets elements by a specific attribute
`$(document).on('click', '[some-attribute]', function() {...});`

You have a custom element that uses html5 templates to create a simple list of items.
Each item has a button to remove the parent item itself from the list.

**What happens**
The click on a items button removes the element before the event reaches the jQuery event handler. At the time the event arrives, the item has been removed and placed in a document fragment. Now jQuery tests the event selector on the document fragment and that fails of course.

I guess the fix would simply be to check if the element tested on the selector actually has a attribute api?

gibson042:
I am unable to reproduce with your demo. Can you provide a reduced test case (e.g., without Polymer) against the [git build](http://code.jquery.com/jquery-git.js), preferably somewhere like [JS Bin](http://jsbin.com/) so it's easy for us to tinker with it?

pkaske:
Haha, sorry. I accidentally uploaded my temp. workaround... (With is a timeout that waits till the event bubbled up and then remove the element).

Here is a demo on JS BIN:
http://jsbin.com/puvifoxebi/edit?html,console,output

Sorry, I didn't prepare a demo without Polymer, because my knowledge of HTML5 Templates is to narrow to set something like this up.

But I posted the same Issue at the polymer repository and got a explanation from them. This maybe provides you some clues on what's going on. There is also a better workaround for the problem in the conversion.

https://github.com/Polymer/polymer/issues/2210

dmethvin:
Here's a simple repro: http://jsbin.com/bobuhovici/1/edit?html,js,output

gibson042:
[...]

Issue #2134, opened 2015-03-09
Standardize "invalid input" behavior


gibson042:
Many jQuery collection operations are only meaningful on elements or (somewhat less commonly) DOM nodes, but the collections can logically contain _anything_. Such operations should **always** assume valid input, potentially throwing exceptions when they encounter an invalid member—especially given the ease of manually cleaning like `$collection.filter("*")`.
* Single-member read (e.g., `.attr( name )`, `.css( name )`, `.offset()`)
* Whole-collection read (e.g., `.hasClass( clazz )`)
* Whole-collection write (e.g., `.attr( name, val )`, `.css( props )`, `.offset( n )`, `.addClass( clazz )`)

Special consideration is warranted for single-member reads responding to empty collections. Current behavior usually returns `undefined`, which seems reasonable to standardize on (including changing those methods that return `null`, like `.scrollTop()` and `.width()`). There are cases where this collides with real output, but such collisions seem to be sane—`$el.css( unknownProperty )`, `$el.prop( nonexistent )`, `$hasNoData.data()`, etc.

Related:
* #2114
* #2115

dmethvin:
I'm all for this in theory, but have the queasy feeling that there are probably cases where existing code is depending on undocumented fail-silent or other quirky behavior.

timmywil:
Per today's meeting, we agreed that we'd like to try out some changes and see how it goes. Let's nail down all of the places we'd like to see this change.

timmywil:
I think we all agree here, but we'll need to address each change on a case-by-case basis and assess its impact on the community. Rather than making sweeping changes across the board and asking everyone to make lots of adjustments to their code, we're going to do this incrementally, in stages. The niche use cases can be done first with a few of the more common cases being introduced every major release. We already have some changes in the pipeline. For any other changes, let's have specific issues opened for each one. I'm closing this issue because we've reached a consensus and now it's too broad.

gibson042:
You closed this without making more specific tickets... are we just going to independently address unsolicited pull requests removing guards?

And even if the answer to the above is affirmative, I think we should include the principle—"assume function input matches a documented signature"—somewhere in the repository for future reference. Maybe [CONTRIBUTING.md](https://github.com/jquery/jquery/blob/master/CONTRIBUTING.md)?

timmywil:
[...]

PR #2782, opened 2015-12-17
Attributes:Fix get attribute on Document Fragments

Brooooooklyn:
should never trying to call getAttribute on DocumentFragment

http://jsbin.com/qoficotogo/1/edit?html,js,output

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -33,8 +33,8 @@ jQuery.extend( {
 		var ret, hooks,
 			nType = elem.nodeType;
 
-		// Don't get/set attributes on text, comment and attribute nodes
-		if ( nType === 3 || nType === 8 || nType === 2 ) {
+		// Don't get/set attributes on text, comment and attribute nodes and document fragments
+		if ( nType === 3 || nType === 8 || nType === 2 || nType === 11 ) {
 			return;
 		}
 


timmywil:
Thanks for submitting a PR. However, first open an issue with a test case from http://jsfiddle.net or http://jsbin.com so we can review. Once the issue has been verified, this PR will need to add tests.

gibson042:
Also note that our general trend has been to allow exceptions from invalid input:
* #2320
* #2134
* #2248

So generally speaking, this would be better fixed in the code that calls jQuery.

Brooooooklyn:
```
    $(document).ready(function() {
      $(document).on('click', '[data-time="123"]', function() {
        console.log('never reached');
      });
      var frag = $(document.createDocumentFragment());
      var ele = $('<div class="wrap"><button class="btn btn-danger btn-cancel">click me!!!!!!!</button></div>');
      $(document.body).append(ele);
      $('.btn-cancel').on('click', function() {
        frag.append(ele);
      });
    });
```

consider about these code, how could I avoid this issue? I'm wondering if I shounldn't use documentFragment to cache my DOM, or I shouldn't use attributes selector in event listener?

I think jQuery shouldn't getAttribute from a documentFragment anytime.

This is the online DEMO: `http://jsbin.com/qoficotogo/1/edit?html,js,output`

gibson042:
Ah, that looks like a duplicate of #2529, which was fixed by 9d820fbde6d89bc7a06e2704be61cf6c0b4d6e3c (but not yet released). You can confirm by loading jQuery from https://code.jquery.com/jquery-git.js .

PR #2786, opened 2015-12-21
Update serialize.js

MathijsvVelde:
Some folks like to work with Objects instead of Arrays...

### DIFF
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -88,9 +88,35 @@ jQuery.param = function( a, traditional ) {
 	return s.join( "&" );
 };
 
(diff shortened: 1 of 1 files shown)


mzgol:
Thanks but it's a duplicate of #2772 and the rationale for rejecting this PR is identical as for the mentioned one.

MathijsvVelde:
It's not a duplicate of #2772, they do not include support for multi-select and checkboxes.
But @dmethvin closed that one and therefore you closed mine because there are several good plugins available for it.

thanks anyway!!

PR #2789, opened 2015-12-29
removed proto properties.

Gcaufy:
This is the test code
```javascript
Object.prototype.__test = function () {};
var t = {};
$.each(t, function (k, v) {console.log(k, v); });
```
This is not what I want.


mgol:
Modifying `Object.prototype` is not supported: https://contribute.jquery.org/wont-fix/#object-prototype-issues. If you do it, you're on your own (it's generally a bad idea, though).

Issue #2792, opened 2015-12-30
css('animation-iteration-count', 5) doesn't apply style to element


ufologist:
When I use ```css('animation-iteration-count', 5)``` I expect element has ```style="-webkit-animation-iteration-count: 5;"``` (this vendor prefix is just a example), but it doesn't.

So I debug jQuery.js, I found when I use ```css('animation-iteration-count', 5)``` actually become ```style[ 'WebkitAnimationIterationCount' ] = 5 + 'px'```.
Browser consider it a illegal value, because ```animation-iteration-count``` only accept ```infinite | <number>```, so this style don't have effect.

This is a [reproduceBug demo](http://output.jsbin.com/gupugawibe)

So that ```cssNumber``` shoud add ```animationIterationCount``` property to prevent automatically add "px" to the value.

ufologist:
#2793

mgol:
Reopening, this shouldn't have been closed.

PR #2793, opened 2015-12-30
CSS: Add animation-iteration-count to cssNumber jquery/jquery#2792

ufologist:
#2792

### DIFF
--- a/test/unit/css.js
+++ b/test/unit/css.js
@@ -848,21 +848,21 @@ QUnit.test( "Do not append px (#9548, #12990, #2792)", function( assert ) {
 	$div.css( "fill-opacity", 1 );
 
 	// Support: Android 2.3 (no support for fill-opacity)
-	if ( $div.css( "fill-opacity" ) ) {
+	if ( $div.css( "fill-opacity" ) !== undefined ) {
 		assert.equal( $div.css( "fill-opacity" ), 1, "Do not append px to 'fill-opacity'" );
 	} else {
 		assert.ok( true, "No support for fill-opacity CSS property" );
 	}
 
 	$div.css( "column-count", 1 );
-	if ( $div.css( "column-count" ) ) {
+	if ( $div.css( "column-count" ) !== undefined ) {
 		assert.equal( $div.css( "column-count" ), 1, "Do not append px to 'column-count'" );
 	} else {
 		assert.ok( true, "No support for column-count CSS property" );
 	}
 
 	$div.css( "animation-iteration-count", 2 );
-	if ( $div.css( "animation-iteration-count" ) ) {
+	if ( $div.css( "animation-iteration-count" ) !== undefined ) {
 		// if $div.css( "animation-iteration-count" ) return "1",
 		// it actually return the default value of animation-iteration-count
 		assert.equal( $div.css( "animation-iteration-count" ), 2, "Do not append px to 'animation-iteration-count'" );
--- a/test/unit/css.js
+++ b/test/unit/css.js
@@ -840,8 +840,8 @@ if ( jQuery.fn.offset ) {
 	} );
 }
 
-QUnit.test( "Do not append px (#9548, #12990)", function( assert ) {
-	assert.expect( 2 );
+QUnit.test( "Do not append px (#9548, #12990, #2792)", function( assert ) {
+	assert.expect( 3 );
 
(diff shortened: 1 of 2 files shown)


ufologist:
I only want to fix a litter bug, why not?

[CLA Verification Results](http://contribute.jquery.org/CLA/status/?repo=jquery&sha=12515b5a27b744c4614eaf5ddabc8bc6b5becb84)
> The name ufologist requires manual verification.

I have done below
1. Signing the CLA
2. Updating Git Config
3. git commit --amend --reset-author
4. git push -f origin master

mgol:
Is ufologist your full name? The CLA check requires a full name (first & last) unless someone has a reason to hide it.

ufologist:
Thanks @mgol
ufologist is my nickname, but even I config and sign my full name, it does not work too.

[CLA Verification Results](http://contribute.jquery.org/CLA/status/?repo=jquery&sha=4f1a6b283fa03ed867f50c1b6f94cf55e6a84cdc)
> ufologist <xxx>: Name on signature (xxx) doesn't match name on commit.

I need more help. :(

mgol:
@ufologist Your commits must be signed by your full name as well. You can modify it via:
```
git commit --amend --author "FirstName LastName <email@address>"
```

You have 3 commits, though, every one of them would have to be modified. You might want to squash them to one commit first.

mgol:
@jquery/core This is a mess, we're adding more & more stuff to this array and it'll always be a problem as CSS is expanding. I'd like to get rid of this `px`-appending but it might be too big a breaking change.

This is probably too late for 3.0 but how about switching the blacklist to a whitelist? If jQuery doesn't append `px` for you it's easier to fix by yourself than if it does append erroneously. My guess would be 99% of the uses are common properties like `(min-|max-|)(height|width)` or `top`, `left` etc. We could identify those most popular ones and append `px` only to them. Migrate could restore the previous behavior but warn if someone passes a numeric value to properties that are not on the current blacklist.

markelog:
Would you mind creating another issue about it? Otherwise we might clutter this one.

mgol:
> Would you mind creating another issue about it? Otherwise we might clutter this one.

Right, created #2795 for that.

mgol:
I'm looking at the added tests and the ones before it and they seem wrong... We're checking for truthiness of `$div.css( "aProperty" )` just after we set it but if jQuery was erroneously adding `px` to values for this property then the browser would reject it and set an empty string which is falsy; this would lead us to incorrectly treat the browser as not supporting the property and, as a result, pass the test. Shouldn't we check:
```js
$div.css( "aProperty" ) !== undefined
```
instead?

ufologist:
Thanks @mgol

I understand that set 'animation-iteration-count' by css() is not a common use.
But if not anyone find the same problem except me?

> My guess would be 99% of the uses are common properties like (min-|max-|)(height|width) or top, left etc.

I accpet that
> warn if someone passes a numeric value to properties that are not on the current blacklist.

Than I will know 'animation-iteration-count' property need added to ```jQuery.cssNumber``` by myself, or I can do ```css('animation-iteration-count', '2')``` passby the rule(use a string instead a numeric).

markelog:
@ufologist Why this was closed? 

I don't think @mgol comment was implying that we don't want your changes.

mgol:
This PR is fine, I proposed changing an approach but that's a separate duscussion. This PR just needs changes I mentioned in my last comment. I'm reopening. M

test/unit/css.js
```diff
@@ -860,6 +860,15 @@ QUnit.test( "Do not append px (#9548, #12990)", function( assert ) {
 	} else {
 		assert.ok( true, "No support for column-count CSS property" );
 	}
+
+	$div.css( "animation-iteration-count", 2 );
```
mgol: @ufologist Would you like to change this line (and a few previous ones) according to what I wrote in https://github.com/jquery/jquery/pull/2793#issuecomment-168035588? I think it would be good to merge after that's done.

ufologist: Thanks @mgol.
I have pass the CLA and change the test with your suggestion.

mgol:
Landed on df822caff079177d1840d67e03d6b24a93ea99a5 (`master`), b9a695865b4eb3916f7320026d028a2866da6e4e (`2.2-stable`) & 01fb17be63d3d08789303f2e0eb7b6fcccc8fd56 (`1.12-stable`). Thanks!

PR #2807, opened 2016-01-10
Add global window.jquery object

RobLoach:
This introduces the ability to reference jQuery through window.jquery in addition to the original window.$ or window.jQuery. This makes the AMD and CommonJS module name (`jquery`) match the possible global name (`window.jquery`).

### DIFF
--- a/src/jquery.js
+++ b/src/jquery.js
@@ -32,6 +32,6 @@ define( [
 	"./exports/amd"
 ], function( jQuery ) {
 
-return ( window.jQuery = window.$ = jQuery );
+return ( window.jQuery = window.jquery = window.$ = jQuery );
 
 } );


mgol:
Why is this needed?

RobLoach:
Hmm, looking into this further, jQuery is on AMD, so the global namespace should never actually happen?

> Why is this needed?

Thinking was to match the canonical module name across AMD, CommonJS, and the global namespace. When compiling the follow code with Babel and [`transform-es2015-modules-umd`](http://babeljs.io/docs/plugins/transform-es2015-modules-umd/):

``` javascript
let $ = require('jquery');
```

The result is:

``` javascript
'use strict';

(function (global, factory) {
  if (typeof define === "function" && define.amd) {
    define(['jquery'], factory);
  } else if (typeof exports !== "undefined") {
    factory(require('jquery'));
  } else {
    var mod = {
      exports: {}
    };
    factory(global.jquery);
    global.packagename = mod.exports;
  }
})(this, function ($) {});
```

As you see, it gives us `global.jquery`, which doesn't currently exist. But, if AMD is used across it, would the global namespace ever be used?

timmywil:
Thanks for opening a PR, but I'd like to avoid this particular solutions. This day and age, 2 globals is almost too many. However, I can answer the AMD question. When AMD is used, `$` and `jQuery` are still exposed to the window. The only environment that does not expose a global is one that uses `module.exports` and has a proper window with a document (i.e. browserify). There are several reasons for this; we deliberated for a long time and it's not likely to change.

Issue #2808, opened 2016-01-11
Tripple negation and text nodes


vanderlee:
A while back, A.Wolff found an inconsistency in jQuery and posted it on StackOverflow.
I thought A.Wolff would submit this issue, but I can't find he did, so here it is:

http://stackoverflow.com/questions/34593399/triple-negation-needed-to-retrieve-text-node/34598245

In short; non-element (e.g. text) nodes in recursive `:not(*)` (and probably other) pseudo-selectors are treated differently for `.filter()` and `.not()`.

I dug into the issue and found the cause. I'm not sure whether to consider this a bug or not, but there is an easy fix (in the answer) to make it atleast consistent. The cost of the fix is quite low.

timmywil:
Fiddle: https://jsfiddle.net/m62syb56/

PR #2809, opened 2016-01-11
Fix for #2808; inconsistent .not/.filter

vanderlee:
Inconsistent ehaviour for non-element nodes when using multiple recursive pseudo-selectors on `*` in `.not()` and `.filter()`.

Fixes gh-2808

### DIFF
--- a/src/traversing/findFilter.js
+++ b/src/traversing/findFilter.js
@@ -33,7 +33,7 @@ function winnow( elements, qualifier, not ) {
 	}
 
 	return jQuery.grep( elements, function( elem ) {
-		return ( indexOf.call( qualifier, elem ) > -1 ) !== not;
+		return ( indexOf.call( qualifier, elem ) > -1 ) !== not && elem.nodeType === 1;
 	} );
 }
 
--- a/test/unit/traversing.js
+++ b/test/unit/traversing.js
@@ -430,6 +430,23 @@ QUnit.test( "not(jQuery)", function( assert ) {
 	assert.deepEqual( jQuery( "p" ).not( jQuery( "#ap, #sndp, .result" ) ).get(), q( "firstp", "en", "sap", "first" ), "not(jQuery)" );
 } );
 
+QUnit.test( "contents().not(':not(:not(*)')", function( assert ) {
+	assert.expect( 1 );
+
+	var elem = jQuery( "<div>foo<b>bar</b></div>" );
+
+	assert.deepEqual( elem.contents().not( ":not(:not(*))" ).get(), [ ], "contents().not(':not(:not(*))')" );
+} );
+
+QUnit.test( "contents().not(':not(:not(:not(*))')", function( assert ) {
+	assert.expect( 2 );
+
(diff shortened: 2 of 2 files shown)


src/traversing/findFilter.js
```diff
@@ -33,7 +33,7 @@ function winnow( elements, qualifier, not ) {
 	}
 
 	return jQuery.grep( elements, function( elem ) {
-		return ( indexOf.call( qualifier, elem ) > -1 ) !== not;
+		return ( indexOf.call( qualifier, elem ) > -1 ) !== not && elem.nodeType === 1;
```
staabm: the cheaper check should be evaled first

vanderlee: It may be a cheaper check on it's own, but it's a far less discriminating check in most use cases as it would only ever be `false` when using non-element nodes.

As far as I know, only the `.contents()` function can return non-element nodes, so not a very common situation.

Placing the `nodeType` check last would shortcut it in most cases.

Placing the `nodeType` check first would mean both checks would have to be evaluated in most cases.

vanderlee:
I think I signed the CLA with "Martijn W. van der Lee" instead of "Martijn van der Lee" name, but I can't seem to change the CLA.

gibson042:
The code change looks good, but it needs `jQuery.filter` and/or `jQuery.not` unit tests backing it up.

vanderlee:
Unittests should be simple enough.
I'll look into it somewhere today or tomorrow.

vanderlee:
Added two unittests which cover the issue.
They both fail without the fix, but work with the fix.

src/traversing/findFilter.js
dmethvin: True, the `nodeType` check is pretty rarely going to fail so the order isn't too important here.

test/unit/traversing.js
```diff
@@ -430,6 +430,23 @@ QUnit.test( "not(jQuery)", function( assert ) {
...
 
+QUnit.test( "contents().not(':not(:not(*)')", function( assert ) {
+	assert.expect( 1 );
+
+	var elem = jQuery( "<div>foo<b>bar</b></div>" );
+
+	assert.deepEqual( elem.contents().not( ":not(:not(*))" ).get(), [ ], "contents().not(':not(:not(*))')" );
+} );
+
+QUnit.test( "contents().not(':not(:not(:not(*))')", function( assert ) {
```
gibson042: Can you combine these tests together into something like `"not(Selector) excludes non-element nodes (gh-2808)"`, and use simpler selectors? E.g.,
```js
var mixedContents = jQuery( "#nonnodes" ).contents(),
	childElements = q( "nonnodesElement" );
assert.deepEqual( mixedContents.not( "*" ).get(), [], "not *" );
assert.deepEqual( mixedContents.not( "[id=a],[id=b]" ).get(), childElements, "not [id=a],[id=b]" );
assert.deepEqual( mixedContents.not( "[id=a],*,[id=b]" ).get(), [], "not [id=a],*,[id=b]" );
```
(https://github.com/vanderlee/jquery/blob/patch-1/test/index.html#L210)

Issue #2654, opened 2015-10-16
Restore previous show/hide behavior


timmywil:

mzgol:
We need to remember to re-close all solved tickets as won't fix (which means removing the milestones from them).

timmywil:
It sounds like @markelog wants to take this, if that's alright with @gibson042. Note that after restoring the old behavior, we'll still want to make a couple changes...

1. only cache inline display values
2. only cache the display value if it was set by the user
3. Ensure we are initially trying to show elements by removing the display value

gibson042:
I'm fine with yielding this, but make no mistake about what a monster it is. I'd like to see the following tables filled in, agreed upon, and _then_ fixed and asserted in a PR:

# Expected inline `display`
## Non-animated
### Initially `block`
Initial state | `.show()` | `.hide()` or `.toggle()` | `.hide().show()` or `.toggle().toggle()` | `.show().hide()`
---- | ---- | ---- | ---- | ----
default `block`, no inline | 
cascade `block`, no inline | 
default `list-item`, inline `block` | 
cascade `list-item`, inline `block` | 
cascade `none`, inline `block` | 

### Initially `inline`
Initial state | `.show()` | `.hide()` or `.toggle()` | `.hide().show()` or `.toggle().toggle()` | `.show().hide()`
---- | ---- | ---- | ---- | ----
default `inline`, no inline | 
cascade `inline`, no inline | 
default `block`, inline `inline` | 
cascade `block`, inline `inline` | 
cascade `none`, inline `inline` | 

### Initially `list-item`
Initial state | `.show()` | `.hide()` or `.toggle()` | `.hide().show()` or `.toggle().toggle()` | `.show().hide()`
---- | ---- | ---- | ---- | ----
default `list-item`, no inline | 
cascade `list-item`, no inline | 
default `inline`, inline `list-item` | 
cascade `inline`, inline `list-item` | 
cascade `none`, inline `list-item` | 

### Initially `none`
Initial state | `.show()` or `.toggle()` | `.hide()` | `.hide().show()` | `.show().hide()` or `.toggle().toggle()`
---- | ---- | ---- | ---- | ----
~~default `none`, no inline~~ | 
cascade `none`, no inline | 
default `block`, inline `none` | 
default `inline`, inline `none` | 
default `list-item`, inline `none` | 
cascade `block`, inline `none` | 
cascade `inline`, inline `none` | 
cascade `list-item`, inline `none` | 
[...]

PR #2810, opened 2016-01-11
CSS: Restore cascade-override behavior in .show

gibson042:
Fixes gh-2654
Fixes gh-2308
Ref 86419b1

```
   raw     gz Compared to master @ fb9472c7fbf9979f48ef49aff76903ac130d0959    
  +954   +271 dist/jquery.js                                                   
  +310    +99 dist/jquery.min.js
```

### DIFF
--- a/src/css/showHide.js
+++ b/src/css/showHide.js
@@ -6,19 +6,27 @@ define( [
 
 var defaultDisplayMap = {};
 
-function setDefaultDisplay( elem ) {
-	var doc = elem.ownerDocument,
+function getDefaultDisplay( elem ) {
+	var temp,
+		doc = elem.ownerDocument,
 		nodeName = elem.nodeName,
-		temp = doc.body.appendChild( doc.createElement( nodeName ) ),
-		display = jQuery.css( temp, "display" );
+		display = defaultDisplayMap[ nodeName ];
+
+	if ( display ) {
+		return display;
+	}
+
+	temp = doc.body.appendChild( doc.createElement( nodeName ) ),
+	display = jQuery.css( temp, "display" );
 
 	temp.parentNode.removeChild( temp );
 
 	if ( display === "none" ) {
 		display = "block";
 	}
+	defaultDisplayMap[ nodeName ] = display;
 
-	return ( defaultDisplayMap[ nodeName ] = display );
+	return display;
 }
 
 function showHide( elements, show ) {
@@ -47,7 +55,7 @@ function showHide( elements, show ) {
 				}
 			}
 			if ( elem.style.display === "" && jQuery.css( elem, "display" ) === "none" ) {
-				values[ index ] = defaultDisplayMap[ elem.nodeName ] || setDefaultDisplay( elem );
+				values[ index ] = getDefaultDisplay( elem );
 			}
 		} else {
 			if ( display !== "none" ) {
--- a/src/css/showHide.js
+++ b/src/css/showHide.js
@@ -4,6 +4,23 @@ define( [
 	"../css/var/isHidden"
 ], function( jQuery, dataPriv, isHidden ) {
 
+var defaultDisplayMap = {};
+
+function setDefaultDisplay( elem ) {
+	var doc = elem.ownerDocument,
+		nodeName = elem.nodeName,
+		temp = doc.body.appendChild( doc.createElement( nodeName ) ),
+		display = jQuery.css( temp, "display" );
+
+	temp.parentNode.removeChild( temp );
+
+	if ( display === "none" ) {
+		display = "block";
+	}
+
+	return ( defaultDisplayMap[ nodeName ] = display );
+}
+
 function showHide( elements, show ) {
 	var display, elem,
 		values = [],
@@ -19,23 +36,30 @@ function showHide( elements, show ) {
 
 		display = elem.style.display;
 		if ( show ) {
-			if ( display === "none" ) {
 
-				// Restore a pre-hide() value if we have one
-				values[ index ] = dataPriv.get( elem, "display" ) || "";
+			// Since we force visibility upon cascade-hidden elements, an immediate (and slow)
+			// check is required in this first loop unless we have a nonempty display value (either
+			// inline or about-to-be-restored)
+			if ( display === "none" ) {
+				values[ index ] = dataPriv.get( elem, "display" ) || null;
+				if ( !values[ index ] ) {
+					elem.style.display = "";
+				}
+			}
+			if ( elem.style.display === "" && jQuery.css( elem, "display" ) === "none" ) {
+				values[ index ] = defaultDisplayMap[ elem.nodeName ] || setDefaultDisplay( elem );
 			}
 		} else {
 			if ( display !== "none" ) {
 				values[ index ] = "none";
 
-				// Remember the value we're replacing
+				// Remember what we're overwriting
 				dataPriv.set( elem, "display", display );
 			}
 		}
 	}
 
-	// Set the display of the elements in a second loop
-	// to avoid the constant reflow
+	// Set the display of the elements in a second loop to avoid constant reflow
 	for ( index = 0; index < length; index++ ) {
 		if ( values[ index ] != null ) {
 			elements[ index ].style.display = values[ index ];
--- a/test/unit/css.js
+++ b/test/unit/css.js
@@ -661,19 +661,29 @@ QUnit.test( "show/hide 3.0, default display", function( assert ) {
 	var i,
 		$elems = jQuery( "<div/>" )
 			.appendTo( "#qunit-fixture" )
-			.html( "<div title='block'/><span title='inline'/><ul><li title='list-item'/></ul>" )
-			.find( "[title]" );
+			.html( "<div data-expected-display='block'/>" +
+				"<span data-expected-display='inline'/>" +
+				"<ul><li data-expected-display='list-item'/></ul>" )
+			.find( "[data-expected-display]" );
 
 	$elems.each( function() {
 		var $elem = jQuery( this ),
-			name = this.nodeName + this.className.replace( /(?=.)/, "." ) +
-				( this.getAttribute( "style" ) || "" ).replace( /.+/, "[style='$&']" ),
+			name = this.nodeName,
+			expected = this.getAttribute( "data-expected-display" ),
 			sequence = [];
+
+		if ( this.className ) {
+			name += "." + this.className;
+		}
+		if ( this.getAttribute( "style" ) ) {
+			name += "[style='" + this.getAttribute( "style" ) + "']";
+		}
+		name += " ";
 
 		for ( i = 0; i < 3; i++ ) {
 			sequence.push( ".show()" );
 			$elem.show();
-			assert.equal( $elem.css( "display" ), this.title,
+			assert.equal( $elem.css( "display" ), expected,
 				name + sequence.join( "" ) + " computed" );
 			assert.equal( this.style.display, "", name + sequence.join( "" ) + " inline" );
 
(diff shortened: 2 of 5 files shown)


src/css/showHide.js
```diff
@@ -4,6 +4,23 @@ define( [
 	"../css/var/isHidden"
 ], function( jQuery, dataPriv, isHidden ) {
 
+var defaultDisplayMap = {};
+
+function setDefaultDisplay( elem ) {
```
gibson042: If the "force visibility" behavior is here to stay, we should consider a public `jQuery.defaultDisplay( element )` interface so this function doesn't go the way of its predecessor.

src/css/showHide.js
```diff
@@ -4,6 +4,23 @@ define( [
...
+	var doc = elem.ownerDocument,
+		nodeName = elem.nodeName,
+		temp = doc.body.appendChild( doc.createElement( nodeName ) ),
+		display = jQuery.css( temp, "display" );
+
+	temp.parentNode.removeChild( temp );
+
+	if ( display === "none" ) {
+		display = "block";
```
markelog: Now this is bold move... don't now how i feel about it. 

We definitely need tests from https://github.com/jquery/jquery/commit/60f546acb1c7136092b4fd01cccff052e468cc72

test/unit/css.js
```diff
@@ -664,6 +654,190 @@ QUnit.test( "show() after hide() should always set display to initial value (#14
 	assert.equal( div.css( "display" ), "list-item", "should get last set display value" );
 } );
 
+QUnit.test( "show/hide 3.0, default display", function( assert ) {
```
markelog: These tests looks pretty complicated, i would suggest following some sort of [AAA](http://c2.com/cgi/wiki?ArrangeActAssert) strategy.

This could be done latter, but i would highly encourage you do it

markelog:
How this correlates with your table there - https://gist.github.com/scottgonzalez/5aadcd7e37cccced1239

In the meeting we discuss the possibility of documenting it. I'd say we need to tests for every value of it.
Not a blocker though, but i think we shouldn't put it under the bed

src/css/showHide.js
```diff
@@ -19,23 +36,30 @@ function showHide( elements, show ) {
...
+			// check is required in this first loop unless we have a nonempty display value (either
+			// inline or about-to-be-restored)
+			if ( display === "none" ) {
+				values[ index ] = dataPriv.get( elem, "display" ) || null;
+				if ( !values[ index ] ) {
+					elem.style.display = "";
+				}
+			}
+			if ( elem.style.display === "" && jQuery.css( elem, "display" ) === "none" ) {
+				values[ index ] = defaultDisplayMap[ elem.nodeName ] || setDefaultDisplay( elem );
```
markelog: So `setDefaultDisplay` is a getter? I would suggest to do write it like 
```js
values[ index ] = getDefaultDisplay(elem);
```

And incapsulate everything else

markelog:
With your
```js
if ( display === "none" ) {
  display = "block";
}
```
you drained all fun from it :-), but i guess we can check out users reaction from the beta release.

src/css/showHide.js
markelog: That's because of the `if none the block` thing? So if there would be a case, when, say inline element would be added to the hidden body, user could re-define it.

I like it, i would even consider it to be a blocker for the release

test/unit/css.js
```diff
@@ -664,6 +654,190 @@ QUnit.test( "show() after hide() should always set display to initial value (#14
...
+	var i,
+		$elems = jQuery( "<div/>" )
+			.appendTo( "#qunit-fixture" )
+			.html( "<div title='block'/><span title='inline'/><ul><li title='list-item'/></ul>" )
+			.find( "[title]" );
+
+	$elems.each( function() {
+		var $elem = jQuery( this ),
+			name = this.nodeName + this.className.replace( /(?=.)/, "." ) +
+				( this.getAttribute( "style" ) || "" ).replace( /.+/, "[style='$&']" ),
```
markelog: Yeah, please make it more simpler, pretty convoluted. I'd say tests is not the place where we need to be clever

src/css/showHide.js
gibson042: Nah, such an extension could easily be 3.1.0.

src/css/showHide.js
gibson042: I suppose the clarity is worth 4 bytes. Updating.

test/unit/css.js
gibson042: Do you have a suggestion? The assertions really are dependent upon methods that have already been called.

src/css/showHide.js
markelog: To to be clear, without that hook we loosing two use-cases:
1. When `body`/`html` is hidden
2. When all specific node-types are hidden, like 
```css
li {display: none}
```

First might exist in at least worlds like SPA or iframes, there is probably more weird use-cases out.
As of all specific elements being hidden, there might be some legit cases for very simple pages.

I think it would be good to provide some workarounds for it.

test/unit/css.js
markelog: Okay, let me dig deeper

gibson042:
@markelog Updated.

timmywil:
LGTM

timmywil:
I'd like to land this today if possible.

test/unit/css.js
markelog: Not sure what these lines supposed to do `className` and `style` are always empty

src/css/showHide.js
dmethvin: As you said above @markelog it misses a couple of cases, but if those aren't super common this might just work well enough. If the person writing a rule like `li { display: none }` is the author of the page they can just rearrange the HTML and CSS a bit to fix that. I think the case that would kill things would be plugins that need to force-hide content. There, hiding by tag name isn't as likely?

bold-strategy

test/unit/css.js
gibson042: Copypasta. They are the same in each test of this family.

gibson042:
> How this correlates with your table there - https://gist.github.com/scottgonzalez/5aadcd7e37cccced1239

> In the meeting we discuss the possibility of documenting it. I'd say we need to tests for every value of it.
Not a blocker though, but i think we shouldn't put it under the bed

I intentionally excluded elements like `<style>` and `<script>` that are `display: none` by default, but tried to include everything else. Did I miss something?

test/unit/css.js
markelog: I would move elements to their own test and then divide it again on `show` and `hide` tests too.

This sounds bad, since this is a lot of work and we don't do stuff like that in jQuery suite, but i would really want our tests to became more focused and clean.

Like moving this in their own `QUnit.module` (not into to separate file but into the module), and start using `beforeEach` here, like
```js
QUnit.module('css:show/hide default display', {
	beforeEach: function() {
		this.$div = jQuery( "<div></div>" ).appendTo( "#qunit-fixture" );
		this.$li = jQuery( "<ul><li></li></ul>" ).appendTo( "#qunit-fixture" ).find( "li" );
		// ... and so on
	}
});

QUnit.test( "hide sequence of `div` (block) element", function( assert ) {
	assert.expect( 3 );
	for ( i = 0; i < 3; i++ ) {
		this.$div.show().hide();
		assert.equal( this.$div.css('display'), 'none' );
	}
});

QUnit.test( "show sequence of `div` (block) element", function( assert ) {
	assert.expect( 3 );
	for ( i = 1; i < 4; i++ ) {
		this.$div.hide().show();
		assert.equal( this.$div.css('display'), 'block' );
	}
});
```
Here, we are still writing tests like we (well, maybe not "we" but @dmethvin :-) did literally ten years ago, 
whereas now js testing has evolved so much.

Yeah, this kinda a different topic, wouldn't hurt to start somewhere, hurt if we would start to do it right though. 

All these thoughts shouldn't block this pr though.

src/css/showHide.js
markelog: Test from https://github.com/jquery/jquery/commit/60f546acb1c7136092b4fd01cccff052e468cc72 should actually pass, this is why i mention it

Whereas `li { display: none }` wouldn't

markelog:
@gibson042  Oh, yeah, this table only for non-animated elements, can't find the full one though, which includes animated elements too.

@timmywil if everyone okay with `jQuery.defaultDisplay` being added only in 3.1.

src/css/showHide.js
markelog: Initially, i have wrote "outrages use-case" there - https://github.com/jquery/jquery/pull/2810#discussion_r49628474, but yeah totally, they really are like that i suppose, my biggest concern is hidden iframe though, which also shouldn't be that common.

timmywil:
done-good

http://jsperf.com/old-vs-new-show-hide

src/css/showHide.js
```diff
@@ -19,23 +44,30 @@ function showHide( elements, show ) {
...
+			// check is required in this first loop unless we have a nonempty display value (either
+			// inline or about-to-be-restored)
+			if ( display === "none" ) {
+				values[ index ] = dataPriv.get( elem, "display" ) || null;
+				if ( !values[ index ] ) {
+					elem.style.display = "";
+				}
+			}
+			if ( elem.style.display === "" && jQuery.css( elem, "display" ) === "none" ) {
+				values[ index ] = getDefaultDisplay( elem );
```
gibson042: Correct. And notably, the input is an element rather than a tag name—it allows for context-sensitivity, but doesn't incur any significant extra cost for a simple cached global treatment like ours.

markelog:
@timmywil the reason for speed bump, is that we are dropping the third case - all re-attached elements will have incorrect display if `show()`/`hide()` applied while they were detached, for reference 
https://bugs.jquery.com/ticket/10416
https://bugs.jquery.com/ticket/10006

Now those has real legit use-cases, no question, if you feel strongly about dropping them, i think we should explicitly notice all the no longer supported use-cases, why we are dropping them and provide documentation for it too.

Gotta say, this is reason why should have brought back previous tests, without claiming them "bad", since we didn't know we doing that, they are regressions by definition.

markelog:
fixed http://jsperf.com/old-vs-new-show-hide/2, but this also doesn't reflect real use-case when perf questions were started, "wikipedia" one, when elements changed their display on the first run.

timmywil:
@markelog It's not invalid. You missed the point. It wasn't so much testing what happens when hiding elements, it was testing the codepath for hiding, which is faster. It doesn't matter whether they were already hidden or not. We were checking display regardless before. That's the reason for the speed boost.

markelog:
I don't understand how it could be correct, if you test the code path that would never be taken in real application.  It doesn't matter what code path would be taken and how fast it is for the bogus use-case. How would hide already hidden element?

> We were checking display regardless before. That's the reason for the speed boost.

I don't understand this statement. And you didn't answer to my suggestion in https://github.com/jquery/jquery/pull/2810#issuecomment-171573946

timmywil:
The use case is not that relevant. It was testing the hide path in showHide accurately. And I can't view the tickets you linked (502 Bad Gateway).

markelog:
> The use case is not that relevant.

I really don't get it, how it could be "not relevant" if that code path would never be taken? How in the world does it matter

> And I can't view the tickets you linked (502 Bad Gateway).

Oh, common -

http://webcache.googleusercontent.com/search?q=cache:3FSB7KkOQyAJ:https://bugs.jquery.com/ticket/10006+&cd=1&hl=en&ct=clnk&gl=ru

http://webcache.googleusercontent.com/search?q=cache:puR0v9SuaoUJ:https://bugs.jquery.com/ticket/10416+&cd=1&hl=en&ct=clnk&gl=ru

timmywil:
> I really don't get it

It is hitting the codepath whether the elements are hidden or not. http://jsperf.com/old-vs-new-show-hide/3 Re-show the elements in teardown instead and you can see that you get the same results.

As for the links, I always forget about webcache.

timmywil:
Ok, for the cases you mentioned, I think I agree that they seem rare enough. Let's discuss in chat, either in -dev or in a meeting, but do you agree they could be addressed after the beta?

timmywil:
@markelog and I talked this over in irc. We probably would have been commenting here for a while. We agreed that hide is faster and while the perf demonstrates that, what I didn't realize was that show might be faster for the reasons @markelog mentioned. We're figuring out what to do about the other tickets in -dev.

timmywil:
@gibson042 There's an issue with disconnected elements in FF. It looks like it's respecting the cascade even when disconnected. For instance, [this test](https://github.com/jquery/jquery/blob/master/test/unit/css.js#L557) fails because FF gets a computed display of "none" instead of empty string, gets a proper default display of block and applies it.

gibson042:
Confirmed the failing tests. I'll take a look later today.

Issue #2710, opened 2015-11-12
$.when + progress


timmywil:
http://jsfiddle.net/timmywil/608mu479/1/

We currently don't have tests for progress in $.when and there are some funky behaviors. [@dmethvin proposed](https://github.com/jquery/jquery/issues/2018#issuecomment-143330148) we drop "the current progress pass-thru and generate one progress notification for each resolved subordinate promise".

PR #2816, opened 2016-01-12
Deferred: Remove undocumented progress notifications in $.when

dmethvin:
Fixes gh-2710

Although we could add a new feature to notify on the new Deferred once each thenable resolved/rejected, it seemed like that would encourage more Deferred-specific behavior and I wasn't sure I wanted to go there. The important 3.0 behavior change here is removing the notification pass-through, which was [undocumented](http://api.jquery.com/jquery.when/). We can always add new functionality in a 3.x.0 if it's requested.

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -295,7 +295,7 @@ jQuery.extend( {
 
 	// Deferred helper
 	when: function() {
-		var method,
+		var method, resolveContexts,
 			i = 0,
 			resolveValues = slice.call( arguments ),
 			length = resolveValues.length,
@@ -306,20 +306,19 @@ jQuery.extend( {
 			// the master Deferred.
 			master = jQuery.Deferred(),
 
-			// Update function for both resolve and progress values
-			updateFunc = function( i, contexts, values ) {
+			// Update function for both resolving subordinates
+			updateFunc = function( i ) {
 				return function( value ) {
-					contexts[ i ] = this;
-					values[ i ] = arguments.length > 1 ? slice.call( arguments ) : value;
+					resolveContexts[ i ] = this;
+					resolveValues[ i ] = arguments.length > 1 ? slice.call( arguments ) : value;
 					if ( !( --remaining ) ) {
 						master.resolveWith(
-							contexts.length === 1 ? contexts[ 0 ] : contexts,
-							values
+							resolveContexts.length === 1 ? resolveContexts[ 0 ] : resolveContexts,
+							resolveValues
 						);
 					}
 				};
-			},
-			resolveContexts;
+			};
 
 		// Add listeners to promise-like subordinates; treat others as resolved
 		if ( length > 0 ) {
@@ -331,7 +330,7 @@ jQuery.extend( {
 					jQuery.isFunction( ( method = resolveValues[ i ].promise ) ) ) {
 
 					method.call( resolveValues[ i ] )
-						.done( updateFunc( i, resolveContexts, resolveValues ) )
+						.done( updateFunc( i ) )
 						.fail( master.reject );
 
 				// Other thenables
@@ -340,11 +339,11 @@ jQuery.extend( {
 
 					method.call(
 						resolveValues[ i ],
-						updateFunc( i, resolveContexts, resolveValues ),
+						updateFunc( i ),
 						master.reject
 					);
 				} else {
-					updateFunc( i, resolveContexts, resolveValues )( resolveValues[ i ] );
+					updateFunc( i )( resolveValues[ i ] );
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


test/unit/deferred.js
```diff
@@ -710,13 +710,12 @@ QUnit.test( "jQuery.when", function( assert ) {
 
 QUnit.test( "jQuery.when - joined", function( assert ) {
 
-	assert.expect( 195 );
+	assert.expect( 81 );
```
markelog: > We currently don't have tests for progress in $.when

So we do have tests?

dmethvin: Yes we did. However, the docs say nothing about it, and it seems strange to pass through the notifications from the subordinates all mixed into the master deferred.

src/deferred.js
```diff
@@ -321,29 +319,29 @@ jQuery.extend( {
 					}
 				};
 			},
-			progressValues, progressContexts, resolveContexts;
+			resolveContexts;
```
gibson042: Uninitialized variables should precede initialized variables.

src/deferred.js
```diff
@@ -321,29 +319,29 @@ jQuery.extend( {
...
 
-		// Add listeners to Deferred subordinates; treat others as resolved
+		// Add listeners to promise-like subordinates; treat others as resolved
 		if ( length > 0 ) {
-			progressValues = new Array( length );
-			progressContexts = new Array( length );
 			resolveContexts = new Array( length );
 			for ( ; i < length; i++ ) {
```
gibson042: `length` and `i` can be merged into a single variable if you reverse the direction of this loop (e.g., `while ( i-- ) {`).

src/deferred.js
```diff
@@ -321,29 +319,29 @@ jQuery.extend( {
...
+
+				// Other thenables
 				} else if ( resolveValues[ i ] &&
 					jQuery.isFunction( ( method = resolveValues[ i ].then ) ) ) {
 
 					method.call(
 						resolveValues[ i ],
 						updateFunc( i, resolveContexts, resolveValues ),
-						master.reject,
-						updateFunc( i, progressContexts, progressValues )
```
gibson042: Every call to `updateFunc` now passes the same arguments, and every branch inside this loop body invokes it. It doesn't seem to be a useful abstraction any longer.

gibson042:
This looks pretty good to me.

src/deferred.js
timmywil: Wouldn't that change the method call order?

dmethvin: It would, although I'm not sure that would be observable? Since we use Deferred with `.done()` it will cause the resolveValues to be filled in a different order if it happens synchronously, but we don't do anything with it externally until all the subordinates resolve.

src/deferred.js
dmethvin: Well it does capture the loop variable `i` in a closure, which is needed for the async resolves. The other args can be moved out though.

src/deferred.js
dmethvin: I pushed some updates but left this as-is.

src/deferred.js
```diff
@@ -306,47 +306,44 @@ jQuery.extend( {
...
-					if ( values === progressValues ) {
-						master.notifyWith( contexts, values );
-					} else if ( !( --remaining ) ) {
+					resolveContexts[ i ] = this;
+					resolveValues[ i ] = arguments.length > 1 ? slice.call( arguments ) : value;
+					if ( !( --remaining ) ) {
 						master.resolveWith(
-							contexts.length === 1 ? contexts[ 0 ] : contexts,
-							values
+							resolveContexts.length === 1 ? resolveContexts[ 0 ] : resolveContexts,
```
dmethvin: Technically this could just be `length` but it seemed clearer to leave it this way.

src/deferred.js
```diff
@@ -306,47 +306,44 @@ jQuery.extend( {
...
 			for ( ; i < length; i++ ) {
+
+				// jQuery.Deferred - treated specially to get resolve-sync behavior
 				if ( resolveValues[ i ] &&
 					jQuery.isFunction( ( method = resolveValues[ i ].promise ) ) ) {
 
 					method.call( resolveValues[ i ] )
-						.progress( updateFunc( i, progressContexts, progressValues ) )
-						.done( updateFunc( i, resolveContexts, resolveValues ) )
+						.done( updateFunc( i ) )
```
dmethvin: This could also be done by wrapping the loop body in a function if it seems better.  Preferences?

Or we could move to ES6 :smiling_imp:

gibson042:
LGTM.

Issue #2796, opened 2016-01-04
closest: remove exception for pos selectors


kingjiv:
Further discussion here:  http://stackoverflow.com/questions/34580013/jquery-closest-default-context

The documentation for closest states: 

>  If no context is passed in then the context of the jQuery set will be used instead.

However, it seems that this is often not the case.  You can see this illustrated here:

https://jsfiddle.net/pjbg4398/

When the context is passed explicitly, closest works as expected and only finds an element if it is within the context element.  However, if the jquery set has a context and closest is called without explicitly passing a context, no context is used (contradicting the documentation).  However, if the selector passed to closest contains non-css selectors (gt, eq, etc.), then the set's context **is** used.

I believe the fix should be to change the loop condition as follows:

>  for (cur = this[i]; cur && **cur !== context**; cur = cur.parentNode) {

should become:

>  for (cur = this[i]; cur && **cur !== (context || this.context)**; cur = cur.parentNode) {

which would be in line with the non-css code which does a similar thing with context:

>  pos = rneedsContext.test(selectors) || typeof selectors !== "string" ? jQuery(selectors, **context || this.context**) : 0;

dmethvin:
Hmmm. The `.context` property was deprecated as of 2.0 and will be removed in 3.0, see gh-1908, so it looks like the docs need to be changed in that respect

Even if the property was still present, `Sizzle.matchesSelector` doesn't take a `context` so any string selectors passed to `.closest()` with a context wouldn't be rooted at the context the way they should be which could cause incorrect results. The only place we use the `context` arg there is for positional selectors, I think it would be best to document that and leave the implementation as-is.

kingjiv:
Deprecation aside, you wouldn't need to be able to pass context to matchesSelector for it to work as documented but rather just update the loop condition as mentioned.  The documentation implies that this.context will be used in place of context if context is not passed.  Since passing context works as expected, you could certainly do the same with this.context.

That said, I think most people would be confused by it actually working as documented.  Since it generally doesn't work as documented (and hasn't for many releases--if ever), I would think it would be better to remove the "this.context" portion of the "pos" creation and just remove that line from the documentation. 

As it is now, it is inconsistent as using ":first" in your selector will use the context but without it it won't, etc.

dmethvin:
[...]

PR #2818, opened 2016-01-12
Traversing: Never let .closest() match positional selectors

gibson042:
Fixes gh-2796

cf. https://github.com/jquery/jquery/issues/2796#issuecomment-168841302
This changes a unit-tested behavior, but is not semver-major since the behavior was undocumented. Still, it probably should be included in 3.0 just in case.

Also adds a test for the acceptance of non-string input (e.g., `$elem.closest( $possibleParents )`).

Best viewed with [`-w`](2818/files?w=1).

### DIFF
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -39,23 +39,24 @@ jQuery.fn.extend( {
 			i = 0,
 			l = this.length,
 			matched = [],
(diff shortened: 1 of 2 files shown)


dmethvin:
LGTM. I agree this is a good thing to change in 3.0, even though it's undocumented, just in case someone is using positional selectors in some crazy way.

timmywil:
:+1:

PR #2821, opened 2016-01-13
Core: use interactive to evaluate dom ready, barring IE6-10.

dracos:
This is a copy of @timmywil's commit, applied to the 1.x branch. This shouldn't have any effect on the extra browsers 1.x supports due to the doScroll check, so I was hoping it could be added in order to provide the same benefit to other browsers as that commit :)

Fixes gh-2822

### DIFF
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -97,11 +97,10 @@ jQuery.ready.promise = function( obj ) {
 
 		// Catch cases where $(document).ready() is called
 		// after the browser event has already occurred.
-		// we once tried to use readyState "interactive" here,
-		// but it caused issues like the one
-		// discovered by ChrisS here:
-		// http://bugs.jquery.com/ticket/12282#comment:15
-		if ( document.readyState === "complete" ) {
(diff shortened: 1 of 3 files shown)


dmethvin:
Oh yeah, seems like we missed this one. 1.12 and 2.2 should definitely behave the same here.

timmywil:
Needs the test that was added as well. https://github.com/jquery/jquery/commit/dabd5ba96c05279b3ffb052db5b8d17f75996694

timmywil:
But I agree. Thanks for finding this!

dmethvin:
It would be better to cherry pick the change if possible, just to simplify auditing.

dracos:
"It would be better to cherry pick the change if possible, just to simplify auditing." - happy to do this, though it will need editing the 'Support' line - would you prefer that be done by editing the commit, or a second commit afterwards?

timmywil:
Editing the commit should be fine.

timmywil:
Thanks! Merged.

Issue #2795, opened 2015-12-30
Stop appending `px` to everything except a blacklist?


mgol:
We're currently appending `px` when using the `.css()` setter if the value given is numeric and the CSS property is not on the [`jQuery.cssNumber` blacklist](http://api.jquery.com/jQuery.cssNumber/).

From https://github.com/jquery/jquery/pull/2793#issuecomment-167990441:

>  This is a mess, we're adding more & more stuff to this array and it'll always be a problem as CSS is expanding. I'd like to get rid of this `px`-appending but it might be too big a breaking change.
> 
> This is probably too late for 3.0 but how about switching the blacklist to a whitelist? If jQuery doesn't append `px` for you it's easier to fix by yourself than if it does append erroneously. My guess would be 99% of the uses are common properties like `(min-|max-|)(height|width)` or `top`, `left` etc. We could identify those most popular ones and append `px` only to them. Migrate could restore the previous behavior but warn if someone passes a numeric value to properties that are not on the current blacklist.

timmywil:
I'd be interested to see how big the whitelist would be, but I like the idea.

markelog:
> I'd be interested to see how big the whitelist would be, but I like the idea.

Yeah, i guess that is the key, what list would be bigger.

> This is a mess, we're adding more & more stuff to this array and it'll always be a problem as CSS is expanding. 

The same argument could be applied to the blacklist too :/

mgol:
Not exactly the same as:

1. As I mentioned, if sth is not on a whitelist, 'px' would not be auto-appended which is trivial to fix - just append 'px' by yourself. If, on the other hand, it's erroneously appended, the fix is not that simple: one has to first figure out jQuery is appending 'px', one has to discover the blacklist is modifiable at jQuery.cssNumber and one has to modify the jQuery global state by appending the property name to this array. This is a much bigger mental overhead.
1. I'd like to not list all the properties that don't accept bare numbers but only a few popular ones, just to avoid the biggest breakages. I'm basically treating this "feature" as a mis-feature and keeping back compat only as little as necessary so that all hell doesn't break loose. In the future we might even drop this whole "feature" completely.

markelog:
Sounds reasonable, although, i don't think we had a lot of issues with that list, there isn't a lot of those properties, burden that we have to update it once in a while.

So i think whitelist strategy would be easier for the user to fix value of the specific property, as you said, but there should be a lot of widely used items in it, otherwise if not, it might confuse a lot of people.

[...]

PR #2823, opened 2016-01-13
Update CSS prefixes in css.js

amatecha:
Webkit's DOM element "style" prefixes start with a lowercase "w", not uppercase.

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -33,7 +33,7 @@ var
 		fontWeight: "400"
 	},
 
-	cssPrefixes = [ "Webkit", "Moz", "ms" ],
+	cssPrefixes = [ "webkit", "Moz", "ms" ],
 	emptyStyle = document.createElement( "div" ).style;
 
 // Return a css property mapped to a potentially vendor prefixed property


amatecha:
Filled out CLA, not sure why it's not verifying...

mgol:
Thanks but WebKit uses both styles of prefixing; otherwise it wouldn't work and we would have many bug reports about that. Moreover, the `Webkit` style seems more correct as you just change `-` + letter in CSS into a capital letter in JS. It's the `ms` prefix that's the weird one here.

amatecha:
So... how will .css() apply my prefix-only style when:  http://i.imgur.com/sfpwxSp.jpg  ?

Filing a bug for the problem that just prompted me to submit this PR, then...

mgol:
For some reason Chrome DevTools prefer the small letter version and suggest that one but both work. Check it yourself, `d.style.WebkitAppRegion` will be an empty string, not `undefined`.

amatecha:
Ahh, that's true!  With further investigation I see the reason the rule isn't being applied is because the .css() function is converting "2" to "2px"...

mgol:
Oh boy, another one. :( I agree it's a problem, subscribe to #2795 to follow progress on that matter.

mgol:
And thanks for the report! What you can do now is:
```js
jQuery.cssNumber.appRegion = true;
```
after including jQuery `1.12` or `2.2` but before your client code. See [`jQuery.cssNumber` docs](https://api.jquery.com/jQuery.cssNumber/) for more info.

amatecha:
Thanks @mgol  :)

Issue #2812, opened 2016-01-11
Chaining .on() broken for null handlers in jQuery 2.2.0 (inconsistency with 1.12.0)


dwilson6:
It looks like a bug was introduced in jquery 2.2.0 with the 'on' function.

If you have code that registers a bunch of handlers in a chain and one of the callbacks is null, the on function will return the window object when the callback is null which causes the chaining to fail.  This works with jquery 2.1.4.  After a little investigation, it looks like the 'on' method was refactored to be global within jquery's scope and returns this (which used to be the jquery object but now is window).

//example from jquery impromptu plugin
$('.selector').on("keydown",keyDownEventHandler)
						.on('impromptu:loaded', opts.loaded)
						.on('impromptu:close', opts.close)
						.on('impromptu:statechanging', opts.statechanging)
						.on('impromptu:statechanged', opts.statechanged);

dmethvin:
A callback can't be `null`, it's not a documented input. Previously that silently did nothing but it seems like that makes it harder to find the bug causing a null input.

mgol:
It's weird, though, we brought the check for `null` back in `2.2.0`, there's a test for the `click` event:
https://github.com/jquery/jquery/blob/33b548c8e3d43b2ebdfb129fd8086a3b0c905919/test/unit/event.js#L8-L21.

dwilson6:
Correct.  The callback doesn't necessarily have to be null but if its not a function, window is returned

dwilson6:
After looking more at the source code, it seems like this reverted commit is the culprit (the on function was broken out so 'this' has a different meaning now): https://github.com/jquery/jquery/commit/3cc496ceecdbd76b6c556e85b12696f7c4c09dd5

mgol:
A function or `false` which is treated specially. As @dmethvin mentioned, this is not a supported input so we can technically do what we want here but I'd like to see what's really happening here. I don't immediately see why the `this` leaks the global object since the functions in which it appears are supposed to be bound to the proper elements.

Thanks for the report!

mgol:
@dwilson6 
> After looking more at the source code, it seems like this reverted commit is the culprit (the on function was broken out so 'this' has a different meaning now): 3cc496c

Good point, that's it!

Note that we've removed this check on `master`, using `.on(event, null)` will throw in jQuery 3.0.0. We've only restored it here so that both 1.12.0 & 2.2.0 don't throw... but the fix is buggy as it seems.

[...]

PR #2825, opened 2016-01-14
Event: Fix chaining .on() with null handlers

dwilson6:
This is to fix https://github.com/jquery/jquery/issues/2812

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -71,7 +71,7 @@ function on( elem, types, selector, data, fn, one ) {
 	if ( fn === false ) {
 		fn = returnFalse;
 	} else if ( !fn ) {
-		return this;
+		return elem;
 	}
 
 	if ( one === 1 ) {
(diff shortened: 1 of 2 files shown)


mgol:
Thanks. I'd normally want a unit test for that but since it's officially unsupported it LGTM. If we wanted to start officially support it we'd need to add tests.

markelog:
We can add test, which would check that return value is not `window` object or that it is a jQuery object.

mgol:
Yeah, we could. @dwilson6 could you modify the test at https://github.com/jquery/jquery/blob/df31b88135f94f903c8a887085d30d085a9c62f6/test/unit/event.js#L8-L21 to not test whether this method throws but to check if the returned value is the original element? We would backport the test to the `1.12-stable` branch as well.

P.S. In the future, it's a good idea to create a separate branch for each patch you create. `master`, `2.2-stable` etc. should be used only to pull from the main jQuery repository, you shouldn't modify them. Otherwise if you needed to pull the newest changes from our repository it may be non-trivial to do.

dwilson6:
Yeah, I was wondering if it was worth a test. I can do that. And thanks for the tip.

timmywil:
Merged.

Issue #2781, opened 2015-12-16
Do you guys support signed string hex for jQuery.isNumeric


stevemao:
```js
$.isNumeric(-0xFF)
true
$.isNumeric('-0xFF')
false
```

Is it the right behaviour?

dmethvin:
In the first case, the JS parser converts the hex before execution. In the second, you're expecting the string to be converted at the point of execution. Hex constants aren't supported by `parseInt` or implicit numeric conversion. So I would say this is the right behavior. If you want to convert or validate hex/binary/octal numbers write your own validator.

stevemao:
Thanks @dmethvin , looks like this needs `parseInt`, `parseFloat`, `isFinite` and `isNaN` combined.

dmethvin:
I think this can be fixed in docs. Negative hex constants aren't a really common use case.

stevemao:
Just note: `parseInt` supprts hex but not binary or octal :(

Also I think `( obj - parseFloat( obj ) + 1 ) >= 0;` isn't semantically correct and it's a hack of not using `isFinite` etc...

its harder to understand it now as the useful comments are removed in https://github.com/jquery/jquery/pull/2663.

gibson042:
With respect to the initial question, `$.isNumeric('-0xFF')` should always be `false` because `+"-0xFF"` is `NaN`. So I think this ticket should be closed, and the `isNumeric` code (including comments removed by 15ac848868e993dfe5ccd7751a94f5c8edc288bc) considered separately.

stevemao:
@gibson042 I'm fine if `'-0xFF'` is not supported but ' +"-0xFF" is NaN' should not be reason. Its just one API that does not recognize signed string hex.

gibson042:
The purpose of `jQuery.isNumeric` is (was?) to indicate that a numeric-cast operation on a value yields meaningful (read: finite) results—`parseInt` and `parseFloat` are irrelevant, except for possible implementation utility. Numeric casting of `"-0xFF"` is not meaningful, therefore `jQuery.isNumeric("-0xFF")` should return `false`. Cf. http://code.jquery.com/jquery-1.11.0.js .

Personally, I would be comfortable with an implementation like `isFinite( obj - parseFloat( obj ) )`, suitably guarded for Symbols¹ or—per https://bugs.jquery.com/ticket/14179 and gh-2662 :disappointed:—non-primitives².

----
[...]

Issue #2780, opened 2015-12-16
Add tests for New integer literals in es2015


stevemao:
```js
$.isNumeric('0oFF')
false
$.isNumeric(0o11)
true
```

timmywil:
Duplicate of #2781

PR #2827, opened 2016-01-14
Core: improve isNumeric logic

stevemao:
Add back accidentally deleted comments about the implementation.
Add tests for new number literals.

Ref #2663
Ref #2781
Ref #2780

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -217,7 +217,11 @@ jQuery.extend( {
 		// that can be coerced to finite numbers (gh-2662)
 		var type = jQuery.type( obj );
 		return ( type === "number" || type === "string" ) &&
-			( obj - parseFloat( obj ) + 1 ) >= 0;
+
+			// parseFloat NaNs numeric-cast false positives ("")
+			// ...but misinterprets leading-number strings, particularly hex literals ("0x...")
+			// subtraction forces infinities to NaN
+			!isNaN( obj - parseFloat( obj ) );
 	},
 
 	isPlainObject: function( obj ) {
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -496,7 +496,7 @@ QUnit.test( "isNumeric", function( assert ) {
 	assert.ok( t( 8e5 ), "Exponential notation" );
 	assert.ok( t( "123e-2" ), "Exponential notation string" );
 
-	assert.equal( t( new ToString( "42" ) ), false, "Custom .toString returning number" );
+	assert.equal( t( new ToString( "42" ) ), false, "Only limited to strings and numbers" );
 	assert.equal( t( "" ), false, "Empty string" );
 	assert.equal( t( "        " ), false, "Whitespace characters string" );
 	assert.equal( t( "\t\t" ), false, "Tab characters string" );
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -449,7 +449,7 @@ QUnit.test( "isFunction", function( assert ) {
 } );
 
 QUnit.test( "isNumeric", function( assert ) {
-	assert.expect( 38 );
+	assert.expect( 41 );
 
 	var t = jQuery.isNumeric,
 		ToString = function( value ) {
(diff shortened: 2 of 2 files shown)


test/.jshintrc
```diff
@@ -19,6 +19,9 @@
 	"devel": true,
 	"wsh": true,
 
+	// Support: New integer literals
```
dmethvin: We usually reserve `// Support:`-style comments for supporting environments.

test/.jshintrc
```diff
@@ -19,6 +19,9 @@
...
+	"esnext": true,
```
dmethvin: This would enable esnext for all the source code and unit tests, I don't think we want to do that because the source should be plain ES5. Can we just enable for the one test where we need it?

test/unit/core.js
```diff
@@ -464,8 +464,11 @@ QUnit.test( "isNumeric", function( assert ) {
 	assert.ok( t( -16 ), "Negative integer number" );
 	assert.ok( t( 0 ), "Zero integer number" );
 	assert.ok( t( 32 ), "Positive integer number" );
+	assert.ok( t( "0b111110" ), "Binary integer literal" );
```
dmethvin: How do these behave in environments like IE9 which is still supported but not ESNext?

stevemao: This change should not add to the compact version

timmywil: I think you meant compat, but we still support IE9 on master.

test/.jshintrc
timmywil: Agreed, this should be removed.

src/core.js
```diff
@@ -217,7 +217,11 @@ jQuery.extend( {
 		// that can be coerced to finite numbers (gh-2662)
 		var type = jQuery.type( obj );
 		return ( type === "number" || type === "string" ) &&
-			( obj - parseFloat( obj ) + 1 ) >= 0;
+
+			// parseFloat NaNs numeric-cast false positives ("")
+			// ...but misinterprets leading-number strings, particularly hex literals ("0x...")
+			// subtraction forces infinities to NaN
+			!isNaN( obj - parseFloat( obj ) );
```
timmywil: Wasn't the magic math doing something that isNaN doesn't do?

gibson042: The change is derived from https://github.com/jquery/jquery/issues/2781#issuecomment-165633427 , and works as follows:
* since `obj - parseFloat( obj )` includes both a numeric cast and an explicit parse, any non-numeric input will generate at least one `NaN` operand
* infinite values survive the conversion, but subtracting an infinite value from itself yields `NaN`
* `isNaN` (or `isFinite`) discriminate `NaN`, and therefore also numeric input

timmywil: I see, so it seems this is safe.

test/unit/core.js
stevemao: thanks. typo... What do you think we should do here?

timmywil: Good question. If the test won't run in IE9, we could just test for IE and skip it if so. As for linting, it's okay to use jshint comments for cases like this.

dmethvin: The eslint problem can be fixed with inline comments like `// j`, but the [JSHint docs](http://jshint.com/docs/options/) say that the `exnext` option is deprecated and `esversion: 6` should be used instead. So 
```js
// jshint esversion: 6
   ... code that uses es6 ...
// jshint esversion: 5
```
I don't really  like explicitly setting esversion back to 5 since it makes an assumption, is there a way to simply revert back to the previous setting? Is the scope of this change only the current test/block in JS?

dmethvin: It might be easier to break out the ES6 tests separately and used `QUnit.skip` to skip them. We just need a nice way to detect when we're in an ES6 env or accept the inevitable and look at the userAgent.

stevemao: I'm happy to do that. Will propose a change soon.

stevemao: Can we use feature detection instead of userAgent? Something like `0o1 === 1`?

src/core.js
stevemao: @gibson042 digging through the tickets, you proposed the change to be `!isNaN( obj - parseFloat( obj ) )` a few years back. But you finally went with `( obj - parseFloat( obj ) + 1 ) >= 0;`. Can you remember any reason?

test/unit/core.js
dmethvin: Hey whatever works. I think you'd need something like `+"0o1" === 1` for non-ES6 browsers though?

src/core.js
gibson042: Mostly [code golfing for smaller size](https://github.com/jquery/jquery/pull/1352#discussion-diff-6190735). It took awhile before the [`+ 1` was added](https://github.com/stevemao/jquery/commit/e1192af8726bcd25da58aba64b20d89e919230b9), which tipped the scales in favor of the original.

For what it's worth, which wins here between `!isNaN` and `isFinite`?

stevemao: `!isNaN`? smaller and semantic. (Your original propose was `!isNaN`)
I didn't know `isFinite(NaN)` is`false` I guess it makes it a bit hard to read too.

gibson042: I'd argue that `isFinite` is more semantic, but it really doesn't matter much. `!isNaN` works just as well.

stevemao:
I have made the changes. Please have a look :)

timmywil:
Thanks @stevemao

stevemao:
My pleasure :)

PR #2830, opened 2016-01-14
Update core.js

mattbrundage:
Added apostrophe for possessive case.


timmywil:
Thank you for submitting a PR, but since the CLA wasn't signed and this is such a low priority PR, I'm going to close. We'll throw in this change with another commit.

Issue #2337, opened 2015-05-20
Deprecate `which` property of the event object


markelog:
In the https://api.jquery.com/category/events/event-object/, we have 

> The following properties are also copied to the event object, though some of their values may be undefined depending on the event:
> 
> altKey, bubbles, button, cancelable, charCode, clientX, clientY, ctrlKey, currentTarget, data, detail, eventPhase, metaKey, offsetX, offsetY, originalTarget, pageX, pageY, relatedTarget, screenX, screenY, shiftKey, target, view, which

`which` property is deprecated by the DOM specification - https://w3c.github.io/uievents/#widl-KeyboardEvent-which, it seems we need to add a note about it (and possible for the other properties too), deprecated it and show message in the migrate?

/cc @dmethvin

timmywil:
Hmm, I'm not sure that's necessary. Doesn't hurt to copy it if it's there.

timmywil:
Or should this be moved to tickets in api/migrate?

dmethvin:
Remember there is a `which` for both KeyboardEvent and MouseEvent. Both are pretty screwy.

For a keyboard event you're supposed to use `key` instead of `which` but for a long time the preferred property was `keyCode`. We copy both of them over to the normalized event if they are present, but neither will be there in IE8.

The W3C created [another confusing set](http://www.w3.org/TR/DOM-Level-3-Events/#interface-MouseEvent) of `button` and `buttons` properties for MouseEvent that also took a few years for browsers to get right. IE8 only supports `which` natively. At the time the other two properties were screwed up with other browsers we decided to normalize `which`. The docs say that we don't copy `buttons` but we do, so a lot of people may be using `button` and `buttons` already if they don't care about IE8.

We could deprecate `which` in 3.0 and have Migrate warn about the better choices here. The problem would be in the compat branch where IE8 needs it and we never normalized `button`, `buttons`, `keyCode`, and/or `key`. Also note that there may be a high perf penalty for that, we'll have to see. Events like mousemove or mouseover happen a lot.

mzgol:
The copying already causes performance penalty in Firefox in some cases, see https://github.com/mozilla/treeherder-ui/pull/486 that @fitzgen made me aware of. I didn't have time to measure what's exactly slowing everything down but in the linked PR the jQuery is patched to use getters instead of copying as it significantly speeds up Firefox from what I understand.

dmethvin:
[...]

PR #2834, opened 2016-01-14
Make event dispatch optimizable by engines

topaxi:
The event dispatch method is not optimizable by modern JS engines. This is especially crucial if a lot of events are dispatched (mousemove and touchmove come to mind).

no-param-reassign
Fixed by not reassigning function param

dont-pass-arguments
Fixed by not pass arguments object to other Array#slice()

Now optimizable by v8, spidermonkey and most probably all the other engines:
optimizable

\o/

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -279,19 +279,24 @@ jQuery.event = {
 		}
 	},
 
-	dispatch: function( event ) {
+	dispatch: function( nativeEvent ) {
 
 		// Make a writable jQuery.Event from the native event object
-		event = jQuery.event.fix( event );
+		var event = jQuery.event.fix( nativeEvent );
 
 		var i, j, ret, matched, handleObj,
 			handlerQueue = [],
-			args = slice.call( arguments ),
+			args = [],
 			handlers = ( dataPriv.get( this, "events" ) || {} )[ event.type ] || [],
 			special = jQuery.event.special[ event.type ] || {};
 
 		// Use the fix-ed jQuery.Event rather than the (read-only) native event
 		args[ 0 ] = event;
+
+		for ( i = 1; i < arguments.length; i++ ) {
+			args[ i ] = arguments[ i ];
+		}
+
 		event.delegateTarget = this;
 
 		// Call the preDispatch hook for the mapped type, and let it bail if desired
--- a/src/event.js
+++ b/src/event.js
@@ -284,9 +284,8 @@ jQuery.event = {
 		// Make a writable jQuery.Event from the native event object
 		var event = jQuery.event.fix( nativeEvent );
 
-		var i, j, ret, matched, handleObj,
-			handlerQueue = [],
-			args = [],
+		var i, j, ret, matched, handleObj, handlerQueue,
+			args = new Array( arguments.length ),
 			handlers = ( dataPriv.get( this, "events" ) || {} )[ event.type ] || [],
 			special = jQuery.event.special[ event.type ] || {};
 


dmethvin:
Nice. I know that has been mentioned in https://github.com/petkaantonov/bluebird/wiki/Optimization-killers, do you have some perf data?

topaxi:
No real perf data yet. Can look into it tomorrow :)

topaxi:
So I'm not sure how to build actual meaningful perf benchmarks.
I used jsperf.com with the following setup: http://jsperf.com/jquery-optimized-event-dispatch

While in the devtools the `event.dispatch` method shows up a lot less and faster, the overall test seems to run at the same speed or even slower in the "optimized" version.

Chrome Profile:
jsperf-jquery-event-dispatch chrome

Firefox Profile:
jsperf-jquery-event-dispatch ff

markelog:
I really, **really** hate this kind of hacks.

What articles like https://github.com/petkaantonov/bluebird/wiki/Optimization-killers forget to tell you, is that you need to improve performance only those parts of code that are slow, but not those parts that could be optimized - if you can, doesn't mean you should, because if you do, then that would be a definition of premature opmitization. 

Like what if we do this - increase the size, introduce additional complexity to the code all that. But the next day, new Chrome comes out and with it, new v8 and with it, new Crankshaft/TurboFan, which do the optimization for arguments value but in in parallel new FF comes out and with it, new spiderMonkey and with it new IonMonkey/<or any other monkey>, which doesn't use fast compilier for functions where `arguments` argument are used in loops or something.

So now we increase the size, logic is clouded the hell out of, we didn't improve the speed in Chrome and decrease the speed in spiderMonkey.

Bluebird is the most obfuscated, weird and unsupported code there is on JS, pretty much no one understand it, except for his author... and that for lib that doesn't have to be **that** fast because it deals with async operations anyways, just like jQuery doesn't have to **that** fast because it deals with DOM and DOM API is always slower part then any front-end JS there is, this is probably because you don't see any "improvements" in your devtools.

--

Since there is no argument for this change i.e. -

> run at the same speed or even slower in the "optimized" version

I'm closing it

topaxi:
> this is probably because you don't see any "improvements" in your devtools

Well I _do_ see improvements in the devtools, I just don't see them in my pseudo-benchmark on jsperf.com. I'm not even sure if the jsperf.com benchmarks are reliable for these kind of things.

When I use this change with `mousemove` events for drawing a rectangle in my browser I even see more frames in Firefox.

I'm currently programming a "selectable" like extension for ember here: https://github.com/topaxi/ember-snip

Using https://github.com/mrdoob/stats.js/ I see the following FPS on my notebook while randomly dragging around on the screen:

Firefox:
master: 10-25fps
this PR: 25-40fps

Chrome:
Both: 60fps

I suspect this change might even have a bigger impact on mobile devices.

If these "hacks" are not welcome, feel free to leave this PR closed.

markelog:
> I'm not even sure if the jsperf.com benchmarks are reliable for these kind of things.

It is reliable it is specifically build for these kind of things.

> I see the following FPS on my notebook while randomly dragging around on the screen:

If you could put together something that we could check ourselves? 

I'm steel strongly against these changes, but it is enough evidence for re-opining

topaxi:
After reading through http://blog.methvin.com/2013/12/please-stop-jsperfcom-abuse.html by @dmethvin I don't feel very confident about making a real test-case based on jsperf.com.

@dmethvin Would you have time to create some benchmark with me this weekend or some time next week? You can reach me on twitter ([@topaxi](https://twitter.com/topaxi)) or irc.freenode.org (topaxi).

markelog:
Nice job of figuring out how to do a bench test, your results are compelling, added couple more browsers to your data, they all show either same results (however i suspect this is `benchmark.js` infelicity
) or improved performance. 

However, `jQuery.event.dispatch` is still private API (since it is not document). Prime consumers of which is `jQuery#trigger` and real events.

We can't test real events (if would like to came up with some test-suite that would be great though), we can test `jQuery#trigger` though and those results are quite the opposite of yours - http://jsperf.com/trigger-vs-opt-trigger/3 or more so, since now all browsers show speed decrease from 15-50%

topaxi:
Yes this is the same result I mentioned in my first (well second really) comment. This is why I added a `jQuery.event.dispatch()` only test. While I know that actual events should be measured and our tests probably aren't near real-world enough. This is why I asked @dmethvin for help, to do a proper benchmark.

The `jQuery#trigger()` slowdown is quite interesting, I guess the engines are trying too hard/long to optimize?

topaxi:
@markelog I updated your bench using a mousemove event (which I observed the real-world speedups in my browsers), as triggering a click event (which occurs not very much, usually...) has a special case in https://github.com/jquery/jquery/blob/master/src/event.js#L505.

http://jsperf.com/trigger-vs-opt-trigger/4

Chrome seems slower on my desktop, but faster on mobile.
Firefox again seems much much faster.

dmethvin:
I can't look at this in detail right now but we would definitely want to see the full path with real events and/or `.trigger()`. Just for kicks I did a profiler run on this page (the Github one you're on now) and `jQuery.event.dispatch` does end up high on the list. 
image

markelog:
Added couple more browsers to your example, on my chrome mobile (although "mouse" event on mobile?) it doesn't show any improvements, latest IE (those three dots), Safari (just like FF but in the opposite way) (and even sometimes FF) mobile Safari (expectedly, {although "mouse" event on mobile?}) still show significant speed decrease.

Whenever or not or our `click` hook  in anywayis responsable for difference is a mystery, we also have `mouseEvents` might they responsible for speed improve in FF?

That sort of inconsistency is what i was ranting about, you did chrome optimization and yet you saw FF only improvements. "Tomorrow" it might be the opposite. Just shot in the dark it seems.

Doing fast compiler optimization, 

I agree we need to improve performance of event handling if we can, but i'd argue that fast compiler optimization is not the way.

markelog:
> While I know that actual events should be measured and our tests probably aren't near real-world enough

Actually, triggering events is pretty common use-case, i would think they are reflect the "real world".

topaxi:
> Actually, triggering events is pretty common use-case, i would think they are reflect the "real world".

Yes, but not as frequent as mousemove or touchmove events, they fire very rapidly. If a click event or similar costs even 10ms more and a mouse- or touchmove costs 5ms less, it might actually be worth it.

markelog:
I was questioning the 

> aren't near real-world enough

part, not whenever or not they worth it. That would be a different discussion.

topaxi:
Anyways, what does the core team need to merge or close this request? I'll be happy to work out some approved benchmarks.

On all my observations and profiling, `jQuery.event.dispatch()` shows up way less and way faster in my branch.
There are cases where this slows down, but I have no deeper knowledge of how big this impact is.
It might be more interesting to find out what causes the other cases to slow down, I haven't noticed any other functions slowing down by these changes. Maybe more time gets spent in the compilers?

I think it should be worthwhile to remove code that actually prevents engines from optimization. I wasn't targeting any engine as all of them have troubles with those two changes I made.
While the performance in this PR might not matter, by removing deoptimizations, engines could improve performance in future versions, without overcoming these (probably, I'm no JIT programmer) complex deoptimizations.
If adding one loop is too much complexity, then feel free to close this immediately. Personally I feel a `for` loop is way clearer than calling `Array#slice()` on `arguments`.

On a side-note, other projects like Ember.js even put this in their guidelines (I respect your choice to not add such rules to yours).
[Ember.js STYLEGUIDE](https://github.com/emberjs/ember.js/blob/master/STYLEGUIDE.md#function-arguments)

markelog:
> Anyways, what does the core team need to merge or close this request?

I'd say proof of speed increase, right now, we definitely established that it will cause speed decrease.

If real events would work faster, we would need to decide what is more important -
increase the speed of events dispatch or convolute the logic, increase the size and worsen the `trigger()` case.

Now of "how", is the part where you can get creative :-). There is no universally good way to do it.

One might argue that you can take real site, profile it, with latest version of jQuery with and without yours changes, on all browsers you have access to. Or you can use the `benchmarks.js` with selenium and your layer, basically writing your bench suite that would be activated with real user events (wow!).

> I think it should be worthwhile to remove code that actually prevents engines from optimization.

That is that part i disagree with, i don't care whenever code is optimized or not, i care if that code is fast enough. 

> If adding one loop is too much complexity, then feel free to close this immediately. Personally I feel a for loop is way clearer than calling Array#slice() on arguments.

> On a side-note, other projects like Ember.js even put this in their guidelines (I respect your choice to not add such rules to yours).
Ember.js STYLEGUIDE

This is why i bring up Bluebird, a lot of people praised and impress with it.
`event` module is one of the complicated modules we have and if we would add engine optimization code in there, like this - https://github.com/petkaantonov/bluebird/blob/dec7f0296209a8552c887421ecbec3d4b99565e6/src/util.js#L196 (Basically you need to write the whole article to explain every line of code, which also did happened) it might just became unmaintainable.

> not add such rules to yours

We might just do that, if you will able to prove you point :-)

dmethvin:
The mousemove/pointermove scenarios (and maybe window.scroll) are the most plausible scenarios where the event subsystem might cause a bottleneck. Once you get all the other parts of the execution path involved, though, it may be that deoptimization of jQuery.event.dispatch isn't enough of a difference to be noticed. In that case it would be good to find the real bottlenecks and fix those, either in addition to or in place of this patch. 

One real perf problem that we know of is that with *real events* we currently copy and normalize the event object as soon as the event arrives, through `jQuery.event.fix`. This incurs a constant-time overhead up front, which can be large for mouse/pointer events because referencing some of those properties can force layout. We should really try to get #2337 landed, @jbedard had a great approach for fixing that.

I would *love* to see some kind of real-world repeatable benchmark we could use to judge whether we're really improving things or just optimizing code for no reason.

topaxi:
Just a small update, I took the benchmark of @markelog and reversed the examples (http://jsperf.com/trigger-vs-opt-trigger/5), seems like the second one always is slower. Even putting the same two test-cases causes the second one to slow down by around 20%. I guess some GC behaviour is interfering here.

I'll look into some other way of doing the benchmarks later this evening. Even if this means I have to do one test-case at a time and do a Ops/sec table by hand.

markelog:
http://jsperf.com/trigger-vs-opt-trigger/9
http://jsperf.com/trigger-vs-opt-trigger/10

Now they are the same.
I think we can say you prove (or i proved :-) your theory. Although profile results would be also preferable.

Now there is question of whenever we want this.

markelog:
Gotta say, i hated proving opposite point :/ it is hard to be impartial

markelog:
Hm, runned results couple more times, they started to differ o_O.

@jdalton little help here? 

How it could be that those tests results are different - [1](http://jsperf.com/trigger-vs-opt-trigger/9), [2](http://jsperf.com/trigger-vs-opt-trigger/10)

topaxi:
So I made a single test for each case and ran each test 3 times. This is the result I got on my notebook:

## Direct call to event dispatch
Original: http://jsperf.com/jquery-optimized-event-dispatch/6
Chrome: 39,013 (3 Runs)
Firefox: 47,880 (3 Runs)

Optimized: http://jsperf.com/jquery-optimized-event-dispatch/7
Chrome: 43,820 (3 Runs)
Firefox: 65,056 (3 Runs)

Diff for direct call to dispatch:
Chrome: 12.3% faster
Firefox: 35.8% faster

## Trigger click
Original: http://jsperf.com/jquery-optimized-event-dispatch/8
Chrome: 10,367 (3 Runs)
Firefox: 8,493 (3 Runs)

Optimized: http://jsperf.com/jquery-optimized-event-dispatch/9
Chrome: 14,120 (3 Runs)
Firefox: 12,638 (3 Runs)

Diff for trigger click:
Chrome: 36.2% faster
Firefox: 48.8% faster

## Trigger mousemove
Original: http://jsperf.com/jquery-optimized-event-dispatch/10
Chrome: 9,921 (3 Runs)
Firefox: 7,242 (3 Runs)

Optimized: http://jsperf.com/jquery-optimized-event-dispatch/11
Chrome: 15,732 (3 Runs)
Firefox: 20,139 (3 Runs)

Diff for trigger mousemove:
Chrome: 58.6%
Firefox: 178.1%

timmywil:
@dmethvin What you do think? Should we land?

dmethvin:
So the con to landing it is that it's a Chrome-specific optimization, I don't know if other browsers  disable optimization when `arguments` passed to another function. There is a danger that we end up with cargo-cult patterns that we can't explain later, which is I think was the concern @markelog had. 

On the pro side, we already do browser-specific things like isolate try/catch into a function because of Chrome, even though other browsers like Edge will optimize such functions. Our team has been stable enough (yay us!) that we'll have enough organizational memory to come back and change things if it turns out it is no longer needed. Plus it's nice to see the dispatch speed double for cases like mousemove and pointermove which can be bottlenecks.

So what the heck, let's land it. Nice work @topaxi!

Since @topaxi already has a framework to test perf for these cases, I'd love to see [#2337 comment](https://github.com/jquery/jquery/issues/2337#issuecomment-104029050) happen in 3.0 so that `jQuery.event.fix` could be faster and we could support more properties directly on our `event` without up-front copy overyead. @jbedard do you see any problems with trying to freshen up your old code? Maybe @topaxi could give you some help there.

jbedard:
Here's an [updated version](657bf57cd1d4c4a33ff70ed17c8017cb00529fbe) of my change which @dmethvin mentioned if anyone wants to try it out.  There were definitely still a few questions about that though. Especially if the perf hit of getter/setters is worth avoiding the fix-loop... I seem to recall the [set on first get](https://github.com/jbedard/jquery/commit/657bf57cd1d4c4a33ff70ed17c8017cb00529fbe#diff-031bb62d959e7e4949d1847c82507f33R110) was also questionable.

There also appears to be an [argument assignment](https://github.com/jquery/jquery/blob/17f0e26ad9eba67ab274d12274cf7c23c8c688fd/src/event.js#L464) in `jQuery.event.fix` that might be worth changing in this PR...

dmethvin:
@jbedard your link isn't working for me. Can you open a new PR? we can discuss/experiment there. Nice catch on the arg assignment!

jbedard:
Opps, I've fixed the link in the comment. Also opened #2860.

markelog:
I would ask for some clarification comment before the loop. 

Also, i'm not sure if actually see speed increase, right now, proof seems inconclusive, in one test we see speed decrease in the other speed increase but for Firefox only. Without any explanation.

Would still love some explanation from the @jdalton and his overall opinion about it. 
@dmethvin if you certain this is right move, would you mind profile, i would feel much more certain about it.

jbedard:
I'd be curious if only removing the argument assignment would make a difference, but keeping the `slice.call( arguments )`. Those are separate "not optimized" warnings and might deserve separate fixes/discussions...

topaxi:
@jbedard param assignments are only bad if you do this while also using an `arguments` object in your function (at least that's how I'm understanding it and observed so far, but I'm definitely not 100% sure here).

The issue is that the `arguments` and param interaction in JavaScript is quite weird and imho very dangerous.

A small example:
```js
function foo(bar, baz) {
  console.log(arguments[0] === 'bar')
  console.log(arguments[1] === 'baz')
  // so far so good.. but if we assign something to arguments
  arguments[0] = 'foo'
  console.log(arguments[0] === 'foo') // still good
  console.log(bar === 'foo') // Oops, by assigning arguments[0] we also reassigned bar
  // this also works the other way around
  baz = 'foo'
  console.log(baz === 'foo') // Okay, this is what we wanted
  console.log(arguments[1] === 'foo') // Oops, we somehow changed something in arguments
}

foo('bar', 'baz') // this logs 6 times `true`
```

And this is why _all_ engines (to date) deoptimize if we pass arguments around or assign to params when an arguments object is being used.

So to avoid these mistakes, some guidelines agreed to avoid param reassigned at all times, while it (at least to my knowledge) is only bad while using `arguments` in the same function.

To summarize params and arguments optimization issues:

```js
// OK, param reassignment but no arguments object
function(foo) {
  foo = somethingelse()
}
// Bad, param reassignment with arguments object
function(foo) {
  foo = somethingelse()
  console.log(arguments[0]
}
// Bad, passing arguments to an other function
function(foo) {
  var args = Array.prototype.slice.call(arguments)
  console.log(args[0])
}
// Good, converting arguments to array
function(foo) {
  var args = new Array(arguments.length)
  for (var i = 0; i < arguments.length; i++) {
    args[i] = arguments[i]
  }
  var args = Array.prototype.slice.call(arguments)
  console.log(args[0])
}
```

jbedard:
Still sounds like a good thing to do, and it can be done without the `slice` change...

dmethvin:
I'd like to land this, with a small addition of a comment that says we're doing the explicit loop to allow JS engines to optimize the function. Any objections?

mgol:
@dmethvin SGTM.

Issue #2732, opened 2015-11-19
.prop("selected", true) does not work as expected in IE11


kennyk-peplink:
By selecting option with .prop("select", true), the behavior is not as expected under IE11

Testing with Firefox / Chrome / Safari, the behavior should be consistent,
the following example should have "Two" being selected:

http://jsfiddle.net/zksm1o5q/2

In IE, however, the result is "Four".

You may comment out Line 11, 12, and/or 14 for more test cases with expected result (last prop being selected) in other browsers, but IE11.

dmethvin:
To remove a red herring here, the special characters in the values don't change the outcome. Here is a simpler case, with the last selection changing the text to show that it was able to select the right element: http://jsfiddle.net/zksm1o5q/5/

Here's a bare DOM example without jQuery that exhibits the same problem: http://jsfiddle.net/15ytbys8/

timmywil:
This is a browser bug, but perhaps we should work around it in a propHook that sets all other option elements' selected values to `false`?

dmethvin:
Does that fix the problem? Is it only a bug with options that exist in the select already, or would we also need to guard against ones appended from elsewhere?

This seems to work on Edge so it's only IE11 (and older IEs) affected. 

In any event it seems like that might really be a performance killer, especially on selects with a lot of options. I guess it could be limited by a feature detect to only IE and only when changing the `selected` property, but silently trying to cure it might lead to a real performance trap. I'm also wondering if properties like `readonly` or `disabled` misbehave the same way?

timmywil:
What about setting `selectedIndex` on the select element instead of the all the `selected` properties?

timmywil:
That wouldn't be a big perf hit – or code size hit – and it seems to work: http://jsfiddle.net/timmywil/15ytbys8/1/.

timmywil:
I don't think `disabled` or `readonly` would have the same problem because multiple options can be disabled and readonly.

dmethvin:
Yeah if that works it sounds pretty good. So if we're setting selected on an `option` and it's already in a `select` that is not `multiple`, set the `selectedIndex` property. #youmaynotneedjquery

mgol:
This is a non-breaking change; should we backport it to 2.2.2?

PR #2840, opened 2016-01-15
Attributes: fix setting selected on an option in IE<=11

timmywil:
- Same as the getter fix. All that was required was accessing the selectedIndex property.
- Also added support for optgroups

Fixes gh-2732

### DIFF
--- a/src/attributes/prop.js
+++ b/src/attributes/prop.js
@@ -79,7 +79,7 @@ jQuery.extend( {
 	}
 } );
 
-// Support: IE<=11
+// Support: IE <=11 only
 // Accessing the selectedIndex property
 // forces the browser to respect setting selected
 // on the option
--- a/src/attributes/prop.js
+++ b/src/attributes/prop.js
@@ -83,6 +83,8 @@ jQuery.extend( {
 // Accessing the selectedIndex property
 // forces the browser to respect setting selected
 // on the option
+// The getter ensures a default option is selected
+// when in an optgroup
 if ( !support.optSelected ) {
 	jQuery.propHooks.selected = {
 		get: function( elem ) {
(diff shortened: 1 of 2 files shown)


src/attributes/prop.js
```diff
@@ -79,6 +79,12 @@ jQuery.extend( {
 	}
 } );
 
+// Support: IE<=11
```
mgol: What about Edge? I'd love all new IE-specific stuff to be tested on Edge and if the bug is not present there to write `IE <= 11 only`.

Adding spaces around `<=` would be good as well, that's where the conventions [seem to go currently](https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-69379197).

timmywil: Edge is ok. Updated comment, but the syntax says there's no space after the `<=`, so it's `IE <=11 only`.

gibson042:
Can the `get` and `set` bodies be made even more similar? gzip will love you...

dmethvin:
Going even further, can get and set use the same method?

timmywil:
> Going even further, can get and set use the same method?

I suppose they can, but notice that the getter doesn't access `selectedIndex` on the select unless there's an `optgroup`. That said, if we're okay with it, I'll just make the getter and setter the same function. I don't think it'll hurt anything to go back to accessing `selectedIndex` twice in the getter.

dmethvin:
Yeah, I doubt an extra access will make that much difference. Since it's already guarded by a support test, only IE will pay that penalty anyway.

timmywil:
Actually, they can't be the same function. The getter needs to return `null` and the setter needs to return `undefined`. I tried a factory to create similar functions, but that was +15 bytes. Even using the same function (even though it wouldn't work right) was +4 bytes. Just for kicks, I tried making the getter identical to the setter – with the extra `return` statement, and was exactly the same. I think what was there is still the best option.

Issue #2370, opened 2015-06-01
Traversing: use jQuery's push to avoid an unnecessary loop in .find


timmywil:
See https://github.com/jquery/jquery/issues/2365#issuecomment-107726209

gibson042:
@timmywil I think we need a jsPerf verifying performance impact before committing to this. I could easily see the improvement from doing less work in `pushStack` being overwhelmed by using Array methods on non-Arrays, and there's no reason to change back if it actually slows us down.

timmywil:
Fair enough. I'll do a perf.

mzgol:
@timmywil This seems like sth that can be done for 3.0.1. Please revert the milestone change if you disagree.

PR #2848, opened 2016-01-19
Traversing: restore jQuery push behavior in .find

timmywil:
It's a minor improvement, but seems to still be an [improvement](http://jsperf.com/find-with-pushstack).

Fixes gh-2370


markelog:
LGTM

gibson042:
There doesn't seem to be a significant change on the sorting case, either: http://jsperf.com/find-with-pushstack/2

:+1:

Issue #2488, opened 2015-07-21
Incorrect the 'slideDown' method behavior with element with height 0


bset:
Here is an example of this issue. Steps to reproduce it:
1. Open example using link: http://jsfiddle.net/j53j1mLq/
2. Click on the 'Toggle' button (maybe several times)
3. As you can see, element with height 0 become visible for a short time and it's hidden after that
I'm waiting for no reaction method call because of element height.

Description: on every button click the 'slideDown' and the 'slideUp' methods are called for the 'box' element in rotation.

gibson042:
This stems from a [rather curious height/width exception](https://github.com/jquery/jquery/blob/b3b2d6c3dd51fbdc69e1942e9af75cc99a1834c2/src/effects.js#L231), ultimately dating all the way back to [jQuery 1.0](https://github.com/jquery/jquery/commit/29a059142123a8bde16de2e579e7708a86a8de78#diff-3f7c2d96161eb7e4cf32df26fd18a923R246). I doubt it's still relevant, and would be comfortable removing it. Any objections, @jquery/core?

gnarf:
Seems safe to remove to me - So long as IE whatever doesn't throw problems if we try to set width/height to 0 (but I think we were guarding that somewhere else anyway recently) -  Also, lol at that comment on that line back in 1.0 :)

mzgol:
If it works fine in IE8 without this hack, :+1: for removing.

gibson042:
Great, so who can identify what the problematic behavior actually was? "IE flicker" isn't quite precise enough for me.

gnarf:
only @jeresig probably

jeresig:
Lord knows - probably for IE 5 or something. Feel free to nuke if you can't reproduce!

PR #2849, opened 2016-01-19
Effects: remove width/height exception for oldIE

timmywil:
- I think as long as existing tests pass, we're good.

Fixes gh-2488


gnarf:
I'm okay with this, but it scares the hell out of me.

Such a subtle difference such that all animations will now start at 0 instead of 1 for width and height, how long before someone "depends" on the `1` behavior... `¯\_(ツ)_/¯`

Issue #1778, opened 2014-10-21
Remove core/ready hard dependency on $.Deferred


mgol:
Originally reported by m_gol at: http://bugs.jquery.com/ticket/15174
<p>
If <tt>$.Deferred</tt> is unavailable (e.g. because it was removed via custom compilation), the core/ready implementation should fall back to the standard Promise.
</p><p>Issue reported for jQuery 2.1.1</p>

timmywil:
Hmm, I don't think this is necessary. Deferred is simply a hard dependency of core/ready. If Deferred is removed from a custom build, so core/ready should be too (which we could easily add to the Gruntfile). Those making custom builds are usually also accustomed to placing scripts at the end of the body, making core/ready wasted space. Maybe this is a docs issue instead.

scottgonzalez:
`.ready()` is a pretty important API. I don't think it should go away just because someone doesn't want `$.Deferred`. Also, I thought there were still some cases where just putting scripts at the end of the body isn't enough. I think @gibson042 might remember when that's the case?

timmywil:
I used to think that too. We had this conversation a while ago and it turned out it is sufficient.

Also, Deferred is an important API too. It is used internally in other areas besides core/ready. core/ready is probably the least important of them all.

timmywil:
I did think of a limitation. On a very large document, you'd probably prefer to start downloading the script immediately, but still defer so the browser can download and parse the DOM in parallel. In this case, doc ready can still play a role just in case the script loads before the document is ready, but I think script loaders often have that covered.

timmywil:
Ah, [here](http://bugs.jquery.com/ticket/14084#comment:12) is the conversation I was thinking of. That only overlaps because I don't remember finding evidence that putting scripts at the end if the body wouldn't work.

timmywil:
You know what, I think I've already gone over to the other side. Sorry for the brain dump of notifications. This could be useful to users. My only question is, how do we handle the lack of native Promise?

mzgol:
Up until *very* recently we used to divide support tests into two parts:
one that didn't require attaching anything to document that was invoked at
start and one that did that was invoked on doc ready. Therefore, if you use
any jQuery earlier than 1.11/2.1 (a.k.a. latest),you have to run your code
on doc ready or you risk stumbling upon bugs as the one linked to by
timmywil in this thread.

Now we have this second part of support tests lazy; in 2.x they have
elements attached to document.documentElement so this can happen anytime;
[...]

Issue #1823, opened 2014-10-30
When an error thrown from ready handler all subsequent handlers are not being invoked


crmouli:
Jquery version: 2.1.1
Browser: on any browser:
Testcase: 
Please take a look at this jsfiddle
Example with jquery2.1.1
http://jsfiddle.net/q7gbzrbu/2/

Example with jquery1.5(where it works fine)
http://jsfiddle.net/q7gbzrbu/1/

[Edit: OLD and improper test case] http://jsfiddle.net/8b3a92yo/1/

Problem:
When an error is thrown from one of the handlers registered through $(docuement).ready() then all subsequent handler are not being invoked.

Addl info:
This is what happening when an error is thrown from a handler function which is registered with $.ready().  jQuery.Callbacks.fire()(refer line 3073 in jquery-2.1.1.js) will invoke the user's ready Handler function.  Below is related piece of code. When an error was thrown, "firing" flag will not be reset to false. Due to this any handlers registered after error will not be triggered.(refer line 3113 in jquery-2.1.1.js)

firing = true;
            for ( ; list && firingIndex < firingLength; firingIndex++ ) {
                if ( list[ firingIndex ].apply( data[ 0 ], data[ 1 ] ) === false && options.stopOnFalse ) {  //this statement invokes our ready handler
                    memory = false; // To prevent further calls using add
                    break;
                }
            }
firing = false;

It looks like "firing" flag need to be reset in a finally block.

Workaround:
Obvious workaround is handler code need to be wrapped with try/catch to avoid this issue.

timmywil:
For posterity, the behavior change here is that ready handlers are now executed async. In other words, if any code after adding a ready handler expected the handler to be executed synchronously
(i.e. `jQuery( fn ); expectsSync();`), that code will break. Hopefully, that isn't very common.

PR #2850, opened 2016-01-19
Core: implement ready without Deferred

timmywil:
- Keeps it Promise-compatible
- +39 bytes
- Keeps `jQuery.ready` to avoid breakage on wikipedia.

Fixes gh-1778

### DIFF
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -52,16 +52,21 @@ jQuery.extend( {
 
 			if ( !readyFiring ) {
 				readyFiring = true;
-				while ( readyCallbacks.length ) {
-					fn = readyCallbacks.shift();
-					if ( jQuery.isFunction( fn ) ) {
 
-						// Prefer sync with no try/catch here
-						// Backwards-compatible, maintain execution order
-						fn.call( document, jQuery );
+				// Prevent errors from freezing future callback execution (gh-1823)
+				try {
+					while ( readyCallbacks.length ) {
+						fn = readyCallbacks.shift();
+						if ( jQuery.isFunction( fn ) ) {
+
+							// Prefer sync with no try/catch here
+							// Backwards-compatible, maintain execution order
+							fn.call( document, jQuery );
+						}
 					}
+				} finally {
+					readyFiring = false;
 				}
-				readyFiring = false;
 			}
 		};
 
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -1,17 +1,16 @@
 define( [
 	"../core",
-	"../var/document",
-	"../deferred"
+	"../var/document"
 ], function( jQuery, document ) {
 
-// The deferred used on DOM ready
-var readyList;
+var readyCallbacks = [],
+	readyFiring = -1,
+	whenReady = function( fn ) {
+		readyCallbacks.push( fn );
+	};
 
 jQuery.fn.ready = function( fn ) {
-
-	// Add the callback
-	jQuery.ready.promise().done( fn );
-
+	whenReady( fn );
 	return this;
 };
 
@@ -33,7 +32,6 @@ jQuery.extend( {
 		}
 	},
 
-	// Handle when the DOM is ready
 	ready: function( wait ) {
 
 		// Abort if there are pending holds or we're already ready
@@ -49,10 +47,27 @@ jQuery.extend( {
 			return;
 		}
 
-		// If there are functions bound, to execute
-		readyList.resolveWith( document, [ jQuery ] );
+		whenReady = function( fn ) {
+			readyCallbacks.push( fn );
+
+			if ( !++readyFiring ) {
+				while ( readyCallbacks.length ) {
+					fn = readyCallbacks.shift();
+					if ( jQuery.isFunction( fn ) ) {
+						fn.call( document, jQuery );
+					}
+				}
+			}
+
+			readyFiring--;
+		};
+
+		whenReady();
 	}
 } );
+
+// Make jQuery.ready promise compatible (gh-1778)
+jQuery.ready.then = jQuery.fn.ready;
 
 /**
  * The ready event handler and self cleanup method
@@ -63,34 +78,23 @@ function completed() {
 	jQuery.ready();
 }
 
-jQuery.ready.promise = function( obj ) {
-	if ( !readyList ) {
+// Catch cases where $(document).ready() is called
+// after the browser event has already occurred.
+// Support: IE9-10 only
+// Older IE sometimes signals "interactive" too soon
+if ( document.readyState === "complete" ||
+	( document.readyState !== "loading" && !document.documentElement.doScroll ) ) {
 
-		readyList = jQuery.Deferred();
+	// Handle it asynchronously to allow scripts the opportunity to delay ready
+	window.setTimeout( jQuery.ready );
 
-		// Catch cases where $(document).ready() is called
-		// after the browser event has already occurred.
-		// Support: IE9-10 only
-		// Older IE sometimes signals "interactive" too soon
-		if ( document.readyState === "complete" ||
-			( document.readyState !== "loading" && !document.documentElement.doScroll ) ) {
+} else {
 
-			// Handle it asynchronously to allow scripts the opportunity to delay ready
-			window.setTimeout( jQuery.ready );
+	// Use the handy event callback
+	document.addEventListener( "DOMContentLoaded", completed );
 
-		} else {
-
-			// Use the handy event callback
-			document.addEventListener( "DOMContentLoaded", completed );
-
-			// A fallback to window.onload, that will always work
-			window.addEventListener( "load", completed );
-		}
-	}
-	return readyList.promise( obj );
-};
-
-// Kick off the DOM ready check even if the user does not
-jQuery.ready.promise();
+	// A fallback to window.onload, that will always work
+	window.addEventListener( "load", completed );
+}
 
 } );
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -70,7 +70,9 @@ jQuery.extend( {
 					// If there was an error in a ready callback,
 					// continue with the rest (gh-1823)
 					if ( readyCallbacks.length ) {
-						whenReady();
+
+						// Retry async to allow the error to propagate to console
+						window.setTimeout( whenReady );
 					}
 				}
 			}
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -55,6 +55,9 @@ jQuery.extend( {
 				while ( readyCallbacks.length ) {
 					fn = readyCallbacks.shift();
 					if ( jQuery.isFunction( fn ) ) {
+
+						// Prefer sync with no try/catch here
+						// Backwards-compatible, maintain execution order
 						fn.call( document, jQuery );
 					}
 				}
@@ -66,7 +69,7 @@ jQuery.extend( {
 	}
 } );
 
-// Make jQuery.ready promise compatible (gh-1778)
+// Make jQuery.ready Promise consumable (gh-1778)
 jQuery.ready.then = jQuery.fn.ready;
 
 /**
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -59,8 +59,8 @@ jQuery.extend( {
 						fn = readyCallbacks.shift();
 						if ( jQuery.isFunction( fn ) ) {
 
-							// Prefer sync with no try/catch here
-							// Backwards-compatible, maintain execution order
+							// For backwards compatibility,
+							// invoke synchronously and with document context
 							fn.call( document, jQuery );
 						}
 					}
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -66,6 +66,12 @@ jQuery.extend( {
 					}
 				} finally {
 					readyFiring = false;
+
+					// If there was an error in a ready callback,
+					// continue with the rest (gh-1823)
+					if ( readyCallbacks.length ) {
+						whenReady();
+					}
 				}
 			}
 		};
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -4,7 +4,7 @@ define( [
 ], function( jQuery, document ) {
 
 var readyCallbacks = [],
-	readyFiring = -1,
+	readyFiring = false,
 	whenReady = function( fn ) {
 		readyCallbacks.push( fn );
 	};
@@ -50,16 +50,16 @@ jQuery.extend( {
 		whenReady = function( fn ) {
 			readyCallbacks.push( fn );
 
-			if ( !++readyFiring ) {
+			if ( !readyFiring ) {
+				readyFiring = true;
 				while ( readyCallbacks.length ) {
 					fn = readyCallbacks.shift();
 					if ( jQuery.isFunction( fn ) ) {
 						fn.call( document, jQuery );
 					}
 				}
+				readyFiring = false;
 			}
-
-			readyFiring--;
 		};
 
 		whenReady();
--- a/test/unit/ready.js
+++ b/test/unit/ready.js
@@ -29,10 +29,12 @@ QUnit.module( "ready" );
 	// Bind to the ready event in every possible way.
 	jQuery( makeHandler( "a" ) );
 	jQuery( document ).ready( makeHandler( "b" ) );
+	jQuery.when( jQuery.ready ).done( makeHandler( "c" ) );
 
 	// Do it twice, just to be sure.
-	jQuery( makeHandler( "c" ) );
-	jQuery( document ).ready( makeHandler( "d" ) );
+	jQuery( makeHandler( "d" ) );
+	jQuery( document ).ready( makeHandler( "e" ) );
+	jQuery.when( jQuery.ready ).done( makeHandler( "f" ) );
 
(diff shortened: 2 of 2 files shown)


src/core/ready.js
```diff
@@ -49,11 +47,28 @@ jQuery.extend( {
...
+
+			readyFiring--;
+		};
+
+		whenReady();
 	}
 } );
 
+// Make jQuery.ready promise compatible (gh-1778)
+jQuery.ready.then = jQuery.fn.ready;
```
dcherman: I think you may want to add some unit tests using a Promise lib, or just plain old native Promises.  Since the implementation of `jQuery.fn.ready` returns `this`, and you're assigning that function to your thenable object, you're essentially getting into this scenario:

```js
var myCrazyPromise = { then: function() { return this; };
Promise.resolve(myCrazyPromise).then(console.log.bind(console));
```

If you run that, you'll see that the `then` handler is never invoked since this is infinitely recursive

timmywil: Test added, but no changes required. Notice that while `jQuery.ready` is a thenable, `jQuery.fn.ready` is not.

timmywil: Hmm, thinking about this wrong. Maybe it's only a problem pre-ready event.

timmywil: I was thinking about it wrong (`context` is still `jQuery.ready`), but I don't see the problem you're describing whether it's before the ready event or not. Do you see anything wrong with my test?

src/core/ready.js
```diff
@@ -49,11 +47,28 @@ jQuery.extend( {
 			return;
 		}
 
-		// If there are functions bound, to execute
-		readyList.resolveWith( document, [ jQuery ] );
+		whenReady = function( fn ) {
+			readyCallbacks.push( fn );
+
+			if ( !++readyFiring ) {
```
mgol: This is not the easiest to read. Would breaking it up increase the size? What size penalty would it bring to just use booleans?

I'd prefer to such "clever coding" to be performed by Uglify, not in source.

test/unit/ready.js
```diff
@@ -67,4 +70,20 @@ QUnit.module( "ready" );
 			"Argument passed to fn in jQuery(document).ready( fn ) should be jQuery" );
 	} );
 
+	QUnit.test( "Promise.resolve(jQuery.ready)", function( assert ) {
+		assert.expect( 2 );
+
+		var beforeReady = assert.async();
+		var afterReady = assert.async();
```
mgol: Betterr:
```js
var done = assert.async( 2 );
```
and then just call `done()` in both callbacks.

That is, if we have new enough QUnit.

src/core/ready.js
timmywil: Oh, blurp, I think is a misunderstanding of `Promise.resolve(thenable)`. `Promise.resolve(thenable)` follows the thenable and eventually defers to its state. This means that the thenable is responsible for calling `.then`. Notice that your example doesn't resolve even when you don't `return this`.

test/unit/ready.js
timmywil: Unfortunately, that doesn't work with 1.17.1.

src/core/ready.js
dcherman: Yep, I agree with that.  Apparently I'm not as familiar with thenables as I thought.

FWIW part of my confusion may be attributed to something that seems weird in Chrome, but I have no idea if this is a legitimate thing.

If you run `Promise.resolve(jQuery.ready)`  then inspect the internal `[[PromiseValue]]` the returned promise, you can see that the value the promise was resolved with is not what was passed to the `.then` handler; the resolved value is `jQuery.ready` as I had initially expected.  Might be an implementation and/or devTools bug, but it doesn't look like it has any observable side effects other than confusion.

test/unit/ready.js
mgol: Hmm. @gibson042, to what would we have to bump?

It's probably better to fix all those cases at the same time. But, @timmywil, to be consistent with the rest of the tests and to make it easier to find all such code in one find+replace you could use the pattern from:
https://github.com/jquery/jquery/blob/250fd800a171cac9ab713cbbe1a70a90a348e3dc/test/unit/deferred.js#L463
and:
https://github.com/jquery/jquery/blob/250fd800a171cac9ab713cbbe1a70a90a348e3dc/test/unit/deferred.js#L470
i.e.:
```js
var done = jQuery.map( new Array( 2 ), function() { return assert.async(); } );
done.pop().call();
```

timmywil: Sure I can do that.

gibson042: > @gibson042, to what would we have to bump?

1.20.0

src/core/ready.js
```diff
@@ -49,11 +47,28 @@ jQuery.extend( {
...
-		readyList.resolveWith( document, [ jQuery ] );
+		whenReady = function( fn ) {
+			readyCallbacks.push( fn );
+
+			if ( !readyFiring ) {
+				readyFiring = true;
+				while ( readyCallbacks.length ) {
+					fn = readyCallbacks.shift();
+					if ( jQuery.isFunction( fn ) ) {
+						fn.call( document, jQuery );
```
gibson042: Because the invocations are synchronous and not wrapped in a `try` block, `jQuery.ready` is not a conforming [Promises/A+](https://promisesaplus.com/) promise. This should either include the explanatory comment from https://github.com/jquery/jquery/issues/1778#issuecomment-72764833 or be corrected (not that _that_ would be sufficient, but it would prevent callback lockups).

gibson042:
Other than the missing comment or missing safeguards (depending on your preference), this looks good.

Also, what of the deprecations from https://github.com/jquery/jquery/issues/1778#issuecomment-72764833 ? Have we done those?

> * Deprecate `jQuery( document ).ready( fn )` etc. in favor of `jQuery( fn )`
* Deprecate `jQuery.holdReady` (a nasty global behavior modifier better covered by case-specific deferred/promise patterns)

src/core/ready.js
```diff
@@ -49,11 +47,28 @@ jQuery.extend( {
...
+				}
+				readyFiring = false;
+			}
+		};
+
+		whenReady();
 	}
 } );
 
+// Make jQuery.ready promise compatible (gh-1778)
```
gibson042: I'd update this to "promise-consumable"; it's more clear and might head off too much effort being spent on bugs like `jQuery.ready.then( fn ).then(…)` not working as expected.

timmywil:
I'm not in favor of deprecating `jQuery( document ).ready( fn )`. It existed before `jQuery( fn )` and imo is more semantic. I am in favor of deprecating `jQuery.holdReady` tho.

test/unit/ready.js
markelog: Just tried 1.20.0 on evergreens, it seems we can easily do that

markelog:
Given how code is started to look like without the `Deferred` dep and because custom build is rariry in our user based (that would be an assumption) and since no one actually asked for it - i'm against that change.

> I'm not in favor of deprecating jQuery( document ).ready( fn )

Me too, but i would wait long time to remove it, next majors.

> I am in favor of deprecating jQuery.holdReady tho.

Why? It seems it still has valid use-cases?

timmywil:
> Me too, but i would wait long time to remove it, next majors.

I think you misread. I don't want to deprecate that.

> Why? It seems it still has valid use-cases?

@gibson042 pointed out those cases can be covered by promises.

markelog:
> I think you misread. I don't want to deprecate that.

Oh, i did, sorry.

> @gibson042 pointed out those cases can be covered by promises.

Like?

gibson042:
> I'm not in favor of deprecating `jQuery( document ).ready( fn )`. It existed before `jQuery( fn )` and imo is more semantic.

Really? Because I contend that all semantic meaning is lost by ignoring the argument to `jQuery`—the following are all equivalent:
* `jQuery( document ).ready( fn )`
* `jQuery().ready( fn )`
* `jQuery( "<div/>" ).ready( fn )`
* `jQuery( [] ).ready( fn )`
* `jQuery( jQuery.noop ).ready( fn )`
* `jQuery.fn.ready( fn )`

gibson042:
> @gibson042 pointed out those [`jQuery.holdReady`] cases can be covered by promises.

`jQuery.when( jQuery.ready, jQuery.Deferred( waitForMyDependencies ) )`

timmywil:
> Because I contend that all semantic meaning is lost by ignoring the argument to jQuery.

Agreed, but I don't see that often. I was referring to how `jQuery( document ).ready()` reads, and I like that.

dmethvin:
Yeah, `jQuery( document ).ready() ` certainly seems to be the most poetic to me. I concede that it's mostly in the eye of the beholder though. :smile:

markelog:
> jQuery.when( jQuery.ready, jQuery.Deferred( waitForMyDependencies ) )

And what if you want other plugins that depend on `ready` event to hold?

gibson042:
Then those plugins would appear in `waitForMyDependencies`, and theirselves pay attention to `jQuery.ready`.

dmethvin:
Although there are a lot of dups here, seems like `.holdReady` is pretty commonly used. One of those is stealjs, i think they said they've moved away from it but still a lot out there.

https://github.com/search?q=jquery.holdReady+true&type=Code&utf8=✓

gibson042:
That's fine. This is a deprecation, not a removal.

markelog:
> Then those plugins would appear in waitForMyDependencies, and theirselves pay attention to jQuery.ready.

User usually download their deps with builded file, which means we affect how app should be loaded and initiated, it seems we became opinionated here, besides that would unite

markelog:
Whereas `waitForMyDependencies` approach is really not common these days

markelog:
> Whereas waitForMyDependencies approach is really not common these days

Buy that i mean, as i understand, you want user to push plugins load callback and data on which those plugins operate in `waitForMyDependencies` object

dmethvin:
Looking at some of the usage it seems like many people have taken the action-at-a-distance behavior of `.holdReady` as a feature rather than a programming atrocity. I'd actually be for deprecating this just to let people know they should find a better way, and show that better way in the docs. Who knows when it could actually be removed.

markelog:
> Looking at some of the usage it seems like many people have taken the action-at-a-distance behavior of .holdReady as a feature rather than a programming atrocity.

Well, probably because we didn't use word "atrocity" but "advanced feature" in the [docs](http://api.jquery.com/jquery.holdready/)

timmywil:
I think we agree that `holdReady` is best avoided, so let's go with the deprecation, but no set time for removal.

gibson042:
> you want user to push plugins load resolve  and data on which those plugins operate under `waitForMyDependencies` promise

I'm not sure what you mean... any dependencies either expose their own thenable (valid as a `jQuery.when` argument), or the conditions can be manually checked in a `waitForMyDependencies` _function_ as shown above (responsible for resolving a Deferred or Promise once they are satisfied).

timmywil:
Added a fix for gh-1823.

src/core/ready.js
```diff
@@ -49,11 +47,36 @@ jQuery.extend( {
...
+						fn = readyCallbacks.shift();
+						if ( jQuery.isFunction( fn ) ) {
+
+							// Prefer sync with no try/catch here
+							// Backwards-compatible, maintain execution order
+							fn.call( document, jQuery );
+						}
+					}
+				} finally {
+					readyFiring = false;
```
gibson042: As coded, this _still_ halts execution until a subsequent handler is added. But you could fix it with a `setTimeout( whenReady )` here.

src/core/ready.js
```diff
@@ -49,11 +47,36 @@ jQuery.extend( {
...
+				readyFiring = true;
+
+				// Prevent errors from freezing future callback execution (gh-1823)
+				try {
+					while ( readyCallbacks.length ) {
+						fn = readyCallbacks.shift();
+						if ( jQuery.isFunction( fn ) ) {
+
+							// Prefer sync with no try/catch here
+							// Backwards-compatible, maintain execution order
```
gibson042: Since there's now a `try`, I'd change these comments to
```js
// For backwards compatibility, invoke synchronously and with document context
```

src/core/ready.js
timmywil: As you say, this only fixes the case where ready could never be called again. I'm not sure we need to go beyond that.

src/core/ready.js
timmywil: Agreed

src/core/ready.js
gibson042: It would certainly be more Promise-compatible to have

```js
jQuery.ready.then( throws );
jQuery.ready.then( shouldRun );
```
work without waiting for a subsequent `jQuery.ready.then` invocation. And fixes don't get much simpler than `setTimeout( whenReady )`.

timmywil: Seems like we'd have to add a `catch`. Wouldn't putting the `whenReady` call in the finally block cause an infinite loop?

gibson042: We don't need a `catch`, and the `whenReady` doesn't go infinite because the throwing callback has already been `.shift()`ed off the list.

timmywil: I don't see how it wouldn't be a loop. A finally is called regardless of whether an error is thrown, so it would just keep calling itself.

dmethvin: So by not having a catch, it also barfs immediately on the console right? That's a good thing.

src/core/ready.js
```diff
@@ -1,17 +1,16 @@
...
-	"../deferred"
+	"../var/document"
 ], function( jQuery, document ) {
 
-// The deferred used on DOM ready
-var readyList;
+var readyCallbacks = [],
+	readyFiring = false,
+	whenReady = function( fn ) {
+		readyCallbacks.push( fn );
```
dmethvin: Would it help for the check for `fn` being a function be done here?

src/core/ready.js
gibson042: It keeps calling itself, but will terminate when the callbacks list is empty: https://jsfiddle.net/ewjschzw/

> So by not having a catch, it also barfs immediately on the console right? That's a good thing.

Yes.

src/core/ready.js
gibson042: As long as you duplicate it in the redefinition of `whenReady`. Doing it this way is just conceptually simpler, and it's not like the function test is expensive.

src/core/ready.js
gibson042: Oops; I see what you mean @timmywil... you just need to move the `try` into the loop (I thought it was already there).

timmywil: Ah, yes, that would work.

src/core/ready.js
timmywil: Agreed, given that it's inexpensive, I'd want to do whichever is smaller. Right now, that would be what we currently have.

src/core/ready.js
```diff
@@ -49,11 +47,42 @@ jQuery.extend( {
...
+							fn.call( document, jQuery );
+						}
+					}
+				} finally {
+					readyFiring = false;
+
+					// If there was an error in a ready callback,
+					// continue with the rest (gh-1823)
+					if ( readyCallbacks.length ) {
+						whenReady();
```
gibson042: Retrying synchronously here swallows all but the last exception generated by a block of `readyCallbacks` (e.g., `jQuery.ready.then( thrower ); jQuery.ready.then( thrower )`): https://jsfiddle.net/ewjschzw/1/

I think a `setTimeout` is necessary.

timmywil: Thanks, didn't notice that.

test/unit/ready.js
```diff
@@ -67,4 +91,32 @@ QUnit.module( "ready" );
 			"Argument passed to fn in jQuery(document).ready( fn ) should be jQuery" );
 	} );
 
+	QUnit.test( "Promise.resolve(jQuery.ready)", function( assert ) {
+		assert.expect( 2 );
+		var done = jQuery.map( new Array( 2 ), function() { return assert.async(); } );
```
Krinkle: `var done = [ assert.async(), assert.async() ];`

timmywil: This is a pattern we already use in several places. Regardless, we'll switch patterns when we update QUnit.

markelog: You want to somehow detect such pattern? You mean update to 2.0... or?

mgol: @Krinkle `assert.async` now accepts a number specifying how many times it should get called so this is sort of blessed by upstream currently... We planned to switch to this pattern once we update QUnit.

markelog: I have [updated](https://github.com/jquery/jquery/commit/b97c8d30c5aedace75dc17056d429f28e41b20c1) QUnit to 1.20.0, so after rebase, you can start using new signature in this pull

markelog:
Even though we unanimously decided to choose a different [approach](https://github.com/jquery/jquery/pull/2891) it's still would be cool to hear @scottgonzalez opinion.

scottgonzalez:
I don't have much to add here. I'll just say that I think `jQuery.ready` is a nice feature, though easily implemented in user land; and `jQuery.holdReady()` can go away and I'd wouldn't think twice about it.

timmywil:
Closing in favor of https://github.com/jquery/jquery/pull/2891

PR #2857, opened 2016-01-22
Set Edge's expected support for clearCloneStyle to true

zackhall:
The Microsoft Edge team noticed the following test failure in Edge from test/unit/support.js: "Verify that support tests resolve as expected per browser." If you're running the tests locally, I isolated the test with the query param `?testId=e1b3f3c8`.

Previous versions of Edge and IE had a bug in which cloning an element and then clearing the cloned element's style would also clear the original element's style. See #886 for an excellent and thorough analysis.

Turns out, somewhere along the line Edge fixed this bug but the unit tests expect Edge to fail still. 

I've copied the code from [support.js#L20](https://github.com/jquery/jquery/blob/487d5ca913c237aafe9efa1179749b46382fddbf/src/css/support.js#L20) that tests the browser for this bug into a [jsbin snippet](http://jsbin.com/gijaquzuwo/2/edit?js,console). If you run it in Edge, you can now see that jQuery determines `$.support.clearCloneStyle == true`.

### DIFF
--- a/test/unit/support.js
+++ b/test/unit/support.js
@@ -53,16 +53,17 @@ testIframeWithCallback(
 );
 
 ( function() {
-	var expected,
+	var expected, version,
 		userAgent = window.navigator.userAgent;
 
 	if ( /edge\//i.test( userAgent ) ) {
+		version = userAgent.match( /edge\/(\d+)/i )[1];
 		expected = {
 			"ajax": true,
 			"boxSizingReliable": true,
 			"checkClone": true,
 			"checkOn": true,
-			"clearCloneStyle": false,
+			"clearCloneStyle": version >= 13,
 			"cors": true,
 			"createHTMLDocument": true,
 			"focusin": false,


mgol:
This will break our CI as it runs on Edge 12. In which version of Edge has this been fixed? This will need version branching like what we had for Chrome some time ago, see 48a3549bc572bebb5dc581e4818e15f1443b3de2.

dmethvin:
By the way @zackhall , glad to see this bug was fixed, it was a bizarre one!

zackhall:
@dmethvin: Yeah, when I was reading back through the history of the old bug it got more and more odd.

@mgol: I'll try to pinpoint which specific version of Edge fixed this. It is passing on my current Edge browser which is Edge v25.11073, EdgeHTML v13.11073.

mgol:
@zackhall Is this a stable EdgeHTML 13 version that was released to the public or a newer build? If the stable one then branching at `version >= 13` should be enough.

Thanks!

markelog:
Relevant - https://github.com/jquery/jquery/pull/1036

> See #886 for an excellent and thorough analysis.

"Thank you" like that, shouldn't go unnoticed! So thank you @elijahmanor!

zackhall:
The currently released, stable build of Edge is [13.10586](https://dev.windows.com/en-us/microsoft-edge/platform/changelog/) and we've verified that `clearCloneStyle` is supported on that build. I will update the PR to branch at `version >= 13` as suggested.

zackhall:
Does the CLA signing take some time to propogate?

I just signed the CLA on the site, and it still says author has not signed the CLA. I also tried using the commit amend strategy on the CLA page.

mgol:
@zackhall You need to use the same email address to sign the CLA that you use in commits; they're different now.

mgol:
LGTM now, thanks!

mgol:

zackhall:
Thanks for the update, @mgol.

Issue #2700, opened 2015-11-09
Add CSS tests for setting negative padding and borderWidth


gibson042:
They should be clipped to zero. See https://github.com/jquery/jquery/pull/2695#discussion_r44318911

vitaliyterziev:
Hi, this is still not fixed- https://bugs.jquery.com/ticket/14038, the example here is working the same.

I checked and came to this in ```setPositiveNumber()```  ```Math.max( 0, matches[ 1 ] - ( subtract > 0 ? subtract : 0 || 0 ) ) + ( matches[ 2 ] || "px" ) :``` instead of ```Math.max( 0, matches[ 1 ] - ( subtract || 0 ) ) + ( matches[ 2 ] || "px" ) :```, because when negative number is returned from ```augmentWidthOrHeight()```, this calculation will add more px,em instead of subtract the number. Sorry if i am misunderstood then you can ignore it :)

PR #2864, opened 2016-01-25
#2700 negative border and padding tests

vitaliyterziev:
additional tests and fixing typos in the old ones


vitaliyterziev:
this is strange, the pull is from vitaliyterziev and i signed the CLA with the very same username.

timmywil:
@vitaliyterziev Thanks for doing this! Keep in mind that your github full name and email must match those in the CLA. There are also some code style errors, which can be viewed by running `grunt` locally – more info on that in the README. If you need any help, reply here.

PR #2865, opened 2016-01-25
Hidden attribute solution for show/hide/toggle

Nettsentrisk:
A simpler solution for show/hide/toggle based on the HTML5 hidden attribute.

To avoid ambiguity regarding how to write CSS to ensure hidden elements should be hidden, the hidden attribute value is set to hidden instead of empty.

This solution is more in line with best practices regarding how to show/hide elements, as it leaves it up to the author to ensure it's hidden properly via CSS (essentially the same as the class-based recommendation, removing/adding classes to control display). 

It will also leave the inline style intact, in case this is needed to overrule jQuery's toggling.

For responsive web sites, this will also provide a better way to control whether or not the element should be shown, as it can be controlled through the hidden attribute selector within the context being styled.

Not to mention: this probably performs better than having to remember the computed styles and resetting these when showing the element.

Another option is to add a method "unhide" that removes the hidden attribute, whereas "show" retains the functionality of the current code where its inline style is changed to force it to show.

### DIFF
--- a/src/css/showHide.js
+++ b/src/css/showHide.js
@@ -24,7 +24,7 @@ function showHide( elements, show ) {
 				elements[ index ].removeAttribute('hidden');
 			}
 			else {
-				elements[ index ].setAttribute('hidden', 'hidden';
+				elements[ index ].setAttribute('hidden', 'hidden');
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


timmywil:
Thanks for opening a PR. The show/hide code has gone through extensive revision and is a product of long debates. We discussed the possibility of using the hidden attribute, but ultimately decided against it. See the [many show/hide tickets](https://github.com/jquery/jquery/search?p=1&q=show+hide&type=Issues&utf8=%E2%9C%93) for more info.

Issue #2636, opened 2015-10-07
Add tests for contextmenu event


markelog:
It is not just an alias, since we fix `event` object for it

jha-naman:
I am interested in helping out here.
Please let me know if any changes are required, this stuff is new to me .

timmywil:
@markelog What did you mean it's not just an alias? Seems to me it is just an alias, since the event object will get fixed whether you use `.contextmenu(fn)` or `.on('contextmenu', fn)`.

mgol:
This issue isn't entirely clear and we don't really do more things to `contextmenu` than we do for e.g. `mousedown` so I'm going to close it. If @markelog replies with more concrete info, we can reopen.

markelog:
We apply special fixes for all events looking like `/^(?:mouse|pointer|contextmenu|drag|drop)|click/`, but don't test it.

For other events we at least have some tests, for `contextmenu` we don't have anything at all. Point of this ticket was to provide tests for `contextmenu` event specifically for the `which` property

PR #2874, opened 2016-01-27
Tests: adds test for contextmenu

jha-naman:
Fixes jquery/jquery#2636

### DIFF
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -2967,3 +2967,18 @@ if ( !( /firefox/i.test( window.navigator.userAgent ) ) ) {
 		}, 50 );
 	} );
 }
+
+QUnit.test( "contextmenu", function( assert ) {
+	assert.expect( 1 );
+
+	var times = 0,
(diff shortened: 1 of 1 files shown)


test/unit/event.js
```diff
@@ -2967,3 +2967,18 @@ if ( !( /firefox/i.test( window.navigator.userAgent ) ) ) {
...
+
+QUnit.test( "contextmenu", function( assert ) {
+	assert.expect( 1 );
+
+	var times = 0,
+	    handler = function() { ++times; };
+
+	jQuery( "#firstp" )
+	    .contextmenu( handler )
+	    .contextmenu()
```
timmywil: We actually already have [a test for adding and triggering "contextmenu"](https://github.com/jquery/jquery/blob/master/test/unit/event.js#L2861). @markelog What tests were you thinking of? Ensuring the event object is fixed for contextmenu?

markelog: I think so, but i honestly don't remember, since it was like four months ago

PR #2891, opened 2016-02-01
Custom builds: use different ready code when excluding Deferred

timmywil:
- Modules depending on deferred are excluded when deferred is excluded (ajax, effects, queue, and core/ready). Whenever deferred is excluded, core/ready is replaced by core/ready-no-deferred.
- `jQuery.ready.promise` has been removed and `jQuery.ready` made a thenable in core/ready.
- #1823 (error resiliency) has been addressed by giving up our synchronicity guarantee in ready handler execution after the ready event has been fired. This has multiple advantages. Besides error resiliency, it makes use of our standards-compatible `.then` in the `.ready` method.

### DIFF
--- a/src/core/ready-no-deferred.js
+++ b/src/core/ready-no-deferred.js
@@ -1,18 +1,16 @@
 define( [
 	"../core",
-	"../var/document",
-	"../deferred"
+	"../var/document"
 ], function( jQuery, document ) {
 
-// The deferred used on DOM ready
-var readyList = jQuery.Deferred(),
-	readyPromise = readyList.promise();
+var readyCallbacks = [],
+	readyFiring = false,
+	whenReady = function( fn ) {
+		readyCallbacks.push( fn );
+	};
 
 jQuery.fn.ready = function( fn ) {
-
-	// Add the callback
-	readyPromise.done( fn );
-
+	whenReady( fn );
 	return this;
 };
 
@@ -34,7 +32,6 @@ jQuery.extend( {
 		}
 	},
 
-	// Handle when the DOM is ready
 	ready: function( wait ) {
 
 		// Abort if there are pending holds or we're already ready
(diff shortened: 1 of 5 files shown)


src/core/ready.js
```diff
@@ -5,12 +5,13 @@ define( [
 ], function( jQuery, document ) {
 
 // The deferred used on DOM ready
-var readyList;
+var readyList = jQuery.Deferred(),
+	readyPromise = readyList.promise();
```
gibson042: I'm pretty sure you can get rid of `readyPromise` and just use `readyList` directly.

src/core/ready.js
```diff
@@ -5,12 +5,13 @@ define( [
...
 
 jQuery.fn.ready = function( fn ) {
 
 	// Add the callback
-	jQuery.ready.promise().done( fn );
+	readyPromise.done( fn );
```
gibson042: This `.done` is the source of #1823... if we're willing to give up a synchronicity guarantee, then it could become `.then`, the [synchronous invocations](https://github.com/jquery/jquery/pull/2891/files#diff-80299fe5124dec96e0bf1341cfcceb49R64) could get wrapped in `setTimeout` (removing `try...finally` in the process), and those [error resiliency tests](https://github.com/timmywil/jquery/commit/08d6e04d43b257284b9306fbb85a3e6423bac972#diff-11bce668df1353212ad70cea7ca6ceed) could be restored.

What do you think? Is anyone counting on `jQuery( invokedFirst ); invokedSecond()`?

mgol: I think it's much less likely to depend on that than to depend on synchronous `then` on a resolved promise and yet we made that change. Aligning those behaviors only seems good to me.

timmywil: I actually tried switching to `then`, but didn't get the behavior I expected. I'll keep digging.

mgol: @timmywil Did you find sth?

timmywil: @mgol I'll work on it this week.

src/core/ready-no-deferred.js
```diff
@@ -0,0 +1,109 @@
...
+	readyFiring = false,
+	whenReady = function( fn ) {
+		readyCallbacks.push( fn );
+	},
+	executeReady = function( fn ) {
+
+		// Prevent errors from freezing future callback execution (gh-1823)
+		// Not backwards-compatible as this does not execute sync
+		window.setTimeout( function() {
+			fn.call( document, jQuery );
```
gibson042: Do we want to keep the `document` context? It's another backwards-compatible but non-Promise "feature". And if we do keep it, there should be verifying assertions in test/unit/ready.js.

timmywil: Perhaps we can leave that subject to change/undocumented. I don't want to verify it if we're going to change it in a future version.

gibson042: I'm suggesting that we change it now, in both places.

timmywil: Let's open another ticket then. This PR is already 2 tickets. I'd rather not hold it up. But I'm indifferent as to what context we use here. `document` still makes sense since it's the document ready handler.

gibson042: This PR is the _introduction_ of `jQuery.ready.then`... why start off on the wrong foot when we don't have to? It doesn't even make sense to create a ticket referencing behavior that hasn't landed. But I guess I'll put some text here just in case:
> Per [Promises/A+](https://promisesaplus.com/#point-35) and [ES2015](http://www.ecma-international.org/ecma-262/6.0/#sec-promisereactionjob), `then` callbacks should be called without a `this` context. And although `jQuery.Deferred#*With` methods provide context to callbacks, we are moving in the direction of increased standards compliance, and should stop providing `document` context to `jQuery.ready` callbacks while we're already making backwards-incompatible changes to `jQuery.ready`.

dmethvin: I'm fine with not passing `document`, we just need to document that and the fact that it's async now. I can do that in the release notes once this lands.

timmywil: @gibson042 The change wouldn't just affect `jQuery.ready.then`, and I'm familiar with the spec on Promises. Like you said, the document context is for back-compat, as we still have `$(function(){})` and `$(document).ready()`, and while some may prefer `Promise.resolve(jQuery.ready)`, I imagine that `$(document).ready()` and `$(function() {})` will still be the most common, and a `document` context still makes sense to me on the latter 2. Like I said tho, I'm not opposed to changing that, but we don't have to change it now.

gibson042: Ok, I can accept that.

Issue #2914, opened 2016-02-10
Don't manually lowercase attribute name passed to `setAttribute`


mgol:
In #2730 we changed the lowercasing logic of the `.attr()` setter to better conform to the spec (we were lowercasing too much) but we had to lowercase at all only because of IE<9. We don't need that logic in `3.0` and removing it may even "accidentally" make this API sort-of work on SVGs (see #2910) so why not.

mgol:
PR: #2916

Issue #2910, opened 2016-02-09
.attr( attributeName, value ) turns attribute name to lowercase


agcolom:
It seems that when using .attr( attributeName, value ), the attributeName is turned to lowercase. This causes an issue when trying to set the viewBox attribute using jQuery. (the attribute has an upper case 'B').

Test cases: 
- setting the attribute without jQuery, using ```setAttribute```:
  - http://jsfiddle.net/jkzvquLw/4/, works as expected
- setting the attribute using jQuery and ```attr()```
  - http://jsfiddle.net/jkzvquLw/1/, doesn't work as expected, because the attribute set is ```viewbox``` (all lowercase) instead of ```viewBox```

agcolom:
Sorry just updated the url for ```setAttribute``` to [http://jsfiddle.net/jkzvquLw/3/](http://jsfiddle.net/jkzvquLw/3/)

mgol:
Thanks for the report. AFAIK this has always been the case. Since 1.12.0/2.2.0 we've made it more conforming to the spec, see my detailed description in this comment: https://github.com/jquery/jquery/issues/2730#issuecomment-157828916. The spec explicitly requires lowercasing the attribute name.

This description applies to HTML so we're doing the right thing but this algorithm doesn't apply to SVGs. jQuery doesn't support SVG manipulation (with rare exceptions), though, so I don't think we have to change anything here, determining current context and applying a different algorithm would be messy.

However, it seems in 3.0.0 we might just as well avoid lowercasing and let the browsers do their magic as it was mostly a problem for old IE. This would make it work automatically.

Thoughts?

P.S. @agcolom Keep in mind, though, that if you use jQuery on SVGs you'll find more & more such cases... And we're not going to support all of them. :) We're not an SVG library.

agcolom:
@mgol yes, thanks for the info. I've just checked the list of html attributes at https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes and all appear to be lowercase. I wonder whether it might be worth adding a note in the docs?

mgol:
We might add a note there but I wouldn't want to end up in a situation where we have explicit warnings against SVG on every other API page. :)

But I'd still want to investigate whether we may drop the lowercasing in jQuery 3.0, regardless of the lack of official SVG support here.

dmethvin:
See #2730 for the trail of tears.

mgol:
[...]

PR #2916, opened 2016-02-10
Attributes: remove the lower-casing logic for attribute names

mgol:
jQuery used to lower-case the attribute names passed to the .attr setter
to workaround an old IE issue. This is no longer in jQuery 3.0 and
removing it may even "accidentally" make this API sort-of work on SVGs
(see gh-2910) so why not.

Manual lowercasing had to be added to the place where the proper
attrHook is retrieved so that it works regardless of the casing of the
provided name. A regular `toLowerCase()` is enough there as those few
attributes don't contain any non-ASCII characters.

Fixes gh-2914

-18 bytes

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -7,14 +7,7 @@ define( [
 ], function( jQuery, access, support, rnotwhite ) {
 
 var boolHook,
-	attrHandle = jQuery.expr.attrHandle,
-
-	// Exclusively lowercase A-Z in attribute names (gh-2730)
-	// https://dom.spec.whatwg.org/#converted-to-ascii-lowercase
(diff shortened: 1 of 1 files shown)


mgol:
I didn't add unit tests for that as we officially don't support SVG attribute manipulation but I've just noticed we [have been testing that it works in some cases](https://github.com/jquery/jquery/blob/3f46cec169addb096fe576313e5cc6dde8618c04/test/unit/attributes.js#L460-L468)... So I'm not sure how to follow up on that one.

dmethvin:
Somehow I missed (twice!) that we back-ported this to 1.12.0/2.2,0, sorry! It makes sense to remove it for 3.0.0 since we were only lowercasing for old IE.

PR #2923, opened 2016-02-15
SVGElement.offsetWidth and offsetHeight will be removed from Chrome April 2016

eivindga:
Using getBoundingClientRect() to find width and height instead of element directly.

This is the same way it is done on master branch. (jquery 3.0)

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -118,22 +118,23 @@ function augmentWidthOrHeight( elem, name, extra, isBorderBox, styles ) {
 function getWidthOrHeight( elem, name, extra ) {
 
 	// Start with offset property, which is equivalent to the border-box value
-	var valueIsBorderBox = true,
-		val = name === "width" ? elem.offsetWidth : elem.offsetHeight,
+	var val,
+		valueIsBorderBox = true,
 		styles = getStyles( elem ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
+
+	// Support: IE <= 11 only
+	// Running getBoundingClientRect on a disconnected node
+	// in IE throws an error.
+	if ( elem.getClientRects().length ) {
+		val = elem.getBoundingClientRect()[ name ];
+	}
 
 	// Support: IE11 only
 	// In IE 11 fullscreen elements inside of an iframe have
 	// 100x too small dimensions (gh-1764).
 	if ( document.msFullscreenElement && window.top !== window ) {
-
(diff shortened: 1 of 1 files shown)


eivindga:
Please restart the CLA check, my github profile did not contain a full name when I first signed it.

mgol:
Thanks but we consider such a change risky as it changes `.width()`, `.height()` & related methods to return fractional data that not all code handles properly, that's why we decided to do it only in 3.0.0.

We don't officially support SVG (with a few exceptions) so this is not a required change for us to make and it creates a compatibility risk for people using jQuery for the regular HTML DOM.

For this to be non-breaking we'd have to round the resulting numbers. On the `1.12-stable` branch we'd need to have a fallback for IE<9. I'd rather we didn't do it.

mgol:
The CLA check fails because the commit is still signed by your nick instead of the full name. Click on the "Details" link of the failed CLA check to see what needs to be done to fix that.

mgol:
Thank you again for your contribution but at the today's Core meeting we decided to not backport this patch to jQuery<3.0. Changes like that are always risky so we'd rather not add them in a patch release, especially that they're only needed for SVG attribute manipulation which we don't support. If you want that to work, you'll need to upgrade to jQuery 3.0 which should arrive before Chrome 50 is released. We won't give promises it will work in 3.0, though, as SVG attribute manipulation remains officially unsupported by jQuery.

eivindga:
Ok, I see! Thanks for the quick reply! 

I thought it would be pretty safe to do on 2.2-stable, since it isn't supposed to work with IE8 anyway. 
http://caniuse.com/#search=getBoundingClientRect

But of course, it does change core functionality so it might lead to other issues...

Upgrading to Jquery 3.0 at this point is not an option for us, as we depend on multiple different jquery plugins that haven't been tested with 3.0. (And we need to ship our product to our client very soon, so we are kind of stuck with what is available today...)

mgol:
> I thought it would be pretty safe to do on 2.2-stable, since it isn't supposed to work with IE8 anyway. 
> http://caniuse.com/#search=getBoundingClientRect

While jQuery 2.x itself doesn't contain code needed only by IE8 & older, we strive to preserve API compatibility between 1.x & 2.x. So we'd only add fractional outputs for those methods on both branches or none of them.

This would still be risky without rounding, though; browsers couldn't make `offsetWidth` return fractional values because it was breaking live sites so it could be the same with jQuery changes.

PR #2925, opened 2016-02-16
Fix typos

wllm-rbnt:


dmethvin:
The first two files are external dependencies and not source input files, so they shouldn't be changed. The second two in `/test/unit` are indeed typos. Can you update your PR to remove the first two? Since you didn't do a branch and are using `master` it's probably best to start a new PR on a branch. You should be able to salvage your edits something like this:
```sh
git reset --soft HEAD~
git checkout -b fix-typos
git reset HEAD external/qunit/qunit.js test/data/jquery-1.9.1.js
git commit -a "Tests: fix typos"
git push origin fix-typos
```
Then make a new pull request against `fix-typos` and close this one. Since you've signed the CLA now that PR should pass with no problem!

PR #2933, opened 2016-02-18
Set statusText to "timeout" for 504

### LINKED ISSUE #2932: .ajax() doesn't emit textStatus: 'timeout' error when a 504 Gateway Timeout occurred
When 504 error occurred, callback `error`'s parameter `textStatus` is just an **`'error'`**.
(prototype: ` error: function( jqXHR jqXHR, String textStatus, String errorThrown )` )

`textStatus` should be `'timeout'` according to the RFC2616's description for 504 error. Though there's some conflicts when I try to comprehend how jQuery.ajax() should do when there's a gateway timeout in [jQuery API documentation](http://api.jquery.com/jQuery.ajax/):
> #### error
> Type: Function( jqXHR jqXHR, String textStatus, String errorThrown )
> A function to be called if the request fails. The function receives three arguments: The jqXHR (...) object, **a string describing the type of error that occurred** and an optional exception object, if one occurred. Possible values for the second argument (besides `null`) are **`"timeout"`**, `"error"`, `"abort"`, and `"parsererror"`. When an HTTP error occurs, `errorThrown` receives the textual portion of the HTTP status, such as "Not Found" or "Internal Server Error." As of jQuery 1.5, ... . Note: This handler is not called for cross-domain ... .

`504 Gateway Timeout` error usually occurs when it use a long-time-pending request to wait for some events but there's a **proxy / reverse proxy**, especially reverse proxy such as *nginx vhost*. That means: once the proxy configuration of gateway set a smaller timeout limit than the browser or ajax's `timeout `option, never a `timeout` error can get.

Here's part of [RFC2616 Hypertext Transfer Protocol -- HTTP/1.1](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.5.5) for reference.
> ### 10.5.5 504 Gateway Timeout
> 
> The server, while acting as a gateway or proxy, did not receive a timely response from the upstream server specified by the URI (e.g. HTTP, FTP, LDAP) or some other auxiliary server (e.g. DNS) it needed to access in attempting to complete the request.
> 
>       Note: Note to implementors: some deployed proxies are known to
>       return 400 or 500 when DNS lookups time out.

LionNatsu:
statusText = "timeout" when 504 Gateway Timeout occurs

Fixes gh-2932

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -772,6 +772,9 @@ jQuery.extend( {
 				}
 			} else {
 
+				if ( status === 504 ) {
+					statusText = "timeout";
+				}
 				// Extract error from statusText and normalize for non-aborts
 				error = statusText;
 				if ( status || !statusText ) {
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -772,9 +772,11 @@ jQuery.extend( {
 				}
 			} else {
 
+				// HTTP Error 504 Gateway Timeout
 				if ( status === 504 ) {
 					statusText = "timeout";
 				}
+
 				// Extract error from statusText and normalize for non-aborts
 				error = statusText;
 				if ( status || !statusText ) {


timmywil:
Thanks for the PR, but we've decided not to do this.

Issue #2868, opened 2016-01-25
Remove code for legacy browsers from tests


mgol:
We have some legacy code in tests kept mostly to keep `1.x` & `2.x` branches in sync. Now that we have only one branch it's high time to clean up all that. Most of that will probably be support comments for various ancient browsers.

mgol:
PR: #2949

PR #2949, opened 2016-02-25
Docs:Tests: Remove obsolete code from tests, update support comments

mgol:
Support comments that were lacking the final IE/Edge version that exhibits
the bug were checked & updated. Links to the Chromium bug tracker were updated.
Code in tests related to unsupported browsers (like Android 2.3 in non-basic
tests) has been removed.

Fixes gh-2868

The first commit is PR #2948, I just didn't want to introduce conflicts between my 2 PRs. I'll rebase once that lands.

### DIFF
--- a/src/ajax/parseXML.js
+++ b/src/ajax/parseXML.js
@@ -9,7 +9,8 @@ jQuery.parseXML = function( data ) {
 		return null;
 	}
 
-	// Support: IE9
+	// Support: IE 9-11 only
+	// IE throws on parseFromString with invalid input.
 	try {
 		xml = ( new window.DOMParser() ).parseFromString( data, "text/xml" );
 	} catch ( e ) {
--- a/src/data/Data.js
+++ b/src/data/Data.js
@@ -141,7 +141,7 @@ Data.prototype = {
 			// Support: Chrome <= 35-45+
 			// Webkit & Blink performance suffers when deleting properties
 			// from DOM nodes, so set to undefined instead
-			// https://code.google.com/p/chromium/issues/detail?id=378607
+			// https://bugs.chromium.org/p/chromium/issues/detail?id=378607 (bug restricted)
 			if ( owner.nodeType ) {
 				owner[ this.expando ] = undefined;
 			} else {
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -249,9 +249,8 @@ QUnit.module( "ajax", {
 				"Nullable": null,
 				"undefined": undefined
 
-				// Support: Firefox
+				// Support: IE 9 - 11, Edge 12 - 13+
 				// Not all browsers allow empty-string headers
-				// https://bugzilla.mozilla.org/show_bug.cgi?id=815299
 				//"Empty": ""
 			},
 			success: function( data, _, xhr ) {
@@ -1699,16 +1698,6 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
 }
 
 	QUnit.asyncTest( "#11743 - jQuery.ajax() - script, throws exception", 1, function( assert ) {
-
-		// Support: Android 2.3 only
-		// Android 2.3 doesn't fire the window.onerror handler, just accept the reality there.
-		if ( /android 2\.3/i.test( navigator.userAgent ) ) {
-			assert.ok( true, "Test skipped, Android 2.3 doesn't fire window.onerror for " +
-				"errors in dynamically included scripts" );
-			QUnit.start();
-			return;
-		}
-
 		var onerror = window.onerror;
 		window.onerror = function() {
 			assert.ok( true, "Exception thrown" );
@@ -1768,13 +1757,6 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
 				var parsedXML = jQuery( jQuery.parseXML( "<tab title=\"Added\">blibli</tab>" ) ).find( "tab" );
 				ajaxXML = jQuery( ajaxXML );
 				try {
-
-					// Android 2.3 doesn't automatically adopt nodes from foreign documents.
-					// (see the comment in test/manipulation.js)
-					// Support: Android 2.3
-					if ( /android 2\.3/i.test( navigator.userAgent ) ) {
-						parsedXML = jQuery( ajaxXML[ 0 ].adoptNode( parsedXML[ 0 ] ) );
-					}
 					ajaxXML.find( "infowindowtab" ).append( parsedXML );
 				} catch ( e ) {
 					assert.strictEqual( e, undefined, "error" );
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -431,20 +431,13 @@ QUnit.test( "on bubbling, isDefaultPrevented, stopImmediatePropagation", functio
 	$anchor2.off( "click" );
 	$main.off( "click", "**" );
 
-	// Android 2.3 doesn't support stopImmediatePropagation; jQuery fallbacks to stopPropagation
-	// in such a case.
-	// Support: Android 2.3
-	if ( /android 2\.3/i.test( navigator.userAgent ) ) {
-		assert.ok( true, "Android 2.3, skipping native stopImmediatePropagation check" );
-	} else {
-		$anchor2.on( "click", function( e ) {
-			e.stopImmediatePropagation();
-			assert.ok( true, "anchor was clicked and prop stopped" );
-		} );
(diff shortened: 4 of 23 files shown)


mgol:
I could update the support comments to conform to @gibson042's proposal at https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-69379197 since I'm already modifying a lot of them. Do we have a consensus on that? (This would involve changing some `IE 9-11 only` to `IE <=9 - 11 only` as the issues often exist in older IE versions as well).

src/manipulation/wrapMap.js
```diff
@@ -3,12 +3,11 @@ define( function() {
...
 var wrapMap = {
 
-	// Support: IE9
-	option: [ 1, "<select multiple='multiple'>", "</select>" ],
-
 	// XHTML parsers do not magically insert elements in the
 	// same way that tag soup parsers do. So we cannot shorten
 	// this by omitting <tbody> or other required elements.
+	option: [ 1, "<select multiple='multiple'>", "</select>" ],
+
```
gibson042: Extraneous line?

mgol: I wanted to separate `option` from the table-related entries.

src/manipulation/wrapMap.js
```diff
@@ -3,12 +3,11 @@ define( function() {
 // We have to close these tags to support XHTML (#13200)
 var wrapMap = {
 
-	// Support: IE9
```
mgol: I removed the `IE9` mention as I believe it's not an IE9 issue but just a general need for wrapping. This comment was added by you, @markelog, could you elaborate?

src/ajax/parseXML.js
```diff
@@ -9,7 +9,8 @@ jQuery.parseXML = function( data ) {
 		return null;
 	}
 
-	// Support: IE9
+	// Support: IE 9-11 only
```
timmywil: No need for `<=` to here since we don't support <9.

gibson042: It's not about what we support (which changes over time), it's about where the problematic behavior exists (which doesn't, except regarding not-yet-released versions).

timmywil: It's not, though. To me, the `<=` implies we are writing code that deals with those browsers, but we aren't.

mgol: @timmywil Do you think the same about comments like "// Support: IE 8-11 only"? We have a couple of those and I didn't remove IE8 mentions.

If we took such a stance to the extreme, we'd have to also change comments like `Chrome 40-48+` to `Chrome 48+` or `Chrome 47-48+` as we don't support older versions of Chrome than the previous one.

mgol: (I don't care too much about this aspect, though. I just want it settled so that we don't continuously argue about it, changing support comments back & forth)

dmethvin: I can see the benefit of either approach and I'm not sure it matters. It's sometimes handy to know whether older browsers exhibited this as well, and our official browser support is mentioned elsewhere. In a practical lazy-me sense, though, I wouldn't want to go back and edit Support comments when we drop support for older browsers, seems kind of tedious.

src/manipulation/wrapMap.js
markelog: > I removed the IE9 mention as I believe it's not an IE9 issue but just a general need for wrapping.

No its not, if you remove that wrapping, tests would not break in any other browser, in general, i would advise you to provide proof if you claim something

src/manipulation/wrapMap.js
markelog: I'm not sure why you would like to move code around, that just creates trash in the blame

test/unit/ajax.js
```diff
@@ -249,9 +249,8 @@ QUnit.module( "ajax", {
 				"Nullable": null,
 				"undefined": undefined
 
-				// Support: Firefox
+				// Support: IE 9 - 11, Edge 12 - 13+
 				// Not all browsers allow empty-string headers
```
markelog: Perhaps we should change that comment?

test/unit/event.js
```diff
@@ -2876,7 +2867,7 @@ QUnit[ jQuery.fn.click ? "test" : "skip" ]( "Event aliases", function( assert )
 	} );
 } );
 
-// Support: IE9 (remove when IE9 is no longer supported)
+// Support: IE9 only
```
markelog: So [this](https://github.com/jquery/contribute.jquery.org/issues/95) is not landed, still open and yet we already using it as standard :/

test/unit/ajax.js
mgol: To what?

src/manipulation/wrapMap.js
mgol: We don't run tests in XHTML so them not failing doesn't prove anything. As the comment below states, XHTML parsers require certain structure, e.g. omitting `<tbody>` is not allowed. I'm just asking - isn't that the same case for `<option>`? Isn't that required to be contained inside `<select>` in XHTML?

src/manipulation/wrapMap.js
mgol: I'm not "moving code around", I just thought the XHTML-related comment applies to this line as well. Doesn't it?

test/unit/event.js
mgol: We need this standard as we need to know what to use so TBH I'd like to document it before we have a JSCS plugin etc. We can always update the description if during the development of the JSCS plugin we discover the spec is incomplete/inconsistent but IMO it'd be better to have at least such an agreed description than none.

src/ajax/parseXML.js
markelog: I thought we discussed that a while ago at the meeting, it seems conclusion was to mention even those browsers that we do not support

src/data/Data.js
```diff
@@ -141,7 +141,7 @@ Data.prototype = {
 			// Support: Chrome <= 35-45+
 			// Webkit & Blink performance suffers when deleting properties
 			// from DOM nodes, so set to undefined instead
-			// https://code.google.com/p/chromium/issues/detail?id=378607
+			// https://bugs.chromium.org/p/chromium/issues/detail?id=378607 (bug restricted)
```
markelog: I wonder why it is restricted, like maybe at least chrome fixed that

src/manipulation/wrapMap.js
markelog: I don't know about such restriction, but if you removed that comment, shouldn't you have some grounds for it?

src/manipulation/wrapMap.js
markelog: As i recall, we know that would break only in table elements

test/unit/event.js
markelog: I think we should provide pull and land that issue

src/data/Data.js
markelog: Oh, hell, @paulirish could you help us out? Why is it restricted? Is it fixed?

src/manipulation/wrapMap.js
dmethvin: That's why he's asking and pointed it out.

markelog: I contacted at irc with @mgol, after this interaction, i believe our discussion was productive.

Yet, from my point of view, if something is proposed to be changed, it should be tested before such proposal is submitted, i believe this is widely accepted practise, including our workflow too, in the same way, we ask all prs to have tests or good reason not to have it.

markelog: In any case, i should have been more friendly, sorry.

src/ajax/parseXML.js
gibson042: Right, and that's the point of the `<=`... we don't know when it was introduced, but it exists as of a known version (in this case, 9), which we don't need to update when our browser support changes.

timmywil: I'm not adamant, but I agree with @dmethvin that it'd be nice not to be so strict that we need to go back and edit all Support comments.

test/unit/ajax.js
mgol: @markelog Ping? What did you mean?

markelog: > Not all browsers allow empty-string headers

Since now it's only browser

mgol: At least two: IE & Edge. ;) I don't care too much either way here.

markelog: haha, okay, dealer choice :)

src/manipulation/wrapMap.js
mgol: I reverted changes in this file as I didn't have time to research that properly.

Issue #2073, opened 2015-02-07
Standardize custom pseudo object


gibson042:
cc @timmywil @scottgonzalez

The Sizzle pseudos object is exposed on `jQuery.expr` as three distinct properties:
* `jQuery.expr.pseudos`
* `jQuery.expr.filters`
* `jQuery.expr[":"]`

Sizzle defines the first two, and jQuery defines the third.

None of them appear in the current jQuery documentation, though Sizzle documents the object as [`pseudos`](https://github.com/jquery/sizzle/wiki#sizzleselectorspseudosname--function-elem--) (and separately documents the related [`setFilters`](https://github.com/jquery/sizzle/wiki#sizzleselectorssetfilterslowercase_name--function-elements-argument-not--)), and I'm _sure_ we have published instructions on how to define a custom pseudo at least once, which would necessarily reference one of them. We also have internal uses in at least two libraries:
* core
  * defines `jQuery.expr[":"]` from `jQuery.expr.pseudos`
  * defines :hidden, :visible, and :animated by extending `jQuery.expr.filters`
  * calls `jQuery.expr.filters.hidden` in the :visible pseudo
* UI
  * defines :data, :focusable, :tabbable, and widget pseudos by extending `$.expr[ ":" ]`
  * replaces the :animated filter at `$.expr.filters.animated`
  * calls `$.expr.filters.visible` and `$.expr.filters.hidden`

So now that this issue affects three source repositories and [jquery/api.jquery.com](https://github.com/jquery/api.jquery.com), here's the crux of it: can we standardize on one internally and deprecate the others? All things considered, my vote would be on `filters`.

timmywil:
I actually would like to keep them all. `[":"]` is probably the most prevalent in user code and needs to stay regardless. `filters` has been there a long time, but it wasn't used as much. And `pseudos` is what NWMatchers' name for filters, so it's a more intuitive name to some.

Regardless, I'd like to continue recommending the use of the [.createPseudo() method](https://github.com/jquery/sizzle/wiki#sizzleselectorscreatepseudofunction) for creating custom selectors, making the choice of one property somewhat moot – except for `setFilter`. jQuery UI already uses it where available.

gibson042:
> I actually would like to keep them all.

In the sense of not moving to deprecated, or of not standardizing?

>  `[":"]` is probably the most prevalent in user code and needs to stay regardless.

[...]

PR #2969, opened 2016-03-03
Selector: filters -> pseudos

timmywil:
Fixes gh-2073

### DIFF
--- a/src/selector-sizzle.js
+++ b/src/selector-sizzle.js
@@ -5,6 +5,8 @@ define( [
 
 jQuery.find = Sizzle;
 jQuery.expr = Sizzle.selectors;
+
+// Deprecated
 jQuery.expr[ ":" ] = jQuery.expr.pseudos;
 jQuery.uniqueSort = jQuery.unique = Sizzle.uniqueSort;
 jQuery.text = Sizzle.getText;
--- a/src/css/hiddenVisibleSelectors.js
+++ b/src/css/hiddenVisibleSelectors.js
@@ -3,10 +3,10 @@ define( [
 	"../selector"
 ], function( jQuery ) {
 
-jQuery.expr.filters.hidden = function( elem ) {
-	return !jQuery.expr.filters.visible( elem );
+jQuery.expr.pseudos.hidden = function( elem ) {
+	return !jQuery.expr.pseudos.visible( elem );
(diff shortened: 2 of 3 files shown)


markelog:
Ref jquery-migrate [issue](https://github.com/jquery/jquery-migrate/issues/146).

I wonder if we should add tests not only for the jQuery constructor for these selectors, but specifically and isolated for these functions

dmethvin:
If we'll start warning about this soon (as in potentially Migrate 3.0) is there a UI ticket for this?

timmywil:
@markelog I'm not sure what you're suggesting. We have tests for the added selectors.

markelog:
@timmywil 

we have stuff like - 
```js
assert.equal( jQuery( "#table td:visible" ).length, 1, "hidden cell is not perceived as visible (#4512). Works on table elements" );
```

but should we have 
```js
  assert.equal(jQuery.expr.pseudos.visible(td), false);
```
Should we have direct tests? Cause tests through jQuery constructor or `jQuery#(find | is)` are proxies, if they work, doesn't mean the `jQuery.expr.pseudos.visible` methods works as expected, maybe jQuery constructor as unindent decorator added something along the way and positive result is just a side-effect.

In other words i'm talking about perfectionism :)

src/selector-sizzle.js
```diff
@@ -5,6 +5,8 @@ define( [
 
 jQuery.find = Sizzle;
 jQuery.expr = Sizzle.selectors;
+
+// Deprecated
 jQuery.expr[ ":" ] = jQuery.expr.pseudos;
```
gibson042: Shouldn't this move to the deprecated module?

timmywil: I thought it would be odd considering we have selector-native, which doesn't include this.

timmywil:
> if they work, doesn't mean the jQuery.expr.pseudos.visible methods works as expected

I think it does with the tests we have.

markelog:
> I think it does with the tests we have.

Some coverage tools would also agree with me in that assessment, but yeah, i guess we can take care of that when we implement it.

timmywil:
> Some coverage tools would also agree with me in that assessment

How so? It's still hitting the codepaths.

Issue #2968, opened 2016-03-03
isPlainObject difference between v2.2.1 and v1.11.2


mitar:
It seems that there is a difference for the following object between v2.2.1 (it returns that it is the plain object) and v1.11.2 (it returns that it is not):

```javascript
var parentObj = {foo: "bar"};
var childObj = Object.assign(Object.create(parentObj), {bar: "foo"});
jQuery.isPlainObject(childObj);
```

I think that 1.11.2 is correct.

mgol:
Confirmed. The [API page](https://api.jquery.com/jQuery.isPlainObject/) says:

> Check to see if an object is a plain object (created using "{}" or "new Object").

and the object presented here can't be created in this way.

The problem is present both on `master` and `2.2-stable` but not present on `1.12-stable` (so 1.12.1 is OK as well).

Thanks for the report!

PR #2970, opened 2016-03-03
Core: restore enumeration behavior in isPlainObject

timmywil:
Fixes gh-2968

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -225,6 +225,7 @@ jQuery.extend( {
 	},
 
 	isPlainObject: function( obj ) {
+		var key;
 
 		// Not plain objects:
 		// - Any object or value whose internal [[Class]] property is not "[object Object]"
@@ -239,9 +240,11 @@ jQuery.extend( {
 			return false;
 		}
 
-		// If the function hasn't returned already, we're confident that
-		// |obj| is a plain object, created by {} or constructed with new Object
-		return true;
+		// Own properties are enumerated firstly, so to speed up,
+		// if last one is own, then all properties are own
+		for ( key in obj ) {}
+
+		return key === undefined || hasOwn.call( obj, key );
 	},
 
 	isEmptyObject: function( obj ) {
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -286,7 +286,7 @@ QUnit.test( "type for `Symbol`", function( assert ) {
 });
 
 QUnit.asyncTest( "isPlainObject", function( assert ) {
-	assert.expect( 17 );
+	assert.expect( 18 );
 
 	var pass, iframe, doc, parentObj, childObj,
 		fn = function() {};
@@ -296,8 +296,10 @@ QUnit.asyncTest( "isPlainObject", function( assert ) {
 	assert.ok( jQuery.isPlainObject( new window.Object() ), "new Object" );
 
 	parentObj = { foo: "bar" };
-	childObj = Object.assign( Object.create( parentObj ), { bar: "foo" } );
+	childObj = Object.create( parentObj );
 
+	assert.ok( !jQuery.isPlainObject( childObj ), "isPlainObject(Object.create({}))" );
+	childObj.bar = "foo";
 	assert.ok( !jQuery.isPlainObject( childObj ), "isPlainObject(Object.create({}))" );
 
 	// Not objects shouldn't be matched
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -286,13 +286,19 @@ QUnit.test( "type for `Symbol`", function( assert ) {
 });
 
 QUnit.asyncTest( "isPlainObject", function( assert ) {
-	assert.expect( 15 );
+	assert.expect( 17 );
 
-	var pass, iframe, doc,
+	var pass, iframe, doc, parentObj, childObj,
 		fn = function() {};
 
 	// The use case that we want to match
 	assert.ok( jQuery.isPlainObject( {} ), "{}" );
+	assert.ok( jQuery.isPlainObject( new Object() ), "new Object" ); // jshint ignore:line
+
+	parentObj = { foo: "bar" };
+	childObj = Object.assign( Object.create( parentObj ), { bar: "foo" } );
+
+	assert.ok( !jQuery.isPlainObject( childObj ), "isPlainObject(Object.create({}))" );
 
 	// Not objects shouldn't be matched
 	assert.ok( !jQuery.isPlainObject( "" ), "string" );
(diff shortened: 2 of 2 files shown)


test/unit/core.js
```diff
@@ -286,13 +286,19 @@ QUnit.test( "type for `Symbol`", function( assert ) {
...
-	assert.expect( 15 );
+	assert.expect( 17 );
 
-	var pass, iframe, doc,
+	var pass, iframe, doc, parentObj, childObj,
 		fn = function() {};
 
 	// The use case that we want to match
 	assert.ok( jQuery.isPlainObject( {} ), "{}" );
+	assert.ok( jQuery.isPlainObject( new Object() ), "new Object" ); // jshint ignore:line
```
markelog: Why jshint comment there?

mgol: Because of `new Object()`, I guess. Is that test necessary considering that it's equivalent to `{}` if one doesn't modify `Object.prototype` which we don't support?

test/unit/core.js
```diff
@@ -286,13 +286,19 @@ QUnit.test( "type for `Symbol`", function( assert ) {
...
+
+	parentObj = { foo: "bar" };
+	childObj = Object.assign( Object.create( parentObj ), { bar: "foo" } );
```
markelog: `Object.assign`? IE < 12?

test/unit/core.js
markelog: `new window.Object`? If we talking about the same thing? `jshint ignore:line` is bit radical, it seems we should disable the rule, not the whole thing

markelog:
It seems `Object.assign` belongs in its own test with possible `skip`, i would go further though, and move all in different `QUnit.test`, like this one is even an `asyncTest` which doesn't make sense for these one liners i think

test/unit/core.js
markelog: Wow, that `new window.Object()` does work! Didn't even know why would jshint highlight that - just pointed in the sky in made it rain i guess :)

markelog: Which a jshint bug of course :)

src/core.js
```diff
@@ -239,9 +240,11 @@ jQuery.extend( {
 			return false;
 		}
 
-		// If the function hasn't returned already, we're confident that
-		// |obj| is a plain object, created by {} or constructed with new Object
-		return true;
+		// Own properties are enumerated firstly, so to speed up,
+		// if last one is own, then all properties are own
```
markelog: That's clever!

src/core.js
```diff
@@ -239,9 +240,11 @@ jQuery.extend( {
...
+		for ( key in obj ) {}
+
+		return key === undefined || hasOwn.call( obj, key );
```
markelog: How about just doing
```js
  return key === undefined
```
?

test/unit/core.js
mgol: I'd just change it to jQuery.extend.

src/core.js
markelog: But it seems we can't, cause of some **other** core tests are failing, but not the `isPlainObject` ones, which means that this pull is missing tests

test/unit/core.js
markelog: Personally, i don't like quick and dirty approaches, that would mean that behaviour of `jQuery.extend` exactly like `Object.assign`, which it isn't

markelog: >  i don't like quick and dirty approaches

Not that i don't use them or anything :)

gibson042: I don't see value in `Object.assign` _or_ `jQuery.extend`... the point of the test is inheritance, and own properties can easily be created without a function call (which would also allow for testing the non-plain object both before and after an assignment). But it has to be guarded anyway, since IE <=8 don't have `Object.create`.

markelog: Use case was presented with `Object.assign` so it seems thorough to check with it, but test without it wouldn't hurt either, i for one, not 100% if `Object.assign` doesn't change behaviour of extend object, so i would leave it, just to be sure.

gibson042: Or you could confirm that `Object.assign` is irrelevant: https://jsfiddle.net/meb6dd0g/

markelog: Two things: if now it is misidentified it doesn't mean it will be identified with `Object.assign` is used and that assert doesn't have to be proof of the fix, but proof of absence of potential regression.

Those are weak arguments, yeah, but i don't get why it is problematic to check with and without `Object.assign` just in case

src/core.js
gibson042: I like this restoration, but would _love_ to see the whole function simplified (_especially_ for 3.0): http://jsbin.com/hezeqilega
```js
isPlainObject: function ( obj ) {
	var proto, Ctor;

	// Catch obvious negatives
	if ( !obj || toString.call( obj ) !== "[object Object]" ) {
		return false;
	}

	// Rely on Object.getPrototypeOf where available
	if ( getPrototypeOf ) {
		proto = getPrototypeOf( obj );

		// Objects with no prototype (e.g., `Object.create( null )`) are plain
		if ( !proto ) {
			return true;
		}

		Ctor = hasOwn.call( proto, "constructor" ) && proto.constructor;

    // Pre-3.0 fallback
	} else {
		Ctor = obj.constructor;
	}

	// All remaining plain objects were constructed by a global Object
	return typeof Ctor === "function" && fnToString.call( Ctor ) === ObjectFunction;
}
```

test/unit/core.js
gibson042: Because it doesn't exist in all supported browsers, making it not just superfluous but also cumbersome.

mgol: I agree with @gibson042, let's just take Object.assign out and do the assignment manually.

As for guarding against IE8, it's not necessary on master but we might want to backport the tests to 1.12-stable and it will be needed there.

src/core.js
mgol: We can simplify it later on master, for 2.2-stable it'd be good to land quickly & release as users of our AMD modules are waiting for a critical fix.

mgol: But I agree with @markelog we should add more tests so that all the code is covered just by isPlainObject tests.

test/unit/core.js
gibson042: To clarify, only removal of this `Object.assign` should be considered blocking. My other comments can be addressed in a followup PR against master.

src/core.js
markelog: @gibson042 nice! I suppose this could be done as different pr

test/unit/core.js
markelog: Yeah, it doesn't exist, just like a lot of new es features, that is why people love polyfills, anyhow if that is such big cumberstone, i can take of that in follow-up commit.

markelog:
>  fixup! add test for object containing an object

Nice

Issue #2880, opened 2016-01-28
Limit all test suite selection operations to #qunit-fixture


gibson042:
For those that are explicitly document-scope, wrap with `supportjQuery( $jQueryResults ).filter( "#qunit-fixture *" )` (cf. https://github.com/jquery/jquery/issues/2877#issuecomment-175838039 ).

markelog:
I believe i have fixed that, there might be more, but they are not causing issues at the moment.

Note: that should be done for sizzle too.

gibson042:
Why did you close this? It is a general hygiene ticket that should stay open until we are reasonably confident that there are no "others".

markelog:
I believe i have addressed your comment mentioned in this ticket description, aside from resolved cases there is no proof that such tests even exist, they *might*, but it is not determined, so i think we reached "reasonable" criteria, otherwise i don't understand how you want determine if those selectors are there.

Also, i think bad hygiene is also to have tickets that are open forever, without any possibility of closing except for subjective feeling of confidence.

However, if you strongly feel about this, please feel free to re-open.

gibson042:
> i don't understand how you want determine if those selectors are there

By looking through all the unit tests.

> i think bad hygiene is also to have tickets that are open forever, without any possibility of closing except for subjective feeling of confidence.

I agree, but closing it now is indistinguishable from "wontfix". The point here is to avoid re-occurrence, and if you're looking for an objective characterization, here you go:

* Someone comments on this issue certifying that they have examined all unit tests and found no selections scoped so broadly that they could match QUnit output, or
* A commit lands on the `master` branch claiming to correct all such selections

gibson042:
> * Someone comments on this issue certifying that they have examined all unit tests and found no selections scoped so broadly that they could match QUnit output, or
* A commit lands on the `master` branch claiming to correct all such selections

The first of these should be discounted, because I've already found some offending examples:
* [attributes.js#L223](https://github.com/jquery/jquery/blob/f1300f18877f0140994190bb1b3a945fa4b2fd3d/test/unit/attributes.js#L223)
* [attributes.js#L261](https://github.com/jquery/jquery/blob/f1300f18877f0140994190bb1b3a945fa4b2fd3d/test/unit/attributes.js#L261)
* [attributes.js#L494-L496](https://github.com/jquery/jquery/blob/f1300f18877f0140994190bb1b3a945fa4b2fd3d/test/unit/attributes.js#L494-L496)
[...]

Issue #2877, opened 2016-01-27
Update npm dependencies


mgol:
Many of our dependencies got pretty out of date; it's time to bump them.

mgol:
@leobalter @gibson042 I wanted to update QUnit to the latest 1.20.0 as well but I only managed to use 1.18.0; versions 1.19.0 & 1.20.0 are introducing race conditions to our tests, making a few of them fail randomly on every full test run. Could you have a look? I'm seeing things like:
<img width="426" alt="screen shot 2016-01-27 at 12 41 59" src="https://cloud.githubusercontent.com/assets/1758366/12612771/749211b4-c4f5-11e5-9c28-cac84693656e.png">
<img width="487" alt="screen shot 2016-01-27 at 12 42 26" src="https://cloud.githubusercontent.com/assets/1758366/12612770/74919aa4-c4f5-11e5-99d1-5f34022708ff.png">

gibson042:
Hmmm, are you sure they're race conditions? Because it looks to me like the kind of thing that would pop up in response to _other_ failures, and ultimately stems from the test results HTML appearing in the very document being used as a test fixture. I've thought about this before and never really come to a good approach, but I think the best way is leveraging `supportjQuery`. So for assertions that rely upon document context, e.g. [selector: child and adjacent: All Grandchildren](https://github.com/jquery/jquery/blob/a268f5225cad9ab380494e61a10105cc9eb107e7/test/unit/selector.js#L112), we can do something like
```js
assert.deepEqual(
	supportjQuery( jQuery( "p > * > *" ) ).filter( "#qunit-fixture *" ).get(),
	q( "anchor1", "anchor2" ),
	"All Grandchildren"
);
```

mgol:
There are only a few failures like that but they have to start somewhere, don't they? Something had to have changed in QUnit 1.19.0.

gibson042:
But I'm pretty sure that what changed is the output format. jQuery's broad selection is picking up new elements from the QUnit HTML reporter.

PR #2973, opened 2016-03-05
Tests: restrict "p > * > *" test in selector.js to #qunit-fixture

all3fox:
Fix [`p > * > *`][1] assert from selector.js
Ref #2880

I am not sure if I should change any other `assert.t`'s from `selector.js` that do not use an id. The same goes for `css.js` and `manipulation.js`.

This change stops the `selector` test module from failing on Chrome 49 when `<All modules>` are run. It did not use to fail and it does not fail now, if the `selector` module is run alone.

[1]: https://github.com/jquery/jquery/issues/2877#issuecomment-175582442

### DIFF
--- a/test/data/testinit.js
+++ b/test/data/testinit.js
@@ -32,14 +32,14 @@ this.q = function() {
 
 /**
  * Asserts that a select matches the given IDs
- * @param {String} a - Assertion name
- * @param {String} b - Sizzle selector
- * @param {String} c - Array of ids to construct what is expected
- * @example t("Check for something", "//[a]", ["foo", "bar"]);
- * @result returns true if "//[a]" return two elements with the IDs 'foo' and 'bar'
+ * @param {String} message - Assertion name
+ * @param {String} selector - Sizzle selector
+ * @param {String} expectedIds - Array of ids to construct what is expected
+ * @param {(String|Node)=document} context - Selector context
+ * @example match("Check for something", "p", ["foo", "bar"]);
  */
-QUnit.assert.t = function( a, b, c ) {
-	var f = jQuery( b ).get(),
+function match( message, selector, expectedIds, context ) {
+	var f = jQuery( selector, context ).get(),
 		s = "",
 		i = 0;
 
@@ -47,7 +47,31 @@ QUnit.assert.t = function( a, b, c ) {
 		s += ( s && "," ) + '"' + f[ i ].id + '"';
 	}
 
-	this.deepEqual( f, q.apply( q, c ), a + " (" + b + ")" );
+	this.deepEqual( f, q.apply( q, expectedIds ), message + " (" + selector + ")" );
+}
+
+/**
+ * Asserts that a select matches the given IDs.
+ * The select is not bound by a context.
+ * @param {String} message - Assertion name
+ * @param {String} selector - Sizzle selector
+ * @param {String} expectedIds - Array of ids to construct what is expected
+ * @example t("Check for something", "p", ["foo", "bar"]);
+ */
+QUnit.assert.t = function( message, selector, expectedIds ) {
+	match( message, selector, expectedIds, undefined );
+};
+
+/**
+ * Asserts that a select matches the given IDs.
+ * The select is performed within the `#qunit-fixture` context.
+ * @param {String} message - Assertion name
+ * @param {String} selector - Sizzle selector
+ * @param {String} expectedIds - Array of ids to construct what is expected
+ * @example selectInFixture("Check for something", "p", ["foo", "bar"]);
+ */
+QUnit.assert.selectInFixture = function( message, selector, expectedIds ) {
+	match( message, selector, expectedIds, "#qunit-fixture" );
 };
 
 this.createDashboardXML = function() {
--- a/test/unit/selector.js
+++ b/test/unit/selector.js
@@ -109,7 +109,7 @@ QUnit.test( "child and adjacent", function( assert ) {
 	assert.t( "Child", "p>a", [ "simon1","google","groups","mark","yahoo","simon" ] );
 	assert.t( "Child w/ Class", "p > a.blog", [ "mark","simon" ] );
 	assert.t( "All Children", "code > *", [ "anchor1","anchor2" ] );
-	assert.t( "All Grandchildren", "p > * > *", [ "anchor1","anchor2" ] );
+	assert.selectInFixture( "All Grandchildren", "p > * > *", [ "anchor1","anchor2" ] );
 	assert.t( "Adjacent", "p + p", [ "ap","en","sap" ] );
 	assert.t( "Adjacent", "p#firstp + p", [ "ap" ] );
 	assert.t( "Adjacent", "p[lang=en] + p", [ "sap" ] );


all3fox:
Mind you, I also do not like that `assert.tqf` is essentially a copy of `assert.t`. Let me know if there is a way "not to repeat myself".

mgol:
You can create an internal helper function that would be based on `QUnit.assert.t` but that would take an additional optional argument for providing context. Then it'd be a simple as calling this function with proper arguments in both `QUnit.assert.t` & `QUnit.assert.tbf`.

all3fox:
What if instead of three functions (one worker and two wrappers) I were to add just one more parameter to existing `assert.t`?

mgol:
What would this parameter be?

all3fox:
It can be any context, but it is supposed to be either `undefined` or `#qunit-fixture`.

I have already done it in the latest git push above, in case you did not look at it.

mgol:
I like how it looks right now. :) Thought, @gibson042?

test/data/testinit.js
```diff
@@ -48,6 +49,32 @@ QUnit.assert.t = function( a, b, c ) {
...
+};
+
+/**
+ * Asserts that a select matches the given IDs.
+ * The select is performed with the `#qunit-fixture` context.
+ * @param {String} a - Assertion name
+ * @param {String} b - Sizzle selector
+ * @param {String} c - Array of ids to construct what is expected
+ * @example t("Check for something", "//[a]", ["foo", "bar"]);
+ * @result returns true if "//[a]" return two elements with the IDs 'foo' and 'bar'
```
markelog: What `@result` means in jsdoc? Can't find a reference to it

test/data/testinit.js
```diff
@@ -48,6 +49,32 @@ QUnit.assert.t = function( a, b, c ) {
...
+ */
+QUnit.assert.u = function( a, b, c ) {
```
markelog: Maybe we could switchto more explicit naming?

test/data/testinit.js
all3fox: @markelog I copied that jsdoc from the original `Qunit.assert.t`. For a second there I though to myself "there is no return statement -- that function must always return `undefined`". However, I do not know javascript well enough and decided that whoever wrote those doc lines knew better than me.

test/data/testinit.js
all3fox: @markelog now I also see that I missed adjusting function names at lines [39][1] and [73][2] in the jsdoc. Once you decide if you want new names or not, I will fix that.

[1]: https://github.com/jquery/jquery/pull/2973/files#diff-289e07cdd24b4895b9f305d02f7ad25bR39
[2]: https://github.com/jquery/jquery/pull/2973/files#diff-289e07cdd24b4895b9f305d02f7ad25bR39

markelog: With that comment, i have also invited you to this discussion, what do you think?

test/data/testinit.js
markelog: Yeah, i notice that, if i recall this method was ported from sizzle, maybe @timmywil have more info?

test/data/testinit.js
all3fox: It looks to me like `assert.t` does not mean anything in particular, apart from that it's short. 

If we rename `assert.t` into `assert.selectMatchesIds` (or something descriptive like that), the painfully long lines in unit tests will become even more painfully long (i.e. way beyond 80 chars). I would leave the name as is.

test/unit/selector.js
```diff
@@ -109,7 +109,7 @@ QUnit.test( "child and adjacent", function( assert ) {
 	assert.t( "Child", "p>a", [ "simon1","google","groups","mark","yahoo","simon" ] );
 	assert.t( "Child w/ Class", "p > a.blog", [ "mark","simon" ] );
 	assert.t( "All Children", "code > *", [ "anchor1","anchor2" ] );
-	assert.t( "All Grandchildren", "p > * > *", [ "anchor1","anchor2" ] );
+	assert.u( "All Grandchildren", "p > * > *", [ "anchor1","anchor2" ] );
```
gibson042: I agree with others on wanting a more meaningful name, e.g. `selectInFixture`.

test/data/testinit.js
```diff
@@ -35,11 +35,12 @@ this.q = function() {
  * @param {String} a - Assertion name
  * @param {String} b - Sizzle selector
  * @param {String} c - Array of ids to construct what is expected
+ * @param {String} d - Context selector, either `undefined` or `#qunit-fixture`
  * @example t("Check for something", "//[a]", ["foo", "bar"]);
  * @result returns true if "//[a]" return two elements with the IDs 'foo' and 'bar'
  */
-QUnit.assert.t = function( a, b, c ) {
-	var f = jQuery( b ).get(),
+function match( a, b, c, d ) {
```
gibson042: I know this isn't your fault, but I'd rather not exacerbate these meaningless parameter names:
* `a` → `message` (cf. [`deepEqual`](http://api.qunitjs.com/deepEqual/))
* `b` → `selector` (cf. [`jQuery`](http://api.jquery.com/jQuery/))
* `c` → `expectedIds` (cf. [`deepEqual`](http://api.qunitjs.com/deepEqual/))
* `d` → `context` (cf. [`jQuery`](http://api.jquery.com/jQuery/))

test/unit/selector.js
all3fox: @gibson042 and perhaps rename `assert.t` to `assert.select`?

gibson042: > and perhaps rename `assert.t` to `assert.select`?

I'm open to a multi-PR approach here, leaving `assert.t` alone in this one and removing it entirely in another. @jquery/core?

mgol: I agree with @gibson042. Changing `assert.t` will require lots of changes so it'd be better done separately to not add too much work to this PR.

all3fox:
Did a rebase and squash, changed title. Hopefully, did not forget anything along the way.

all3fox:
Sorry, misclicked.

test/data/testinit.js
```diff
@@ -32,14 +32,14 @@ this.q = function() {
...
  * Asserts that a select matches the given IDs
- * @param {String} a - Assertion name
- * @param {String} b - Sizzle selector
- * @param {String} c - Array of ids to construct what is expected
- * @example t("Check for something", "//[a]", ["foo", "bar"]);
- * @result returns true if "//[a]" return two elements with the IDs 'foo' and 'bar'
+ * @param {String} message - Assertion name
+ * @param {String} selector - Sizzle selector
+ * @param {String} expectedIds - Array of ids to construct what is expected
+ * @param {String} context - Expected `undefined` or `#qunit-fixture`
```
gibson042: The type should be `{(String|Node)=document}` (cf. [`jQuery`](http://api.jquery.com/jQuery/)), and the description should merely document it as "selection context".

test/data/testinit.js
```diff
@@ -32,14 +32,14 @@ this.q = function() {
...
+ * @example match("Check for something", "//[a]", ["foo", "bar"]);
```
gibson042: `//[a]` is not a valid selector (it's XPath).

test/data/testinit.js
all3fox: @gibson042 I agree that I missed that a context is not necessarily a string. Can be a jQuery object too. Which is what you mean by `(String|Node)`. But I do not understand the `=document)` part of your comment. Looking at the linked page did not help, please explain?

all3fox: Does it mean "defaults to entire document"? I also see that `Node` not only means a jQuery object but also a bare DOM element?

gibson042: > But I do not understand the `=document)` part of your comment. Looking at the linked page did not help, please explain?
> …
> Does it mean "defaults to entire document"?

Correct: http://usejsdoc.org/tags-type.html

gibson042:
Landed; thanks!

Issue #2975, opened 2016-03-05
Eliminate the intro/outro nuisance tickets


dmethvin:
Good news! The development best practice of using JSHint or ESLint to check source files is becoming a lot more common now.

Bad news! People are running JSHint and ESLint on jQuery source files and getting error messages on intro.js and outro.js since they are not actually complete files. Examples are #2747, #2819, #2953, and (I think based on Google Translate) #2971.

We should be able to fix this by changing the extension of those two files from `.js` to `.js_partial` or some such. That way they won't be linted by default and the name should provide a clue that they are not complete files.

mgol:
I've been thinking about that recently. If that's possible, I'd prefer to change [the wrapping logic](https://github.com/jquery/jquery/blob/5943f1d7ffef41cc1a99bffe189f6d111fbccfd0/build/tasks/build.js#L30-L33) so that we have a template js file containing both intro & outro that would have a placeholder comment to be replaced by the rest of the build file. That would mean those files, consolidated, could keep their JS extension.

dmethvin:
That would let us lint the combined file; currently the parts are obviously excluded although they are linted once the combined file is created. :+1:

mgol:
> That would let us lint the combined intro/outro file

Yup, that was my thinking. :)

timmywil:
I like it.

jaubourg:
I came to the same conclusion myself! Go for it! :)

markelog:
Don't want to be a downer here, but those linters would still throw error on `wrapper` file :/, since there like unused arguments and a "dead code" for them.

It seems we just complicated the build process :/, again sorry, for the bad mood about this

timmywil:
@markelog We aren't setting eslint rules, so if they're running lint without modifying their own rules, it's going to break regardless. If they're using jshint, it will work because of the pragma at the top.

PR #2976, opened 2016-03-07
Build: combine intro and outro

timmywil:
Fixes gh-2975

### DIFF
--- a/src/wrapper.js
+++ b/src/wrapper.js
@@ -45,6 +45,7 @@
 //"use strict";
 
 // @CODE
+// build.js inserts compiled jQuery here
 
 return jQuery;
 } ) );
--- a/src/wrapper.js
+++ b/src/wrapper.js
@@ -1,3 +1,4 @@
+/*jshint unused:false */
 /*!
  * jQuery JavaScript Library v@VERSION
  * https://jquery.com/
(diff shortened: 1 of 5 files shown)


mgol:
:+1:

src/wrapper.js
```diff
@@ -35,10 +36,15 @@
...
-}(typeof window !== "undefined" ? window : this, function( window, noGlobal ) {
+}( typeof window !== "undefined" ? window : this, function( window, noGlobal ) {
 
 // Support: Firefox 18+
 // Can't be in strict mode, several libs including ASP.NET trace
 // the stack via arguments.caller.callee and Firefox dies if
 // you try to trace through "use strict" call chains. (#13335)
 //"use strict";
+
+// @CODE
```
dmethvin: Just to preempt any further tickets being opened, maybe this could say to the effect of, "see build.js" as well?

dmethvin:
Other than the clarification pointing to build.js, this LGTM.

src/wrapper.js
timmywil: Sure

src/wrapper.js
```diff
@@ -35,10 +36,16 @@
...
+// build.js inserts compiled jQuery here
```
mgol: Could we get this comment in the same line as `@CODE` so that it disappears in the built file? It's weird to stil have this comment in the full file.

dmethvin: Agreed, that may confuse people.

timmywil: It gets removed.

markelog:
Wish i would have time to review this, couldn't we use underscore templates or something like this?

timmywil:
@markelog Not really. We're using r.js to build. Besides, it's a simple split.

Issue #2979, opened 2016-03-09
Published jQuery versions should have `@VERSION` etc. replaced in `src/` as well


mgol:
If someone uses our AMD modules in a published version, they shouldn't get `$.fn.version === "@VERSION"` but the true version instead. We should do the replacement not only in `dist/jquery.js` but also in `src/` (but only for published versions, not on regular `grunt` runs).

timmywil:
I suppose we can do that.

PR #2981, opened 2016-03-09
Release: set version in source for releases

timmywil:
Fixes gh-2979

### DIFF
--- a/build/release.js
+++ b/build/release.js
@@ -9,7 +9,8 @@ module.exports = function( Release ) {
 			"dist/jquery.min.map",
 			"dist/jquery.slim.js",
 			"dist/jquery.slim.min.js",
-			"dist/jquery.slim.min.map"
+			"dist/jquery.slim.min.map",
+			"src/core.js"
 		],
 		cdn = require( "./release/cdn" ),
 		dist = require( "./release/dist" ),
--- a/build/release.js
+++ b/build/release.js
@@ -1,3 +1,4 @@
+var fs = require( "fs" );
 
 module.exports = function( Release ) {
 
@@ -27,6 +28,15 @@ module.exports = function( Release ) {
 			ensureSizzle( Release, callback );
 		},
 		/**
+		 * Set the version in the src folder for distributing AMD
+		 */
+		_setSrcVersion: function() {
+			var corePath = __dirname + "/../src/core.js",
+				contents = fs.readFileSync( corePath, "utf8" );
+			contents = contents.replace( /@VERSION/g, Release.newVersion );
+			fs.writeFileSync( corePath, contents, "utf8" );
+		},
+		/**
 		 * Generates any release artifacts that should be included in the release.
 		 * The callback must be invoked with an array of files that should be
 		 * committed before creating the tag.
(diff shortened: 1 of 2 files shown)


mgol:
Could you also replace [`@DATE`](https://github.com/jquery/jquery/blob/59ec78e6020cc963c1f95cb96a28eaaf20e37b3e/src/wrapper.js#L13)?

build/release.js
```diff
@@ -1,3 +1,4 @@
+var fs = require( "fs" );
```
markelog: Looks inconsistent, other requires in the `module.exports` scope

build/release.js
```diff
@@ -27,6 +29,15 @@ module.exports = function( Release ) {
 			ensureSizzle( Release, callback );
 		},
 		/**
+		 * Set the version in the src folder for distributing AMD
+		 */
+		_setSrcVersion: function() {
+			var corePath = __dirname + "/../src/core.js",
+				contents = fs.readFileSync( corePath, "utf8" );
+			contents = contents.replace( /@VERSION/g, Release.newVersion );
```
markelog: Feels so weird that we use those regexes all over, while not using something builder, package or something

build/release.js
timmywil: I make a distinction between modules that should be required when the exported function is called and ones that don't need to be.

build/release.js
timmywil: I ain't afraid of regex replaces. :)

timmywil:
@mgol I think it's okay to publish files that aren't used. If anyone tries to use wrapper.js, they'll find out quick it doesn't work, but I doubt anyone will actually try that. I don't mind leaving it out, but not copying it would mean switching to a list of all of the files in the src folder, rather than just copying the src folder (at least for bower), and I'd rather not do that.

mgol:
> I don't mind leaving it out, but not copying it would mean switching to a list of all of the files in the src folder, rather than just copying the src folder (at least for bower), and I'd rather not do that.

You could use glob patterns & negate this one file out without major changes. But I don't insist. ;)

So, are you for leaving it inside the published package with the non-replaced `@DATE`?

build/release.js
markelog: <xkcd commic>

build/release.js
markelog: Feels very specific as "gotcha" style rule i.e. hard to follow, don't recall we do this in anywhere else.

build/release.js
dmethvin: :smile_cat: True but it's a question of complexity. I know how a regex works very well, but if you pull in a template package I will have to look up how it works. At some point I agree it might make sense to put one in, but to me it doesn't look like we are at that point yet. The code is all very understandable and we haven't had problems with it.

markelog: Somewhere i suggested to use underscore templates since they used in grunt itself and lodash (obviously) also has them, which make them kind of standard in some sense.

The bigger question for me is that we have a lot of custom logic for the build, i don't know, but it seems this should be a common issue and therefore should have a common solution...

But of course i need to put my money where the mouth is, i.e. do something about this, otherwise it is just unproductive critic from my side.

timmywil:
> So, are you for leaving it inside the published package with the non-replaced @DATE?

For now, I'd be open to a later patch excluding the file, but it's not high priority.

PR #2988, opened 2016-03-12
Build: try insight package to get info about custom builds

### LINKED ISSUE #2890: Try out insight package to get info about custom builds
See https://github.com/yeoman/insight

markelog:
Fixes #2890

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -10,6 +10,8 @@ module.exports = function( grunt ) {
 
 	var fs = require( "fs" ),
 		requirejs = require( "requirejs" ),
+		Insight = require( "insight" ),
+		pkg = require( "../../package.json" ),
 		srcFolder = __dirname + "/../../src/",
 		rdefineEnd = /\}\s*?\);[^}\w]*$/,
 		read = function( fileName ) {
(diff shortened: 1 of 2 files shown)


build/tasks/build.js
```diff
@@ -312,10 +314,47 @@ module.exports = function( grunt ) {
 	//   grunt build:*:*:+ajax:-dimensions:-effects:-offset
 	grunt.registerTask( "custom", function() {
 		var args = this.args,
-			modules = args.length ? args[ 0 ].replace( /,/g, ":" ) : "";
+			modules = args.length ? args[ 0 ].replace( /,/g, ":" ) : "",
+			done = this.async(),
+			insight = new Insight( {
+				trackingCode: "UA-74908151-1",
```
markelog: What to put here?

build/tasks/build.js
```diff
@@ -312,10 +314,47 @@ module.exports = function( grunt ) {
...
 		grunt.log.writeln( "Creating custom build...\n" );
 
-		grunt.task.run( [ "build:*:*" + ( modules ? ":" + modules : "" ), "uglify", "dist" ] );
+		// Ask for permission the first time
+		if ( insight.optOut === undefined ) {
+			insight.askPermission( "", function( error, result ) {
+				exec( result );
+			} );
+		} else {
+			exec( true );
```
gibson042: Shouldn't this be `exec( !insight.optOut )`?

build/tasks/build.js
```diff
@@ -312,10 +314,47 @@ module.exports = function( grunt ) {
...
+			grunt.task.run( [ "build:*:*" + ( modules ? ":" + modules : "" ), "uglify", "dist" ] );
+			done();
+		}
 
 		grunt.log.writeln( "Creating custom build...\n" );
 
-		grunt.task.run( [ "build:*:*" + ( modules ? ":" + modules : "" ), "uglify", "dist" ] );
+		// Ask for permission the first time
+		if ( insight.optOut === undefined ) {
+			insight.askPermission( "", function( error, result ) {
```
gibson042: A `null` first argument would use the default message: https://www.npmjs.com/package/insight#askpermissionmessage-callback

markelog: ok-ok, will change

build/tasks/build.js
```diff
@@ -312,10 +314,47 @@ module.exports = function( grunt ) {
...
+				exec( result );
+			} );
```
gibson042: I'd like to see something like `setTimeout( exec, grunt.config( "promptTimeout" ) )` here (defaulting to 60 seconds), so we don't break automated workflows.

gibson042:
This is really cool. :clap:

build/tasks/build.js
markelog: That was kinda the point to do this in 3.0, so we could be back it up, if needed, but no fear, everything is already [thought](https://github.com/yeoman/insight/blob/cca7b5902cda58970a69b7825cea2c62cc79917d/lib/index.js#L127) through, as it turns out.

build/tasks/build.js
markelog: yeah, you right, but why we need to coerce?

gibson042: We don't need to coerce, we need to invert: `optOut` is the opposite of `trackingAllowed`.

build/tasks/build.js
gibson042: That's interesting, but I note that they permanently store the automatic opt-out, whereas I would want it to remain undefined. At any rate, I'll take it up [there](https://github.com/yeoman/insight/issues/39) and this won't need to change.

build/tasks/build.js
timmywil: Good question. @dmethvin Which account do we use for GA on jQuery sites?

dmethvin: Use `-4` instead of `-1` in the code, that will separate the results from the web sites and make it easier to find. So, UA-74908151-4

markelog: aam, `UA-74908151` is my account, can we use `jquery.com`?

dmethvin: Sorry i'm horrible at long numbers! `UA-1076265-4`

markelog:
Okay, so it seems to be ready

Issue #2986, opened 2016-03-11
Simplify isPlainObject


gibson042:
We can skip property enumeration and just look for prototype/constructor directly (cf. https://github.com/jquery/jquery/pull/2970#discussion_r54991520 ).

markelog:
3.0.1? It seems you already have a draft for the refactoring is there any concerns for postponing?

mgol:
We can be adding more & more small items to the milestone & never release; I think it's good to not add new things to `3.0.0` now if not necessary. We can always change the milestone down to `3.0.0` *at the moment of merging*.

wisdommm:
1

发自我的 iPhone

wisdommm:
1


发自我的 iPhone

markelog:
Sounds more like invitation for bureaucratic discussion, since we can also do the opposite - make it a `3.0.0` and move it to `3.0.1` if necessary, if code is ready and tests are there i'm not sure why it can't be an easy fix.

In any way, i'm not suggesting to add a `blocker` label, but asking @gibson042 if there is any concerns with proposed code.

mgol:
> Sounds more like invitation for bureaucratic discussion, since we can also do the opposite - make it a `3.0.0` and move it to `3.0.1` if necessary

This adds the burden of having to make an actual decision of moving it to a later milestone which can delay the release. This release has already dragged on for way too long.

> if code is ready and tests are there

There is no PR so it's not ready yet.

markelog:
I think you making an issue where there isn't one, if we can land it, i think we should, i can create pr right now, this discussion feels redundant and highly unproductive

markelog:
Discussed at the meeting, if code will be submitted before 3.0.0 release, it will land, if not, then waiting until the next one

PR #2998, opened 2016-03-14
Core: Simplify isPlainObject

gibson042:
Fixes gh-2986

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,6 +1,7 @@
 define( [
 	"./var/arr",
 	"./var/document",
+	"./var/getProto",
 	"./var/slice",
 	"./var/concat",
 	"./var/push",
@@ -8,10 +9,13 @@ define( [
 	"./var/class2type",
 	"./var/toString",
 	"./var/hasOwn",
+	"./var/fnToString",
+	"./var/ObjectFunctionString",
 	"./var/support",
 	"./core/DOMEval"
-], function( arr, document, slice, concat,
-	push, indexOf, class2type, toString, hasOwn, support, DOMEval ) {
+], function( arr, document, getProto, slice, concat, push, indexOf,
+	class2type, toString, hasOwn, fnToString, ObjectFunctionString,
+	support, DOMEval ) {
 
 var
 	version = "@VERSION",
@@ -23,10 +27,6 @@ var
 		// Need init if jQuery is called (just allow error to be thrown if not included)
 		return new jQuery.fn.init( selector, context );
 	},
-
-	getProto = Object.getPrototypeOf,
-	fnToString = jQuery.toString,
-	ObjectFunctionString = fnToString.call( Object ),
 
 	// Support: Android<4.1
 	// Make sure we trim BOM and NBSP
--- a/src/core.js
+++ b/src/core.js
@@ -24,6 +24,10 @@ var
 		return new jQuery.fn.init( selector, context );
 	},
 
+	getProto = Object.getPrototypeOf,
+	fnToString = jQuery.toString,
+	ObjectFunctionString = fnToString.call( Object ),
+
 	// Support: Android<4.1
 	// Make sure we trim BOM and NBSP
 	rtrim = /^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g,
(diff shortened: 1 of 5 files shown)


dmethvin:
Hmmm. @gibson042 did you change your git sig? If not this may be a recent change to the CLA checker by @csnover.

gibson042:
I haven't changed my signature, but was wondering if the failure was caused by opening a PR with two commits.

csnover:
I suspect I know what the problem is, apologies. One moment while I fix that.

src/var/getProto.js
```diff
@@ -0,0 +1,3 @@
+define( function() {
+	return Object.getPrototypeOf;
```
mgol: Is it to protect against overwriting this field after loading jQuery?

src/core.js
```diff
@@ -225,28 +229,24 @@ jQuery.extend( {
...
-		for ( key in obj ) {}
+		// Objects with no prototype (e.g., `Object.create( null )`) are plain
+		if ( !proto ) {
+			return true;
+		}
 
-		return key === undefined || hasOwn.call( obj, key );
+		// Objects with prototype are plain iff they were constructed by a global Object function
+		Ctor = hasOwn.call( proto, "constructor" ) && proto.constructor;
+		return typeof Ctor === "function" && fnToString.call( Ctor ) === ObjectFunctionString;
```
mgol: Is it safe? ES6 doesn't fully specify [how `Function.prototype.toString` should behave](http://www.ecma-international.org/ecma-262/6.0/#sec-function.prototype.tostring). There's [a proposal](https://tc39.github.io/Function-prototype-toString-revision/) to make it more specific but that's still work in progress.

src/var/getProto.js
gibson042: No; just because it's smaller.

src/core.js
gibson042: What risks do you anticipate? This isn't relying on any specific behavior, just consistency. And there are already tests covering that functions aren't recognized as plain objects.

src/var/ObjectFunctionString.js
```diff
@@ -0,0 +1,5 @@
+define( [
+	"./fnToString"
+], function( fnToString ) {
+	return fnToString.call( Object );
```
mgol: This `.call` is unnecessary, isn't it?

mgol: It isn't, disregard.

src/core.js
mgol: You're right it only depends on consistency which makes the risk small but this consistency is not specced as well as I understand it. Even if all envs we test on behave like that it still makes me a little uneasy to rely on such unspecced behavior.

gibson042: No, I mean that this technique will work on any environment that is _internally_ consistent in yielding different string representations of `Object` (e.g., "function Object() { [native code] }") and `Function` (or any other function, for that matter), regardless of the specifics on other environments—which is guaranteed by unit tests. Did you have a specific concern that isn't covered?

mgol: I wouldn't be surprised if some env behaved differently in its `toString` when applied to objects from another realm.

Unit tests only guarantee that it works in currently tested browsers but that itself doesn't mean those browsers may not change in the future, that other not-officially tested envs behave differently etc. If passing unit tests was the only requirement we'd be free to use proprietary APIs that are implemented widely enough and yet we do look at standards before using something.

gibson042: With the caveat that variation that variation is always a risk, this logic works on all known environments (including weird ones like Rhino) and will be even more solid if https://tc39.github.io/Function-prototype-toString-revision/ is accepted. And if this logic fails, or any environment _does_ change, we'll get a bug report and can react accordingly.

Note also that the current logic isn't just inefficient, it is incomplete, and fails on empty-object prototypes (PR updated). And as for the cross-realm concern, that's [already tested](https://github.com/gibson042/jquery/blob/2986/test/unit/core.js#L359-L364), but could be expanded if you like.

mgol: The `Function#toString` revision doesn't specify the `BindingIdentifier` that's being used; an implementation where `Function.prototype.toString.call(ObjectFromAnotherFrame) === function surprise() { [native code] }` would be conformant as well.

That said, I can see how edge-casey this all seems so since it improves the logic I won't object to landing this in its current state unless there's a way to not rely on `Function#toString` for now and still keep the improvements.

gibson042: https://github.com/tc39/Function-prototype-toString-revision/pull/10

mgol: OK, you win. :smile:

dmethvin: That's some Kobayashi Maru level stuff there @gibson042, if the spec doesn't fully support the code, change the spec! :100:

jaubourg: (imagines @gibson042 twenty years from now, biting in an apple while explaining the trick to a youngster)

Issue #2978, opened 2016-03-08
Cannot set `val()` for select options containing tabs (`\t`)


carols10cents:
I have a select element that contains options whose values contain tab characters. My original use case was a trailing tab (`string\t`), but I've also tested with leading tabs (`\tstring`) and tabs in the middle of non-whitespace (`str\ting`).

I called `.val('value with tab')` on the select element and expected the corresponding option to be selected since the values are identical, but no option is selected.

A few interesting observations:
- I can work around this issue for the leading and trailing cases by removing the leading or trailing tab characters from the string I pass to `.val()`, but this is unexpected since the values don't match exactly.
- Removing a middle tab character and passing that to `.val()` does not select the value, so there is no workaround for strings that have tab characters in the middle.
- The behavior is the same with jquery 2.2.1, 3.0.0-alpha1 and edge.
- With jquery 1.12.1, the leading and trailing tabs have the bug behavior but having a tab in the middle of the word and passing a matching string with a tab in the middle works as expected, so there was a regression at some point in this case.

JSFiddle: https://jsfiddle.net/onnucjx7/2/ You'll see in the console what I expect the behavior to be compared to what the behavior actually is. 

I searched for "tab" and "whitespace" and didn't see anything directly relevant; I did find this issue https://github.com/jquery/jquery/issues/1902 but it only mentions IE10 and I am seeing the unexpected behavior in Firefox 44 and Chrome 48.

dmethvin:
Yeah, [this bug fix](https://github.com/jquery/jquery/blob/b43a3685b60b307d61f41f0c94412380ed46ab22/src/attributes/val.js#L85) seems responsible. [Here's the PR and discussion](https://github.com/jquery/jquery/pull/1531). If we pull out that fix it seems like the code you're trying to use may not work on IE9 or 10, but then again we seem to have broken it consistently everywhere with this.

gibson042:
Actually, it looks to me like f6302b0b53d61dfe1adbfaf6612be5cbced5bbc1 was overzealous in expanding 541e7349b6533eb533c15d17e3e9e432e4a719ea to include trimming of _values_ rather than just text. I think a partial revert is in order.

mgol:
Is it really broken, though? From what I see we do the same that Chrome/Firefox does with one exception that we return `null` when no value is selected instead of an empty string: https://jsfiddle.net/m_gol/onnucjx7/3/

gibson042:
[...]

Issue #2886, opened 2016-01-29
Consider increasing modularization of test entities


markelog:
See https://github.com/jquery/jquery/pull/2810#discussion_r49638025

dmethvin:
I agree we could use better correspondence between modules and have better focus with test markup. For a lot of the project we still use a big messy global document that is really hard to change because it risks breaking tests elsewhere. This would be a great PR for someone who wanted to learn about the jQuery project structure. It's mostly rearranging code but you'd learn a lot in the process.

markelog:
I'd like to improve those show/hide tests, if that approach would be acceptable, we can consider writing tests like that from now on

gibson042:
I'm already concerned that those highly related tests span 240 lines, and am not sure that separating them will help things. _However_, QUnit 1.20 supports [nested modules via callback to `module`](http://api.qunitjs.com/QUnit.module/), which would assuage those concerns. So... have at it!

markelog:
> I'm already concerned that those highly related tests span 240 lines, and am not sure that separating them will help things.

It does for every other project, according to many books on the subject and well, other projects code. Also, i don't think it matter how many lines of code it has, i think it matters how well it could be understood and supported.

--

I would like to note, that this ticket not only proposes to rewrite some of the "bad" parts, but i also suggests to write in that manner for every new test.

gibson042:
I would like to refer to https://github.com/jquery/jquery/issues/2880#issuecomment-177578378 as a distinct counterexample. It's fine if we're sufficiently [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), but it's a nightmare while we're not. The problem is "highly related tests" without sufficient abstraction, of which I'm guilty as recently as the tests in question here (but in my defense, we hadn't yet upgraded to a version of QUnit that encourages it).

markelog:
> It's fine if we're sufficiently DRY, but it's a nightmare while we're not

So on one side of the spectre it is fine and on the other it is a nightmare :)

> The problem is "highly related tests" without sufficient abstraction

That an issue too, related, but a bit a different kind i'd say.

> The problem is "highly related tests" without sufficient abstraction, of which I'm guilty as recently as the tests in question here (but in my defense, we hadn't yet upgraded to a version of QUnit that encourages it).

Actually, you could do http://api.qunitjs.com/QUnit.module/#example-0 for long time now, new way is more convenient/explicit but seems it doesn't provide much more to existing functionality.

[...]

PR #3002, opened 2016-03-15
Attributes: strip/collapse whitespace for set values on selects

timmywil:
Fixes gh-2978

### DIFF
--- a/src/attributes/val.js
+++ b/src/attributes/val.js
@@ -5,7 +5,7 @@ define( [
 ], function( jQuery, support ) {
 
 var rreturn = /\r/g,
-	rspaces = /\s/g;
+	rspaces = /[\x20\t\r\n\f]+/g;
 
 jQuery.fn.extend( {
 	val: function( value ) {
--- a/src/attributes/val.js
+++ b/src/attributes/val.js
@@ -81,9 +81,15 @@ jQuery.extend( {
 		option: {
 			get: function( elem ) {
 
-				// Support: IE<11
-				// option.value not trimmed (#14858)
-				return jQuery.trim( elem.value );
+				var val = jQuery.find.attr( elem, "value" );
+				return val != null ?
+					val :
+
+					// Support: IE10-11+
+					// option.text throws exceptions (#14686, #14858)
+					// Strip and collapse whitespace
+					// https://html.spec.whatwg.org/#strip-and-collapse-whitespace
+					jQuery.trim( jQuery.text( elem ) ).replace( rspaces, " " );
 			}
 		},
 		select: {
@@ -129,18 +135,13 @@ jQuery.extend( {
 			set: function( elem, value ) {
 				var optionSet, option,
 					options = elem.options,
-
-					// Strip and collapse whitespace to normalize set values
-					// https://html.spec.whatwg.org/#strip-and-collapse-whitespace
-					values = jQuery.map( jQuery.makeArray( value ), function( val ) {
-						return jQuery.trim( val ).replace( rspaces, " " );
-					} ),
+					values = jQuery.makeArray( value ),
 					i = options.length;
 
 				while ( i-- ) {
 					option = options[ i ];
 					if ( option.selected =
-							jQuery.inArray( jQuery.valHooks.option.get( option ), values ) > -1
+						jQuery.inArray( jQuery.valHooks.option.get( option ), values ) > -1
 					) {
 						optionSet = true;
 					}
--- a/src/attributes/val.js
+++ b/src/attributes/val.js
@@ -4,7 +4,8 @@ define( [
 	"../core/init"
 ], function( jQuery, support ) {
 
-var rreturn = /\r/g;
+var rreturn = /\r/g,
+	rspaces = /\s/g;
 
 jQuery.fn.extend( {
 	val: function( value ) {
@@ -128,7 +129,12 @@ jQuery.extend( {
 			set: function( elem, value ) {
 				var optionSet, option,
 					options = elem.options,
-					values = jQuery.makeArray( value ),
+
+					// Strip and collapse whitespace to normalize set values
+					// https://html.spec.whatwg.org/#strip-and-collapse-whitespace
+					values = jQuery.map( jQuery.makeArray( value ), function( val ) {
+						return jQuery.trim( val ).replace( rspaces, " " );
+					} ),
 					i = options.length;
 
 				while ( i-- ) {
--- a/test/unit/attributes.js
+++ b/test/unit/attributes.js
@@ -1107,6 +1107,48 @@ QUnit.test( "val(select) after form.reset() (Bug #2551)", function( assert ) {
 	jQuery( "#kk" ).remove();
 } );
 
+QUnit.test( "select.val(space characters) (gh-2978)", function( assert ) {
+	assert.expect( 15 );
+
+	var $select = jQuery( "<select/>" ).appendTo( "#qunit-fixture" ),
+		spaces = {
+			"\\t": {
+				html: "&#09;",
+				val: "\t"
+			},
+			"\\n": {
+				html: "&#10;",
+				val: "\n"
+			},
+			"\\r": {
+				html: "&#13;",
+				val: "\r"
+			},
+			"\\f": "\f",
+			"space": " "
+		},
+		html = "";
+	jQuery.each( spaces, function( key, obj ) {
+		var value = obj.html || obj;
+		html += "<option value='bar" + value + "'>bar" + value + "</option>";
+		html += "<option value='te" + value + "st'>te" + value + "st</option>";
+		html += "<option value='" + value + "baz'>" + value + "baz</option>";
+	} );
+	$select.html( html );
+
+	jQuery.each( spaces, function( key, obj ) {
+		var val = obj.val || obj;
+		$select.val( "bar" + val );
+		assert.equal( $select.val(), "bar", "Value ending with space character (" + key + ") selected" );
+
+		$select.val( "te" + val + "st" );
+		assert.equal( $select.val(), "te st", "Value with space character (" + key + ") in the middle selected" );
+
+		$select.val( val + "baz" );
+		assert.equal( $select.val(), "baz", "Value starting with space character (" + key + ") selected" );
+	} );
+} );
+
 var testAddClass = function( valueObj, assert ) {
 	assert.expect( 9 );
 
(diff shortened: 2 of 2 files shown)


src/attributes/val.js
```diff
@@ -4,7 +4,8 @@ define( [
 	"../core/init"
 ], function( jQuery, support ) {
 
-var rreturn = /\r/g;
+var rreturn = /\r/g,
+	rspaces = /\s/g;
```
gibson042: [HTML whitespace](https://html.spec.whatwg.org/#space-character) is a smaller set than that matched by `\s`, and this doesn't account for longer runs. I think you want `/[\x20\t\r\n\f]+/g`, matching various other regular expressions in our codebase.

timmywil: Ha, I was just going off your fiddle.

src/attributes/val.js
```diff
@@ -128,7 +129,12 @@ jQuery.extend( {
 			set: function( elem, value ) {
 				var optionSet, option,
 					options = elem.options,
-					values = jQuery.makeArray( value ),
+
+					// Strip and collapse whitespace to normalize set values
+					// https://html.spec.whatwg.org/#strip-and-collapse-whitespace
+					values = jQuery.map( jQuery.makeArray( value ), function( val ) {
+						return jQuery.trim( val ).replace( rspaces, " " );
+					} ),
```
gibson042: I'm confused... what is the point of this? Fixing gh-2978 requires an update to `valHooks.option.get`, not to `valHooks.select.set`: https://github.com/jquery/jquery/issues/2978#issuecomment-194447965 . Specifically, strip and collapse whitespace is necessary to get the value of an option when it has no `value` attribute, and AFAICT nowhere else.

timmywil: The value coming in has to match the value coming out. `get` seems to be doing the right thing already.

test/unit/attributes.js
```diff
@@ -1107,6 +1107,48 @@ QUnit.test( "val(select) after form.reset() (Bug #2551)", function( assert ) {
...
+		html += "<option value='bar" + value + "'>bar" + value + "</option>";
+		html += "<option value='te" + value + "st'>te" + value + "st</option>";
+		html += "<option value='" + value + "baz'>" + value + "baz</option>";
+	} );
+	$select.html( html );
+
+	jQuery.each( spaces, function( key, obj ) {
+		var val = obj.val || obj;
+		$select.val( "bar" + val );
+		assert.equal( $select.val(), "bar", "Value ending with space character (" + key + ") selected" );
```
gibson042: This assertion is incorrect (as are the others). `$select.val()` should equal `"bar" + val`, because that is the value explicitly specified by the `value` attribute of the option element.

src/attributes/val.js
timmywil: Note the new tests to confirm. `elem.value` does indeed return text when there is no `value` content attribute.

test/unit/attributes.js
```diff
@@ -1523,17 +1565,22 @@ QUnit.test( "option value not trimmed when setting via parent select", function(
...
+QUnit.test( "Insignificant white space returned for $(option).val() (#14858, gh-2978)", function( assert ) {
+	assert.expect( 16 );
 
 	var val = jQuery( "<option></option>" ).val();
 	assert.equal( val.length, 0, "Empty option should have no value" );
 
-	val = jQuery( "<option>  </option>" ).val();
-	assert.equal( val.length, 0, "insignificant white-space returned for value" );
+	jQuery.each( [ " ", "\n", "\t", "\f", "\r" ], function() {
+		var val = jQuery( "<option>" + this + "</option>" ).val();
```
gibson042: `this` abuse!

test/unit/attributes.js
```diff
@@ -1523,17 +1565,22 @@ QUnit.test( "option value not trimmed when setting via parent select", function(
...
+		assert.equal( val.length, 0, "insignificant white-space returned for value" );
 
-	val = jQuery( "<option>  test  </option>" ).val();
-	assert.equal( val.length, 4, "insignificant white-space returned for value" );
+		val = jQuery( "<option>" + this + "test" + this + "</option>" ).val();
+		assert.equal( val.length, 4, "insignificant white-space returned for value" );
```
gibson042: Why not `assert.equal( val, "test", … )`?

test/unit/attributes.js
timmywil: So, `text` should be stripped and collapsed, but `value` shouldn't? Taking out that trim might be a big break.

test/unit/attributes.js
timmywil: I just kept the old. It's accomplishing the same thing.

test/unit/attributes.js
timmywil: ?

gibson042:
I'd like to see some more testing, both for non-HTML whitespace included in `\s` and/or `jQuery.trim()`, but more importantly proving a fix for the _real_ regression revealed by gh-2978, e.g.:
```js
assert.equal(
	jQuery( "<option value='&#13;foo&#9;bar&#10;'>foo bar</option>" ).val(),
	"\rfoo\tbar\n"
);
```

test/unit/attributes.js
gibson042: Everywhere else we `each`/`map`/etc. over primitive values, we refer to them by callback arguments rather than the `Object`-boxed `this`.

test/unit/attributes.js
gibson042: Correct. https://html.spec.whatwg.org/multipage/forms.html#concept-option-value specifies the `value` IDL attribute as the `value` content attribute, falling back on the [`text` IDL attribute](https://html.spec.whatwg.org/multipage/forms.html#dom-option-text), which is the element's content, concatenated, with whitespace stripped and collapsed.

timmywil: Nvm, I see that's what we were doing before.

timmywil:
@gibson042 The regression was not the only thing reported, but I see you're right about not stripping/collapsing the value.

src/attributes/val.js
gibson042: To summarize, `jQuery.valHooks.select.set` should be left alone (the changes here reverted), but `jQuery.valHooks.option.get` as it exists in master (and here) has two problems:
* It is erroneously trimming `value` when derived from a content attribute.
* It is misimplementing [strip and collapse whitespace](https://html.spec.whatwg.org/multipage/infrastructure.html#strip-and-collapse-whitespace) when `value` is not derived from a content attribute.

timmywil:
@gibson042 "non-space" whitespace tests added, and content attr/text behavior separated. This still isn't perfect as jQuery.trim is stripping more than we want, but perhaps it's not worth being that nitpicky.

timmywil:
I was thinking we could normalize setting the value whether or not the option value was from a content attribute or text, but after the adjustments, I think we'd need a 2nd `.inArray` check, and I don't think it's worth it.

test/unit/attributes.js
timmywil: [Not](https://github.com/jquery/jquery/blob/master/src/serialize.js#L74) [everywhere](https://github.com/jquery/jquery/blob/master/src/attributes/val.js#L154), but I think I agree that naming what I'm after is better.

gibson042:
> This still isn't perfect as jQuery.trim is stripping more than we want, but perhaps it's not worth being that nitpicky.

I get what you're saying, but I just checked, and think _all_ our internal uses of `jQuery.trim` should actually use HTML's "strip and collapse whitespace", because there are other latent edge case bugs ([example](https://jsfiddle.net/5Lhfgd3r/)). Note also that we're [already close](https://github.com/jquery/jquery/blob/a4d16a26abe3ea36337f7bfc4d1bfe13b893cd17/src/attributes/classes.js#L8), so taking the final step probably won't be that big a hit:
```js
var rhtmlSpace = /[\x20\t\r\n\f]+/g,
	stripAndCollapse = function( match, pos, str ) {
		return pos === 0 || pos === str.length - match.length ? "" : " ";
	};
…
// formerly jQuery.trim( value )
value.replace( rhtmlSpace, stripAndCollapse );
```
or
```js
var rhtmlSpace = /[\x20\t\r\n\f]+/g;
…
// formerly jQuery.trim( value )
(" " + value + " ").replace( rhtmlSpace, " " ).slice( 1, -1 );
```

test/unit/attributes.js
```diff
@@ -1107,6 +1107,71 @@ QUnit.test( "val(select) after form.reset() (Bug #2551)", function( assert ) {
 	jQuery( "#kk" ).remove();
 } );
 
+QUnit.test( "select.val(space characters) (gh-2978)", function( assert ) {
```
markelog: I guess the old way of writing tests stays forever, oh well :/

timmywil: Well, there's the old way, the really old way, and the ancient way. I suspect you're talking about a newer way?

timmywil:
I'd like to finish this one up today or tomorrow. @gibson042 Would you say this is good to go for now and we can address the #3003 later?

test/unit/attributes.js
timmywil: Ah, I remember now you're talking about #2886. I guess I just stuck with what I was used to.

gibson042:
> @gibson042 Would you say this is good to go for now and we can address the #3003 later?

:+1:

test/unit/attributes.js
markelog: I'm just sad about this, you know? We write tests like we did 10 years ago, tests should be obvious, stateless, and isolated, now it is just ugly, like -
```js
assert.equal( $select.val(), "attr" + val, "Value ending with space character (" + key + ") selected (attr)" );
```
I can't isolate that, i don't understand what is `attr` or `key` and that thing depend on other actions, while violating every standard there is about testing.

Like ember project uses the same old QUnit, but they write them in much more structured way, like compare - https://github.com/emberjs/ember.js/blob/ed298ced885b2f4b4435fbccb26b24b9a24bb68f/tests/node/app-boot-test.js
with what we have here.

timmywil: I guess I don't see the similarity in purpose between those `app-boot` tests and this. Looks like they have lots of helpers and they are establishing that HTML is generated as expected, or that certain callbacks are called/not called. We have some tests in ajax [that are similar](https://github.com/jquery/jquery/blob/master/test/unit/ajax.js#L40). I think it would help me if I could see how you'd like to write some of our existing tests to compare something closer to home.

timmywil: That said, I'll land this for now so we can get 1.12.2/2.2.2 out, but I'm still interested in what you're proposing.

markelog: Yeah, ajax looks better, i have shown that [before](https://github.com/jquery/jquery/pull/2810#discussion_r49638025), right now though, i'm low on motivation about this, since most of the team don't see as importance in it, which is a shame.

dmethvin: Cleaning up the tests would be great, @markelog, I don't think anyone disagrees although we may not see the same degree of urgency. Are you looking for someone else on the team to show that it is important by taking it on? We all have limited resources to spend here. It seems like #2886 stalled waiting for you to give a good example of what a refactor would look like. If we had such an example and it were a "good project for a beginner" kind of thing we could keep a ticket open for that, but it probably will take someone with experience to do it.

As I said in #2886 if it's not a clear bug or feature, I'd prefer to *not* have an open ticket and instead put it on our wish list in the roadmap as @timmywil has been doing lately. We can create a ticket when we have a volunteer and a plan. Long lists of "it works now but wouldn't it be nice to do this one day" tickets are often demotivating, at least for me.

markelog: > It seems like #2886 stalled waiting for you to give a good example of what a refactor would look like

I was waiting for the @mgol changes on removal tests for unsupported environments, that happened only very recently, i was planning to do that after the show/hide changes, which now never come.

> not see the same degree of urgency

The more tests we write the harder it is to change things, i would advise to check out some articles about tests writing technics (i can come with some if anyone would need that), because this should feel important, personally, i think tests sometimes even more important then the code itself.

All and all, i'm not trying to be preachy about these things without doing anything about it but it is kinda hard to fight for it when it seems no one sees why this should matter that much, like am tenager who is never understood by their parents or some stuff like that.

Issue #3005, opened 2016-03-16
$.param() returns "null" and "undefined" literal if attribute is function


jtrumbull:
```javascript
var param = $.param({
  a: function () {},
  b: function () { return null; }
});
```
Will result in `a=undefined&b=null`

[jsfiddle](https://jsfiddle.net/joeycakes/s3txrgfu/1/)

The `value == null` check on [Line 61](https://github.com/jquery/jquery/blob/master/src/serialize.js#L61) should happen after the ternary statement, since both `value` and the result of `value()` can be null. i.e.

```javascript
		add = function( key, value ) {

			// If value is a function, invoke it and return its value
			value = jQuery.isFunction( value ) ? value() : value;
			value = value == null ? "" : value;
			s[ s.length ] = encodeURIComponent( key ) + "=" + encodeURIComponent( value );
		};
```

gibson042:
Makes sense to me. Would you like to submit a PR, @jtrumbull?

jtrumbull:
@gibson042 Yeah, just wanted to confirm that there wasn't reasoning behind the current setup.

dmethvin:
This is a breaking change so it either needs to land ASAP or wait for 4.0.

mgol:
The PR looks good to me. Should we just land it?

dmethvin:
:+1: from me

gibson042:
I don't think this is a breaking change, since "the return value of a function is used instead of the function as a String" still holds (http://api.jquery.com/jQuery.param/ ). But please make an issue or PR in https://github.com/jquery/api.jquery.com if you disagree.

mgol:
Hmm, OK. But maybe it's still best to not backport this to 1.12/2.2.

gibson042:
I'm fine with that.

dmethvin:
I see your point, let's call this a 3.0 feature/fix.

PR #3007, opened 2016-03-17
Serialize: set empty string if value function returns null

jtrumbull:
https://github.com/jquery/jquery/issues/3005

### DIFF
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -59,7 +59,9 @@ jQuery.param = function( a, traditional ) {
 
 			// If value is a function, invoke it and return its value
 			value = jQuery.isFunction( value ) ? value() : value;
-			value = value == null ? "" : value;
+			if (value == null) {
+				value = "";
+			}
 			s[ s.length ] = encodeURIComponent( key ) + "=" + encodeURIComponent( value );
 		};
 
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -58,7 +58,8 @@ jQuery.param = function( a, traditional ) {
 		add = function( key, value ) {
 
 			// If value is a function, invoke it and return its value
-			value = jQuery.isFunction( value ) ? value() : ( value == null ? "" : value );
+			value = jQuery.isFunction( value ) ? value() : value;
+			value = value == null ? "" : value;
 			s[ s.length ] = encodeURIComponent( key ) + "=" + encodeURIComponent( value );
 		};
 
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -59,7 +59,7 @@ jQuery.param = function( a, traditional ) {
 
 			// If value is a function, invoke it and return its value
 			value = jQuery.isFunction( value ) ? value() : value;
-			if (value == null) {
+			if ( value == null ) {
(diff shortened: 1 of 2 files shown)


mgol:
@jtrumbull Thanks for the PR! This will need unit tests to prove that this PR fixes the linked issue. Tests for this module are located in https://github.com/jquery/jquery/blob/master/test/unit/serialize.js.

jtrumbull:
@mgol Full test suite is passing.

test/unit/serialize.js
```diff
@@ -73,13 +73,16 @@ QUnit.test( "jQuery.param()", function( assert ) {
...
+	params = { "param1": function() {}, "param2": function() { return null; } };
+	assert.equal( jQuery.param( params, false ), "param1=&param2=", "object with method properties return null" );
+	
 	params = { "test": { "length": 3, "foo": "bar" } };
 	assert.equal( jQuery.param( params, false ), "test%5Blength%5D=3&test%5Bfoo%5D=bar", "Sub-object with a length property" );
 
 	params = { "test": [ 1, 2, null ] };
 	assert.equal( jQuery.param( params, false ), "test%5B%5D=1&test%5B%5D=2&test%5B%5D=", "object with array property with null value" );
-
+
```
mgol: Please revert the change in this line. We disallow trailing spaces.

mgol: This applies to a few other changed lines as well.

jtrumbull:
@mgol I fixed the whitepace issue. However, this made me curious as to why an error wasn't being thrown by `jscs`. After looking at the gruntfile, I noticed that it's because only parts of the test directory are being run through. More out of curiosity than anything, do you happen to know why this is?

mgol:
Ideally we'd check every file but those test files are very old, they contain lots of errors so fixing them is not an easy task... We're enabling JSCS file by file but we've got a long way ahead to lint everything.

Source is linted completely, though, only tests are lacking.

src/serialize.js
```diff
@@ -58,7 +58,8 @@ jQuery.param = function( a, traditional ) {
 		add = function( key, value ) {
 
 			// If value is a function, invoke it and return its value
-			value = jQuery.isFunction( value ) ? value() : ( value == null ? "" : value );
+			value = jQuery.isFunction( value ) ? value() : value;
+			value = value == null ? "" : value;
```
gibson042: This may be smaller as `if ( value == null )`.

test/unit/serialize.js
```diff
@@ -74,6 +74,9 @@ QUnit.test( "jQuery.param()", function( assert ) {
 	params = { "param1": null };
 	assert.equal( jQuery.param( params, false ), "param1=", "Make sure that null params aren't traversed." );
 
+	params = { "param1": function() {}, "param2": function() { return null; } };
+	assert.equal( jQuery.param( params, false ), "param1=&param2=", "object with method properties return null" );
```
mgol: `object with method properties return null` -> `Object with method properties that return null`

mgol:
LGTM.

PR #3014, opened 2016-03-23
Remove legacy code from tests, update support comments

mgol:
I divided this PR into 3 commits as it made sense that way but I didn't want to file them separately. Please see commit messages for specific commits to see what was their scope.

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -535,12 +535,13 @@ jQuery.extend( {
 		if ( s.crossDomain == null ) {
 			urlAnchor = document.createElement( "a" );
 
-			// Support: IE8-11+
-			// IE throws exception if url is malformed, e.g. http://example.com:80x/
+			// Support: IE <=8 - 11, Edge 12 - 13
+			// IE throws exception on accessing the href property if url is malformed,
+			// e.g. http://example.com:80x/
 			try {
 				urlAnchor.href = s.url;
 
-				// Support: IE8-11+
+				// Support: IE <=8 - 11 only
 				// Anchor's host property isn't correctly set when s.url is relative
 				urlAnchor.href = urlAnchor.href;
 				s.crossDomain = originAnchor.protocol + "//" + originAnchor.host !==
--- a/src/core.js
+++ b/src/core.js
@@ -274,7 +274,7 @@ jQuery.extend( {
 	},
 
 	// Convert dashed to camelCase; used by the css and data modules
-	// Support: IE9-11+
+	// Support: IE <=9 - 11, Edge 12 - 13
 	// Microsoft forgot to hump their vendor prefix (#9572)
 	camelCase: function( string ) {
 		return string.replace( rmsPrefix, "ms-" ).replace( rdashAlpha, fcamelCase );
--- a/src/core.js
+++ b/src/core.js
@@ -24,7 +24,7 @@ var
 		return new jQuery.fn.init( selector, context );
 	},
 
-	// Support: Android<4.1
+	// Support: Android <=4.0 only
 	// Make sure we trim BOM and NBSP
 	rtrim = /^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g,
 
@@ -262,7 +262,7 @@ jQuery.extend( {
 			return obj + "";
 		}
 
-		// Support: Android<4.0 (functionish RegExp)
+		// Support: Android <=2.3 only (functionish RegExp)
 		return typeof obj === "object" || typeof obj === "function" ?
 			class2type[ toString.call( obj ) ] || "object" :
 			typeof obj;
@@ -305,7 +305,7 @@ jQuery.extend( {
 		return obj;
 	},
 
-	// Support: Android<4.1
+	// Support: Android <=4.0 only
 	trim: function( text ) {
 		return text == null ?
 			"" :
@@ -334,7 +334,7 @@ jQuery.extend( {
 		return arr == null ? -1 : indexOf.call( arr, elem, i );
 	},
 
-	// Support: Android<4.1, PhantomJS<2
+	// Support: Android <=4.0 only, PhantomJS 1 only
 	// push.apply(_, arraylike) throws on ancient WebKit
 	merge: function( first, second ) {
 		var len = +second.length,
@@ -458,7 +458,7 @@ function( i, name ) {
 
 function isArrayLike( obj ) {
 
-	// Support: iOS 8.2 (not reproducible in simulator)
+	// Support: real iOS 8.2 only (not reproducible in simulator)
 	// `in` check used to prevent JIT error (gh-2145)
 	// hasOwn isn't used here due to false negatives
 	// regarding Nodelist length in IE
--- a/src/css/var/getStyles.js
+++ b/src/css/var/getStyles.js
@@ -1,7 +1,7 @@
 define( function() {
 	return function( elem ) {
 
-		// Support: IE<=11 only, Firefox<=30+ (#15098, #14150)
+		// Support: IE <=11 only, Firefox <=30 (#15098, #14150)
 		// IE throws on elements created in popups
 		// FF meanwhile throws on frame elements through "defaultView.getComputedStyle"
 		var view = elem.ownerDocument.defaultView;
--- a/src/deferred/exceptionHook.js
+++ b/src/deferred/exceptionHook.js
@@ -9,7 +9,7 @@ var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
 
 jQuery.Deferred.exceptionHook = function( error, stack ) {
 
-	// Support: IE9 only
+	// Support: IE 8 - 9 only
 	// Console exists when dev tools are open, which can happen at any time
 	if ( window.console && window.console.warn && error && rerrorNames.test( error.name ) ) {
 		window.console.warn( "jQuery.Deferred exception: " + error.message, stack );
--- a/src/effects.js
+++ b/src/effects.js
@@ -142,7 +142,7 @@ function defaultPrefilter( elem, props, opts ) {
 	// Restrict "overflow" and "display" styles during box animations
 	if ( isBox && elem.nodeType === 1 ) {
 
-		// Support: IE 9 - 11
+		// Support: IE <=9 - 11, Edge 12 - 13
 		// Record all 3 overflow attributes because IE does not infer the shorthand
 		// from identically-valued overflowX and overflowY
 		opts.overflow = [ style.overflow, style.overflowX, style.overflowY ];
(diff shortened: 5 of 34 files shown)


src/core.js
```diff
@@ -262,7 +262,7 @@ jQuery.extend( {
 			return obj + "";
 		}
 
-		// Support: Android<4.0 (functionish RegExp)
+		// Support: Android 2.3 only (functionish RegExp)
```
gibson042: I think all lines like these, including "IE 9 only" and "IE 9 - 11 only", should include the initial `<=`. Unless, of course, they are confirmed to _not_ be relevant in e.g. Android 2.2 or IE 8. I wouldn't ask except that it's part of the reason for this PR.

mgol: Makes sense, I'll apply those where needed (which is almost everywhere).

src/css/var/getStyles.js
```diff
@@ -1,7 +1,7 @@
 define( function() {
 	return function( elem ) {
 
-		// Support: IE<=11 only, Firefox<=30+ (#15098, #14150)
+		// Support: IE <=11 only, Firefox <=30+ (#15098, #14150)
```
gibson042: `Firefox <=30+` isn't valid. If the issue still exists, it should be `Firefox <=30 - 45`.

src/event.js
```diff
@@ -347,11 +347,11 @@ jQuery.event = {
 			delegateCount = handlers.delegateCount,
 			cur = event.target;
 
-		// Support (at least): Chrome, IE9
+		// Support (at least): IE 9
```
gibson042: Might as well drop the parenthetical here.

src/event.js
```diff
@@ -469,7 +469,7 @@ jQuery.event = {
 			event[ prop ] = originalEvent[ prop ];
 		}
 
-		// Support: Safari 6-8+
+		// Support: Safari 6 - 8+
```
gibson042: Drop the `+`.

src/event/focusin.js
```diff
@@ -11,7 +11,7 @@ define( [
 // Firefox doesn't have focus(in | out) events
 // Related ticket - https://bugzilla.mozilla.org/show_bug.cgi?id=687787
 //
-// Support: Chrome 48+, Safari 9.0+
+// Support: Chrome 49+, Safari 9.1+
```
gibson042: `Support: Chrome <=48 - 49, Safari <=9.0 - 9.1`

src/event.js
mgol: Is it fixed in Safari 9? I didn't check that one.

src/manipulation.js
```diff
@@ -292,13 +292,13 @@ jQuery.extend( {
 						}
 					}
 
-					// Support: Chrome <= 35-45+
+					// Support: Chrome <=35 - 45+
```
gibson042: `Chrome <=35 - 45`

src/manipulation.js
```diff
@@ -292,13 +292,13 @@ jQuery.extend( {
...
 					// Assign undefined instead of using delete, see Data#remove
 					elem[ dataPriv.expando ] = undefined;
 				}
 				if ( elem[ dataUser.expando ] ) {
 
-					// Support: Chrome <= 35-45+
+					// Support: Chrome <=35 - 45+
```
gibson042: `Chrome <=35 - 45`

src/css/var/getStyles.js
mgol: I'll need to verify first if it's still broken in current Firefox. I mostly tested IE/Edge in these commits.

src/event.js
mgol: I'll need to first check if Edge/Safari/Firefox are affected, won't I?

src/offset.js
```diff
@@ -203,7 +203,7 @@ jQuery.each( { scrollLeft: "pageXOffset", scrollTop: "pageYOffset" }, function(
 	};
 } );
 
-// Support: Safari<7-9.0+, Chrome<37-48+
+// Support: Safari <=7 - 9.1+, Chrome <=37 - 49+
```
gibson042: Drop the `+` characters.

src/manipulation.js
mgol: Ah, because the proposal drops the pluses.

test/unit/deferred.js
```diff
@@ -535,7 +535,7 @@ QUnit[ window.console ? "test" : "skip" ]( "jQuery.Deferred.exceptionHook", func
 
 	window.console.warn = function( msg ) {
 
-		// Support: Chrome < 42
+		// Support: Chrome <=41
```
gibson042: need ` only`.

test/unit/deferred.js
```diff
@@ -590,7 +590,7 @@ QUnit[ window.console ? "test" : "skip" ]( "jQuery.Deferred.exceptionHook with s
 
 	window.console.warn = function( msg, stack ) {
 
-		// Support: Chrome < 42
+		// Support: Chrome <=41
```
gibson042: Needs ` only`.

test/unit/event.js
```diff
@@ -1382,7 +1382,7 @@ QUnit.test( "Submit event can be stopped (#11049)", function( assert ) {
 	form.remove();
 } );
 
-// Support: iOS 7-9+
+// Support: iOS 7 - 9+
```
gibson042: Drop the `+`.

test/unit/offset.js
```diff
@@ -540,7 +540,7 @@ QUnit.test( "fractions (see #7730 and #7885)", function( assert ) {
 
 	result = div.offset();
 
-	// Support: Chrome 45-46+
+	// Support: Chrome 45 - 46+
```
gibson042: Drop the `+`.

src/ajax.js
```diff
@@ -535,12 +535,13 @@ jQuery.extend( {
 		if ( s.crossDomain == null ) {
 			urlAnchor = document.createElement( "a" );
 
-			// Support: IE8-11+
-			// IE throws exception if url is malformed, e.g. http://example.com:80x/
+			// Support: IE 8 - 11, Edge 12 - 13+
```
gibson042: Drop the `+`.

src/core.js
```diff
@@ -274,7 +274,7 @@ jQuery.extend( {
 	},
 
 	// Convert dashed to camelCase; used by the css and data modules
-	// Support: IE9-11+
+	// Support: IE 9 - 11, Edge 12 - 13+
```
gibson042: Drop the `+`.

src/effects.js
```diff
@@ -142,7 +142,7 @@ function defaultPrefilter( elem, props, opts ) {
 	// Restrict "overflow" and "display" styles during box animations
 	if ( isBox && elem.nodeType === 1 ) {
 
-		// Support: IE 9 - 11
+		// Support: IE 9 - 11, Edge 12 - 13+
```
gibson042: Drop the `+`.

src/manipulation.js
```diff
@@ -29,7 +29,7 @@ define( [
 var
 	rxhtmlTag = /<(?!area|br|col|embed|hr|img|input|link|meta|param)(([a-z][^\/\0>\x20\t\r\n\f]*)[^>]*)\/>/gi,
 
-	// Support: IE 10-11, Edge 10240+
+	// Support: IE 10 - 11, Edge 12 - 13+
```
gibson042: Drop the `+`.

src/event.js
gibson042: It would be valuable, but you don't _need_ to. Support comments document what we know, not what we don't.

src/event.js
gibson042: Doesn't matter. Fixes are documented with ` only`.

gibson042:
Thanks so much for this, @mgol! :clap:

src/event.js
mgol: OK, I dropped it. However, I think there's some value in stating explicitly that a particular scenario wasn't tested in at least the few major desktop browsers we support. Otherwise it's easy to remove the code when support for a particular browser is dropped only to be surprised later.

OTH, 98% of such code should have unit tests so we'd catch that immediately...

mgol:
@gibson042 I think I fixed everything but you never know so PTAL.

src/deferred/exceptionHook.js
```diff
@@ -9,7 +9,7 @@ var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
 
 jQuery.Deferred.exceptionHook = function( error, stack ) {
 
-	// Support: IE9 only
+	// Support: IE 8-9 only
```
gibson042: `IE 8 - 9 only`

src/event.js
```diff
@@ -469,7 +469,7 @@ jQuery.event = {
 			event[ prop ] = originalEvent[ prop ];
 		}
 
-		// Support: Safari 6-8+
+		// Support: Safari <=6 - 7 only
```
gibson042: The behavior doesn't exist in Safari 8?

src/event/focusin.js
```diff
@@ -7,11 +7,11 @@ define( [
 	"./trigger"
 ], function( jQuery, dataPriv, support ) {
 
-// Support: Firefox 44+
+// Support: Firefox <=44+
```
gibson042: Drop the `+`.

src/manipulation.js
markelog: @gibson042 for this rule to became official, you gotta submit pr to the contribute repo. Wouldn't want contributors to follow invisible guidelines

gibson042: https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-69626900 :
> It may turn out that I'm the one to do it, but certainly not before core 3.0.

src/event.js
mgol: Yup, I've checked v8 & v9 and it doesn't break there as it used to.

src/event/focusin.js
mgol: Dropped.

src/deferred/exceptionHook.js
mgol: Fixed.

mgol:
PR rebased over latest master, fixes applied. Hopefully it's ready to land this time.

src/core.js
```diff
@@ -458,7 +458,7 @@ function( i, name ) {
 
 function isArrayLike( obj ) {
 
-	// Support: iOS 8.2 (not reproducible in simulator)
+	// Support: iOS 8.2 only (not reproducible in simulator)
```
gibson042: I think this would be better as `Support: real iOS 8.2 only` or `Support: non-simulated iOS 8.2 only`.

gibson042:
Just the one last comment, @mgol, and a non-blocker at that. :+1: from me!

mgol:
PR updated. If no one opposes, I'd like to land it this weekend.

PR #3015, opened 2016-03-23
Ajax: make jqXHR an instance of jQuery.Deferred

bluekvirus:
So that `$.ajax() instanceof $.Deferred` will be True. This is to enable accurate async grouping (all-done-timing) based on the return object types of individual callbacks. Right now, the returned jqXHR object is of no class at all but does have full promise interfaces. Would like to suggest a class for it to enable a rigid type switching on callback queue returns, please consider. 

I am using 
```
//callback n 
{
   return $.ajax();
} 
```
then
```
result = callback();
if(result && result.done)
  ...
```
now for detecting a return of the jqXHR object.

Cheers,
Tim.

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -444,80 +444,80 @@ jQuery.extend( {
 			// Default abort message
 			strAbort = "canceled",
 
-			// Fake xhr
-			jqXHR = {
-				readyState: 0,
+			// Fake xhr (as a true Deferred instance)
+			jqXHR = deferred.promise();
 
-				// Builds headers hashtable if needed
-				getResponseHeader: function( key ) {
-					var match;
-					if ( completed ) {
-						if ( !responseHeaders ) {
-							responseHeaders = {};
-							while ( ( match = rheaders.exec( responseHeadersString ) ) ) {
-								responseHeaders[ match[ 1 ].toLowerCase() ] = match[ 2 ];
-							}
-						}
-						match = responseHeaders[ key.toLowerCase() ];
-					}
-					return match == null ? null : match;
-				},
+		// Attach xhr capabilities
+		jQuery.extend( jqXHR, {
+			readyState: 0,
 
-				// Raw string
-				getAllResponseHeaders: function() {
-					return completed ? responseHeadersString : null;
-				},
-
-				// Caches the header
-				setRequestHeader: function( name, value ) {
-					if ( completed == null ) {
-						name = requestHeadersNames[ name.toLowerCase() ] =
-							requestHeadersNames[ name.toLowerCase() ] || name;
-						requestHeaders[ name ] = value;
-					}
-					return this;
-				},
-
-				// Overrides response content-type header
-				overrideMimeType: function( type ) {
-					if ( completed == null ) {
-						s.mimeType = type;
-					}
-					return this;
-				},
-
-				// Status-dependent callbacks
-				statusCode: function( map ) {
-					var code;
-					if ( map ) {
-						if ( completed ) {
-
-							// Execute the appropriate callbacks
-							jqXHR.always( map[ jqXHR.status ] );
-						} else {
-
-							// Lazy-add the new callbacks in a way that preserves old ones
-							for ( code in map ) {
-								statusCode[ code ] = [ statusCode[ code ], map[ code ] ];
-							}
+			// Builds headers hashtable if needed
+			getResponseHeader: function( key ) {
+				var match;
+				if ( completed ) {
+					if ( !responseHeaders ) {
+						responseHeaders = {};
+						while ( ( match = rheaders.exec( responseHeadersString ) ) ) {
+							responseHeaders[ match[ 1 ].toLowerCase() ] = match[ 2 ];
 						}
 					}
-					return this;
-				},
-
-				// Cancel the request
-				abort: function( statusText ) {
-					var finalText = statusText || strAbort;
-					if ( transport ) {
-						transport.abort( finalText );
-					}
-					done( 0, finalText );
-					return this;
+					match = responseHeaders[ key.toLowerCase() ];
 				}
-			};
+				return match == null ? null : match;
+			},
 
-		// Attach deferreds
-		deferred.promise( jqXHR );
+			// Raw string
+			getAllResponseHeaders: function() {
+				return completed ? responseHeadersString : null;
+			},
+
+			// Caches the header
+			setRequestHeader: function( name, value ) {
+				if ( completed == null ) {
+					name = requestHeadersNames[ name.toLowerCase() ] =
+						requestHeadersNames[ name.toLowerCase() ] || name;
+					requestHeaders[ name ] = value;
+				}
+				return this;
+			},
+
+			// Overrides response content-type header
+			overrideMimeType: function( type ) {
+				if ( completed == null ) {
+					s.mimeType = type;
+				}
+				return this;
+			},
+
+			// Status-dependent callbacks
+			statusCode: function( map ) {
+				var code;
+				if ( map ) {
+					if ( completed ) {
+
+						// Execute the appropriate callbacks
+						jqXHR.always( map[ jqXHR.status ] );
+					} else {
+
+						// Lazy-add the new callbacks in a way that preserves old ones
+						for ( code in map ) {
+							statusCode[ code ] = [ statusCode[ code ], map[ code ] ];
+						}
+					}
+				}
+				return this;
+			},
+
+			// Cancel the request
+			abort: function( statusText ) {
+				var finalText = statusText || strAbort;
+				if ( transport ) {
+					transport.abort( finalText );
+				}
+				done( 0, finalText );
+				return this;
+			}
+		} );
 
 		// Add protocol if not provided (prefilters might expect it)
 		// Handle falsy url in the settings object (#10093: consistency with old signature)
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -431,7 +431,7 @@ jQuery.extend( {
 					jQuery.event,
 
 			// Deferreds
-			deferred = jQuery.Deferred(),
+			deferred = new jQuery.Deferred(),
 			completeDeferred = jQuery.Callbacks( "once memory" ),
 
 			// Status-dependent callbacks
@@ -444,8 +444,8 @@ jQuery.extend( {
 			// Default abort message
 			strAbort = "canceled",
 
-			// Fake xhr (as a true Deferred instance)
-			jqXHR = deferred.promise();
+			// Fake xhr (ref the true Deferred instance above)
+			jqXHR = deferred;
 
 		// Attach xhr capabilities
 		jQuery.extend( jqXHR, {
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -301,7 +301,7 @@ jQuery.extend( {
 		}
 
 		// All done!
-		return deferred;
+		return this === jQuery ? deferred : jQuery.extend( this, deferred );
 	},
 
 	// Deferred helper
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -1966,6 +1966,11 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
 		};
 	} );
 
+	QUnit.test( "jQuery.ajax() returns jqXHR as an instance of jQuery.Deferred", function( assert ) {
+		assert.expect( 1 );
+		assert.ok( jQuery.ajax( "data/json_obj.js" ) instanceof jQuery.Deferred, "jqXHR should be an instance of jQuery.Deferred." );
+	} );
+
 //----------- jQuery.ajaxPrefilter()
 
 	ajaxTest( "jQuery.ajaxPrefilter() - abort", 1, function( assert ) {


bluekvirus:
Creating a new one since the signing email isn't changing after making profile email public...come on github...:(

bluekvirus:
Reopen to try refresh email signing.

dmethvin:
@bluekvirus you just need to make sure that the signature exactly matches the one you used in your commit. Same name and email. Click the failed test for more info. You don't need to close the PR. If you need to update your Github info just `commit --amend` and force-push the PR.

bluekvirus:
@dmethvin thanks, reopened. Also, I pulled my patch branch and saw the first 2 commits still under the private @users.noreply.github.com email, any quick way of changing these?

dmethvin:
Looks like progress! The latest commit has a valid address but the first two do not. What you can do is `git rebase -i HEAD~3` to squash those three commits, check that the email is right on that commit, use `commit --amend --author="Your name <your@email.com>"` if needed to fix it, and  then `push --force patch-1`. Does that all make sense?

bluekvirus:
@dmethvin thanks a lot for the prompt guidance. Commits squashed and re-signed with public email.

gibson042:
Thanks for the pull request, @bluekvirus. I'm not exactly clear on what you're trying to do here... can you add a unit test to demonstrate?

bluekvirus:
@gibson042 thanks for review, the change I'm proposing is to let jqXHR be of class Deferred, so that we can use `instanceof` to detect the object as any other promise objects if it is returned by possible application initializer callbacks. The current implementation create the jqXHR first as a plain object and use Deferred.promise() to patch it with promise apis. 

In the pull request I altered the process to be getting an empty Deferred object as jqXHR first and use $.extend() to patch it with the old jqXHR object code. A possible unit test could be as the following:
```
QUnit.test( "jQuery.ajax() returns jqXHR as an true instance of jQuery.Deferred", function( assert ) {
   assert.expect( 1 );
   assert.ok( jQuery.ajax('data/json_obj.js') instanceof jQuery.Deferred, "jqXHR should be a true instance of jQuery.Deferred." );
} );
``` 

It is quite important for me as an SPA developer for embedded and mobile devices to have the freedom of crafting special customized asynchronize application boot-up sequence. Having jqXHR as true jQuery.Deferred objects greatly enhance the tidiness of the code especially when mixing $.ajax() calls with other promise based workers. Thanks in advance for considering.

dmethvin:
There should definitely be some unit tests added to ensure it does what it claims to do.

The design of `jQuery.Deferred` allows a usage where arbitrary objects can be "deferrified" if you will, but I wouldn't object to having `jQuery.ajax` return an actual instance of a `jQuery.Deferred`. I haven't looked at the changes in depth or thought about what other implications that might have.

gibson042:
That is an appropriate unit test, and if you add it to this code as-is, I think you will find that it fails. There is no such thing as a "true jQuery.Deferred"; [they're _all_ just plain objects](https://github.com/jquery/jquery/blob/aae44111e2d30d908b48f6711272efddb84e6b79/src/deferred.js#L245) with own properties.

bluekvirus:
@gibson042 thanks for pointing out, I was assuming that jQuery.Deferred would be a properly defined pseudo-class so didn't even check... I am a bit disappointed to find out about it now. It seems that a more concrete patch is needed if we are to pursuit this thread.

bluekvirus:
Further modified 3 lines of codes to employ the following idea for this patch to complete, let B be jQuery (luckily), X be Deferred and p as deferred in our case:
```
B.X = function(options){
   var p = {a:1, b:$.noop}; 
   if(this === B) return p; 
   return $.extend(this, p); 
};
```
Test result:
```
var x = B.X();
// => x is Object { a=1,  b=function()}

var y = new B.X()
// => y is Object { a=1,  b=function()}

x instanceof B.X
// => false

y instanceof B.X
// => true
```

I have already modified the patch and submitted together with the unit test case. Let me know if a tidy up squashing is needed. (2 commits wasted on the style check)

src/deferred.js
```diff
@@ -301,7 +301,7 @@ jQuery.extend( {
 		}
 
 		// All done!
-		return deferred;
+		return this === jQuery ? deferred : jQuery.extend( this, deferred );
```
gibson042: Testing like `this instanceof jQuery.Deferred` would be more in line with our code style ([example](https://github.com/jquery/jquery/blob/9b086888b803db872e8906cbdbf0de458231f225/src/event.js#L547)), though honestly I'd be happier if all our auto-constructors were named rather than anonymous (e.g., `Deferred: function Deferred( func ) { … }`).

src/ajax.js
```diff
@@ -444,80 +444,80 @@ jQuery.extend( {
 			// Default abort message
 			strAbort = "canceled",
 
-			// Fake xhr
-			jqXHR = {
-				readyState: 0,
+			// Fake xhr (ref the true Deferred instance above)
+			jqXHR = deferred;
```
gibson042: I'm sorry I missed this earlier, but the reason we're returning a promise right now is that Deferred objects have private methods (`resolveWith`, `reject`, `notify`, etc.). It's not acceptable to return an object that can circumvent the normal flow (e.g., `var jqXHR = jQuery.ajax( url ); jqXHR.resolve( fakeData, fakeStatus, jqXHR )`), and I'm not sure the benefits of this ticket are worth the cost of manually deleting them.

bluekvirus:
Deleting the extra api of a Deferred object to form a Promise object is probably not a good idea, it also shouldn't be used as an excuse for creating a headless object (the `jqXHR` unfortunately) with patched api like a tmp work around. 

After all, the purpose of this ticket is to prevent the un-typed promise object (the `jqXHR` object) from running around in 3.0, this kind of ugliness should be rectified if jQuery were to be the goto DOM bridger that powers IoT and mobile device UIs (those with screens of course). Promise objects have a general type in each modern browser, maybe named differently but if jqXHR by definition is of the Promise type it should pass the `instanceof` SomeClass type check so callback chains can be formed with both sync and async routines without checking `result && result.always` for each of the async returns like some kind of workaround of another workaround... 

The $.Deferred was never a properly defined pseudo class in the first place, using it to hack out a promise object is kind of brilliant and stupid at the same time. While a few lines of code in this ticket could potentially extend its life-time in a backward compatible way, I would rather see it gets rewritten. 

As for the public or private apis debate, I thought we were exempted from it from 1995 since this is Javascript we are talking about... But again, like I said in the beginning, manual deletion for cleaning up a hack or a hack is never acceptable if time is not a serious concern in a project. 

So, what's it gonna be? Give jqXHR a properly groomed head or let it be a rogue in the promised object domain? (similar to ajax, device status, peripheral i/o control and system api calls can all be promise based on an IoT/Mobile device, and they can all appear in the DOM as pseudo objects just like XMLHttpRequest)

jaubourg:
> Deleting the extra api of a Deferred object to form a Promise object is probably not a good idea, it also shouldn't be used as an excuse for creating a headless object (the jqXHR unfortunately) with patched api like a tmp workaround.

Except the Deferred has a resolve and reject method meaning you cannot safely pass it around to code that shouldn't have the capacity of resolving or rejecting the underlying promise. If anything, the mistake in the original implementation was to put the Promise (ie read-only) methods on the Deferred object in the first place.

> [the Promise type] should pass the instanceof SomeClass type check so callback chains can be formed with both sync and async routines without checking result && result.always for each of the async returns like some kind of workaround of another workaround

`$.when( result ).then( /*...*/ )`... problem solved.

> So, what's it gonna be? Give jqXHR a properly groomed head or let it be a rogue in the promised object domain?

Any code that accepts a promise as an input has to accept a thenable (that is an object that has a method called `then`). That's the standard. It does not need to be of a specific class.

dmethvin:
Given that thenables don't have to be a specific class, and that a jQuery promise is a subset of Deferred methods, I doubt this change can happen. Sorry for leading you astray, I didn't understand all the implications until about halfway through and really should have notice it right off.

bluekvirus:
Thanks for the `$.when()` tip. It is not solving the problem but does offer improvement visually. (on cleaner syntax level compare to me explicitly checking .alway() or .then() or wrapping with `Deferred.promise()` on every returned object first regardlessly in that case...)

Again, the purpose of this ticket is to have a way of detecting the promised object (if you want to call it `.then()` objects like A+ that's fine), not to have it merged my way (my hack was just an example). The problem is always with Javascript since we can fake pseudo class to put a type on objects but never really can detect interface/mix-in boosted objects native-ly (not if we always use `$.extend` directly on plain objects to do this). You guys obviously agree to the fact that A+ is an interface, so I'll ask again like this,

1.  What would you do if want to detect that an object has certain interface/mixin without a complete check of available interface/mixin apis on that object? (in JS realm of course)

2. Can we give jqXHR a way of being detected native-ly (through instanceof)? (you gave me `$.when()` which is an api that re-ensures `.then()`... so my object returned from any of the callback in the vector can't have `.then()` if it is not a promise and don't want `.then()` to be called immediately)

For the sake of this argument, please don't ask why I wanted to put a `.then()` in an object which is not a promise, it is not the point... For now it might not be a big deal for promise since it has only 1 api to check and skip, and I can see the kind of hacking mindset is towards privacy of methods instead of declarative typing. 

You are essentially saying that private/public with plain object in jQuery is the way to go forward and for each protected interface/mixin you will have a `$.when()` alike to glue it back in a way that that interface/mixin-ed object is supposed to be used.

@jaubourg Does the *standard* you are looking at demands a hidden A+ object? In every other language that does support interface/mixin, you have a native way of detecting the object type. It is not like *it is not required to be of type Promise*, it is because when you declare Promise as an implemented interface, the objects become of type Promise and their own original type at the same time automatically in that language. 

With Javascript we can't. If using `instanceof` it requires that you have a `new Promise()` in its prototype chain. And you have no way of protecting *private* apis once it is there. If using a check like `$.when()` we have privacy and protection but need one check for each interface... I'd rather we go the first way since we all know that Javascript doesn't have privacy designed in mind at all, you can see that with the free and cheap `this` re-binding routines. Why are we keep saying safe and read-only? 

Neither `new Deferred` nor `.promise()` returns any typed object at the moment and this is 3.0 we are talking about. Asking people to use `$.when()` with Deferred and `.promise()` is really messed up especially with `jqXHR` written as it is (a var name!!) in the online documentation for $.ajax() return. Why would I want to know the name of a return? I need the TYPE!! I know everything is an Object, but didn't I bring up that I need to use `instanceof` in the first place so that I can divert the processing routines accordingly?

Anyway, I'm using it for free, so doesn't really want to complain, just leaving a note here for other people. In case you don't understand where I stand, here are the facts:

a. Like using `$()` to get jQuery objects, you use `$.Deferred()` to get a deferred object, however, the following appears in your console:
```
$('body') instanceof $
true

$.Deferred() instanceof $.Deferred
false
```
The deferred object has a super set of A+ promise apis.

b. There is no $.Promise, neither is the native Promise in your browser used, call .promise() from a deferred object to get a promise object. or just use `$.when(o)` or `$.Deferred.promise(o)` to return a wrap. Either way the returned objects are of type Object. You will see the word Promise as the return object written in the documentation, that's just for your viewing pleasure. Also you can read this 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

in the light that it will confuse you even more as what the `$.Deferred`, `$.Deferred.promise()` and `$.when()` is trying to do when everything could have just use Promise itself in v3.0, also I have tried to convince the developers to just use `$.Deferred` as a proper class like the native Promise but no, they refused, they love having 3 different things around to feel safe and look sophisticated. (Please don't bring up IE11 as we are talking about modern browsers and v3)

c. `jqXHR` returned by `$.ajax` related calls is a plain object, it has A+ interfaces mixed in by `$.Deferred.promise()`, but it is not of type jQuery, `$.Deferred`, nor Promise, is it just an object. use `$.when()` with it if you want to but most of us will just use jqXHR directly without. So don't panic if jqXHR feels like a promise object but is really some hacked object of no type with .then() interfaces, the developers think this is the jQuery way of comply to the A+ standard.

End of note regarding `$.Deferred`, `$.Deferred.promise()` and `$.when()` in an attempt to let jqXHR be of type $.Deferred. (Which should really be just Promise in native JS engines) Please don't think I am in love with `$.Deferred` or jQuery itself so I submitted the ticket. I did it because I came across v3 branch and still see the ugly code in there so I thought I try. Sorry to have disturbed the peaceful minds. I assume that there is no plan of using Promise directly nor fixing $.Deferred in v3 now.

src/ajax.js
bluekvirus: Could you elaborate through when can this **circumvent the normal flow** happen in the browser please? `.resolve` is a perfectly fine Promise API, why would someone call `.resolve` before XMLHttpRequest resolve it for him/her? Assuming the var is always on the call stack (scoped) not globally hackable through a js console.

dmethvin: I think what @gibson042 is saying is that the jQXHR object returned right now does not have the ability to *resolve itself*, whereas if we returned a full `jQuery.Deferred` it would since the object would now have both `.resolve` and `.reject` methods. That violates the encapsulation we created. It also would violate the similar encapsulation provided by ES6 promises, which also cannot self-resolve.

Now you might say, "Who would do that?" but when you have millions of people writing jQuery, someone will do that (either on purpose or by accident) if you give them the ability.

timmywil:
There are 2 patterns when creating objects in JS: factory and instance. We chose the former because it has advantages, like being able to pass around disconnected promise methods without binding them (`e.g. done(promise.then);`). The instance path also has advantages, but we don't think there are enough advantages to make a switch. I don't think it's as confusing as you have made it seem. It's just JavaScript. Did you know `null` is of type `object`?

src/ajax.js
bluekvirus: I understand what he is trying to say, and you are restating that jQuery is trying to protect ppl from stepping on their own foot, or screw someone else's code when loaded together. I, of course, understand and appreciate that concept and I will probably want to do the same whenever I can. The question is why and to what extend. Like if you have C pointers, you can kill your own stack, or in the worst case all the other programs if your OS doesn't protect it. In browser, your tab JS engine is the OS, and JS is C, jQuery is probably some relevant lib. Now, you see what I'm asking right? Why would a lib want to *protect* when it is only providing APIs and unit routines?

But again, the argument is based on my way of giving both Deferred and promise the class $.Deferred when there is no other options in the jQuery code base that I can use to achieve my goal of giving jqXHR a type. So based on my implementation, I ask @gibson042 to give me an example of a hackable situation, not examples of programmers stepping on their own foot. Like if every Chef holding a knife kills him/herself then we are in real trouble. Our purpose and caution when using a tool would be the only guard there, not the tool.

The JS language was built without protection in the first place, I think clear typing is far more important than passing plain *protected* objects around, especially when it comes to Promise. Promise is an important contract, the object that has that contract should have that type. Give me one counter example in good OO standard that's against it then I'll shut up.

A language itself should protect enough and leave room for the programmers, a library can only do so much about additional APIs. And in APIs we need to be clear about returns. This is why we want to hook up *your* APIs in the first place and why we are all in the Open Source community. We communicate clearly to each other, and provide reusable routines with meaningful interfaces (both in fn signature and return types). Not hidden caveat or some hacked policy when there is clearly a better way of interfacing. 

But again, returning a chopped object with minimum api is OK if you merge it with a type first in jQuery, so I do hope you consider giving all the Deferred and promise a proper type in v3. If $.Deferred is considered then having exposed `resolve` and `reject` shouldn't be the reason to drop it, since they are loved by our testers & devs when faking ajax calls in a server-less UI dev setup.

bluekvirus:
Thanks @timmywil, that's a fair comment. I respect the choices made and agree that JS is poorly made. I love JS objects just as much as you guys do, but sometimes there should be limit. That's all I'm saying. It's v3, I really hope we can get out of this plain type factory pattern and have some clearly defined interfacing contract types. @dmethvin Brought up ES6 in the discussion, and I think it is worth mentioning because JS class in ES6 has only functions. So this means jQuery would need to give up the plain type factories and define proper contract classes before $.extend() the returning objects anyway. 

I don't like Java to be honest, so I'm not trying to have full blown class definitions before even starting solving the problem at hand. Plain objects with contracts (interfaces) would be more than enough, just **be clear** (have it on paper with a contract) about what's given and what's returned. If $.Deferred is not up to the task, then by all means use Promise.

dmethvin:
> It's v3, I really hope we can get out of this plain type factory pattern and have some clearly defined interfacing contract types.

The way to get out of this is to adopt ES6 Promise and/or convert other thenables to ES6 Promise as early as possible when they're returned by something. Then deal with ES6 Promise quirks and shortcomings, which are numerous. However, at least the tools and developer knowledge will support those shortcomings better.

The changes made in jQuery 3.0 are intended to improve ES6 Promise interoperability as much as we can without breaking five year's worth of existing jQuery user code. It's v3, I really hope we don't try to change `jQuery.Deferred` in incompatible ways when we can instead just coexist with ES6 Promise.

gibson042:
@bluekvirus I did not realize you had a request of me... can you restate it? If you're asking about our guarantees, we take pains to ensure that promises allow registering callbacks on Deferred objects but _not_ resolving them. We also guarantee that non-progress callbacks are invoked no more than once, and that Deferred state changes no more than once. And as of jQuery 3.0, we further guarantee that both Deferreds and their promises adhere to Promises/A+, so our `then` is what everyone expects.

I did look at implementing inheritance while honoring those guarantees, but decided that the benefit wasn't worth the increased file size... especially when considering the inability of `instanceof` to recognize objects from _other_ jQuery instances. The simple fact is that we (i.e., Javascript developers) already _have_ a means of recognizing promises, even if it has warts: the `then` method. And `Promise.resolve` and `jQuery.when` even provide a means of getting a thenable from any value. But like it or not, this ship has already sailed.

bluekvirus:
Thanks guys. It is good to know how we arrived at the point with some extra details. v3 is a major version release, there might not be a better time to introduce advancement in the price of incompatibilities (or remove some protective precautions in our case, which looks a bit wild I know). But again, I'm not pushing my way of changing nor are the one deciding. I can see and appreciate the extra mile involved in the current decision. 

ES6 has a long way to make it to the browser, and there will be a time when chaotic co-existing backward compatible runtimes offered by different vendors. Whether or not to have detectable *interfaced* objects early is entirely up to the lib developers and I do hope the group that's leading that era is still jQuery. I picked $.Deferred as a start because promise is a well known contract.

For the record, after all these years giving me a JS object that I can't do whatever I want is unacceptable. But there were times when I did desperately need a lock on the object type to get the guaranteed interfaces. And those moments always challenged my feelings about Javascript. If you already asked the question: *What would I do with `A && A.then()` without `$.when` to the rescue?*, you certainly will have its evil twin popped the other way around like: *What if the `$.when` is the problem?* This is how I arrived at the patch.

Anyway, bon voyage!

PR #3021, opened 2016-03-28
Improve support properties computation

markelog:
* Remove div from the memory if it is not needed anymore

* Make `computeStyleTests` method a singleton

### DIFF
--- a/src/css/support.js
+++ b/src/css/support.js
@@ -28,6 +28,12 @@ define( [
 	// Executing both pixelPosition & boxSizingReliable tests require only one layout
 	// so they're executed at the same time to save the second computation.
 	function computeStyleTests() {
+
+		// This is a singleton, we need to execute it only once
+		if ( !div ) {
+			return;
+		}
+
 		div.style.cssText =
 			"box-sizing:border-box;" +
 			"position:relative;display:block;" +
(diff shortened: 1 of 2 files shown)


src/css/support.js
```diff
@@ -38,6 +44,8 @@ define( [
 
 		var divStyle = window.getComputedStyle( div );
 		pixelPositionVal = divStyle.top !== "1%";
+
+		// Support: Android 4.0 - 4.3 only, Firefox <=3 - 44+
```
mgol: Did ypu check Firefox 3? :)

mgol:
What's the size difference?

src/css/support.js
markelog: Did you? Since you are the author of that comment :), i just moved it around

markelog:
The whole byte!

timmywil:
I like it :+1:

src/css/support.js
mgol: Ah, OK then. I didn't seem the whole context on mobile so I thought this
line is new. :)

-- 
Michał Gołębiowski

mgol:
Looks nice!

-- 
Michał Gołębiowski

Issue #2965, opened 2016-03-02
$.parseHTML() creates anchor elements with empty href property in Chrome


mgol:
`document.implementation.createHTMLDocument("")`-created documents interpret the `href` property of anchor elements different in different browsers, e.g. in Chrome returning an empty string. This affects our `jQuery.parseHTML`. The `createHTMLDocument` usage has been backed out on `1.12-stable` & `2.2-stable` but we try to may make it work for 3.0.

In https://github.com/jquery/jquery/issues/2941#issuecomment-190305062 @gibson042 suggested to create a `<base>` element, setting it to the base of the current document.

The original issue: 2941

Elusive138:
I'll just repeat my comment on the proposed fix:

@gibson042's fix looks like the correct way to go about it. Based on the [brief look I had at the specs](https://bugs.chromium.org/p/chromium/issues/detail?id=568886#c5), the default base URL is `about:blank`, so if your intention is to make the behaviour match the [`document` fallback](https://github.com/jquery/jquery/blob/58c24608210c9a9a264a38746628ebc26823f59b/src/core/parseHTML.js#L23) then you'd need to set the base URL anyway, *for all browsers*. Even if the Chromium bug is fixed correctly, you'd *still* need to do this to match `document` behaviour.

Also, if this is going to be a permanent fix it'd probably be a good idea to make sure a `<base>` element does not already exist in the input.

dmethvin:
Reviewing the patched `parseHTML( data, context, keepScripts )`, the thing that struck me was that the behavior varies a *lot* depending on whether `context` is provided. Since a constructor like `$("<input>")` passes `document` to the method by default, it circumvents any do-not-run-script benefit of the missing-`context` case. So if we want to make `parseHTML()` safer I'm wondering if we should try so hard to pass `document` for cases like that.

PR #3022, opened 2016-03-28
Core: set the base href of the context in parseHTML

timmywil:
Fixes gh-2965

### DIFF
--- a/src/core/parseHTML.js
+++ b/src/core/parseHTML.js
@@ -30,7 +30,7 @@ jQuery.parseHTML = function( data, context, keepScripts ) {
 		if ( support.createHTMLDocument ) {
 			context = document.implementation.createHTMLDocument( "" );
 
-			// Support: Chrome <=46-48
+			// Support: Chrome <=46 - 48
 			// Set the base href for the created document
 			// so any parsed elements with URLs
 			// are based on the document's URL (gh-2965)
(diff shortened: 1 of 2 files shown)


src/core/parseHTML.js
```diff
@@ -21,14 +21,29 @@ jQuery.parseHTML = function( data, context, keepScripts ) {
...
+	var base, parsed, scripts;
+
+	if ( !context ) {
+
+		// Stop scripts or inline event handlers from being executed immediately
+		// by using document.implementation
+		if ( support.createHTMLDocument ) {
+			context = document.implementation.createHTMLDocument( "" );
+
+			// Support: Chrome <=46-48
```
gibson042: Spaces around the `-`.

src/core/parseHTML.js
```diff
@@ -21,14 +21,29 @@ jQuery.parseHTML = function( data, context, keepScripts ) {
...
+	var base, parsed, scripts;
+
+	if ( !context ) {
+
+		// Stop scripts or inline event handlers from being executed immediately
+		// by using document.implementation
+		if ( support.createHTMLDocument ) {
+			context = document.implementation.createHTMLDocument( "" );
+
+			// Support: Chrome <=46 - 48
```
mgol: This is not just Chrome but also Firefox as it doesn't provide the resolved URL there. Also, we're really required to do it so I'm not sure if we can claim it's to support some browser as if it had a bug. It's just something we have to do to make it work as we want, the previous way was incorrect.

timmywil: I suppose that's true. So just remove the support comment?

mgol: Yeah, I think it's not needed here.

-- 
Michał Gołębiowski

Issue #2888, opened 2016-01-31
Exception of width/height for oldIE removed incorrectly or/and unfinished


markelog:
Causing test faillure in IE9.

See https://github.com/jquery/jquery/commit/e04e246552c27e872bbf4ae00b55def02b197189
and http://swarm.jquery.org/result/1033320

/cc @timmywil @mgol

markelog:
Can't repro, should fail on http://swarm.jquery.org/result/1084483, but doesn't (including in compat mode), via browserstack

<img width="720" alt="screen shot 2016-02-15 at 22 30 23" src="https://cloud.githubusercontent.com/assets/945528/13058363/d5ca0fac-d433-11e5-86db-29094145e40b.png">

Same thing with saucelabs - 

<img width="536" alt="screen shot 2016-02-15 at 22 34 34" src="https://cloud.githubusercontent.com/assets/945528/13058451/6906fe2e-d434-11e5-8755-798fced2373f.png">

-- 
@mgol could you check if correct browser is spawned in here? If it is, i think someone should verify absence of the failure on the real thing after which we would need to disable this test for ie9.

mgol:
@markelog If it doesn't fail on the BrowserStack IE 9 version then it might be related to being run in an iframe. I'd rather not exclude this test but investigate what's the real reason of the failure as it's easy to mask a real bug in this way.

EDIT: I've joined the swarm manually from IE9 from BrowserStack and it fails, while it doesn't when run outside of TestSwarm.

markelog:
Good catch @mgol! Will check it out

markelog:
Loaded tests from an iframe, still no luck, don't remember how to do

> I've joined the swarm manually from IE9 from BrowserStack

"Joined the swarm", but don't remember what to do next :/. @mgol it seems you have better understanding on how to run this thing, would you mind?

mgol:
When you join the swarm with a recognized (i.e. being in one of the tested browser sets) browser and then schedule a test run on TestSwarm, it will get picked up unless another identical browser gets picked up first. Your one has an advantage, though, of being already spawned. You can go to https://www.browserstack.com/automate and kill the other IE9s if they interfere & then restart the job.

So after joining the swarm you can go to http://swarm.jquery.org/project/jquery, log in (see https://github.com/jquery/infrastructure/wiki/TestSwarm#create-users for login details, this page is private) and restart the failing test. You should see your IE 9 getting the job, running it & failing.

[...]

PR #3034, opened 2016-04-03
Tests: do not run IE9 effect tests if inside testswarm

markelog:
That test doesn't work properly in such environment

Ref e04e246552c27e872bbf4ae00b55def02b197189
Fixes gh-2888


mgol:
Are you sure it's just a TestSwarm flake and not the browser being actually buggy here?

markelog:
Can't reproduce this on the actual test suite nor if run through iframe

Issue #2863, opened 2016-01-25
Commit to how disconnected elements toggle


gibson042:
Ref https://github.com/jquery/jquery/issues/2404#issuecomment-172993424

`.toggle()` uses the newly-renamed `isHiddenWithinTree` to determine whether a `.show()` or `.hide()` is called for, and effects methods use it to determine if they should invoke a preliminary `.show()`. But since `isHiddenWithinTree` does not yet perform exactly as advertised, _all_ disconnected elements are treated as hidden. In code, `jQuery( "<div><a/></div>" ).find( "*" ).addBack().toggle()` locks in inline `defaultDisplay` values for both elements, as does `$visible.detach().toggle().appendTo( inPageParent )`.

This is counterintuitive to me, _especially_ when it comes to element descendants like the above `<a/>`. I'd propose updating `isHiddenWithinTree` to use only inline display values when evaluating disconnected elements. Other possibilities include a) treating elements differently by parent (document or element vs. document fragment or null)—please no, and b) keeping everything as-is. But no matter what, we should either add unit tests asserting the consensus decision, or document that disconnected elements are invalid context for the methods.

markelog:
> but I don't think that should be the case for elements with element parents unless they have inline display: none (e.g., the <p> in jQuery("<div><p></div>"))

Why?! Again, this was working for years, no complaints, fast, small, clear and now we suddenly need to reverse the logic?

markelog:
> But no matter what, we should either add unit tests asserting the consensus decision, or document that disconnected elements are invalid context for the methods.

That was a concision decision, until you removed those tests - https://github.com/jquery/jquery/blob/c4d9eac930b86c192b0f96210d88bb7d91e697e1/test/unit/css.js#L559

gibson042:
> That was a concision decision, until you removed those tests - https://github.com/jquery/jquery/blob/c4d9eac930b86c192b0f96210d88bb7d91e697e1/test/unit/css.js#L559

That is completely unrelated... note the complete _absence_ of `.toggle()`.

markelog:
> That is completely unrelated... note the complete absence of .toggle().

Those tests were done to support show()/hide() logic, and toggle() logic of showing or hiding something should be equivalent to show()/hide(), because toggle() is basically choosing what to call - show() or hide().

gibson042:
[...]

PR #3037, opened 2016-04-04
CSS: Toggle detached elements as visible unless they have display: none

gibson042:
Fixes gh-2863

### DIFF
--- a/src/css/var/isHiddenWithinTree.js
+++ b/src/css/var/isHiddenWithinTree.js
@@ -12,7 +12,12 @@ define( [
 		// isHiddenWithinTree might be called from jQuery#filter function;
 		// in that case, element will be second argument
 		elem = el || elem;
-		return jQuery.css( elem, "display" ) === "none" ||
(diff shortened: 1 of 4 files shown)


markelog:
Do i understand this correctly, we didn't have any tests for `jQuery#fadeTo`, or `jQuery#animation` for hidden elems?

src/css/var/isHiddenWithinTree.js
```diff
@@ -12,7 +12,12 @@ define( [
 		// isHiddenWithinTree might be called from jQuery#filter function;
 		// in that case, element will be second argument
 		elem = el || elem;
-		return jQuery.css( elem, "display" ) === "none" ||
-			!jQuery.contains( elem.ownerDocument, elem );
+		return elem.style.display === "none" ||
```
markelog: I would add a bunch of comments which would explain every line, logic is pretty bizarre for unprepared eyes

gibson042: Good point. Fixed.

PR #3039, opened 2016-04-04
Ajax: execute jQuery#load callback with correct context

### LINKED ISSUE #3035: load() callback "this" breaking in 2.2.0 vs 2.1.4
Hi,
"this" in the callback to the load() function changed from 2.1.4 to 2.2.0 in a breaking way.

The documentation at http://api.jquery.com/load/ says: ".. this is set to each DOM element in turn.".
In 2.2.0 "this" is instead the jQuery instance, and not the DOM element.

I tracked it down to this commit: 
https://github.com/jquery/jquery/commit/a4715f4216ace92fba6991106053415e66289686

Regards
/Fredrik Blomqvist

markelog:
Thanks @blq (Fredrik Blomqvist)

Fixes #3035


mgol:
LGTM

PR #3044, opened 2016-04-07
Update wrap.js

Abdallahozaifa:
Properly Fixed Indentation


markelog:
Thank you for taking the time to contribute to the jQuery project. Indentation here, however, is correct.

Specially for this, we have `jscs` task, which could be triggered with `grunt jscs` command. See more in [readme](https://github.com/jquery/jquery/blob/master/README.md) and/or on http://contribute.jquery.org/.

Thank you!

Issue #3040, opened 2016-04-04
Async ready event may let window.onload run before it


dmethvin:
http://jsbin.com/mecowewibi/4/edit?html,output

This seems like it will cause trouble. Timmy's spidey sense was tingling on this one.

timmywil:
It might be an issue, but I'd be okay with leaving it in for now.

dmethvin:
i can work around it in the unit test and we can see if we get any complaints.

dmethvin:
See https://github.com/jquery/jquery/commit/5cbb234dd3273d8e0bbd454fb431ad639c7242c1#commitcomment-16963004 , I have a PR incoming to fix this in a rather hacky way but still not sure of the implications for code in the field.

dmethvin:
Closing for now, the unit tests no longer have the race condition and we can see if code expects this in the field.

PR #3049, opened 2016-04-08
Tests: Refactor iframe tests for async ready event

dmethvin:
Ref gh-3040

~~Not really happy with this solution for the long term, it seems like should be
having the iframe code signal when it's done. The old code would have broken for
any use of Deferred or setTimeout in the iframe, and it's likely this will break
too.~~ See below.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,32 +340,25 @@ url("data/test.php?foo=bar");
 ```
 
 
-### Load tests in an iframe ###
-
-Loads a given page constructing a url with fileName: `"./data/" + fileName + ".html"`
-and fires the given callback on jQuery ready (using the jQuery loading from that page)
-and passes the iFrame's jQuery to the callback.
-
-```js
-testIframe( fileName, testName, callback );
-```
-
-Callback arguments:
-
-```js
-callback( jQueryFromIFrame, iFrameWindow, iFrameDocument );
-```
-
 ### Load tests in an iframe (window.iframeCallback) ###
 
-Loads a given page constructing a url with fileName: `"./data/" + fileName + ".html"`
-The given callback is fired when window.iframeCallback is called by the page.
-The arguments passed to the callback are the same as the
-arguments passed to window.iframeCallback, whatever that may be.
-
 ```js
-testIframeWithCallback( testName, fileName, callback );
+testIframeWithCallback( testName, fileName,
+  function callback( arg1, arg2, ... assert ) {
+	...
+  } );
 ```
+
+Loads a given page constructing a url with fileName: `"./data/" + fileName + ".html"`
+The iframe page is responsible for determining when `window.parent.iframeCallback`
+should be called, for example at document ready or window.onload time.
+Arguments passed to the callback are the same as the arguments passed
+to `window.parent.iframeCallback` by the iframe, plus the QUnit `assert`
+object from the `QUnit.test()` that this wrapper sets up for you.
+The iframe should send any objects needed by the unit test via arguments, for example
+its `jQuery`, `window`, and `document` objects from the iframe.
+
+
 
 Questions?
 ----------
--- a/README.md
+++ b/README.md
@@ -340,24 +340,32 @@ url("data/test.php?foo=bar");
 ```
 
 
-### Load tests in an iframe (window.iframeCallback) ###
+### Run tests in an iframe ###
+
+Some tests may require a document other than the standard test fixture, and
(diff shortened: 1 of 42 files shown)


mgol:
This seems pretty fragile...

dmethvin:
Completely agree. The better solution is to have the iframe call back to let the parent know it's done. That requires some (simple I think) changes to all the tests run via `testIframe` but I didn't have the time to think through and test that.

We can either merge this to make the problem go away now or someone can do the right thing when they get a chance. (Or revert the async ready code I guess.)

gibson042:
There are only four uses of `testIframe`, and one of them is in documentation:
* [large document dimensions](https://github.com/jquery/jquery/blob/055cb7534e2dcf7ee8ad145be83cb2d74b5331c7/test/unit/dimensions.js#L472)
* a dubious test of [multiple instance independence](https://github.com/jquery/jquery/blob/055cb7534e2dcf7ee8ad145be83cb2d74b5331c7/test/unit/selector.js#L492)
* [HTML5 boolean attributes](https://github.com/jquery/jquery/blob/055cb7534e2dcf7ee8ad145be83cb2d74b5331c7/test/unit/selector.js#L292)
* [README](https://github.com/jquery/jquery/blob/055cb7534e2dcf7ee8ad145be83cb2d74b5331c7/README.md#load-tests-in-an-iframe)

Especially considering that ready is an excludable module, I'd like to do things the right way and drop all four, replacing `testIframe` with `testIframeWithCallback` (which is almost identical anyway). Of those tests, only the first even plausibly cares about document readiness.

dmethvin:
Okay, since @gibson042 did the searching and I have about 90 minutes now, imma try to get this fixed the right way. Hold my coat I'm going in!

timmywil:
@dmethvin I assume someone got your coat. Are you taking anything else off?

dmethvin:
WHY ARE THE FIRST TWO ARGS REVERSED FOR THESE GUYS???!

dmethvin:
WHY DOES testIframe add the `.html` but testIframeWithCallback NOT????

dmethvin:
What a mess! There were a bunch of other `testIframe()` tests in offset.js but I think I got everything.

gibson042:
Huh, I don't know how I missed those. At any rate, thanks @dmethvin. You are a gentleman and a scholar.

README.md
```diff
@@ -340,32 +340,25 @@ url("data/test.php?foo=bar");
...
-testIframe( fileName, testName, callback );
-```
-
-Callback arguments:
+### Load tests in an iframe (window.iframeCallback) ###
 
 ```js
-callback( jQueryFromIFrame, iFrameWindow, iFrameDocument );
+testIframeWithCallback( testName, fileName,
+  function callback( arg1, arg2, ... assert ) {
```
mgol: This syntax looks weird to me. Did you mean:
```js
function callback( arg1, arg2, ..., assert ) {
	...
} );
```
?

mgol:
That's great! :clap:

dmethvin:
I added another commit to clean up a bunch of things. Now each iframe page includes a file that defines `startIframeTest()` which is what the iframe calls to get the test started. The callback function gets `assert` as its first arg rather than it being tacked onto the end, that matches `QUnit.test()` conventions and may be a little less error prone. If this looks good to everyone I'd like to land it tomorrow before it has to be rebased.

test/data/iframeTest.js
```diff
@@ -0,0 +1,9 @@
+// Just used in iframed html tests which jshint doesn't see
+// jshint unused: false
```
markelog: `window.startIframeTest = function() {}` should help with jshint?

README.md
```diff
@@ -340,32 +340,33 @@ url("data/test.php?foo=bar");
...
-Loads a given page constructing a url with fileName: `"./data/" + fileName + ".html"`
-and fires the given callback on jQuery ready (using the jQuery loading from that page)
-and passes the iFrame's jQuery to the callback.
+Some tests may require a document other than the standard test fixture, and
+these can be run in a separate iframe. The actual test code and assertions
+remain in jQuery's main test files; only the minimal test fixture markup
+and setup code should be placed in the iframe file.
 
 ```js
-testIframe( fileName, testName, callback );
```
markelog: Could this be done as `QUnit` plugin?

dmethvin: Possibly? I don't think it will be nicely encapsulated because the iframe test does not include QUnit.

Issue #3043, opened 2016-04-06
Element which is not in page still :visible in IE8


RobertMe:
Hi,

I'm having an issue with the `:visible`/`:hidden` check giving the wrong result in IE8 when the element under test isn't in the page.

A simplified testcase is the following snippet: `$('<div>Test</div>').is(':visible')` In all browsers the result of this is `false` (because the element isn't in the page), except in IE8, for which it actually is `true`.

From what I've investigated it seems like the issue is the final return of the `filterHidden` function: https://github.com/jquery/jquery/blob/1.12-stable/src/css/hiddenVisibleSelectors.js#L19
I think this must check if the resulting `node` is a document fragment. If it is a fragment the element can't be visible, and the result must thus be `true`. And there may possibly be different node types which could be the parent while the element isn't actually in the page, for which the test also returns the wrong result.

dmethvin:
Is this a regression? That is, did it work in a previous version of jQuery? The docs say this which seem to indicate the IE8 result is incorrect:

> Elements that are not in a document are considered hidden; jQuery does not have a way to know if they will be visible when appended to a document since it depends on the applicable styles.

I think we'd need to do a `jQuery.contains()` check to ensure it was disconnected.

RobertMe:
Yes, I forgot to mention that. I'm currently using version 1.9.1 for which it does work as expected.

timmywil:
Confirmed. Test case: http://jsbin.com/puwifu/edit?html,js,console

mgol:
This was working fine in 1.11.3: http://jsbin.com/liziju/edit?html,js,console. Therefore, it's a recent regression.

boonebgorges:
I was also experiencing this issue and bisected it to ecf52b948bcc7c9a28b5b2f93a0764868f09428e.

timmywil:
Fixed in 1.12-stable. c43066c41ec84ac5e5fe7fb3e6cd3a912a395140

PR #3057, opened 2016-04-14
CSS: disconnected elements should be hidden

timmywil:
Fixes gh-3043

### DIFF
--- a/src/css/hiddenVisibleSelectors.js
+++ b/src/css/hiddenVisibleSelectors.js
@@ -1,15 +1,21 @@
 define( [
 	"../core",
+	"../var/document",
 	"./support",
 	"../selector",
 	"../css"
-], function( jQuery, support ) {
+], function( jQuery, document, support ) {
 
 function getDisplay( elem ) {
 	return elem.style && elem.style.display || jQuery.css( elem, "display" );
 }
 
 function filterHidden( elem ) {
+
+	// Disconnected elements are considered hidden
+	if ( !jQuery.contains( elem.ownerDocument || document, elem ) ) {
+		return true;
+	}
 	while ( elem && elem.nodeType === 1 ) {
 		if ( getDisplay( elem ) === "none" || elem.type === "hidden" ) {
 			return true;
--- a/test/unit/css.js
+++ b/test/unit/css.js
@@ -1059,7 +1059,7 @@ QUnit.test( "css opacity consistency across browsers (#12685)", function( assert
 } );
 
 QUnit.test( ":visible/:hidden selectors", function( assert ) {
-	assert.expect( 17 );
+	assert.expect( 19 );
 
 	var $div, $table, $a;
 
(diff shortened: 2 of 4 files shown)


test/unit/css.js
```diff
@@ -1100,6 +1100,9 @@ QUnit.test( ":visible/:hidden selectors", function( assert ) {
 
 	$a = jQuery( "<a href='#'><h1>Header</h1></a>" ).appendTo( "#qunit-fixture" );
 	assert.ok( $a.is( ":visible" ), "Anchor tag with flow content is visible (gh-2227)" );
+
+	assert.ok( !jQuery( "<div>Test</div>" ).is( ":visible" ), "Disconnected element is not visible" );
```
markelog: Could those assertions be in separate test?

src/css/hiddenVisibleSelectors.js
```diff
@@ -1,15 +1,21 @@
...
+], function( jQuery, document, support ) {
 
 function getDisplay( elem ) {
 	return elem.style && elem.style.display || jQuery.css( elem, "display" );
 }
 
 function filterHidden( elem ) {
+
+	// Disconnected elements are considered hidden
+	if ( !jQuery.contains( elem.ownerDocument || document, elem ) ) {
```
markelog: Could you remind me when element doesn't have an `ownerDocument` property

markelog: Maybe we could put this in the comment too?

test/unit/css.js
```diff
@@ -1100,6 +1100,9 @@ QUnit.test( ":visible/:hidden selectors", function( assert ) {
...
+	assert.ok( !jQuery( "<div><div>Test</div></div>" ).find("div").is( ":visible" ), "Disconnected element is not visible" );
```
markelog: Duplicated name assertion, same as in above test

src/css/hiddenVisibleSelectors.js
timmywil: I don't recall all the cases, but this is pretty standard in our codebase. We do it in several places without a comment.

test/unit/css.js
timmywil: I guess they could, but it seems appropriate in this test.

src/css/hiddenVisibleSelectors.js
timmywil: Actually, it may not apply here considering we do a nodeType check in Sizzle.

timmywil: I feel nervous not doing it tho. Seems like whenever I leave it out, I hit a problem eventually.

markelog: I'm usually not in favor of the code which purpose is unclear

mgol:
The patch looks good but we'll most likely need another modification in this code to fix #3065.

timmywil:
Merged in 1.12-stable.

Issue #3029, opened 2016-03-30
$.when( deferred ).done() (serious?) regression


dmethvin:
This was [discussed on IRC](http://irc.jquery.org/%23jquery-dev/default_%23jquery-dev_20160328.log.html#t12:06:48) but it doesn't look like a code ticket was created. There is one for the docs at https://github.com/jquery/api.jquery.com/issues/904 because the example in `$.when()` is pretty bad, but the main problem is that `.done()` now gets an array instead of individual args.

Repro here: http://jsbin.com/civaxelixa/edit?html,console

Marking this as a blocker for the moment.

PR #3059, opened 2016-04-15
Split the two sides of jQuery.when

gibson042:
Fixes gh-3029


gibson042:
```
   raw     gz Sizes
264540  78519 dist/jquery.js
 86225  30013 dist/jquery.min.js

   raw     gz Compared to master @ f4961822160c671fd72f1da7501049aab6cfd56b
    -5    +51 dist/jquery.js
    +7     -6 dist/jquery.min.js
```

src/deferred.js
```diff
@@ -305,64 +322,43 @@ jQuery.extend( {
 	},
 
 	// Deferred helper
-	when: function() {
-		var method, resolveContexts,
-			i = 0,
-			resolveValues = slice.call( arguments ),
-			length = resolveValues.length,
+	when: function( singleValue ) {
+		var // count of uncompleted subordinates
```
dmethvin: The comment position seems a little off-kilter here. Maybe add a newline or tab? Most likely a newline.

dmethvin:
Wow, that makes the code a lot easier to understand as well. :+1:

src/deferred.js
```diff
@@ -13,6 +13,23 @@ function Thrower( ex ) {
 	throw ex;
 }
 
+function adoptValue( value, resolve, reject ) {
+	var method;
+
+	// Check for promise aspect first to privilege synchronous behavior
+	if ( value && jQuery.isFunction( ( method = value.promise ) ) ) {
```
markelog: With assignment inside of the call is attempt to save some size?

src/deferred.js
```diff
@@ -13,6 +13,23 @@ function Thrower( ex ) {
...
+		method.call( value ).done( resolve ).fail( reject );
+
+	// Other thenables
+	} else if ( value && jQuery.isFunction( ( method = value.then ) ) ) {
```
markelog: Did you try
```js 
if (...) {
 ... 
 return;
}
if (...) {
 ... 
 return;
}

...
``

approach

Comment (gibson042) [inline src/deferred.js]: +9 bytes

File: src/deferred.js (hunk 2 above)
Comment (gibson042) [inline src/deferred.js]: Yes, but it's not new. Saves 7 bytes here.

File: src/deferred.js (hunk 1 above)
Comment (gibson042) [inline src/deferred.js]: Fixed to match other instances.

**Comment by gibson042** [comment]
Rebased on the Android 4.0 fix.
```
   raw     gz Sizes
263375  78345 dist/jquery.js
 86192  29971 dist/jquery.min.js

   raw     gz Compared to master @ 7f1e59343b1600e530472a90aa27a2bcc7b72c96
    -7    +55 dist/jquery.js
    +7     -3 dist/jquery.min.js
```

File: src/deferred.js (hunk 2 above)
Comment (markelog) [inline src/deferred.js]: With the 

```js
function adoptValue( value, resolve, reject ) {

	// Check for promise aspect first to privilege synchronous behavior
	if ( value && jQuery.isFunction( value.promise ) ) {
		value.promise.call( value ).done( resolve ).fail( reject );

	// Other thenables
	} else if ( value && jQuery.isFunction( value.then ) ) {
		value.then.call( value, resolve, reject );

	// Other non-thenables
	} else {

		// Support: Android 4.0 only
		// Strict mode functions invoked without .call/.apply get global-object context
		resolve.call( undefined, value );
	}
}
```

There is not diff in the size?

gibson042: Promises/A+ requires accessing `.then` only once; `if ( jQuery.isFunction( value.then ) ) { value.then.call(…); }` is prohibited (cf. [Notes 3.5](https://promisesaplus.com/#point-75)).

markelog: Why we don't fail any tests in this case then?

dmethvin: How would you test that? It's a timing thing.

markelog: Not sure what complication there might be - if accessor called multiple times fail the test

markelog: In the matter of fact https://github.com/promises-aplus/promises-tests/blob/0d409df278345753c9527539a5558db67d42bb02/lib/tests/helpers/thenables.js#L29

markelog: So assignment triggers getter, `isFunction` also does it if called directly on object and does not trigger it if called indirectly. 

So in above example we would have same amount of triggered accessors i.e. i don't see violations here

dmethvin: Okay, yeah, since we just need to be sure it's called once, and not *when* it's called, you're right!

markelog: /cc @domenic

gibson042: @markelog: Commenting it out won't fail any tests because Promises/A+ only specifies `then`, and we only define a [promises-tests adapter](https://github.com/jquery/jquery/blob/055cb7534e2dcf7ee8ad145be83cb2d74b5331c7/test/promises_aplus_adapter.js#L13-L21) for `jQuery.Deferred`. I can define a `jQuery.when` adapter, put in a QUnit test of `jQuery.when( pathologicalThenable )`, or just let it be.

gibson042: And to clarify: `if ( jQuery.isFunction( value.then ) ) { value.then.call(…); }` is prohibited because it accesses `value.then` twice, and the results of those accesses won't be identical in the presence of a pathological getter like https://github.com/promises-aplus/promises-tests/blob/0d409df278345753c9527539a5558db67d42bb02/lib/tests/helpers/thenables.js#L29 .

markelog: >  or just let it be, but the line MUST take that form if we guarantee proper 

I think such changes should be supported and confirmed by the tests, otherwise all of it could be seen as speculation or subjective interpretation, it seems this discussion revealed luck of it, so i'd say either in this pull or right after this one, we need to add those tests for `jQuery.when`.

gibson042: Done.

gibson042:
We're now taking a slight hit, but within tolerances. It would be nice to consolidate the `when` and `then` helpers in the future.

```
   raw     gz Sizes
263788  78457 dist/jquery.js
 86222  29977 dist/jquery.min.js

   raw     gz Compared to master @ 7f1e59343b1600e530472a90aa27a2bcc7b72c96
  +406   +167 dist/jquery.js
   +37     +3 dist/jquery.min.js
```

test/unit/effects.js
```diff
@@ -1028,7 +1028,7 @@ jQuery.each( {
 				jQuery( elem ).remove();
 
 			} );
-			this.clock.tick( 50 );
+			this.clock.tick( 100 );
```
markelog: hmm, why?

src/deferred.js
markelog: you-da-man

test/unit/effects.js
gibson042: Funny you should ask... the test uses [`jQuery.when( elem ).done(…)`](https://github.com/gibson042/jquery/blob/gh-3029-jQuery-when-deferred/test/unit/effects.js#L963), and this PR updates the single-argument form of `jQuery.when` to [use `.then()`](#diff-00843d85b3ae02d8ca45835b8d0a4edfR371) for unwrapping secondary thenables (e.g., `jQuery.when( jQuery.Deferred().resolve( thenable ) )`). Single-argument `jQuery.when` becomes necessarily asynchronous as a side effect, and we therefore require an extra clock tick to invoke the callback.

markelog: ooo, so we have animate(..., 50), but because `when.done` is now async we need value more then `50`, interesting, you probably were surprised a bit :)

gibson042: Yes.

marcandre: > Single-argument jQuery.when becomes necessarily asynchronous as a side effect

It feels to me like placing a distinction between single argument and multiple argument is a really bad idea. I opened #3100 about this.

Issue #3060, opened 2016-04-15
Remove default Deferred callback context


gibson042:
I ran into an interesting obstacle when working on #3029: our longstanding behavior of using a default context for jQuery.Deferred callbacks (specifically, [the deferred's base promise](http://jsbin.com/tugocasose/edit?js,console)). It's not _necessarily_ a blocker, but it does make `$.when( $.Deferred().resolve() ).done( whatsMyContext )` very counterintuitive (in that I'm not even sure what to code for #3029). For sanity in that case—and for increased similarity with native Promises—I'd like to remove the default, making `*With` methods _required_ to provide a meaningful context. In other words, the preceding code would have either `undefined` or global object context, depending on whether or not it was executed in strict mode.

Any objections?

dmethvin:
Reading through the issues you filed I agree, explicit `*With` seems like a reasonable requirement if you really want context.

Krinkle:
For most code I can't imagine any useful purpose in having the promise available as this value. One would always chain on the promise directly, create a new one, or involve another code that produces a promise. One wouldn't call `then()` inside the then-callback for the same promise.

However there does appear to be one fairly common jQuery-specific use case, that applies to inside callbacks: `state()`.

```js
$.Deferred().reject().always(function () {
  console.log(this.state());
});
```
While I wouldn't expect this pattern in more advanced projects, I wouldn't be surprised to see this often in simple jQuery projects.

```js
/** @return {jQuery.Promise} */
function foo() { /* .. */ }

showSpinner();
foo().always(function () {
  hideSpinner();
  if (this.state() == 'resolve') {
    // Stuff...
  }
});
```

It seems awkward to have to store the promise in a variable for that.

Krinkle:
The following use case was rather counter-intuitive to migrate when using only the text on <https://api.jquery.com/jQuery.Deferred/> and <https://jquery.com/upgrade-guide/3.0/>

```js
function postWithToken(type, params) {
  return api.getToken(type).then(function (token) {
    params.token = token;
    return api.post(params).then(null, function fail() {
      if ( code === 'expired' ) {
        return retryOnce(type, params);
      }
      // Let caller handle any other error
      return this;
    });
}
```

Initial attempt:

```diff
-    return api.post(params).then(null, function fail(err) {
+    return api.post(params).then(null, function fail(err) {
       // Let caller handle any other error
-      return this;
+      return err;
```

[...]

PR #3061, opened 2016-04-15
Deferred: Remove default callback context

gibson042:
Fixes gh-3060

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -23,9 +23,6 @@ module.exports = function( grunt ) {
 			baseUrl: "src",
 			name: "jquery",
 
-			// Allow strict mode
-			useStrict: true,
-
 			// We have multiple minify steps
 			optimize: "none",
 
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -23,6 +23,9 @@ module.exports = function( grunt ) {
 			baseUrl: "src",
 			name: "jquery",
 
+			// Allow strict mode
+			useStrict: true,
+
 			// We have multiple minify steps
 			optimize: "none",
 
--- a/src/callbacks.js
+++ b/src/callbacks.js
@@ -76,6 +76,10 @@ jQuery.Callbacks = function( options ) {
 				memory = queue.shift();
 				while ( ++firingIndex < list.length ) {
 
+					// Preserve null/undefined context
+					// (which non-strict `apply` maps to the global object)
+					self.context = memory[ 0 ];
+
 					// Run callback and check for early termination
 					if ( list[ firingIndex ].apply( memory[ 0 ], memory[ 1 ] ) === false &&
 						options.stopOnFalse ) {
@@ -92,7 +96,7 @@ jQuery.Callbacks = function( options ) {
 				memory = false;
 			}
 
-			firing = false;
+			self.context = firing = false;
 
 			// Clean up if we're done firing for good
 			if ( locked ) {
--- a/src/callbacks.js
+++ b/src/callbacks.js
@@ -76,10 +76,6 @@ jQuery.Callbacks = function( options ) {
 				memory = queue.shift();
 				while ( ++firingIndex < list.length ) {
 
-					// Preserve null/undefined context
-					// (which non-strict `apply` maps to the global object)
-					self.context = memory[ 0 ];
-
 					// Run callback and check for early termination
 					if ( list[ firingIndex ].apply( memory[ 0 ], memory[ 1 ] ) === false &&
 						options.stopOnFalse ) {
@@ -96,7 +92,7 @@ jQuery.Callbacks = function( options ) {
 				memory = false;
 			}
 
-			self.context = firing = false;
+			firing = false;
 
 			// Clean up if we're done firing for good
 			if ( locked ) {
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -71,9 +71,11 @@ jQuery.extend( {
 				},
 				then: function( onFulfilled, onRejected, onProgress ) {
 					var maxDepth = 0;
-					function resolve( depth, deferred, handler, special ) {
+					function resolve( depth, deferred, callbacks, handler, special ) {
 						return function() {
-							var that = this,
+
+							// Recover null/undefined context from global `this`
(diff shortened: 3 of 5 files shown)


gibson042:
Deeper and deeper... af85a48 broke the Promises/A+ strict-mode assertions, and there are two ways to fix it. The simpler (and smaller) of the two is just [letting jQuery run in strict mode](https://github.com/jquery/jquery/compare/af85a48908d1c2242fdbe6f76dd710b6fc771213...9776354009fe9dbe7c2a4bce9be42556ae9b5703), and the more complex involves [preserving `Callbacks#fireWith` context](https://github.com/jquery/jquery/compare/af85a48908d1c2242fdbe6f76dd710b6fc771213...3a80ba5694b3b39fdc9b41f01a9c12c0fae1f184). I've left commits for both on this PR so we can make a fully-informed decision.

My personal inclination is to go strict mode, with justification documented in [`97763540`](https://github.com/jquery/jquery/commit/9776354009fe9dbe7c2a4bce9be42556ae9b5703#diff-1d6f10f8402b799d1f593652e845f280R41):
> But as of jQuery 3.0 (2016), strict mode should be common enough that all such attempts [to access strict mode `arguments.callee.caller`] are guarded in a try block.

However, three years ago [trac-13335](https://bugs.jquery.com/ticket/13335) was considered a big enough deal that we removed "use strict" just three weeks after introducing it, and today every browser engine except Blink throws an exception when a non-strict function tries to access a strict caller. The question is, has enough time elapsed that we now feel comfortable with reintroducing it and wontfixing any reports about ill-advised code that tries to naïvely `callee.caller` its way up a stack?

mgol:
Id really love us to go to strict mode if possible. The concerns about tracing strict code from sloppy one are permanent as AFAIK the spec forbids such access. We're thinking about moving to ES6 modules one day; if we're going to expose them similarly to how we expose our AMD modules in `src/` now, we have to remember ES6 modules are always strict so if we didn't put the pragma in the built file, we'd be basically running our code sometimes in strict, sometimes in sloppy mode.

I'd also like to point out a comment from one Chromium issue (https://bugs.chromium.org/p/v8/issues/detail?id=222#c13):
> JavaScript only supports (and is only intended to support) such security concerns for strict code, and for contexts that enforce that all untrusted code must be strict and has no access to any sloppy functions. For these, aFunction.arguments is either poisoned or absent.

More & more libraries use strict mode. Angular, Ember & React do so if you use a JS framework you most likely already have it enabled somewhere. Bootstrap & Foundation do so if you use a popular CSS framework it's taken care of as well. Three.js enables strict mode as well.

I'd argue the majority of our users might be already using strict mode via various libraries that they use. Us not doing that won't help it but it allows external code to mess with jQuery more, introduces potential security concerns and puts us further from ES6 modules.

mgol:
> Deeper and deeper... af85a48 broke the Promises/A+ strict-mode assertions

Although we still support Android 4.0 & IE 9 which don't support strict mode and our test suite should still pass in them so perhaps there's no way around that just via enabling strict mode...

markelog:
We can try it again and release 3.0.1 if users wouldn't like it

@gibson042 

> Deeper and deeper... af85a48 broke the Promises/A+ strict-mode assertions, and there are two ways to fix it.

As i understand those tests expect for context to be `undefined` rather then `window`? It seems like a bug in the tests though.

gibson042:
> The concerns about tracing strict code from sloppy one are permanent as AFAIK the spec forbids such access.

…but doesn't require throwing an exception. Chrome, for example, just returns a `null` caller: https://jsfiddle.net/d0vaf1uL/

> More & more libraries use strict mode.

Right, which serves as good defense for jQuery doing so as well.

> Although we still support IE 9 which doesn't support strict mode

But that's fine; it just means `"use strict"` is a harmless string literal in that environment (and `.call()` results in global-object `window` context).

> As i understand those tests expect for context to be `undefined` rather then `window`? It seems like a bug in the tests though.

It's not a bug. `promise.then(strictModeFunction)` should have `undefined` context, even though `promise.then(sloppyFunction)` has global-object context, because the [spec](https://promisesaplus.com/#point-35) requires callbacks to be invoked as functions and that invocation has observable differences in strict mode: https://promisesaplus.com/#point-69 .

At any rate, it looks like jumping into strict mode is the way to go for 3.0. I'll land it later today.

markelog:
> It's not a bug.

By that i meant that test should check what mode they are in, if strict - then assertion should check for undefined, if sloppy then for global.

Otherwise, those tests suggesting that Promise/A+ spec is not compatible with sloppy mode, which is not true.

So 

> promise.then(strictModeFunction) should have undefined context, even though promise.then(sloppyFunction) has global-object context

That exactly what they should be testing

markelog:
Like it said in https://promisesaplus.com/#point-69

> That is, in strict mode this will be undefined inside of them; in sloppy mode, it will be the global object.

markelog:
In any case, i'm not oppose to land, but if would have issues with strict mode again, then we would have to revert so we would face that issue all over again.

gibson042:
> test should check what mode they are in, if strict - then assertion should check for `undefined`, if sloppy then for global.

They check both, explicitly: https://github.com/promises-aplus/promises-tests/blob/0d409df278345753c9527539a5558db67d42bb02/lib/tests/2.2.5.js

dmethvin:
I'm good with another shot at `"use strict"` for 3.0. This is another upgrade guide item and one that needs to get attention. I've moved it from the Google doc to https://github.com/jquery/jquery.com/blob/master/pages/upgrade-guide/3.0.md I will add this once it lands.

mgol:
Remember that we now need not only to uncomment the pragma in wrapper.js
but also add it to every single AMD module in src.



-- 
Michał Gołębiowski

gibson042:
Landed, and 3a80ba5694b3b39fdc9b41f01a9c12c0fae1f184 should provide a good starting point if it turns out that we do want to revert strict mode without breaking #3060.

Also, and for similar reasons, I decided to separate module-level strict mode adoption into its own issue at #3073.

dmethvin:
I added a note in the upgrade guide.

PR #3069, opened 2016-04-20
master

lexxur:
change line https://github.com/jquery/jquery/blob/master/src/core/init.js#L76

from

```javascript
				} else {
```

to

```javascript
				} else if ( match[ 2 ]) {
```

to fix annoing **notice**
> Empty string can not be argument of method **getElementById()**


mgol:
This PR is broken; if you experience a specific issue, please open an issue with steps to reproduce the problem.

Issue #3077, opened 2016-04-25
Get rid of excessive "use strict" pragmas in the built file


mgol:
Since our AMD modules are now strict and we're not stripping the `"use strict";` pragmas when building, we have a lot of (harmless) excessive `"use strict";` statements in the built file. They all get stripped when minified so it's not a big deal, just slightly confusing to casual readers of the built file contents.

markelog:
I wonder if there more tools like https://github.com/gfranko/amdclean out there

mgol:
I have a PR for that that strips the pragma in non-hacky way: #3078.

PR #3078, opened 2016-04-25
Build: Strip excessive "use strict" pragmas when building

mgol:
### Summary ###

1. Strip excessive "use strict" pragmas when building (1st commit)
2. Strip the strict-mode related comment in exports/global.js (2nd commit)

Only the first commit is relevant to #3077, the second one was added mostly because it may reduce confusion and is also related to strict mode. If it's controversial, I may land only the first one.

### Checklist ###

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -39,7 +39,10 @@ module.exports = function( grunt ) {
 			skipSemiColonInsertion: true,
 			wrap: {
 				start: wrapper[ 0 ].replace( /\/\*jshint .* \*\/\n/, "" ),
-				end: globals + wrapper[ 1 ]
+				end: globals.replace(
+					/\/\*\s*ExcludeStart\s*\*\/[\w\W]*?\/\*\s*ExcludeEnd\s*\*\//ig,
+					""
+				) + wrapper[ 1 ]
 			},
 			rawText: {},
 			onBuildWrite: convert
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -81,7 +81,7 @@ module.exports = function( grunt ) {
 
 			// Remove define wrappers, closure ends, and empty declarations
 			contents = contents
-				.replace( /define\([^{]*?{/, "" )
+				.replace( /define\([^{]*?{\s*(?:("|')use strict\1(?:;|))?/, "" )
 				.replace( rdefineEnd, "" );
 
 			// Remove anything wrapped with
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -1,6 +1,10 @@
+/* ExcludeStart */
+
 // This file is included in a different way from all the others
 // so the "use strict" pragma is not needed.
 /* jshint strict: false */
+
+/* ExcludeEnd */
 
 var
 


mgol:
Note that it fixes #3077 without any additional hacks: we've already been stripping the `define` AMD intro, this just enhances the regex so that it catches an optional `"use strict";` pragma just after the wrapper start.

dmethvin:
LGTM and not hacky at all if you consider this to be part of the wrapper. It sorta is.

markelog:
I'm considering everything involving parsing sources ourselves as not ideal, but yeah, this is in line of what we have now.

If it works and don't break resulting output then LGTM.

But i think we really should consider current approach as suboptimal

mgol:
@markelog Ideally we'd generate a CST, strip the define wrappers together with `"use strict";` pragmas and concatenate afterwards. What we have now works so far, though so I don't feel the urge to change it ASAP.

In any case, I'm definitely not increasing the number of hacks here. :)

mgol:
Does the second commit look fine to everyone?

dmethvin:
I think that's good, that way we can expand the explanation if we get more confused people. :smiling_imp:

mgol:
> that way we can expand the explanation

What do you mean?

markelog:
> ASAP

Oh yeah, definitely post 3.0, will create ticket about it

dmethvin:
> @dmethvin What do you mean?

I just meant we can add more words between those two comments if we continue to get people asking questions.

mgol:
> I just meant we can add more words between those two comments if we continue to get people asking questions.

What questions? :) What I did there is strip those lines from the built file as they don't apply there; they'll only be available in `src/global.js`.

dmethvin:
Right, they are only present in the un-built file since they are stripped, but we can add more text there to explain the purpose of that file and how it works if people are still confused.

mgol:
@dmethvin OK, got it now, thanks.

mgol:
If there are no objections, I'd like to land this tomorrow.

build/tasks/build.js
```diff
@@ -81,7 +84,7 @@ module.exports = function( grunt ) {
 
 			// Remove define wrappers, closure ends, and empty declarations
 			contents = contents
-				.replace( /define\([^{]*?{/, "" )
+				.replace( /define\([^{]*?{\s*(?:"use strict";)?/, "" )
```
gibson042: Given recent talk of changing code style, the directive-matching subexpression should probably be more general, e.g. accepting a trailing [line terminator](http://www.ecma-international.org/ecma-262/6.0/#sec-line-terminators) instead of semicolon and stripping [whitespace](http://www.ecma-international.org/ecma-262/6.0/#sec-white-space) and comments:
```
// was: (?:"use strict";)
(?:("|')use strict\1[\t\v\f \xA0\uFEFF]*(?:\/\*[\w\W]*?\*\/[\t\v\f \xA0\uFEFF]*)*(?:\/\/.*)?[\n\r;])
```

mgol: This is for the purposes of transforming our code only so e.g. catching `'use strict'` is unnecessary. Also, why strip comments?

gibson042: Maybe I got carried away with the comment matching, but https://www.npmjs.com/package/standard requires single-quoted strings and prohibits trailing semicolons. Since jquery/contribute.jquery.org/issues/138 proposes adopting it, we should be ready here (especially given that we have no verification that excess "use strict" expressions are absent from the built file).

mgol: OK, then `(?:"use strict";)` -> `(?:("|')use strict\1;` should work. I think there are enough people on the jQuery Core team with extremely strong preference to not omit semi-colons that I don't expect it to ever happen but if you feel strongly about being secure here `(?:("|')use strict\1(?:;|)` would work as well.

gibson042: :+1:

mgol: PR updated.

markelog: > we should be ready

Oh god, i hope we will really not be ready for it! :).

PR #3090, opened 2016-04-29
Ajax: Make traditional flag explicitly false in options

dmethvin:
### Summary ###

See https://github.com/jquery/jquery-migrate/issues/173

This allows Migrate to detect cases where traditional is not specified to
jQuery.param() but jQuery.ajaxSettings.traditional is set to true.

### Checklist ###

* [ ] ~~New tests have been added to show the fix or feature works~~
 - (New tests will be added in Migrate)


dmethvin:
I realize I can still do a decent job with the warning in Migrate even if we leave this undefined as a default, so I'm closing this PR as unnecessary.

Issue #3080, opened 2016-04-26
`offset` method returns incorrect value in Firefox


anseki:
Hi,
I found a case that `offset` method returns incorrect value in Firefox.
`offset` method subtracts `docElem.clientTop/clientLeft` (i.e. `border-width` of `html` element), but Firefox seems to return `0` as `clientTop/clientLeft` of `html` element always.

For example:
https://jsfiddle.net/xwuc0g6q/

``` html
<html>
  <body>
    <div id="box1"></div>
  </body>
</html>
```

``` css
html {
  border: 10px solid red;
}

body {
  border: 10px solid blue;

  margin: 20px;
  padding: 30px;

  /*position: relative;*/
}

#box1 {
  position: absolute;
  left: 0;
  top: 0;
  width: 50px;
  height: 50px;
  background-color: green;
}
```

``` js
$(function() {

  var elem = document.getElementById('box1');

  // <html>.clientLeft/clientTop
  // from https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/offset.js#L78
  var doc = elem.ownerDocument,
    docElem = doc.documentElement;
  console.log(docElem.tagName + ' > clientLeft: ' + docElem.clientLeft + ' clientTop: ' + docElem.clientTop);

  // document.body.clientLeft/clientTop
  var body = document.body;
  console.log(body.tagName + ' > clientLeft: ' + body.clientLeft + ' clientTop: ' + body.clientTop);

  // getBoundingClientRect
  var box = elem.getBoundingClientRect();
  console.log('BoundingBox > left: ' + box.left + ' top: ' + box.top);

  // jQuery#offset
  var offset = $(elem).offset();
  console.log('offset() > left: ' + offset.left + ' top: ' + offset.top);

});
```

Res in Google Chrome:

```
HTML > clientLeft: 10 clientTop: 10
BODY > clientLeft: 10 clientTop: 10
BoundingBox > left: 0 top: 0
offset() > left: -10 top: -10
```

Res in IE:

```
HTML > clientLeft: 10 clientTop: 10
BODY > clientLeft: 10 clientTop: 10
BoundingBox > left: 0 top: 0
offset() > left: -10 top: -10
```

Res in Firefox:

```
HTML > clientLeft: 0 clientTop: 0
BODY > clientLeft: 10 clientTop: 10
BoundingBox > left: 0 top: 0
offset() > left: 0 top: 0
```

[...]

PR #3096, opened 2016-05-01
Offset: Operate "relative to document" correctly

anseki:
### Summary

Fix: `.offset()` doesn't operate "relative to document" correctly
### Checklist

- [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

Related to: #3080
Remove unnecessary code that subtracts
`document.documentElement.clientTop/clientLeft` (i.e. `border-width` of
`<html>`) from coordinates.
In Firefox, those are not subtracted because `clientTop/clientLeft` are
`0` always by accidentally bug.

### DIFF
--- a/test/data/offset/rel-doc.html
+++ b/test/data/offset/rel-doc.html
@@ -5,35 +5,10 @@
 		<meta http-equiv="Content-type" content="text/html; charset=utf-8">
 		<title>body</title>
 		<style type="text/css" media="screen">
-			html, body { border: solid maroon; } /* for human */
-			html { background-color: green; } /* for human */
-			body { background-color: yellow; } /* for human */
-			#static { background-color: blue; } /* for human */
-			#absolute { background-color: aqua; } /* for human */
-			#child-static { background-color: darkviolet; } /* for human */
-			#child-absolute { background-color: violet; } /* for human */
-			html {
-				margin: 1px;
-				border-width: 2px;
-				padding: 4px;
-			}
-			body {
-				margin: 8px;
-				border-width: 16px;
-				padding: 32px;
-			}
-			#absolute {
-				position: absolute;
-				top: 64px;
-				left: 64px;
-			}
-			#parent { position: relative; }
-			#child-absolute {
-				position: absolute;
-				/* Specific coordinates for avoiding bug of `.position()` */
-				top: 0;
-				left: 0;
-			}
+			html { font-size: 10px; margin: 1em; border: 2em solid red; padding: 4em; }
+			body { margin: 8em; border: 16em solid blue; padding: 32em; }
+			#static { position: static; }
+			#absolute { position: absolute; top: 2em; left: 2em; }
 		</style>
 		<script src="../../jquery.js"></script>
 		<script src="../iframeTest.js"></script>
@@ -46,9 +21,5 @@
 	<body>
 		<div id="static">static</div>
 		<div id="absolute">absolute</div>
-		<div id="parent">
-			<div id="child-static">child-static</div>
-			<div id="child-absolute">child-absolute</div>
-		</div>
 	</body>
 </html>
--- a/test/data/offset/rel-doc.html
+++ b/test/data/offset/rel-doc.html
@@ -5,11 +5,35 @@
 		<meta http-equiv="Content-type" content="text/html; charset=utf-8">
 		<title>body</title>
 		<style type="text/css" media="screen">
-			html, body { border: solid red; }
-			html { background-color: green; }
-			body { background-color: yellow; }
-			#marker1 { background-color: blue; }
-			#marker2 { background-color: orange; position: absolute; }
+			html, body { border: solid maroon; } /* for human */
+			html { background-color: green; } /* for human */
+			body { background-color: yellow; } /* for human */
+			#static { background-color: blue; } /* for human */
+			#absolute { background-color: aqua; } /* for human */
+			#child-static { background-color: darkviolet; } /* for human */
+			#child-absolute { background-color: violet; } /* for human */
+			html {
+				margin: 1px;
+				border-width: 2px;
+				padding: 4px;
+			}
+			body {
+				margin: 8px;
+				border-width: 16px;
+				padding: 32px;
+			}
+			#absolute {
+				position: absolute;
+				top: 64px;
+				left: 64px;
+			}
+			#parent { position: relative; }
+			#child-absolute {
+				position: absolute;
+				/* Specific coordinates for avoiding bug of `.position()` */
+				top: 0;
+				left: 0;
+			}
 		</style>
 		<script src="../../jquery.js"></script>
 		<script src="../iframeTest.js"></script>
@@ -20,7 +44,11 @@
 		</script>
 	</head>
 	<body>
-		<div id="marker1">marker1</div>
-		<div id="marker2">marker2</div>
+		<div id="static">static</div>
+		<div id="absolute">absolute</div>
+		<div id="parent">
+			<div id="child-static">child-static</div>
+			<div id="child-absolute">child-absolute</div>
+		</div>
 	</body>
 </html>
--- a/test/data/offset/rel-doc.html
+++ b/test/data/offset/rel-doc.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
+	"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html>
+	<head>
+		<meta http-equiv="Content-type" content="text/html; charset=utf-8">
+		<title>body</title>
+		<style type="text/css" media="screen">
+			html, body { border: solid red; }
+			html { background-color: green; }
+			body { background-color: yellow; }
+			#marker1 { background-color: blue; }
+			#marker2 { background-color: orange; position: absolute; }
+		</style>
+		<script src="../../jquery.js"></script>
+		<script src="../iframeTest.js"></script>
+		<script type="text/javascript" charset="utf-8">
+			jQuery(function() {
+				startIframeTest();
+			});
+		</script>
+	</head>
+	<body>
+		<div id="marker1">marker1</div>
+		<div id="marker2">marker2</div>
+	</body>
+</html>
--- a/test/unit/offset.js
+++ b/test/unit/offset.js
@@ -568,117 +568,142 @@ QUnit.test( "iframe scrollTop/Left (see gh-1945)", function( assert ) {
 	}
 } );
 
-testIframe( "coordinates relative to document", "offset/rel-doc.html", function( assert, $, iframe, doc ) {
-	assert.expect( 768 );
+(function() {
+	var POSITION_VALUES = [ "static", "relative", "absolute", "fixed" ];
 
-	var docElem = doc.documentElement, // <html> element
-		bodyElem = doc.body,
-		marker1 = doc.getElementById( "marker1" ),
-		marker2 = doc.getElementById( "marker2" ),
-		$marker1 = $( marker1 ),
-		$marker2 = $( marker2 ),
+	supportjQuery.each( POSITION_VALUES, function( i, docPosition ) {
+		supportjQuery.each( POSITION_VALUES, function( i, bodyPosition ) {
+			supportjQuery.each( [ "static", "absolute" ], function( i, markerPosition ) {
 
-		// Pixels
-		DOC_MARGIN = 1,
-		DOC_BORDER = 2,
-		DOC_PADDING = 4,
-		BODY_MARGIN = 8,
-		BODY_BORDER = 16,
-		BODY_PADDING = 32,
-		MARKER_TOP_LEFT = 64,
-		// Bit as Flag
-		FLAG_DOC_MARGIN = 1,
-		FLAG_DOC_BORDER = 2,
-		FLAG_DOC_PADDING = 4,
-		FLAG_BODY_MARGIN = 8,
-		FLAG_BODY_BORDER = 16,
-		FLAG_BODY_PADDING = 32;
+				testIframe( "coordinates relative to document" +
+							" (html{position:" + docPosition + "} body{position:" + bodyPosition + "} div{position:" + markerPosition + "})",
+						"offset/rel-doc.html", function( assert, $, iframe, doc ) {
+					assert.expect( 256 );
 
-	// Test it in each situation that is made by combined properties.
-	function situations( affect, markerTopLeft, exBodyPadding, messageHead ) {
-		var flags, offset1, offset2, sumLen, labels, labelsInMessage;
+					var docElem = doc.documentElement, // `<html>` element
+						bodyElem = doc.body,
+						marker1 = doc.getElementById( "marker1" ),
+						marker2 = doc.getElementById( "marker2" ),
+						$marker1 = $( marker1 ),
+						$marker2 = $( marker2 ),
+						affectProps, markerTopLeft,
 
-		function switchProp( elem, enable, styleProp, pixels, label ) {
-			if ( enable ) {
-				elem.style[ styleProp ] = pixels + "px";
-				if ( !exBodyPadding || elem !== bodyElem || styleProp !== "padding" ) {
-					sumLen += pixels;
-				}
-				labels.push( label );
-			} else {
-				elem.style[ styleProp ] = "0";
-			}
-		}
+						// Bit as Flag of box-properties
+						FLAG_DOC_MARGIN = 1,
+						FLAG_DOC_BORDER = 2,
+						FLAG_DOC_PADDING = 4,
+						FLAG_BODY_MARGIN = 8,
+						FLAG_BODY_BORDER = 16,
+						FLAG_BODY_PADDING = 32,
 
-		for ( flags = 0;
-				flags <= ( FLAG_DOC_MARGIN | FLAG_DOC_BORDER | FLAG_DOC_PADDING | FLAG_BODY_MARGIN | FLAG_BODY_BORDER | FLAG_BODY_PADDING );
-				flags++ ) {
+						// Pixels
+						DOC_MARGIN = 1,
+						DOC_BORDER = 2,
+						DOC_PADDING = 4,
+						BODY_MARGIN = 8,
+						BODY_BORDER = 16,
+						BODY_PADDING = 32,
+						MARKER_TOP_LEFT = 64;
 
-			sumLen = 0;
-			labels = [];
+					// Test it in each situation that is made by combined box-properties.
+					function testWithBoxProps( affectProps, markerTopLeft ) {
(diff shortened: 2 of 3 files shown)


anseki:
I don't know why "CLA:Error"... :cry: 
I already input and submitted on http://contribute.jquery.org/CLA/

gibson042:
@anseki See http://contribute.jquery.org/CLA/status/?owner=jquery&repo=jquery&sha=359da5dd3b080c4c6367ec904fa9325a1d4f0305 . We ask for your real email address and real name (or consistently-used pseudonym), and require that the git commits match the CLA. Is "anseki" the best name for you to use for assigning contributions?

anseki:
Thank you @gibson042.
Yes, I visited that page, and I changed private GitHub email address to my Gmail address by `git config --global user.email`.
And I tried:
```
git checkout offset-html-border
git commit --amend --reset-author
git push -f origin offset-html-border
```
What should I do?

gibson042:
The current exception is about the name field. Do you have a more "traditional" name, or is "AnSeki" the best descriptor for us?

anseki:
I've been using this name anytime, and I hope use this.

test/unit/offset.js
```diff
@@ -568,4 +568,117 @@ QUnit.test( "iframe scrollTop/Left (see gh-1945)", function( assert ) {
...
+				}
+				labels.push( label );
+			} else {
+				elem.style[ styleProp ] = "0";
+			}
+		}
+
+		for ( flags = 0;
+				flags <= ( FLAG_DOC_MARGIN | FLAG_DOC_BORDER | FLAG_DOC_PADDING | FLAG_BODY_MARGIN | FLAG_BODY_BORDER | FLAG_BODY_PADDING );
+				flags++ ) {
```
gibson042: This is a very good start, but is a bit hard to read and hard to reason about. I now think every interesting case here can operate in a page with nonzero, non-pixel (to catch inappropriate `parseFloat`s) values for all of margin, border, and padding on both `<html>` and `<body>`, with the only variation being `position` values for those two elements. You can achieve that by means of a single iframe fixture (setting `position` styles before making any other calls), but I would like them in distinct `testIframe` invocations (see [effects.js](https://github.com/jquery/jquery/blob/master/test/unit/effects.js#L193) for an example of such process-constructed `QUnit.test` input). The cases, and my observations of them, are as follows:

`<html>` position | `<body>` position | `.offset` origin
------------ | ------------- | --------
static | static/relative/absolute/fixed | `document` (i.e., outer edge of `<html>` margin)
relative/absolute/fixed | static/relative/absolute/fixed | inside `<html>` border

That's a total of 16 cases, each case with its own `testIframe`. I'm aware that `<body>` position doesn't actually seem to matter, but it's probably worth covering anyway.

anseki: Ok, I try to rewrite tests.
You said that `situations()` should be called in each case of that 16 cases. Is my interpretation correct?

Also, that 16 cases include two cases in current code "CASE 1" and "CASE 2".
Is "CASE 3" in current code unnecessary?
Or, should two cases `marker1.style.position = "absolute"` and `marker1.style.position = "static"` be tested in each case of that 16 cases? That is, `situations()` is called in each case of 2x16 cases.

anseki:
It seems "CLA:Error" was solved by second commit.

anseki:
I added test cases, then

- 256 cases: `margin`, `border` and `padding` of `<html>` and `<body>`
- 16 cases: `position` of `<html>` and `<body>`
- 2 cases: `position:static` and `position:absolute` of `marker1`

Total: 8192 cases

Also, I tried to change messages.
But, it might be hard to read still.

test/unit/offset.js
```diff
@@ -568,4 +568,142 @@ QUnit.test( "iframe scrollTop/Left (see gh-1945)", function( assert ) {
...
+						$marker2 = $( marker2 ),
+						affectProps, markerTopLeft,
+
+						// Bit as Flag of box-properties
+						FLAG_DOC_MARGIN = 1,
+						FLAG_DOC_BORDER = 2,
+						FLAG_DOC_PADDING = 4,
+						FLAG_BODY_MARGIN = 8,
+						FLAG_BODY_BORDER = 16,
+						FLAG_BODY_PADDING = 32,
```
gibson042: > You said that `situations()` should be called in each case of that 16 cases. Is my interpretation correct?

Not exactly., I'm saying that `situations` and "CASE 1"/"CASE 2"/"CASE 3" should be removed in favor of a simpler approach. The `testIframe` callback shouldn't need to do _too_ much:
```js
// Assume initial conditions include html/document margin/border/padding:
// html { font-size: 10px; margin: 1em; border: 2em solid red; padding: 4em; }
// body { margin: 8em; border: 16em solid blue; padding: 32em; }

// …and a static element and an absolutely-positioned element:
// #static { position: static; }
// #absolute { position: absolute; top: 2em, left: 2em; }

// Establish document-relative origins for children of <body>
docElem.style.position = docPosition;
body.style.position = bodyPosition;
var bodyContentOrigin = 10 + 20 + 40 + 80 + 160 + 320;
var origin = 0 +
	(docPosition === "static" ? 0 : 10 + 20) +
	(bodyPosition === "static" ? 0 : 40 + 80 + 160);

// Check offsets
var absoluteOffset = $( "#absolute" ).offset();
assert.deepEqual(
	$( "#static" ).offset(),
	{ top: bodyContentOrigin, left: bodyContentOrigin },
	"offset of position:static element includes <html> and <body> box styles" );
assert.deepEqual( absoluteOffset, { top: origin + 20, left: origin + 20 },
	"offset of position:absolute element ignores box styles of position:static ancestors" );
$( "#absolute" ).offset( absoluteOffset );
assert.deepEqual( $( "#absolute" ).offset(), absoluteOffset, "offset() round-trips" );

// Reposition html and body, tracking origin adjustments given margin/border/padding
var originAdjust = 0;
$( docElem ).css( { top: "1.5em", left: "1.5em" } );
if ( docPosition !== "static" ) {
	originAdjust += 15 - (docPosition === "relative" ? 0 : 10);
}
$( body ).css( { top: "3em", left: "3em" } );
if ( bodyPosition === "fixed" || bodyPosition === "absolute" && docPosition === "static" ) {

	// html box styles no longer matter
	origin = 80 + 160;
	bodyContentOrigin = origin + 320;
	originAdjust = 30;
} else if ( bodyPosition !== "static" ) {
	originAdjust += 30 - (bodyPosition === "relative" ? 0 : 40);
}

// Recheck offsets
absoluteOffset = $( "#absolute" ).offset();
assert.deepEqual(
	$( "#static" ).offset(),
	{ top: bodyContentOrigin + originAdjust, left: bodyContentOrigin + originAdjust },
	"offset of position:static element respects ancestor positioning" );
assert.deepEqual(
	absoluteOffset,
	{ top: origin + originAdjust + 20, left: origin + originAdjust + 20 },
	"offset of position:absolute respects ancestor positioning" );
$( "#absolute" ).offset( absoluteOffset );
assert.deepEqual( $( "#absolute" ).offset(), absoluteOffset, "offset() still round-trips" );
```

And there could easily be ways to simplify even further.

anseki: Sorry, I have to learn English...
So, please check following:

- I interpreted [your comment in that issue](https://github.com/jquery/jquery/issues/3080#issuecomment-215738621) as that the code should be tested with changing `margin` and `border`. Did I mistake again?

- All `assert.deepEqual()`s that are passed  return value from `.offset()` failed.
I found this:
https://github.com/jquery/jquery/blob/master/external/qunit/qunit.js#L1628
`if ( a.constructor === b.constructor ) {`
Expected values that were made in the test code and return values from `.offset()` that were made in `iframe` have constructor in different namespaces (i.e. window), therefore `a.constructor === b.constructor` is `false`.
Return values from `.offset()` have to be copied to be checked by `assert.deepEqual()`.
`supportjQuery.extend({}, $( "#static" ).offset())`

- This test failed when `html{position}` is `static` and `body{position}` is `non-static`.
```js
assert.deepEqual( absoluteOffset, { top: origin + 20, left: origin + 20 },
    "offset of position:absolute element ignores box styles of position:static ancestors" );
```

For example, in a case `html{position:static} body{position:relative}`:
```
Expected: 

{
  "left": 300,
  "top": 300
}

Result: 

{
  "left": 330,
  "top": 330
}
```

I think, this code is incorrect:

```js
var origin = 0 +
    (docPosition === "static" ? 0 : 10 + 20) +
    (bodyPosition === "static" ? 0 : 40 + 80 + 160);
```

If `body` has `position:non-static`, the element is positioned relative to `body` without being affected by `position` of `<html>`. `body` was already positioned with `margin`, `border` and `padding` of `<html>`.
Therefore the code should be:

```js
var origin =
    bodyPosition !== "static" ? 10 + 20 + 40 + 80 + 160 : // `10 + 20 + 40` must be included
    docPosition !== "static" ? 10 + 20 : // relative to inside `<html>`
    0; // relative to document (i.e. outside `<html>`)
```

- This test failed when `html{position}` is not `static` or `relative`.
```js
assert.deepEqual(
    supportjQuery.extend({}, $( "#static" ).offset()),
    { top: bodyContentOrigin + originAdjust, left: bodyContentOrigin + originAdjust },
    "offset of position:static element respects ancestor positioning" );
```

I think, this code is incorrect:

```js
if ( docPosition !== "static" ) {
    originAdjust += 15 - (docPosition === "relative" ? 0 : 10);
}
```

If `<html>` has `position:non-static`, it is positioned with `margin` even if `potision` is other than `relative`.
Therefore the code should be:

```js
if ( docPosition !== "static" ) {
    originAdjust += 15;
}
```

- This test failed in some cases:

```js
assert.deepEqual(
    absoluteOffset,
    { top: origin + originAdjust + 20, left: origin + originAdjust + 20 },
    "offset of position:absolute respects ancestor positioning" );
```

This is solved by fixing `originAdjust` and `origin`.

anseki: The test code I wrote might do the almost same as your code, by removing code that changes `margin` and `border`.
But I didn't think that it should change `top` and `left` of `<html>` and `body`.
Therefore I think that I can't write PR. Maybe, I can't understand your request well by my very poor English.
Then, I hope that someone write PR.

anseki: BTW, why the test doesn't need to change `border`?
I think that this is PR to fix bug with `clientTop/clientLeft` (i.e. border).
At least, it should test in cases of that `<html>` has `border` and no `border`.

gibson042: > Sorry, I have to learn English...
So, please check following:

No problem. And it's not all your misunderstanding; my suggestions are changing as I learn new details of browser behavior around this issue.

> I interpreted [your comment in that issue](https://github.com/jquery/jquery/issues/3080#issuecomment-215738621) as that the code should be tested with changing `margin` and `border`. Did I mistake again?

It was a partial mistake. I wasn't saying that margin and border need to _change_, just that test cases should cover both zero and nonzero values. Further, we already have tests for the no-margin no-border no-padding cases, and I have since concluded that the "only margin" and "only border" cases are unnecessary, so this PR can focus on situations in which **all** the box styles are nonzero.

> Return values from `.offset()` have to be copied to be checked by `assert.deepEqual()`… `supportjQuery.extend({}, $( "#static" ).offset())`

:+1:

> If `body` has `position:non-static`, the element is positioned relative to `body` without being affected by `position` of `<html>`. `body` was already positioned with `margin`, `border` and `padding` of `<html>`.
Therefore the code should be:

> ```js
var origin =
    bodyPosition !== "static" ? 10 + 20 + 40 + 80 + 160 : // `10 + 20 + 40` must be included
    docPosition !== "static" ? 10 + 20 : // relative to inside `<html>`
    0; // relative to document (i.e. outside `<html>`)
```

I definitely mis-defined `origin`, and I agree that your fix is correct.

> I think, this code is incorrect:

> ```js
if ( docPosition !== "static" ) {
    originAdjust += 15 - (docPosition === "relative" ? 0 : 10);
}
```

> If `<html>` has `position:non-static`, it is positioned with `margin` even if `potision` is other than `relative`.
Therefore the code should be:

> ```js
if ( docPosition !== "static" ) {
    originAdjust += 15;
}
```

Again, a correct fix for my mistake.

> The test code I wrote might do the almost same as your code, by removing code that changes `margin` and `border`.
But I didn't think that it should change `top` and `left` of `<html>` and `body`.
Therefore I think that I can't write PR. Maybe, I can't understand your request well by my very poor English.
Then, I hope that someone write PR.

I really think we're close. My complaint wan't about what you were testing, just how it was being tested. And as shown above, you clearly have the right insights.

Comment (anseki) [inline test/unit/offset.js]: Thank you very much for your very attentive care.
I understood that the test doesn't have to change `border`.

I have more questions.

- Why does the code change the `top/left` of `html/body`?
I think that these properties affect `getBoundingClientRect()` method of browser, and `.offset()` method of jQuery calculates with a return value of the `getBoundingClientRect()` without considering `top/left`.
That is, if the result become incorrect by changing `top/left`, it is bug of `getBoundingClientRect()` method (i.e. browser), not jQuery.
Also, `top/left` of `html/body` are usually not changed.

- Why is the code that passes a return value of `.offset()` to another element unnecessary?
https://github.com/anseki/jquery/blob/offset-html-border/test/unit/offset.js#L650
This is not "round-trip". I thought that the test should compare an element that was positioned by `.offset()` and another element that was positioned by native CSS, for `setOffset()`.
`setOffset()` must position an element at the same position as another element that was positioned by various CSS properties (`position`, `border`, etc.).
I think that a cases of an element that has `position:static` are important because `setOffset()` calculates distance from current coordinates. https://github.com/jquery/jquery/issues/3080#issuecomment-215325352
That is, it compares coordinates that was got without `.offset` and coordinates that was changed by `.offset` that should not be affected by border of `<html>`.

**Comment by anseki** [comment]
I reconsidered about setter.
The setter moves an element by using distance between a current position and a requested position. It doesn't need calculation for getting coordinates relative to the document of requested position. And then, it change coordinates relative to a closest ancestor element that is positioned.
And, it uses getter to get current position.

Therefore if getter works correctly, things that affect setter are:
- Whether target element has `position:static`.
- Whether target element is already positioned. (it has `position:absolute-or-fixed; top:not-auto; left:not-auto`)

I think that the test doesn't need to change `margin` and `border` of `<html>` and `<body>` for testing the setter, after testing the getter.
The test needs only `position:static` element and `position:non-static` element that are positioned relative to another element that is not document.

I will rewrite the test later.

**Comment by anseki** [comment]
BTW, I found another issue.
As it happens, the issue was found by the test I said [above](https://github.com/jquery/jquery/pull/3096#discussion_r61992723) (the test that passes a return value of `.offset()` to `.offset()` with another element). That is, the another element was not positioned at the coordinates of first element.
That test failed in only browsers other than Firefox, and just `2px` (border-width) was lost from the result. Therefore I thought that this seems that bug of Firefox again. But it is not.
The test failed only first time in each `iframe` with `body{position:static}`.
Strange to say, the test succeeded when I specified `top` and `left` of `#marker2` (i.e. "another element" above).
https://github.com/anseki/jquery/blob/8fba23519ca5dfcd219ef2c66fe5eda4d3059023/test/data/offset/rel-doc.html#L12

I found that `setOffset()` gets current position by using `.position()`, when the element is not positioned (i.e. `top/left` are `auto`). And `getComputedStyle()` in Firefox seems to return `top/left` that are `*px` (real pixels, not `auto`) even if those were not specified. This also may be another bug of Firefox.
And I found issue that is unrelated to that in `.position()` and `offsetParent`. Since `.position()` is not used in Firefox, the test succeeded in Firefox.
In https://github.com/jquery/jquery/issues/3080, I said that the issue affects other methods, and I said that `.position()` method also uses `.offset()`. Those will be solved by this PR.
The issue in `.position()` and `offsetParent` that was found today is not that issue.

I send this as "new issue" later.

File: test/unit/offset.js (hunk 2 above)
Comment (gibson042) [inline test/unit/offset.js]: > Why does the code change the `top/left` of `html/body`?
I think that these properties affect `getBoundingClientRect()` method of browser, and `.offset()` method of jQuery calculates with a return value of the `getBoundingClientRect()` without considering `top/left`.
That is, if the result become incorrect by changing `top/left`, it is bug of `getBoundingClientRect()` method (i.e. browser), not jQuery.

If there are such bugs, we need to know so we can work around them (or document our failure to do so).

> Also, `top/left` of `html/body` are usually not changed.

Nor do html or body elements usually have borders. The point is verifying that our functions perform as claimed.

> Why is the code that passes a return value of `.offset()` to another element unnecessary?
…
I think that a cases of an element that has `position:static` are important because `setOffset()` calculates distance from current coordinates. https://github.com/jquery/jquery/issues/3080#issuecomment-215325352

I agree. We should do both `$( "#absolute" ).offset( absoluteOffset )` and `$( "#static" ).offset( absoluteOffset )` and expect the subsequent `.offset()` to match `absoluteOffset`.

**Comment by gibson042** [comment]
> I think that the test doesn't need to change `position` of `<html>` and `<body>` for testing the setter, after testing the getter.
The test needs only `position:static` element and `position:non-static` element that are positioned relative to another element that is not document.

We don't need to _change_ `<html>` or `<body>` position, but we do need to verify correct behavior when they are nonzero, for the same reason that we need to verify correct behavior when those elements have margin/border/padding.

File: test/data/offset/rel-doc.html [hunk 3]
```diff
@@ -0,0 +1,54 @@
...
+			html { background-color: green; } /* for human */
+			body { background-color: yellow; } /* for human */
+			#static { background-color: blue; } /* for human */
+			#absolute { background-color: aqua; } /* for human */
+			#child-static { background-color: darkviolet; } /* for human */
+			#child-absolute { background-color: violet; } /* for human */
+			html {
+				margin: 1px;
+				border-width: 2px;
+				padding: 4px;
```
gibson042: It is important to me that all of these units are not pixels.

anseki:
I see.
Since you updated your test code, I think that replacing my test code with yours is better than adding code which changes `top/left`. It makes the work progress quickly.
As I said, I should not have written PR because I could not read your request precisely many times.
Anyway, I replace the test later.

test/unit/offset.js
```diff
@@ -568,4 +568,72 @@ QUnit.test( "iframe scrollTop/Left (see gh-1945)", function( assert ) {
...
+
+				// Check offsets
+				var absoluteOffset = supportjQuery.extend( {}, $( "#absolute" ).offset() );
+				assert.deepEqual(
+					supportjQuery.extend( {}, $( "#static" ).offset() ),
+					{ top: bodyContentOrigin, left: bodyContentOrigin },
+					"offset of position:static element includes <html> and <body> box styles" );
+				assert.deepEqual( absoluteOffset, { top: origin + 20, left: origin + 20 },
+					"offset of position:absolute element ignores box styles of position:static ancestors" );
+				$( "#absolute, #static" ).offset( absoluteOffset );
```
timmywil: `$` -> `jQuery`

gibson042:
As discussed at today's meeting, we'll hold off on changing the `.offset` implementation until #3107 is fixed. And the critical piece here will be carrying forward the thorough testing of positioning and box properties affecting documentElement and body.

Issue #3071, opened 2016-04-21
Event handlers with invalid selectors should be caught at attach time


fsateler:
Otherwise, they become very difficult to debug.

See [this jsfiddle](https://jsfiddle.net/v4emgtLs/1/) for an example. Attaching an event handler with an invalid selector `div:not` causes an exception to be raised at event time. This makes it very difficult to debug, as the trace of the error will usually only include jquery code.

I suggest jQuery should throw an exception at attach time instead, which makes it fairly easy to detect the faulty code.

gibson042:
I'm normally opposed to explicitly introducing exceptions, but eliminating this arbitrarily long delay between mistake and feedback seems really valuable. I recommend adding `if ( selector ) { jQuery.find.matchesSelector( elem, selector ); }` in [`jQuery.event.add`](https://github.com/jquery/jquery/blob/91850ecbbe04ad8f5d89dc050aa1d9002047c435/src/event.js#L120). @fsateler, care to submit a PR?

markelog:
I think we already received similar issue before - @dmethvin?

I would tend to agree with assumption here though.

fsateler:
@markelog Indeed, I searched but did not find it. I now see it here: https://github.com/jquery/jquery/issues/2720 

The consensus there seems to be that the check should not be added. A claim was made that the console has the erroneous selector. This is not the case if the selector instead triggers a null reference exception or other error in jquery code (as the `div:not` selector does, this is probably a separate bug that needs fixing). Moreover, if the event was supposed to catch a special condition and prevent navigation, debugging becomes doubly hard. 

Should I prepare and submit a PR?

dmethvin:
I think it would be easier to see the advantages and drawbacks if there was a PR, so I wouldn't mind seeing a proposed solution with unit tests. One I can think of offhand is that custom selector extensions might not yet be in place, but that is probably rare.

fsateler:
ok, will prepare a PR

markelog:
@fsateler cool, check it out - http://contribute.jquery.org/

fsateler:
Sorry for the newbie question, but... how do I run the testsuite? `npm test` does not appear to run the unit tests...

mgol:
@fsateler You need to have a working Apache+PHP setup (see https://github.com/jquery/jquery#running-the-unit-tests) and one domain pointed at the jQuery repository. Then open http://your-jquery-domain/test in a browser you want to check.

gibson042:
[...]

Issue #2824, opened 2016-01-14
1.12.0 vs 1.11.2 selectors with # broken


cosmicnet:
The following is a perfectly valid selector under 1.11.2:
$('.class a[href=#anchor]');

But fails under 1.12.0, and requires quotes:
$('.class a[href="#anchor"]');

Likely related to:
https://github.com/jquery/jquery/commit/41f522ace1518f7f6f22f2e227350b0bdf78e62b

dmethvin:
Yes, but it was never a valid standard CSS selector, anything with special characters or spaces has to be quoted in the selector. Accepting it in previous versions was a bug. 

> Attribute values must be CSS identifiers or strings. -- https://www.w3.org/TR/css3-selectors/#attribute-selectors

cosmicnet:
Whether you regard this as a previous bug or not, it is a breaking change.

timmywil:
Fixing a bug almost always breaks something.

cosmicnet:
The release notes specify this is a non-breaking change. But you are changing the way the selector syntax works and has worked for a long time. There will be a LOT of code in the wild that uses this construct.

Your own test suite previous proved the accepted # syntax:
https://github.com/jquery/jquery/commit/41f522ace1518f7f6f22f2e227350b0bdf78e62b
You had to update it to work with the new syntax. Yet nowhere do you indicate that the selector syntax has now changed. Where is the issue for the bug? Where is the release note stating the change in behaviour?

mgol:
@cosmicnet Every behavior change may break code that relies on unspecified behavior; if we treated all of them as breaking changes we couldn't have non-major releases at all. Semver only applies to what's documented, i.e. what's on https://api.jquery.com/. We've never promised there that such a selector works, hence by definition changing that is not a breaking change. We did think that it might break some incorrect code, though and that's why we didn't introduce this change in a patch release but in a minor one, even though semver permits us to do it in a patch release.

dmethvin:
@cosmicnet specifically what do you think we need to do here?

Are you advocating that we revert this change? I don't think we can do that. We are trying to parse valid CSS selectors per the spec and allowing these kind of exceptions (even undocumented) makes that difficult if not impossible. It also causes the selector to be rejected by the `querySelectorAll` path which results in performance degradation.

[...]

PR #3097, opened 2016-05-02
Event: evaluate selector at add time

fsateler:
### Summary ###

This ensures events with invalid selectors throw at attach time instead
of event time. This is done by instantiating a Sizzle for the selector.

`jQuery.find()` is used instead of `jQuery.find.matchesSelector` (as suggested in https://github.com/jquery/jquery/issues/3071#issuecomment-213254660) , because `matchesSelector` requires a `DOMElement`, which won't be if the element is actually document.

Fixes: #3071 

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -120,6 +120,11 @@ jQuery.event = {
 			selector = handleObjIn.selector;
 		}
 
+		// If the selector is invalid, throw any exceptions at attach time
+		if ( selector ) {
+			jQuery.find( selector, elem );
+		}
+
 		// Make sure that the handler has a unique ID, used to find/remove it later
 		if ( !handler.guid ) {
 			handler.guid = jQuery.guid++;
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -1291,21 +1291,13 @@ QUnit.test( "Delegated events in SVG (#10791; #13180)", function( assert ) {
 
 QUnit.test( "Delegated events with malformed selectors (#3071)", function( assert ) {
 	assert.expect( 2 );
-	try {
-		jQuery( "#qunit-fixture" ).on( "click", "div:not", function () { } );
-		assert.ok( false, "malformed selector throws on attach" );
-	}
-	catch ( e ) {
-		assert.ok( true, "malformed selector throws on attach" );
-	}
 
-	try {
-		jQuery( "#qunit-fixture" ).click();
-		assert.ok( true, "malformed selector does not throw on event" );
-	}
-	catch ( e ) {
-		assert.ok( false, "malformed selector does not throw on event" );
-	}
+	assert.throws( function () {
+		jQuery( "#qunit-fixture" ).on( "click", "div:not", function () { } );
+	}, null, "malformed selector throws on attach" );
+
+	jQuery( "#qunit-fixture" ).click();
+	assert.ok( true, "malformed selector does not throw on event" );
 
 	jQuery( "#qunit-fixture" ).off( "click" );
 } );
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -1289,6 +1289,27 @@ QUnit.test( "Delegated events in SVG (#10791; #13180)", function( assert ) {
 	jQuery( "#qunit-fixture" ).off( "click" );
 } );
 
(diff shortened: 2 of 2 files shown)


src/event.js
```diff
@@ -120,6 +120,11 @@ jQuery.event = {
 			selector = handleObjIn.selector;
 		}
 
+		// If the selector is invalid, throw any exceptions at attach time
+		if ( selector ) {
+			jQuery.find( selector, elem );
```
timmywil: This may or may not require a context `elem.ownerDocument`. If it's just testing validity, perhaps not, but can anyone think of a use case where the document would affect validity?

dmethvin: None that I can think of.  The only case I can think of that breaks is if a selector extension was added after the event was attached here, but before an event was fired. That's probably not common?

dmethvin: Also note that this is inside `jQuery.event.add()` and the API surface is `jQuery.fn.on()`. If someone does `$("tr").on("mouseover" ...)` on a table with 100 elements it will check the selector 100 times. The parsed selector is cached and the selection set is small obviously, but is there a downside to moving this up to [function on()](https://github.com/fsateler/jquery/blob/10dc9f515a97c69ac674cfa19298ddd1e5cecbba/src/event.js#L91)?

test/unit/event.js
```diff
@@ -1289,6 +1289,27 @@ QUnit.test( "Delegated events in SVG (#10791; #13180)", function( assert ) {
...
 } );
 
+QUnit.test( "Delegated events with malformed selectors (#3071)", function( assert ) {
+	assert.expect( 2 );
+	try {
+		jQuery( "#qunit-fixture" ).on( "click", "div:not", function () { } );
+		assert.ok( false, "malformed selector throws on attach" );
+	}
+	catch ( e ) {
+		assert.ok( true, "malformed selector throws on attach" );
```
gibson042: I think this would be a little clearer with [`assert.throws`](http://api.qunitjs.com/throws/), but otherwise :+1:.

src/event.js
fsateler: The only problem I see is that one would have to replicate the selector parameter parsing. Maybe a better approach would be to check if there is already an event handler with such a selector in the `handlers` list?

I may be wrong, but I think event handlers with selectors usually attach to a single object, as the whole point is to have to avoid listing all possible elements (which may be a lot, or not exist yet).

test/unit/event.js
fsateler: That would leave this asymmetric with the below call (I don't see a `notThrows` assert method). I you think that doesn't matter, I can change that.

dmethvin: If the test throws and you don't expect it to, you find out pretty quickly. :smile:

fsateler: OK, so I just let qUnit handle the exception.

To fix these issues, should I force-push, or open a new PR ?

gibson042: I'd recommend just adding one or more new commits and pushing... no `-f` required.

dmethvin: Generally you can just add a new commit that says `squash! (comment about the change)` and we will squash it when the PR lands. That way the commits here in the PR and the comments show the history of how we got to the squashed PR.

src/event.js
dmethvin: @fsateler You are right, it won't be common for people to make a call to attach delegated events on a lot of elements at once. Also, this change does have the potential to throw errors where they did not before, in the case where the selector was invalid but no event ever occurred. We learned in gh-2824 that even obscure cases like that can cause trouble. So even if we moved this to `on()` it should be executed only if there is at least one element.

In short, I think you have this check in a good place.

dmethvin:
@fsateler can you make that `assert.throws` tweak to the tests? I think this is good to land otherwise.

Anyone have thoughts on whether and how this should be documented? I am thinking it doesn't need docs, this behavior may have been what everyone expected anyway.

fsateler:
Yes, I've just been busy during this week with $work. I'll push the change during the weekend.

fsateler:
Updated for the `throws` change.

gibson042:
Thanks, @fsateler! As it turns out, `matchesSelector` was necessary after all.

Issue #3102, opened 2016-05-03
Make jQuery.when always async


markelog:
See discussion in https://github.com/jquery/jquery/issues/3100.

Making it a blocker at the moment for 3.0, can provide a pull for that. 

If there strong motivation to not deal with it, feel free to close.

marcandre:
> It's important to realize that jQuery.when is polymorphic—multi-argument calls being analogous to ES6 `Promise.all` and other calls being analogous to ES6 `Promise.resolve`.

Thanks for the replies.

I'd like to point out that `$.when(one_argument)` is *not* really analogous to `Promise.resolve`, in that it currently returns a wrapped promise while `Promise.resolve` doesn't wrap existing promises:

```
p = Promise.resolve(42)
p === Promise.resolve(p) // => true
j = $.when(42)
j === $.when(j) // => false
```

The argument about the "reliance on `.then` to unwrap secondary thenables" seems to apply to `Promise.all` at least as much when it has more than one argument than when it has a single one.

What I mean to say is that if I had gotten ABDC in the above example (instead of ACDB), then I could understand the reasoning (and disagree with the decision), but currently the justification and behavior seems reversed!

ABCD: Synchronous behavior (what I'm secretly hoping for)
ADBC: Asynchronous behavior (I would secretly disapprove)
ABDC: Inconsistent behavior of single argument vs multiple, with single promise treated specially like `Promise.resolve` (I would disapprove)
ACDB: Inconsistent behavior where single promise is treated differently and illogically (current behavior to which I can only object)

I think there's clearly a bug currently:

    $.when(42).state() // => 'pending', should be 'resolved'

In closing, I'd ask naively: why not have two different methods? `$.when` could be `Promise.all`, while `$.resolve` would be `Promise.resolve`.
`resolve` would coerce its sole argument into a promise.
`when` coerces them via `resolve` if need be. It could keep the synchronous behavior of `done` intact when the given promises have `state` method available.

markelog:
> $.when(42).state()

This would execute asynchronously, so it seems `state` returns correct result. 

> ADBC: Asynchronous behavior (I would secretly disapprove)

That would be with the spirit of the spec and in line with other implementations. 

> why not have two different methods?

I guess we could consider adding new method, but definitely not for 3.0

gibson042:
[...]

Issue #3100, opened 2016-05-03
$.when should always be synchronous when possible


marcandre:
IIC, jQuery 3.0 will have `then` be A+ compatible in that callbacks are always executed out of context (e.g with `setTimeout`), while `done` will continue to be executed synchronously whenever possible (eager).

I was double checking that `$.when` kept the eager behavior and discovered that it does, unless there's a single argument:

```
p = $.Deferred()
p.resolveWith(42);
p.done(function() {console.log('A')});
$.when(p).done(function() {console.log('B')});
$.when(p, p).done(function() {console.log('C')});
console.log('D');
```

I was hope to get ABCD, given that the promised is already resolved when the `done` are called, and dreaded ADBC if `$when` was no longer eager. I get worse though. Console is ACDB, which is not acceptable; `$.when(p)` and `$.when(p, p)` should behave the exact same way.

FWIW, I personally disagree with Promise/A+'s rule 2.2.4 and I hope jQuery keeps all behavior eager (except for `then` to comply with A+) and gives us an eager version of `then` (could be `pipe`, but with the correct semantics for callbacks return values).

I did not investigate any further, but it looks related to #3059.

markelog:
@gibson042 Did we discuss this before? Did we want to make `done` async or sync? I agree arguments shouldn't change affect on it.

gibson042:
It's important to realize that `jQuery.when` is polymorphic—multi-argument calls being analogous to ES6 `Promise.all` and other calls being analogous to ES6 `Promise.resolve`. It's unfortunate that their similarity led to a shared function name in jQuery, but they _are_ distinct. And although `done` itself remains eager, general-case `jQuery.when( value )` is necessarily not because of its reliance on `.then` to unwrap secondary thenables as `Promise.resolve` does. And I don't want to introduce a `jQuery.when( nonThenable )` exception, which really would introduce an inconsistency.

I also disagree with Promises/A+ 2.2.4, but that is insufficient justification to abandon compatibility which has been demanded for so long.

dmethvin:
I know @gibson042 has rightly railed about this several times, I'm too lazy to dig it up. 

[...]

PR #3105, opened 2016-05-05
Deferred: Make jQuery.when synchronous when possible

gibson042:
Closes gh-3102
Fixes gh-3100

I'm ambivalent about this, but it doesn't cost much.

```
   raw     gz Sizes
264011  78513 dist/jquery.js
 86318  30014 dist/jquery.min.js

   raw     gz Compared to master @ 7f2ebd2c4dea186d7d981b939e6e2983a9d7f9c1
  +110    +22 dist/jquery.js
   +44    +17 dist/jquery.min.js
```

### DIFF
--- a/test/unit/deferred.js
+++ b/test/unit/deferred.js
@@ -1066,16 +1066,20 @@ QUnit.test( "jQuery.when(...) - opportunistically synchronous", function( assert
 	var when = "before",
 		resolved = jQuery.Deferred().resolve( true ),
 		rejected = jQuery.Deferred().reject( false ),
-		validate = function() {
-			assert.equal( when, "before" );
+		validate = function( label ) {
+			return function() {
+				assert.equal( when, "before", label );
+			};
 		},
 		done = assert.async( 5 );
 
-	jQuery.when().done( validate ).always( done );
-	jQuery.when( when ).done( validate ).always( done );
-	jQuery.when( resolved ).done( validate ).always( done );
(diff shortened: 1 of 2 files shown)


test/unit/deferred.js
```diff
@@ -1049,3 +1049,24 @@ QUnit.test( "jQuery.when - notify does not affect resolved", function( assert )
...
+
+	assert.expect( 5 );
+
+	var when = "before",
+		resolved = jQuery.Deferred().resolve( true ),
+		rejected = jQuery.Deferred().reject( false ),
+		validate = function() {
+			assert.equal( when, "before" );
+		},
+		done = assert.async( 5 );
```
dmethvin: I always forget you can do this.

dmethvin:
I'm okay with it. It actually simplifies the docs and upgrade guide I think?

gibson042:
I think so, just by virtue of changing less.

test/unit/deferred.js
timmywil: Didn't it require a QUnit update?

timmywil:
I guess I'm okay with this.

test/unit/deferred.js
gibson042: We've only had it since January. It was added in QUnit 1.20.

marcandre:
Great! Eager to run the spec suite of Parsley with this in. Thanks for the consideration and good discutions :smile:

PR #3113, opened 2016-05-11
Deferred: Give better stack diagnostics on exceptions

dmethvin:
### Summary ###

Ref gh-2736

The exception stack has the name of the immediately outer function where the
exception occurred, which can be very handy for tracing errors. Since we already
have the exception object we might as well use it.

I was debugging some code that had errors inside a document ready function and this would have been really useful to have.

### Checklist ###

### DIFF
--- a/src/deferred/exceptionHook.js
+++ b/src/deferred/exceptionHook.js
@@ -11,7 +11,7 @@ var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
 
 jQuery.Deferred.exceptionHook = function( error, stack ) {
 
-	// Support: IE 9 only
+	// Support: IE 8 - 9 only
 	// Console exists when dev tools are open, which can happen at any time
 	if ( window.console && window.console.warn && error && rerrorNames.test( error.name ) ) {
 		window.console.warn( "jQuery.Deferred exception: " + error.message, error.stack, stack );
--- a/src/deferred/exceptionHook.js
+++ b/src/deferred/exceptionHook.js
@@ -11,10 +11,10 @@ var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
 
 jQuery.Deferred.exceptionHook = function( error, stack ) {
 
-	// Support: IE 8 - 9 only
+	// Support: IE 9 only
 	// Console exists when dev tools are open, which can happen at any time
 	if ( window.console && window.console.warn && error && rerrorNames.test( error.name ) ) {
-		window.console.warn( "jQuery.Deferred exception: " + error.message, stack );
+		window.console.warn( "jQuery.Deferred exception: " + error.message, error.stack, stack );
 	}
 };
 
--- a/test/unit/deferred.js
+++ b/test/unit/deferred.js
@@ -542,28 +542,30 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
 	} catch ( _ ) {}
 } );
 
+// Test fails in IE9 but is skipped there because console is not active
 QUnit[ window.console ? "test" : "skip" ]( "jQuery.Deferred.exceptionHook", function( assert ) {
 
-	assert.expect( 1 );
+	assert.expect( 2 );
 
 	var done = assert.async(),
 		defer = jQuery.Deferred(),
 		oldWarn = window.console.warn;
 
-	window.console.warn = function( msg ) {
+	window.console.warn = function() {
 
 		// Support: Chrome <=41 only
 		// Some Chrome versions newer than 30 but older than 42 display the "undefined is
 		// not a function" error, not mentioning the function name. This has been fixed
 		// in Chrome 42. Relax this test there.
 		// This affects our Android 5.0 & Yandex.Browser testing.
-		var oldChromium = false;
+		var msg = Array.prototype.join.call( arguments, " " ),
+			oldChromium = false;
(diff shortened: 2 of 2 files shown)


test/unit/deferred.js
```diff
@@ -580,14 +582,17 @@ QUnit[ window.console ? "test" : "skip" ]( "jQuery.Deferred.exceptionHook", func
...
+		jQuery.thisDiesToo();
+	} ).then( null, function( ) {
 		window.console.warn = oldWarn;
 		done();
 	} );
 
 	defer.resolve();
 } );
 
+// Test fails in IE9 but is skipped there because console is not active
```
mgol: Does it fail with DevTools open? If so, why?

mgol:
So, the idea is that this stack trace is nowhere near that useful than the original one which would be passed by your [jquery-deferred-reporter](https://github.com/dmethvin/jquery-deferred-reporter) plugin but it still provides some useful info?

src/deferred/exceptionHook.js
```diff
@@ -11,10 +11,10 @@ var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
 
 jQuery.Deferred.exceptionHook = function( error, stack ) {
 
-	// Support: IE 8 - 9 only
```
mgol: Let's not change this comment. We've decided some time ago to not remove info about workarounds for now non-supported browsers as long as we still need the workaround for other, supported ones.

mgol:
I don't see the name in the stack trace for the handler named `barf` in Chrome 50, that's what I see:
```
jQuery.Deferred exception: jQuery.thisDiesToo is not a function TypeError: jQuery.thisDiesToo is not a function
    at Array.barf (<anonymous>:12:9)
    at mightThrow (http://jquery.l/dist/jquery.js:3505:29)
    at process (http://jquery.l/dist/jquery.js:3573:12) undefined
```
It shows `Array.barf` which seems weird.

mgol:
Firefox, on the other hand, show just the first error (Chrome shows both, with stack traces with your patch):
```
TypeError: jQuery.barf is not a function
```
with no stack trace. I'm doing the testing with this code:
```js
var defer = jQuery.Deferred();

jQuery.when(
	defer.then( function() {
		jQuery.barf();
	} ).then( null, jQuery.noop ),

	defer.then( function() {
		throw new Error( "Make me a sandwich" );
	} ).then( null, jQuery.noop )
).then( function barf( ) {
	jQuery.thisDiesToo();
} );

defer.resolve();
```

mgol:
Interesting; even with:
```js
jQuery.Deferred.exceptionHook = function( error, stack ) {
	console.log("jQuery.Deferred exception: " + error.message, error.name, error.stack, stack );
};
```
I see 3 errors in Chrome with this test but only one error message in Firefox and the Firefox one doesn't seem to come from jQuery as it doesn't begin with `jQuery.Deferred exception:`.

mgol:
It all works fine for me in Firefox beta or newer so perhaps only the current version has problems.

src/deferred/exceptionHook.js
dmethvin: OK, for some reason I thought we were doing the opposite.

test/unit/deferred.js
dmethvin: IE9 doesn't have `error.stack`.

test/unit/deferred.js
```diff
@@ -542,28 +542,30 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
...
+	assert.expect( 2 );
 
 	var done = assert.async(),
 		defer = jQuery.Deferred(),
 		oldWarn = window.console.warn;
 
-	window.console.warn = function( msg ) {
+	window.console.warn = function() {
 
 		// Support: Chrome <=41 only
```
dmethvin: We've dropped Yandex now, but I guess this is still needed for Android 5.0?

mgol: Yeah, although technically that shouldn't matter as Android 5.0 has both auto-updatable browser (Chrome) & WebView. It's just that BrowserStack doesn't update Chrome in Android images (I'm not sure if that's even possible in emulators). The solution would be to move to Android 6.0 and drop 5.0 as we're really interested in testing Chrome for Android, not specific Android versions (it used to be different as Android Browser versions were tied to Android versions as is still the case on iOS) but last time I checked local testing didn't work on Android 6.0 on BrowserStack. I'll need to re-check, if it worked we'd be able to drop these workarounds.

test/unit/deferred.js
mgol: Could you add a check for `(new Error()).stack` here so that the tests work even with IE9 DevTools enabled?

dmethvin: I'll give that a try!

dmethvin: Oh wait, I just remembered:
>Different browsers set this value at different times. For example, Firefox sets it when creating an Error object, while PhantomJS sets it only when throwing the Error, and MSDN docs also seem to match the PhantomJS implementation. -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack

So the test would need to be more elaborate. I think the current check is less likely to skip by accident at least? Everyone but IE9 has a console.

mgol: Yeah, I thought it'd be simpler. 😅 It's OK then.

dmethvin:
Closed by 07c11c03cc7f6d182eab9b50e2f3908ee397f70f

Issue #3111, opened 2016-05-10
[2.2.x] Regression with focusin handling on Firefox


avdv:
** Description**

The cursor is missing or misplaced in an input box with an focusin handler attached when returning `false` from the handler.

Affected is Firefox (I'm using 46.1).

See emberjs/ember.js#13112

** Link to test case **

Courtesy of @Serabe (https://github.com/emberjs/ember.js/issues/13112#issuecomment-217957434):

http://codepen.io/Serabe/pen/PNVXdd?editors=0010

This codepen uses jQuery 2.2.2 but it also breaks with 2.2.3. It works alright with jQuery 2.1.4

mgol:
Confirmed.

mgol:
Confirmed. Thanks for the report!

mgol:
The test case using the dev version of jQuery 2.x: http://codepen.io/mgol/pen/JXzjJV?editors=0010

mgol:
The first bad commit is 6df669f0fb87cd9975a18bf6bbe3c3548afa4fee (cc @markelog).

mgol:
jQuery 1.12.x is fine: http://codepen.io/mgol/pen/ONqPyO?editors=0010

Perhaps we need to backport the workaround from 37c3d087824b51e742126252ced8be592858102d? We should test that as well.

mgol:
(by `2.x-only` I mean it's broken on `master` as well: http://codepen.io/mgol/pen/zqbxrO?editors=0010)

markelog:
Will check it out.

Sounds like a blocker?

mgol:
Yeah, we should get that in 2.2.4 IMO.

mgol:
@markelog Please mention gh-3111 in the PR description so that it's easy to go to the issue.

markelog:
Currently, this fix landed only in master branch i.e. 3.x, not sure if this constitutes new 1.x/2.x release.

Marking it for 3.0 only for now

timmywil:
While this should also be landed in master, this fix is mainly for the 2.x branch.

markelog:
Landed in 2.x

markelog:
I think `2.x-only` is confusing here though

timmywil:
It's just to denote that it does not apply to 1.x. Maybe there's a better name for that, but soon we won't need the label at all.

PR #3121, opened 2016-05-16
Events: don't execute native stop(Immediate)Propagation from simulation

markelog:
In Firefox, called `stop(Immediate)Propagation` methods,
in capturing phase prevents receiving focus

See https://codepen.io/markelog/pen/pyBXOr.

/cc @bzbarsky

Probably not spec violation, but still pretty weird...

Fixes gh-3111

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -558,13 +558,14 @@ jQuery.Event.prototype = {
 	isDefaultPrevented: returnFalse,
 	isPropagationStopped: returnFalse,
 	isImmediatePropagationStopped: returnFalse,
+	isSimulated: false,
 
 	preventDefault: function() {
 		var e = this.originalEvent;
 
 		this.isDefaultPrevented = returnTrue;
 
-		if ( e ) {
+		if ( e && !this.isSimulated ) {
 			e.preventDefault();
 		}
 	},
@@ -573,7 +574,7 @@ jQuery.Event.prototype = {
 
 		this.isPropagationStopped = returnTrue;
 
-		if ( e ) {
+		if ( e && !this.isSimulated ) {
 			e.stopPropagation();
 		}
 	},
@@ -582,7 +583,7 @@ jQuery.Event.prototype = {
 
 		this.isImmediatePropagationStopped = returnTrue;
 
-		if ( e ) {
+		if ( e && !this.isSimulated ) {
 			e.stopImmediatePropagation();
 		}
 
--- a/src/event/trigger.js
+++ b/src/event/trigger.js
@@ -150,6 +150,7 @@ jQuery.extend( jQuery.event, {
 	},
 
 	// Piggyback on a donor event to simulate a different one
+	// Used only for `focus(in | out)` events
 	simulate: function( type, elem, event ) {
 		var e = jQuery.extend(
 			new jQuery.Event(),
@@ -157,27 +158,10 @@ jQuery.extend( jQuery.event, {
 			{
 				type: type,
 				isSimulated: true
-
-				// Previously, `originalEvent: {}` was set here, so stopPropagation call
-				// would not be triggered on donor event, since in our own
-				// jQuery.event.stopPropagation function we had a check for existence of
-				// originalEvent.stopPropagation method, so, consequently it would be a noop.
-				//
-				// But now, this "simulate" function is used only for events
-				// for which stopPropagation() is noop, so there is no need for that anymore.
-				//
-				// For the compat branch though, guard for "click" and "submit"
-				// events is still used, but was moved to jQuery.event.stopPropagation function
-				// because `originalEvent` should point to the original event for the constancy
-				// with other events and for more focused logic
 			}
 		);
 
 		jQuery.event.trigger( e, null, elem );
-
-		if ( e.isDefaultPrevented() ) {
-			event.preventDefault();
-		}
 	}
 
 } );
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -2790,6 +2790,81 @@ QUnit.test( "Donor event interference", function( assert ) {
 	jQuery( "#donor-input" )[ 0 ].click();
 } );
 
+QUnit.test(
+	"native stop(Immediate)Propagation/preventDefault methods shouldn't be called",
+	function( assert ) {
+		var userAgent = window.navigator.userAgent;
+
+		if ( !( /firefox/i.test( userAgent ) || /safari/i.test( userAgent ) ) ) {
+			assert.expect( 1 );
+			assert.ok( true, "Assertions should run only in Chrome, Safari, Fx & Edge" );
+			return;
+		}
+
+		assert.expect( 3 );
+
+		var checker = {};
+
+		var html = "<div id='donor-outer'>" +
+			"<form id='donor-form'>" +
+				"<input id='donor-input' type='radio' />" +
+			"</form>" +
+		"</div>";
+
+		jQuery( "#qunit-fixture" ).append( html );
+		var outer = jQuery( "#donor-outer" );
+
+		outer
+			.on( "focusin", function( event ) {
+				checker.prevent = sinon.stub( event.originalEvent, "preventDefault" );
+				event.preventDefault();
+			} )
+			.on( "focusin", function( event ) {
+				checker.simple = sinon.stub( event.originalEvent, "stopPropagation" );
+				event.stopPropagation();
+			} )
+			.on( "focusin", function( event ) {
+				checker.immediate = sinon.stub( event.originalEvent, "stopImmediatePropagation" );
+				event.stopImmediatePropagation();
+			} );
+
+		jQuery( "#donor-input" ).trigger( "focus" );
+		assert.strictEqual( checker.simple.called, false );
+		assert.strictEqual( checker.immediate.called, false );
+		assert.strictEqual( checker.prevent.called, false );
+
+		// We need to "off" it, since yes QUnit always update the fixtures
+		// but "focus" event listener is attached to document for focus(in | out)
+		// event and document doesn't get cleared obviously :)
+		outer.off( "focusin" );
+	}
+);
+
+QUnit.test(
+	"isSimulated property always exist on event object",
+	function( assert ) {
+		var userAgent = window.navigator.userAgent;
+
+		if ( !( /firefox/i.test( userAgent ) || /safari/i.test( userAgent ) ) ) {
+			assert.expect( 1 );
+			assert.ok( true, "Assertions should run only in Chrome, Safari, Fx & Edge" );
+			return;
+		}
+
+		assert.expect( 1 );
+
+		var element = jQuery( "<input/>" );
+
+		jQuery( "#qunit-fixture" ).append( element );
+
+		element.on( "focus", function( event ) {
+			assert.notOk( event.isSimulated );
+		} );
+
+		element.trigger( "focus" );
+	}
+);
+
 QUnit.test( "originalEvent property for Chrome, Safari, Fx & Edge of simulated event", function( assert ) {
 	var userAgent = window.navigator.userAgent;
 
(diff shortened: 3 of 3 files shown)


markelog:
> capturing phase prevents receiving focus

Or maybe not "focus" exactly, but "caret"

src/event.js
```diff
@@ -573,7 +574,7 @@ jQuery.Event.prototype = {
 
 		this.isPropagationStopped = returnTrue;
 
-		if ( e ) {
+		if ( e && this.executeOriginalPropagation ) {
```
markelog: I could have used `isSimulated` but choose not to, since such use wasn't intended for that property

test/unit/event.js
```diff
@@ -2810,6 +2810,7 @@ QUnit.test( "originalEvent property for Chrome, Safari, Fx & Edge of simulated e
 	jQuery( "#qunit-fixture" ).append( html );
 
 	jQuery( "#donor-outer" ).on( "focusin", function( event ) {
+		console.log(1);
```
staabm: Debug leftover

markelog: Thank you, fixed

src/event/trigger.js
```diff
@@ -150,26 +150,20 @@ jQuery.extend( jQuery.event, {
...
-				// For the compat branch though, guard for "click" and "submit"
-				// events is still used, but was moved to jQuery.event.stopPropagation function
-				// because `originalEvent` should point to the original event for the constancy
-				// with other events and for more focused logic
+				isSimulated: true,
+
+				// Support: Firefox <= 46
+				// In Firefox, called `stop(Immediate)Propagation` methods, in
+				// capturing phase, prevents receiving focus,
+				// so we stop only jQuery propagation not the native one instead
```
mgol: I'd tweak the grammar here:
```
// In Firefox, calling `stop(Immediate)Propagation` methods in
// the capture phase prevent receiving focus so we only stop
// the jQuery emulated propagation and not the native one.
```

test/unit/event.js
```diff
@@ -2821,6 +2821,46 @@ QUnit.test( "originalEvent property for Chrome, Safari, Fx & Edge of simulated e
...
 } );
 
+QUnit.test(
+	"navite stop(Immediate)Propagation methods should not be called from simulated event",
+	function( assert ) {
+		var userAgent = window.navigator.userAgent;
+
+		if ( !( /firefox/i.test( userAgent ) || /safari/i.test( userAgent ) ) ) {
+			assert.expect( 1 );
+			assert.ok( true, "Assertions should run only in Chrome, Safari, Fx & Edge" );
```
mgol: Why is that necessary?

Also, it might be simpler to say where we don't want to run them. :) It seems only in IE?

mgol:
@markelog Please mention gh-3111 in the PR description so that it's easy to go to the issue.

test/unit/event.js
```diff
@@ -2821,6 +2821,46 @@ QUnit.test( "originalEvent property for Chrome, Safari, Fx & Edge of simulated e
...
+			} )
+			.on( "focusin", function( event ) {
+				checker.immediate = sinon.stub( event.originalEvent, "stopImmediatePropagation" );
+				event.stopImmediatePropagation();
+			} );
+
+
+		jQuery( "#donor-input" ).trigger( "focus" );
+		assert.strictEqual( checker.simple.called, false );
+		assert.strictEqual( checker.immediate.called, false );
```
mgol: That's checking our implementation but not the reported issue. Could we add a test that makes sure the input actually received focus?

src/event/trigger.js
markelog: mm, not sure why current version doesn't suffice

test/unit/event.js
markelog: See test above this one

test/unit/event.js
markelog: Only with e2e tests

markelog:
> @markelog Please mention gh-3111 in the PR description so that it's easy to go to the issue.

I did in commit message, github already handles did the referencing

test/unit/event.js
markelog: btw, it actually tests reported issue

src/event/trigger.js
mgol: The current version has grammar errors.

test/unit/event.js
mgol: Ah, it's because it's emulated, ok.

test/unit/event.js
```diff
@@ -2821,6 +2821,46 @@ QUnit.test( "originalEvent property for Chrome, Safari, Fx & Edge of simulated e
 	jQuery( "#donor-input" ).trigger( "focus" );
 } );
 
+QUnit.test(
+	"navite stop(Immediate)Propagation methods should not be called from simulated event",
```
mgol: navite -> native

mgol: "from a simulated event" or "from simulated events"

mgol:
> I did in commit message, github already handles the referencing

To go from a PR to the issue you have to click on the commit tab, open the commit & only then click on the issue link. Having it in the description would make that easier (and that's the default when submitting a PR with the link to the issue already in the description).

test/unit/event.js
mgol: The issue was:

"The cursor is missing or misplaced in an input box with an focusin handler attached when returning false from the handler."

You could translate that to "Input doesn't receive focus if you invoke `preventDefault` or `stopPropagation` in the event handler, this doesn't test that, it only tests that the reason for this issue doesn't exist.

You'd test it only once we have e2e tests because focus-related tests are unreliable or why?

src/event/trigger.js
markelog: What errors?

test/unit/event.js
markelog: > Input doesn't receive focus

It does receive focus, but not fully, it only represented with absent caret, it is happening because if `stopPropagation` or `stopImmediatePropagation`, **not** `preventDefault` is called on original, so here we test that it doesn't happen.

We can't actually trigger the real event, since it can be done only with e2e tests.

markelog:
> To go from a PR to the issue you have to click...

There is reference to the issue, so that's that :), whereas discussion like this would be an example of waisted time

bzbarsky:
@smaug---- for Mozilla event stuff.  That said, if focus is done by a listener and propagation is stopped... then focus won't happen.

src/event.js
gibson042: In what sense is it different, though? The "Firefox <= 46" support comment could just move up here.

markelog: Comments would have to be duplicated in two places.

This is my preference to put it simply, i would like to be explicit on what this is and why it was introduced

mgol:
> There is reference to the issue, so that's that :), whereas discussion like this would be an example of waisted time

It puzzles me you like to spend so much time arguing against a simple thing I asked you to do just because it makes it easier for me (& perhaps others) instead of just adding it. `¯\_(ツ)_/¯` I'll just end here.

markelog:
@mgol you already posted reference to that issue with your **first** comment man, i really don't what else would you want, whereas following such rules which are not documented anywhere feels like nitpicking with the bad side. 

These things really do not matter, what matters is the bug which this pull trying to fix but instead of discussing that, you clobbering this issue with this nonsense - you could just post that link and notify me about your preference privately.

I really would just delete all these comments, feels like spam and waist of time to be honest to all involved.

> I'll just end here.

I would really like to not starting it at all.

test/unit/event.js
markelog: fixed

src/event/trigger.js
markelog: To note: i will happily fix all potential errors, if you would point them out

smaug----:
> That said, if focus is done by a listener and propagation is stopped... then focus won't happen.

This sounds like an old Gecko issue, IIRC there is a focus listener in editor code... Can't find the bug now though.

src/event.js
gibson042: > Then comments would have to be duplicated in two places.

Where else would you have them?

src/event/trigger.js
mgol: Methods (plural) prevent focus, not prevents, there are too many commas in the sentence, the action of calling the method prevents receiving focus and not the methods themselves, `only` sounds weird with `instead`, etc. You'll see that if you compare your comment with mine.

I've seen now that I made a mistake in my version as well: after changing `methods` to `calling` as the subject it's `prevents` again. Fixed.

src/event.js
markelog: `stopImmediatePropagation` and `stopPropagation` methods

src/event/trigger.js
```diff
@@ -150,26 +150,20 @@ jQuery.extend( jQuery.event, {
...
-				// But now, this "simulate" function is used only for events
-				// for which stopPropagation() is noop, so there is no need for that anymore.
-				//
-				// For the compat branch though, guard for "click" and "submit"
-				// events is still used, but was moved to jQuery.event.stopPropagation function
-				// because `originalEvent` should point to the original event for the constancy
-				// with other events and for more focused logic
+				isSimulated: true,
+
+				// Support: Firefox <= 46
```
markelog: @mgol changed, check it out

src/event.js
gibson042: I see. But rereading the [`simulate` comments](https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/event/trigger.js#L166-L167), it seems more like our folly than Mozilla's. I'd actually advocate for dropping the "Support" documentation here altogether in favor uniform simulated event treatment:
```js
stopPropagation: function() {
	var e = this.originalEvent;

	this.isPropagationStopped = returnTrue;

	// Stop propagation of original non-donor events
	if ( !this.isSimulated && e ) {
		e.stopPropagation();
	}
}
```
(and likewise for `stopImmediatePropagation`).

markelog: > But rereading the simulate comments,

If we would go to that route we would need the same thing for the `preventDefault`.

I think it was and it is reasonable, back then i checked with the spec of all donor events in case they had default actions, cancelable, etc; they all supposed to be noop - still holds true now too.

So it just waist of the space and judging by conversation of the parallel thread this is indeed mozilla's bug.

So these lines could be only "just in case" lines, without tests or use case, when/if mozilla people comes up with the fix

src/event/trigger.js
mgol: Sounds good, thx.

src/event.js
markelog: I can go either way though, the only requirement i would want to leave in is to save `originalEvent` property

test/unit/event.js
mgol: OK, got it, I thought it's being blurred in a way but I see I was wrong.

markelog:
Changes are made

test/unit/event.js
```diff
@@ -2790,6 +2790,81 @@ QUnit.test( "Donor event interference", function( assert ) {
...
+		assert.strictEqual( checker.prevent.called, false );
+
+		// We need to "off" it, since yes QUnit always update the fixtures
+		// but "focus" event listener is attached to document for focus(in | out)
+		// event and document doesn't get cleared obviously :)
+		outer.off( "focusin" );
+	}
+);
+
+QUnit.test(
```
dmethvin: Why would you want to assert this? Isn't it an implementation detail? Just thinking about the case where Firefox implements the event and we would want to take out the code, this would fail.

markelog: mm, i test here that property is present with `focus` event which is already implemented in firefox :). Here i wanted to make sure event has always consistent set of properties

dmethvin:
Is this ready to land? I'm okay with it since `simulate` is only used for one thing and this change fixes the issue. We should get the  `.4` releases out this week.

markelog:
Let say if there wouldn't be any objections for next 3-4 hours i will land this, if something comes up, we could always revert

timmywil:
👍

PR #3127, opened 2016-05-24
[WIP] Event: Use one native handler per jQuery handler

dmethvin:
This is definitely a v4 kind of thing with breaking changes and it still needs more work before it passes tests, but I'll post it here in case anyone has thoughts or comments.

EDIT: All my comments from below are in my fork, https://github.com/dmethvin/jquery/commit/113ce671659fea1e39774225cc1900076bcc4e01

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -104,10 +104,18 @@ jQuery.event = {
 
 	add: function( elem, types, handler, data, selector ) {
 
-		var handleObjIn, eventHandle, tmp,
+		var handleObjIn, tmp,
 			events, t, handleObj,
 			special, handlers, type, namespaces, origType,
-			elemData = dataPriv.get( elem );
+			elemData = dataPriv.get( elem ),
+			handleEventCommon = function( e ) {
+
+				// `this` for handleEvent is the handleObj
+				// Discard the second event of a jQuery.event.trigger() and
+				// when an event is called after a page has unloaded
+				return typeof jQuery !== "undefined" && jQuery.event.triggered !== e.type ?
+					jQuery.event.dispatch.apply( this, arguments ) : undefined;
+			};
 
 		// Don't attach events to noData or text/comment nodes (but allow plain objects)
 		if ( !elemData ) {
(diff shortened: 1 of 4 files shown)


src/event.js
```diff
@@ -187,17 +188,18 @@ jQuery.event = {
 
 				// Only use addEventListener if the special events handler returns false
 				if ( !special.setup ||
-					special.setup.call( elem, data, namespaces, eventHandle ) === false ) {
+					special.setup.call( elem, data, namespaces, handleObj ) === false ) {
 
 					if ( elem.addEventListener ) {
-						elem.addEventListener( type, eventHandle );
+						elem.addEventListener( type, handleObj );
```
mgol: This is still guarded by:
```js
if ( !( handlers = events[ type ] ) ) {
```
making it one `addEventListener` per event type... Or am I reading it wrong?

dmethvin: Hmm, I'll definitely look into that. Seems like a LOT of stuff should have broken.

mgol:
@dmethvin was it supposed to get closed? I suppose not as it still has a milestone so I'm reopening it.

dmethvin:
Looks like I accidentally deleted the fork in August which closed it, so yeah it should still be open.

mgol:

src/event.js
```diff
@@ -104,17 +104,26 @@ jQuery.event = {
 
 	add: function( elem, types, handler, data, selector ) {
```
22aam, line 106: all of the code looks good do me

mgol: @22aam stop spamming or you’ll get blocked

src/event.js
```diff
@@ -136,15 +145,6 @@ jQuery.event = {
 		if ( !( events = elemData.events ) ) {
 			events = elemData.events = {};
 		}
-		if ( !( eventHandle = elemData.handle ) ) {
-			eventHandle = elemData.handle = function( e ) {
-
-				// Discard the second event of a jQuery.event.trigger() and
```
22aam, line 142: there is one tiny thing that does not look ok

Issue #3133, opened 2016-05-29
Boolean attribute getters throw if the attribute name is not all lowercase


mgol:
### Description ###
In #2916 we removed our logic that lowercased attribute names. This caused one regression: any attribute getter using a name for boolean attributes but not all lowercased is going into an infinite recursion, exceeding the stack call limit.

Amongst others, this is breaking the AngularJS test suite when tested with jQuery 3.0.0-rc1.

### Link to test case ###
https://jsfiddle.net/shnann6y/2/

Basically, `$('<div>').attr('requiRed')` is enough to trigger the error.

mgol:
PR: #3134.

mgol:
We can discuss but I'm changing the milestone to 3.0.0 for now as it breaks the Angular test suite.

anarcat:
[CVE-2016-10707](https://www.cve.mitre.org/cgi-bin/cvename.cgi?name=2016-10707) was assigned to track this issue.

mgol:
@anarcat The CVE is incorrect. There is no stable jQuery version that suffers from this issue. The regression happened in preparation for jQuery 3.0.0 and it was fixed before 3.0.0 was released.

anarcat:
@mgol well, i based the CVE on the [Snyk database](https://snyk.io/vuln/npm:jquery:20160529) which says "versions <3.0.0 >=2.1.0-beta1". i'm not sure I parsed those version numbers right, but i read that as releases after 2.1 (e.g. 2.2.2) being vulnerable.

i'd be happy to clarify the description in the CVE, once we have a better idea of what's going on here..

do you want to submit an update to the CVE or should i?

mgol:
I've never submitted an update to a CVE before. :) Feel free to do it.

The range in the database is incorrect, the only affected version is `3.0.0-rc.1`. Both the earlier `2.1.x`, `2.2.x` & `3.0.0-beta1` and the later `3.0.0` don't exhibit the issue.

Are CVEs assigned for bugs existing in pre-release versions? If so, the range should be changed to `=3.0.0-rc.1`.

anarcat:
acknowledged, i'll carry that over to mitre. i'm not sure this should have been marked as vulnerability in Snyke, and therefore in Mitre. i should have done my homework better but, arguably, there wasn't much information available to begin with, without digging deep in the source code. :)

[...]

PR #3134, opened 2016-05-29
Attributes: Avoid infinite recursion on non-lowercase attribute getters

mgol:
### Summary ###

Attribute hooks are determined for the lowercase versions of attribute names
but this has not been reflected in the bool attribute hooks. The code that
temporarily removed a handler to avoid an infinite loop was removing an
incorrect handler causing stack overflow.

Fixes gh-3133
Refs gh-2914
Refs gh-2916

+2 bytes

### Checklist ###

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -38,7 +38,7 @@ jQuery.extend( {
 			return jQuery.prop( elem, name, value );
 		}
 
-		// All attributes are lowercase
+		// Attribute hooks are determined by the lowercase version
 		// Grab necessary hook if one is defined
 		if ( nType !== 1 || !jQuery.isXMLDoc( elem ) ) {
 			hooks = jQuery.attrHooks[ name.toLowerCase() ] ||
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -117,16 +117,18 @@ jQuery.each( jQuery.expr.match.bool.source.match( /\w+/g ), function( i, name ) 
 	var getter = attrHandle[ name ] || jQuery.find.attr;
 
 	attrHandle[ name ] = function( elem, name, isXML ) {
-		var ret, handle;
+		var ret, handle,
+			lowercaseName = name.toLowerCase();
+
 		if ( !isXML ) {
 
 			// Avoid an infinite loop by temporarily removing this function from the getter
-			handle = attrHandle[ name ];
-			attrHandle[ name ] = ret;
+			handle = attrHandle[ lowercaseName ];
+			attrHandle[ lowercaseName ] = ret;
 			ret = getter( elem, name, isXML ) != null ?
-				name.toLowerCase() :
+				lowercaseName :
 				null;
-			attrHandle[ name ] = handle;
+			attrHandle[ lowercaseName ] = handle;
 		}
 		return ret;
 	};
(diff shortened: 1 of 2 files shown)


test/unit/attributes.js
```diff
@@ -1633,3 +1633,20 @@ QUnit.test( "SVG class manipulation (gh-2199)", function( assert ) {
...
+QUnit.test( "non-lowercase boolean attribute getters should not crash", function( assert ) {
+	assert.expect( 3 );
+
+	jQuery.each( [
+		"Checked",
+		"requiRed",
+		"SELECTED"
+	], function() {
+	    try {
+			jQuery( "<div>" ).attr( this );
```
dmethvin: Should this be a test that mixed-case attributes are correctly retrieved? This just ensures they don't throw but not that they yield correct results.

dmethvin:
This does seem like it will be more effective at avoiding the infinite loop. :smiley_cat:

test/unit/attributes.js
mgol: You're right, I'll change that.

What do you think about the `try-catch` approach? Or should I just check assertions and rely on QUnit to fail when it throws? I thought the `try-catch` approach is better here where I'm specifically testing that no error is thrown.

mgol:
@dmethvin I was surprised we have no tests for that! (oops)

mgol:
PR updated. I added another commit that fixes an older test that was introducing side effects, in turn causing the test I've just added to fail only on even runs. :D

src/attributes/attr.js
```diff
@@ -114,19 +114,21 @@ boolHook = {
 };
 
 jQuery.each( jQuery.expr.match.bool.source.match( /\w+/g ), function( i, name ) {
-	var getter = attrHandle[ name ] || jQuery.find.attr;
+	var getter = attrHandle[ name.toLowerCase() ] || jQuery.find.attr;
+
+	attrHandle[ name.toLowerCase() ] = function( elem, name, isXML ) {
+		var ret, handle,
+			lowercaseName = name.toLowerCase();
```
gibson042: If there's going to be a variable anyway, why not introduce it earlier?

mgol: Where earlier? That's the beginning of the function, this name is different
than the one before its definition.

-- 
Michał Gołębiowski

gibson042: Only if someone repurposes it, but point taken.

mgol: Yeah, it's only +2 bytes so I didn't want to break that unofficial contract.

src/attributes/attr.js
```diff
@@ -114,19 +114,21 @@ boolHook = {
 };
 
 jQuery.each( jQuery.expr.match.bool.source.match( /\w+/g ), function( i, name ) {
-	var getter = attrHandle[ name ] || jQuery.find.attr;
+	var getter = attrHandle[ name.toLowerCase() ] || jQuery.find.attr;
+
+	attrHandle[ name.toLowerCase() ] = function( elem, name, isXML ) {
```
timmywil: This is just looping over the attrs in the regex. Is it necessary to lowercase here?

mgol: That's true but removing lowercasing from this & previous line... makes the gzipped size larger by one byte. ¯\_(ツ)_/¯

nazar-pc: 1 byte when gzipped is a serious win, for sure :laughing:

timmywil: lol, let's take it out if it's not needed.

mgol: OK. ;)

mgol: PR updated.

mgol:
Any further remarks? Is this good to land?

dmethvin:
LGTM

gibson042:
LGTM

PR #3148, opened 2016-06-08
ESLint

markelog:
### Checklist ###


.eslintignore
```diff
@@ -1,4 +1,6 @@
 external
+node_modules
```
markelog: That file and `.npmignore` feels weird to me, will evaluate them in the next pull

package.json
```diff
@@ -29,15 +29,15 @@
     "commitplease": "2.3.1",
     "core-js": "2.2.2",
     "cross-spawn": "2.2.3",
+    "eslint-config-jquery": "github:markelog/eslint-config-jquery",
```
markelog: Temporary

src/attributes/val.js
```diff
@@ -143,11 +143,16 @@ jQuery.extend( {
 
 				while ( i-- ) {
 					option = options[ i ];
+
+					/* eslint-disable no-cond-assign */
```
markelog: Would like to change the source in the next pull to remove the inline

src/core.js
```diff
@@ -252,7 +256,11 @@ jQuery.extend( {
 	},
 
 	isEmptyObject: function( obj ) {
+
+		/* eslint-disable no-unused-vars */
+		// See https://github.com/eslint/eslint/issues/6125
```
markelog: Have a pull for it - https://github.com/eslint/eslint/pull/6126

src/core.js
```diff
@@ -442,15 +450,9 @@ jQuery.extend( {
 	support: support
 } );
 
-// JSHint would error on this code due to the Symbol not being defined in ES5.
```
markelog: Moved to the top

src/data/var/acceptData.js
```diff
@@ -13,7 +13,6 @@ return function( owner ) {
 	//    - Node.DOCUMENT_NODE
 	//  - Object
 	//    - Any
-	/* jshint -W018 */
```
markelog: Couldn't find analogues rule for eslint, suggestions?

src/deferred.js
```diff
@@ -37,7 +37,7 @@ function adoptValue( value, resolve, reject ) {
 	// For Promises/A+, convert exceptions into rejections
 	// Since jQuery.when doesn't unwrap thenables, we can skip the extra checks appearing in
 	// Deferred#then to conditionally suppress rejection.
-	} catch ( /*jshint -W002 */ value ) {
```
markelog: After IE8 < drop, this wasn't needed

markelog: But maybe for Android?

src/effects.js
```diff
@@ -78,7 +78,6 @@ function createTween( value, prop, animation ) {
 }
 
 function defaultPrefilter( elem, props, opts ) {
-	/* jshint validthis: true */
```
markelog: Analogue rule - http://eslint.org/docs/rules/no-invalid-this yields to many errors, not sure what to do here

test/.eslintrc
```diff
@@ -52,5 +40,20 @@
 		"original$": true,
 		"baseURL": true,
 		"externalHost": true
+	},
+	"rules": {
```
markelog: Like the new structure, in the sense that we now know what rules failing for `test` folder

test/promises_aplus_adapters/.eslintrc
```diff
@@ -0,0 +1,4 @@
+{
+	"extends": "../../.eslintrc",
```
markelog: Since we don't want to inherit from the config from `test` folder

src/data/var/acceptData.js
Kovensky: http://eslint.org/docs/rules/no-negated-in-lhs handles one of the cases W018 would apply to, but not this one...

markelog: If no one would object, i will add this rule to `eslint-config-jquery`.

Thanks

markelog:
Would like to land this ASAP, since if other commits would be presented to the master it would be harder to keep this up to date.

If there is no objections will land this tomorrow

.eslintignore
mgol: We still want to lint `dist/jquery.js`, it did catch bugs in the compilation process in the past for me.

.eslintignore
```diff
@@ -1,4 +1,6 @@
...
+dist
```
mgol: We still want to lint `dist/jquery.js`, it did catch bugs in the compilation process in the past for me.

build/tasks/build.js
```diff
@@ -64,7 +64,11 @@ module.exports = function( grunt ) {
 		// Convert var modules
 		if ( /.\/var\//.test( path.replace( process.cwd(), "" ) ) ) {
 			contents = contents
-				.replace( /define\([\w\W]*?return/, "var " + ( /var\/([\w-]+)/.exec( name )[ 1 ] ) + " =" )
+				.replace(
+					/define\([\w\W]*?return/, "var " +
+					( /var\/([\w-]+)/.exec( name )[ 1 ] ) +
```
mgol: Not very important but the indentation here seems weird to me, I'd expect `"var " +` to see in this line and the continuations indented by one more tab.

mgol:
Will you squash it before landing? Sometimes landing many commits makes sense but the first one doesn't compile to me so it'd be bad to land it without squashing on `master`.

package.json
mgol: Can we get rid of that before ladning? I'd like to avoid having commits on `master` that in some time will almost for sure stop compiling, even on the original environment.

src/.eslintrc
```diff
@@ -0,0 +1,20 @@
+{
+	"env": {
+		"browser": true
```
mgol: This shouldn't be here for the same reason that we haven't specified `"browser": true` in `.jshintrc` for some time. We only assume the presence of the `window` global, the rest of the browser globals have to be taken from `window`. This is the main way in which we guarantee working correctly in configurations like Node+jsdom.

test/.eslintrc
```diff
@@ -1,24 +1,13 @@
...
-
-	"evil": true,
-	"sub": true,
-
-	"browser": true,
-	"devel": true,
-	"wsh": true,
-
+	"env": {
+		"browser": true
```
markelog: @mgol you fine with this though, right?

src/.eslintrc
markelog: You wanted to use `window` in `globals` instead?

markelog:
> Will you squash it before landing?

Not sure, would like keep `Tests:` and `Build:` commits separate, if it will be tricky, i wouldn't want to risk it

package.json
markelog: > I'd like to avoid having commits on master that in some time will almost for sure stop compiling

For sure everything will stop working eventually i assure i do not plan to suddenly remove this repository :).

Didn't want to release `eslint-config-jquery` without conformation of code-style compliance. So i'm either switch to correct name right after this lands or even in here

src/core.js
mgol: It was put here because I wanted to make sure all `Symbol` uses are guarded by a typeof check to not break ES5-only browsers. Could we just change the `jshint ignore` here to `/* eslint-disable no-undef */` & `/* eslint-enable no-undef */`? I assume there's no way to declare a global to exist only between 2 pragmas?

src/data/var/acceptData.js
mgol: @markelog makes sense

src/deferred.js
mgol: What do you mean? The comment is aimed at a linter so the browsers don't matter.

src/effects.js
mgol: I'm fine with adding it later once we resolve the issues with it. Just maybe create a ticket about that so that we don't forget.

test/.eslintrc
mgol: Just to note: here this `env` is fine, lots of tests assume the browser environment. It's just in `src/.eslintrc` where it's important to not specify it.

mgol: @markelog  comment race condition. :-)

test/.eslintrc
mgol: Yes, a very good idea! :)

src/.eslintrc
mgol: Yup, as it was in `.jshintrc`.

src/core.js
markelog: I would prefer this way, since i fear someone might introduce some debug variable or something.

But if you feel strongly about i'm not gonna object

src/deferred.js
markelog: This is https://jslinterrors.com/value-of-a-may-be-overwritten-in-ie8 rule and we do not support IE8 now, so the rule is redundant at least for that env, but not sure for Android

package.json
mgol: I'd prefer if you could do it here if possible. I know you might not plan to remove it but accidents happen and I think there's value in trying to make sure even old commits can be build without depending on data that may change or disappear. The npm registry is good in that regard because it doesn't allow to overwrite package versions and now it even made it much more difficult to unpublish packages.

If there's no time to publish it as `eslint-config-jquery` before landing I think even just publishing a `temporary-markelog-eslint-config-markelog` package and using that for a while would be safer. :) As the package wouldn't disappear and GitHub repos can.

Or just publish `eslint-config-jquery@0.0.1`, I don't think the jQuery team would be opposed to starting from that and just iterating later.

src/effects.js
markelog: There would be 183 errors if we enable it, i'm not sure we want this, since we use seemingly contextless (just coined it!) functions all over, cause we love helpers for methods like `jQuery#forEach` and for them we have to use `call/bind` stuff

test/.eslintrc
markelog: Until we manage to execute those test in jsdom, ha-ha

package.json
markelog: fine, fine, will do that before merge

src/deferred.js
mgol: Ahh, you meant the rule, not the pragma, I see.

package.json
markelog: btw, could check this repo? If there any issues with it

mgol:
>> Will you squash it before landing?
>
> Not sure, would like keep `Tests:` and `Build:` commits separate, if it will be tricky, i wouldn't want to risk it 

I'm mostly asking because of Git rebase. I sometimes use it to narrow down which commit introduced a particular regression, via the cycle of `npm it`, testing the issue and `git bisect good` or `git bisect bad` and if a commit that doesn't compile is landed it makes it harder to bisect.

If it's hard to squash, maybe you could edit the first commit to disable running `eslint` on `grunt` or `grunt test` there (e.g. via commenting out its entry in various tasks) and restore at the end? That would make all commits compile at the cost of having 3 commits with temporarily disabled linting which doesn't sound like a big issue.

mgol:
@markelog OK, I wrote some comments, the most important to me are those related to the browser env in `.eslintrc` and linting `dist/jquery.js`.

It looks good overall, though, thanks for this work! 👏

markelog:
> If it's hard to squash, maybe you could edit the first commit to disable running eslint on grunt or grunt test there (e.g. via commenting out its entry in various tasks) and restore at the end?

Okay, will do.

Will fix issues you pointed out and if no one has anything else, will land this tomorrow

mgol:
@markelog I commented about a few things in the first commit in the preset repo but nothing terribly important.

PR #3151, opened 2016-06-08
Effects: use requestAnimationFrame timestamp if available

### LINKED ISSUE #3143: We should use RAF's timestamp for smoother animations
jQuery 3.0.0-rc1 uses `requestAnimationFrame`, resulting in significantly smoother animations. However, it uses `Date.now()` instead of the [RAF `timestamp` argument](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame). As a result, the animation is *slightly* choppy.

Before you switch to using `timestamp`, I should point out that the Velocity.js library opts not to use `timestamp` ([velocity.js:3366](https://github.com/julianshapiro/velocity/blob/9c9c052a5fb6d5854c9aad931d9410c6790816ee/velocity.js#L3366-L3368)):

```
/* We ignore RAF's high resolution timestamp since it can be significantly offset
   when the browser is
   under high stress; we opt for choppiness over allowing the browser to drop huge
   chunks of frames. */
var timeCurrent = (new Date).getTime();
```

I don't know if this is actually still an issue in modern browsers. @julianshapiro originally wrote this [in 2014](https://github.com/julianshapiro/velocity/commit/580e01711e6d362ad8f1e9b4805e7db76673d686) -- perhaps we can chime in with some details.

I created a test case to demonstrate choppiness resulting from not using `timestamp`: https://codepen.io/joliss/pen/wWKGeV/ (The test case uses Velocity.js, but I believe it's the same for jQuery.)

My hope is that you can use this test case as a debugging aid, and as a motivating example to convince browser vendors to get any remaining `timestamp` issues resolved.

Running the CodePen, we see that there is some slight choppiness with the jQuery's `getTime` implementation:

a pen by jo liss jtbyc

Using the RAF `timestamp` argument instead, the animation feels smoother, and the graph is perfectly smooth as well:

a pen by jo liss her4j

I thought that perhaps using the microsecond-resolution `performance.now()` function would improve smoothness, but it doesn't seem to help much:

a pen by jo liss hft9a

mgol:
### Summary

In some environments that support the requestAnimationFrame timestamp callback
parameter using it results in smoother animations.

Fixes gh-3143

Currently it's +35 bytes, most likely due to the `getTimestamp` function used in `createFxNow()`. Any ideas on how to avoid this size tax?
### Checklist

- [x] ~~New tests have been added to show the fix or feature works~~
- [x] ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~
- [ ] Unit tests need to run on code with rAF enabled, currently we [disable it](https://github.com/mgol/jquery/blob/ee4a067/test/unit/effects.js#L16)
- [ ] Add one test using non-mocked `rAF`/`Date.now()`

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -27,16 +27,23 @@ var
 	rfxtypes = /^(?:toggle|show|hide)$/,
 	rrun = /queueHooks$/;
 
-function schedule() {
+function schedule( timestamp ) {
 	if ( inProgress ) {
-		if ( document.hidden === false && window.requestAnimationFrame ) {
+		if ( document.hidden === false ) {
 			window.requestAnimationFrame( schedule );
 		} else {
 			window.setTimeout( schedule, jQuery.fx.interval );
 		}
 
-		jQuery.fx.tick();
+		jQuery.fx.tick( timestamp );
 	}
+}
+
+// We need to be using Date.now() or performance.now() consistently as they return different
+// values: performance.now() counter starts on page load.
+// Support: IE <10, Safari <8.0, iOS <9, Android <4.4, Node with jsdom 9.4
+function getTimestamp() {
+	return window.performance.now();
 }
 
 // Animations created synchronously will run synchronously
@@ -44,7 +51,7 @@ function createFxNow() {
 	window.setTimeout( function() {
 		fxNow = undefined;
 	} );
-	return ( fxNow = Date.now() );
+	return ( fxNow = getTimestamp() );
 }
 
 // Generate parameters to create a standard animation
@@ -644,12 +651,12 @@ jQuery.each( {
 } );
 
 jQuery.timers = [];
-jQuery.fx.tick = function() {
+jQuery.fx.tick = function( timestamp ) {
 	var timer,
 		i = 0,
 		timers = jQuery.timers;
 
-	fxNow = Date.now();
+	fxNow = timestamp || getTimestamp();
 
 	for ( ; i < timers.length; i++ ) {
 		timer = timers[ i ];
--- a/test/unit/effects.js
+++ b/test/unit/effects.js
@@ -5,29 +5,42 @@ if ( !jQuery.fx ) {
 	return;
 }
 
-var oldRaf = window.requestAnimationFrame,
-	hideOptions = {
-		inline: function() { jQuery.style( this, "display", "none" ); },
-		cascade: function() { this.className = "hidden"; }
-	};
+var hideOptions = {
+	inline: function() { jQuery.style( this, "display", "none" ); },
+	cascade: function() { this.className = "hidden"; }
+};
 
 QUnit.module( "effects", {
 	beforeEach: function() {
 		this.sandbox = sinon.createSandbox();
 		this.clock = this.sandbox.useFakeTimers( 505877050 );
+
 		this._oldInterval = jQuery.fx.interval;
-		window.requestAnimationFrame = null;
 		jQuery.fx.step = {};
 		jQuery.fx.interval = 10;
+
+		// Simulate rAF using the mocked setTimeout as otherwise we couldn't test
+		// the rAF + performance.now code path.
+		this.sandbox
+			.stub( window, "requestAnimationFrame" )
+			.callsFake( function( callback ) {
+				return window.setTimeout( callback, jQuery.fx.interval,
+					Date.now() - 99989.6394 );
+			} );
+
+		// Fake performance.now() returning lower values than Date.now()
+		// and that its values are fractional.
+		this.sandbox.stub( window.performance, "now" ).callsFake( function() {
+			return Date.now() - 99999.6394;
+		} );
 	},
 	afterEach: function() {
 		this.sandbox.restore();
 		jQuery.fx.stop();
 		jQuery.fx.interval = this._oldInterval;
-		window.requestAnimationFrame = oldRaf;
 		return moduleTeardown.apply( this, arguments );
 	}
-} );
+}, function() {
 
 QUnit[ QUnit.jQuerySelectors ? "test" : "skip" ]( "sanity check", function( assert ) {
 	assert.expect( 1 );
@@ -2623,4 +2636,21 @@ QUnit.test( "jQuery.speed() - durations", function( assert ) {
 	jQuery.fx.off = false;
 } );
 
+} );
+
+QUnit.module( "effects-nostubs", function() {
+
+testIframe(
+	"matching timestamp",
+	"effects/matchingTimestamps.html",
+	function( assert, framejQuery, frameWin, frameDoc, success, failureReason ) {
+		assert.expect( 1 );
+
+		assert.ok( success, failureReason );
+	}
+);
+
+} );
+
 } )();
+
--- a/test/unit/tween.js
+++ b/test/unit/tween.js
@@ -5,14 +5,11 @@ if ( !jQuery.fx ) {
 	return;
 }
 
-var oldRaf = window.requestAnimationFrame;
-
 QUnit.module( "tween", {
 	beforeEach: function() {
 		this.sandbox = sinon.createSandbox();
 		this.clock = this.sandbox.useFakeTimers( 505877050 );
 		this._oldInterval = jQuery.fx.interval;
-		window.requestAnimationFrame = null;
 		jQuery.fx.step = {};
 		jQuery.fx.interval = 10;
 	},
@@ -20,7 +17,6 @@ QUnit.module( "tween", {
 		this.sandbox.restore();
 		jQuery.fx.stop();
 		jQuery.fx.interval = this._oldInterval;
-		window.requestAnimationFrame = oldRaf;
 		return moduleTeardown.apply( this, arguments );
 	}
 } );
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -108,6 +108,7 @@ module.exports = function( grunt ) {
 				"deprecated",
 				"dimensions",
 				"effects",
+				"effects-nostubs",
 				"event",
 				"manipulation",
 				"offset",
--- a/test/data/effects/matchingTimestamps.html
+++ b/test/data/effects/matchingTimestamps.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en" dir="ltr" id="html">
+<head>
+	<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+	<title>Effects: matching timestamps</title>
+</head>
+<body>
+	<script src="../../jquery.js"></script>
+	<script src="../iframeTest.js"></script>
+	<div id="test-div" style="height: 1000px;"></div>
+	<script>
+		//<![CDATA[
+		setTimeout( function () {
+
+			// Handle a timeout.
+		    startIframeTest( false, "The test timed out" );
+		}, 5000 );
+
+		var maxNow = 0;
+
+		jQuery( "#test-div" ).animate( {
+			height: '2000px',
+		}, {
+			duration: 300,
+			step: function( now ) {
+				if ( maxNow - now > 100 ) {
+					startIframeTest( false, "Animation is stepping back from " + maxNow + " to " + now );
+				}
+				maxNow = Math.max( now, maxNow );
+			},
+			complete: function() {
+				startIframeTest( true );
+			}
+		} );
+		//]]>
+	</script>
+</body>
+</html>
--- a/test/unit/animation.js
+++ b/test/unit/animation.js
@@ -5,10 +5,9 @@ if ( !jQuery.fx ) {
 	return;
 }
 
-var oldRaf = window.requestAnimationFrame,
-	defaultPrefilter = jQuery.Animation.prefilters[ 0 ],
-	defaultTweener = jQuery.Animation.tweeners[ "*" ][ 0 ],
-	startTime = 505877050;
+var defaultPrefilter = jQuery.Animation.prefilters[ 0 ];
+var defaultTweener = jQuery.Animation.tweeners[ "*" ][ 0 ];
+var startTime = 505877050;
 
 // This module tests jQuery.Animation and the corresponding 1.8+ effects APIs
 QUnit.module( "animation", {
@@ -26,7 +25,6 @@ QUnit.module( "animation", {
 		this.sandbox.restore();
 		jQuery.fx.stop();
 		jQuery.fx.interval = this._oldInterval;
-		window.requestAnimationFrame = oldRaf;
 		return moduleTeardown.apply( this, arguments );
 	}
 } );


scottgonzalez:
There's a typo in your commit message, see the title change I just made.

mgol:
Thanks, fixed.

src/effects.js
```diff
@@ -27,19 +27,27 @@ var
...
 	if ( timerId ) {
 		window.requestAnimationFrame( raf );
-		jQuery.fx.tick();
+		jQuery.fx.tick( timestamp );
 	}
 }
 
+// We need to be using jQuery.now() or performance.now() consistently as they return different
+// values: performance.now() counter starts on page load.
+function getTimestamp() {
```
dmethvin: Can we just use `jQuery.now()` for all the browsers that don't have the rAF timestamp? Which browsers are affected?

mgol: This function is also used in `createFxNow()` which, I think, is not limited to old browsers?

src/effects.js
```diff
@@ -27,19 +27,27 @@ var
...
+    return window.performance && typeof window.performance.now === "function" ?
```
markelog: Would you mind add a `Support` comment

mgol: Sure, will do.

markelog: In what browsers we do not have `now` in performance object?

Also, could you try to use `if..else` instead of a ternar?

test/unit/effects.js
```diff
@@ -17,15 +17,24 @@ QUnit.module( "effects", {
 		this.sandbox = sinon.sandbox.create();
 		this.clock = this.sandbox.useFakeTimers( 505877050 );
 		this._oldInterval = jQuery.fx.interval;
+		this._oldPerformanceNow = window.performance.now;
 		jQuery.fx.step = {};
 		jQuery.fx.interval = 10;
+
+		// Fake performance.now() returning lower values than Date.now()
+		// and that its values are fractional.
+		window.performance.now = function() {
```
markelog: This will break in IE9 right?

markelog:
I am skeptical, since neither gsap nor velocity doing this and there is open tickets in vendor about this

src/effects.js
mgol: > In what browsers we do not have now in performance object?

IE 9, Android <4.4, iOS 7.

And Node.js. :-)

> Also, could you try to use if..else instead of a ternary?

Sure.

test/unit/effects.js
mgol: Nice catch.

markelog: So extrapolating from your comment https://github.com/jquery/jquery/pull/3151/files/c0d05944f93309dc8c0329316d4b6cf699e21d4d#r66414166 it will not break in IE9, but will break in iOS7 and similar envs

mgol: Why not? IE 9 doesn't have the `performance` object.

markelog: mm, i'm confused - https://developer.mozilla.org/en-US/docs/Web/API/Performance#Browser_compatibility

mgol: Ah, you're right. :) But it's good to avoid useless assignments anyway.

src/effects.js
mgol: Done.

test/unit/effects.js
mgol: I wrapped it in an if & added support comments.

test/unit/effects.js
```diff
@@ -17,15 +17,32 @@ QUnit.module( "effects", {
...
+		// Fake performance.now() returning lower values than Date.now()
+		// and that its values are fractional.
+		// Support: IE <10, Safari <8.0, iOS <9, Android <4.4, Node with jsdom 9.4
+		if ( window.performance && typeof window.performance.now === "function" ) {
+			window.performance.now = function() {
+				return Date.now() - 99999.6394;
+			};
+		}
+
 		jQuery.now = Date.now;
```
gibson042: Not directly germane, but I don't think this line is useful.

src/effects.js
```diff
@@ -27,19 +27,29 @@ var
...
 		window.requestAnimationFrame( raf );
-		jQuery.fx.tick();
+		jQuery.fx.tick( timestamp );
 	}
 }
 
+// We need to be using jQuery.now() or performance.now() consistently as they return different
+// values: performance.now() counter starts on page load.
+// Support: IE <10, Safari <8.0, iOS <9, Android <4.4, Node with jsdom 9.4
+function getTimestamp() {
```
gibson042: Given the comments, shouldn't we have something more like this?
```js
getTimestamp = window.performance && typeof window.performance.now === "function" ?

	// or window.performance.now.bind( window.performance )
	jQuery.proxy( window.performance, "now" ) :

	jQuery.now;
```
If the complaint is about testing, then we just need an iframe test in which the sandbox is mocked before loading jQuery.

mgol: I guess it should work, I'll try it.

mgol:
@timmywil This is not a new feature so it could move it to `3.1.1`, couldn't we?

markelog:
Did we already agree on landing it? Can we have conformation from mozilla people on this?

mgol:
I guess there wasn't a definite decision, only +1s from @dmethvin & @gibson042: https://irc.jquery.org/%23jquery-meeting/default_%23jquery-meeting_20160718.log.html#t12:53:43

There's a patch being prepared for Firefox in https://bugzilla.mozilla.org/show_bug.cgi?id=1278408.

dmethvin:
As I understood the results, it didn't make any browser animate any jerkier. It just didn't help Firefox animate any smoother, but it helped Chrome a lot. If that's the case I'd say it's still good for 3.2.

markelog:
My concern with with frames dump, judging by the open tickets and because none of the other animate libraries used it seems issue is still present.

So my understanding is that we want to battle test it and ignore accepted practise and vendor bugs?

dmethvin:
> So my understanding is that we want to battle test it and ignore accepted practise and vendor bugs?

If there's a reason that GSAP or Velocity don't use it, we don't currently know that reason. I don't think that's ignoring accepted practice. Who knows, maybe it was too flakey back when they looked at this. @julianshapiro are there specific reasons Velocity doesn't use the rAF timestamp?

The only vendor bug I see mentioned in https://bugzilla.mozilla.org/show_bug.cgi?id=1278408 which is just saying that right now this change won't improve Firefox.

markelog:
We already received comments from @julianshapiro - https://github.com/jquery/jquery/issues/3143#issuecomment-224105768, you want them to be more descriptive?

Furher more we have explanation in description ticket - https://github.com/jquery/jquery/issues/3143

> We ignore RAF's high resolution timestamp since it can be significantly offset
when the browser is under high stress; we opt for choppiness over allowing the browser to drop huge chunks of frames.

> I don't think that's ignoring accepted practice

This is how currently those animation libs work, i'm not sure if anyone else is doing what we are trying do to, on the contrary, so i'm not sure what do you mean by this

> which is just saying that right now this change won't improve Firefox.

Mm, not sure what do you mean - https://bugzilla.mozilla.org/show_bug.cgi?id=1278408#c0

> ...the jitter is a problem for tweening code like jQuery's $.animate, because they'd ideally like to calculate tweening values based on the timestamp argument. The jitter then causes animations to feel slightly choppy even at 60fps. Here's a discussion on the jQuery bug tracker: https://github.com/jquery/jquery/issues/3143

dmethvin:
> We already received comments from @julianshapiro - #3143 (comment), you want them to be more descriptive?

I interpreted the @joliss reply in https://github.com/jquery/jquery/issues/3143#issuecomment-224113673 as saying that although dropping frames is worse, @joliss didn't see it dropping frames.

timmywil:
@mgol I see this as an enhancement, which is closer to a feature than a bug fix, so I prefer it be landed in 3.2.

markelog:
> @joliss didn't see it dropping frames.

I don't think these conditions were met - 

> it can be significantly offset when the browser is under **high stress**

That's why I would want to have conformation from the browser people - was it happening? How it looks like? Was it fixed?

gibson042:
Before this PR, codepath branches hinge on `window.requestAnimationFrame`:
* if truthy: `window.requestAnimationFrame( function() { …; jQuery.fx.tick(); } )`
* else: `window.setInterval( jQuery.fx.tick, jQuery.fx.interval )`

So even though we don't test both, there are no functional consequences if we assume that `window.requestAnimationFrame` callbacks get invoked once and that `window.setInterval` callbacks get invoked on a schedule (i.e., that the environment is sane).

This PR introduces additional but essentially independent¹ branching:
* if `window.performance.now` is a function: it defines animation timestamps (as real-number milliseconds since page load)
* else: animation timestamps are integer milliseconds since 1970 UTC

Since the former is slightly more complex than the latter _and_ generalizes it, I think the best approach is to run the effects suite against a mocked `window.performance.now` that returns non-integers (as [already appears](https://github.com/mgol/jquery/blob/ee4a0674111386265b70120949d4354b6f530da8/test/unit/effects.js#L24-L31)), and also introduce (iframe?) tests that explicitly validate the environment (cross-checking `window.performance.now` against `Date.now` and rAF callback arguments against `window.performance.now`)—a slight abuse of the jQuery test suite, but analogous to support tests.

¹ The exception being an environment that provides a truthy (`timestamp`) argument to rAF callbacks but does not expose `window.performance.now`, in which animation ticks would set `fxNow` to the wrong origin.

markelog:
> The exception being an environment that provides a truthy (timestamp) argument to rAF callbacks but does not expose window.performance.now, in which animation ticks would set fxNow to the wrong origin.

I extensively checked that when reviewed this PR, I don't think such environment exist

markelog:
Do we plan to land this? It's open for quite some time now

mgol:
Thanks for the reminder. I have to apply changes from @gibson042's comment at https://github.com/jquery/jquery/pull/3151#issuecomment-241485113.

src/effects.js
mgol: Unfortunately, test setup is run after jQuery is sourced so this would freeze `getTimestamp` to use the initial, non-mocked `window.performance`.

mgol:
I'm afraid I won't be able to get back to this PR before Jan 17.

markelog:
No worries :)

src/effects.js
mgol: I could put the `if` outside of the function definition but that only adds 3 bytes and I suppose such a single `if` will not cause any grief for any relatively modern browser.

The fact that we can't reliably mock stuff *before* jQuery is loaded is obviously a problem but this is an issue bigger than this PR.

mgol:
PR updated & rebased. I'm still working on the iframe test but the rest should be ready.

I've introduced one change - in the newest version of the PR `rAF` is not used in browsers that don't support `performance.now()`. From our supported browsers this should only affect iOS <9 (currently less than 5% of all iOS users).

src/effects.js
```diff
@@ -646,12 +657,12 @@ jQuery.each( {
...
 
 jQuery.timers = [];
-jQuery.fx.tick = function() {
+jQuery.fx.tick = function( timestamp ) {
 	var timer,
 		i = 0,
 		timers = jQuery.timers;
 
-	fxNow = jQuery.now();
+	fxNow = timestamp || getTimestamp();
```
markelog: I'm wondering if still needs this, since this change was added for the old ipad zooming thing - https://bugs.jquery.com/ticket/12837, maybe it's not required anymore. Or should we add `Support` comment?

test/unit/effects.js
```diff
@@ -13,20 +13,50 @@ var oldRaf = window.requestAnimationFrame,
...
 	},
 	teardown: function() {
 		this.sandbox.restore();
 		jQuery.now = Date.now;
 		jQuery.fx.stop();
 		jQuery.fx.interval = this._oldInterval;
-		window.requestAnimationFrame = oldRaf;
+
+		// Support: IE <10, Android <4.4, Node with jsdom 9.4
```
markelog: Love the comments

markelog (dismissed):

src/effects.js
mgol: Just removing this line (& the one setting `fxNow` to `undefined` at the end) makes Chrome fail most of effects tests.

mgol: I could optionally change `getTimestamp()` here to `jQuery.now()` as timestamp will always be defined if `tick` is called by jQuery & `rAF` & `perfomance.now()` are supported. But this seems safer, esp. that `tick` is exposed.

src/effects/support.js
```diff
@@ -0,0 +1,15 @@
...
+	"../var/support"
+], function( support ) {
+
+"use strict";
+
+// Support: Safari 8 only
+// In Safari 8 documents created via document.implementation.createHTMLDocument
+// collapse sibling forms: the second one becomes a child of the first one.
+// Because of that, this security measure has to be disabled in Safari 8.
+// https://bugs.webkit.org/show_bug.cgi?id=137337
```
mgol: Ouch, this is a copy-pasted comment.

markelog: Ha, was wondering how this relates to raf :)

mgol: Fixed, I'll commit it in a moment.

test/unit/effects.js
```diff
@@ -13,20 +13,50 @@ var oldRaf = window.requestAnimationFrame,
...
 		this._oldInterval = jQuery.fx.interval;
+		this._oldPerformanceNow = window.performance.now;
 		jQuery.fx.step = {};
 		jQuery.fx.interval = 10;
+
+		// Simulate rAF using the mocked setTimeout as otherwise we couldn't test
+		// the rAF + performance.now code path.
+		// Support: IE <10, Android <4.4, Node with jsdom 9.4
+		if ( window.requestAnimationFrame ) {
+			window.requestAnimationFrame = function( callback ) {
```
mgol: Good idea with sinon stubs, I've just switched to them.

As for the linked libs, I haven't checked them. We need something that will return values much smaller than `Date.now` to test that there is no divergence between various used timestamps.

The problem is I'd like it to hook to the same fake timer API that regular `Date.now()` does so that `this.clock.tick( number )` can be used in tests without changes for both code paths. Does it handle that?

My mock should be enough for our purposes, shouldn't it?

test/unit/effects.js
mgol: It turns out it was useful as `jQuery.now` is defined as `Date.now` so if you mock `Date.now` `jQuery.now` would remain unmocked.

This is yet another manifestation of how broken our current mocking that happens after jQuery has already loaded & run really is. :/

test/unit/tween.js
```diff
@@ -5,11 +5,13 @@ if ( !jQuery.fx ) {
 	return;
 }
 
-var oldRaf = window.requestAnimationFrame;
+var oldRaf = window.requestAnimationFrame,
+	oldPerformance = window.performance;
```
mgol: I do prefer that! Changing...

mgol:
@markelog PR updated.

test/unit/effects.js
markelog: Don't know, never used one before :), I'm thinking we should be aware of the all options before creating our own, that kinda thing

mgol:
I added another fixup that also contains the iframe test.

There's one thing that worries me - in the iframe test I had to use the `maxNow - now > 100` check as just `maxNow > now` sometimes failed which means the animation was going slightly backwards on occasion; test it yourself. I couldn't reproduce the same issue in the old code that doesn't rely on the `rAF` timestamp or `performance.now()`.

mgol:
I've removed the "Needs review" label as I first have to debug the Chrome underlying issue (& perhaps report it upstream).

mgol:

LifeIsStrange:
What's blocking this PR? It seems it would be a great performance/ux improvement

mgol:
@LifeIsStrange mainly the lack of my time to investigate all the issues with using the time stamp, other priorities & cross-browser issues. There were e.g. some ordering issues in Edge Legacy which makes this infeasible to land in v3.

LifeIsStrange:
@mgol I wanted to thank you for your excellent work, not all heroes wear capes :)

PR #3171, opened 2016-06-15
update syntax to ES2015

dshaps10:
### Checklist ###

* [X ] All authors have signed the CLA at https://contribute.jquery.com/CLA/

### DIFF
--- a/build/release.js
+++ b/build/release.js
@@ -1,9 +1,8 @@
-var fs = require( "fs" );
+const fs = require( "fs" );
 
 module.exports = function( Release ) {
 
-	var
-		files = [
+	let files = [
 			"dist/jquery.js",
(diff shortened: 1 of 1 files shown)


dmethvin:
I don't think we can do this until we decide that we are not supporting older node. There are plenty of users who haven't upgraded and they often have decent reasons for that. Also, I think we'd want to decide what style guide we should follow for ES6 files and lint them.

mgol:
Yes, this would fail in Node 0.10 & 0.12. Good news is they EOL at the end of 2016.

I agree with @dmethvin's points about the ES6 style guide etc.

timmywil:
Agreed. This will need more consideration and much broader sweep when it does happen. Thank you for the contribution though!

Issue #3167, opened 2016-06-15
Offset: report offset for zero size elements


poppinlp:
### Description

In your document, i find:

> Note: jQuery does not support getting the offset coordinates of hidden elements or accounting for borders, margins, or padding set on the body element.

And in your code, it's:

``` js
// Make sure element is not hidden (display: none)
if ( rect.width || rect.height ) {
    ...
}
```

So how about element shown but with 0 width and height? Seems it's belong to hidden element in your code and get wrong offset value.
Is that a feature or issue? Since there may be some chance i need to know a element's offset which is shown but has no size.
### Link to test case

http://codepen.io/poppinlp/pen/gMMaEY

In this test case you will see element with no size get wrong offset.

gibson042:
I believe we should report accurate offsets of visible but size zero elements.

poppinlp:
I have open a PR for this issue.
Is that fix OK?
I haven't check `visibility` since it will get correct bounding rect.

PR #3173, opened 2016-06-16
Offset: report offset for zero size elements

poppinlp:
### Checklist ###

Is this fix OK?
I haven't add unit test case for offset method with zero size element since i don't know whether you need that.
For [issue 3167](https://github.com/jquery/jquery/issues/3167)

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -103,7 +103,7 @@ jQuery.fn.extend( {
 		rect = elem.getBoundingClientRect();
 
 		// Make sure element is not hidden (display: none)
-		if ( rect.width || rect.height ) {
+		if ( elem.style.display !== 'none' ) {
 			doc = elem.ownerDocument;
 			win = getWindow( doc );
 			docElem = doc.documentElement;


src/offset.js
```diff
@@ -103,7 +103,7 @@ jQuery.fn.extend( {
 		rect = elem.getBoundingClientRect();
 
 		// Make sure element is not hidden (display: none)
-		if ( rect.width || rect.height ) {
+		if ( elem.style.display !== 'none' ) {
```
dmethvin: What if its parent is `display; none`?

We would also need a few unit tests for this case to be sure that it works now and doesn't break in the future.

src/offset.js
```diff
@@ -103,7 +103,7 @@ jQuery.fn.extend( {
 		rect = elem.getBoundingClientRect();
```
dmethvin: For width and height measurements [we use getClientRects](https://github.com/jquery/jquery/blob/ff1a0822f72d2b39fac691dfcceab6ede5623b90/src/css.js#L124-129) to determine if the element is disconnected, before calling getBoundingClientRect, it seems like this IE11 problem might apply here too?

mgol:
@poppinlp Do you want to finish this PR? We'd need you yo sign our [CLA](https://contribute.jquery.org/CLA/) and respond to @dmethvin's comments.

poppinlp:
@mgol 
Sry for no response since i got something to do the last few days.
Yep, I'll finish this recently. THX.

poppinlp:
@mgol 
Sry ag. I'll fix it this week.
I have a project must be online last week. :(

mgol:
Sure, no problem!

-- 
Michał Gołębiowski

timmywil:
Don't want to leave this open. It needs more work before reopening.

PR #3199, opened 2016-06-25
CSS: Support custom properties, #3144

### LINKED ISSUE #3144: Request: support for CSS custom properties
Now that CSS custom properties are supported by most browsers (except for IE), please consider adding support for them in the .css() method.

Example:

```
<div style="--color: red; color: var(--color)">text</div>
<script>
$('div').css('--color') ; // should return "red"
$('div').css('--color','blue') ; // should change text color to blue

//without jQuery we must currently do it like this:
getComputedStyle($('div')[0]).getPropertyValue('--color') ;
$('div')[0].style.setProperty('--color','blue') ;
</script>
```

[JsFiddle here](https://jsfiddle.net/wkvbq9dm/)

ConnorAtherton:
### Summary

Allows users to get and set the value of [CSS custom properties](https://drafts.csswg.org/css-variables/) on a DOM element through the `.css` method.

It works exactly like the example in the issue, which I am copying here for reference. 

```
<div style="--color: red; color: var(--color)">text</div>
<script>
$('div').css('--color') ; // Returns 'red'
$('div').css('--color' ,'blue') ; // Sets to 'blue'
```

Fixes #3144
### Checklist
- [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com


src/css.js
```diff
@@ -165,6 +165,10 @@ function getWidthOrHeight( elem, name, extra ) {
 	) + "px";
 }
 
+function isCustomProperty( value ) {
+	return value.match( /--(.*)/ );
```
mgol: A few remarks:

1. This will match things like `a--b`; you should check that the string starts with `--`, not just contains it.
1. Why did you introduce the group `(.*)`? You don't use it anywhere.
1. It's better to change that to `/^--.*`/.test( value )` as you only care about it matching or not and what you wrote will return the array with matches; the browser has better chances of optimising the former.

src/css.js
```diff
@@ -255,7 +263,7 @@ jQuery.extend( {
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
 
-				style[ name ] = value;
+				customProperty ? style.setProperty( name, value ) : style[ name ] = value;
```
mgol: I'm quite surprised it passed our linter... cc @markelog 

Please change that to the regular `if-else`, the minifier will optimize it anyway.

src/css.js
```diff
@@ -303,7 +316,8 @@ jQuery.extend( {
 			num = parseFloat( val );
 			return extra === true || isFinite( num ) ? num || 0 : val;
 		}
-		return val;
+
+		return val === undefined ? undefined : jQuery.trim( val );
```
mgol: What's the reason for this change?

mgol:
This PR adds 47 bytes (minified+gzipped). That's not a negligible amount. We should explore whether it wouldn't be better to just use the `setProperty` method everywhere. It might affect performance so we should do some performance testing. https://jsperf.com/ is offline; we can use [benchmark.js](https://benchmarkjs.com/), though.

mgol:
@ConnorAtherton Would you like to try to run some performance tests? You can see my example template in [one of my PRs](https://github.com/jquery/jquery/pull/2439#issuecomment-118898816).

src/css.js
ConnorAtherton: Testing in chrome returned the computed values with a preceding space in the string. I will look into cleaning this up.

ConnorAtherton:
@mgol Thanks for looking at this. All your comments are valid and I will rework my additions based on your feedback. Thanks for providing the link to your benchmarks in your other PR, I will update this thread with similar metrics in the coming days. I like the idea of using `setProperty` everywhere but will hold off on that decision until we have some metrics to consider.

src/css.js
mgol: > Testing in chrome returned the computed values with a preceding space in the string. I will look into cleaning this up.

That's probably as it should be. CSS variables are designed to be very powerful, you should e.g. be able to polyfill new CSS constructs by using them + a JS polyfill. Because of all that, almost everything is a valid value for such a property.

In particular, AFAIK everything between a colon and a semi-colon is the value for the custom property. So if you have the following in your CSS:
```css
div {
    --prop: value;
}
```
then ` value` (with a space) is the correct value. We shouldn't trim it. In 99% of cases it doesn't matter, if you later use `val(--prop)` then the space usually becomes irrelevant but there may be cases where it's expected.

I'd even add that we should have tests ensuring that we don't trim anything. :)

src/css.js
markelog: mm, what rule do you think we should enable or even create for this case?

scottgonzalez: Only allow ternaries if the result is used for assignment or return values.

markelog: Okay, so i have found three related rules in ESLint -

1. `no-nested-ternary` - http://eslint.org/docs/rules/no-nested-ternary
1. `no-ternary` - http://eslint.org/docs/rules/no-ternary
1. `no-unneeded-ternary` - http://eslint.org/docs/rules/no-unneeded-ternary

It seems those rules wouldn't satisfy us, but it seems `no-nested-ternary` would be a good fit? Created - https://github.com/jquery/eslint-config-jquery/issues/2

Didn't found any similar proposals in ESLint issues, will bring this up in ESLint gitter.

mgol: > mm, what rule do you think we should enable or even create for this case?

What @scottgonzalez said. It could be named `no-ternary-statement`.

src/css.js
mgol: To clarify, by a proper test for that I meant dynamically creating a stylesheet with contents like:
```css
.test-custom-props {
    --prop1: val1;
    --prop2:val2;
    --prop3:val3 ;
    --prop4: "val4" ;
}
```
and make sure all spaces/quotes/etc. are there in the computed values (I hope browsers are not buggy here!).

ConnorAtherton:
Following on from my previous comment, I have benchmarked both methods for adding styles to a DOM node. Here is my benchmark (I think this would be a useful template to add to the repo for future PRs) followed by the results:

```html
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8">
    <title>A page for perf tests</title>
    <style>
        /**
         * Any styles for perf start state
         */
    </style>
</head>
<body>
    <div id="test-div"></div>
    <script src="http://code.jquery.com/jquery-git.js"></script>
    <script>
        var $j = jQuery.noConflict();
    </script>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.13.1/lodash.js"></script>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/benchmark/2.1.0/benchmark.js"></script>
    <script>
        var suite = new Benchmark.Suite;

        var currentVersion = $j('#test-div');
        var newVersion = $('#test-div');

        // add tests
        suite
            .add('Adding styles using object assignment', function() {
                currentVersion.css('--color', 'blue');
                currentVersion.css('--color');
            })
            .add('Adding styles using setProperty', function() {
                newVersion.css('--color', 'blue');
                newVersion.css('--color');
            })
            // add listeners
            .on('cycle', function(event) {
                console.log(String(event.target));
            })
            .on('complete', function() {
                // So we can inspect the results ourselves in the console
                window.s = this;
                console.log('Fastest is ' + this.filter('fastest').map('name'));
            })
            // run async
            .run({'async': true});
    </script>
</body>
</html>
```

**Chrome 51**:
```
Adding styles using object assignment x 52,588 ops/sec ±3.13% (52 runs sampled)
Adding styles using setProperty x 38,907 ops/sec ±2.29% (54 runs sampled)
Fastest is Adding styles using object assignment
```

**Firefox 49**:
```
Adding styles using object assignment x 94,883 ops/sec ±5.27% (16 runs sampled)
Adding styles using setProperty x 15,532 ops/sec ±10.38% (15 runs sampled)
Fastest is Adding styles using object assignment
```

**Safari 9**:
```
Adding styles using object assignment x 84,248 ops/sec ±4.56% (41 runs sampled) 
Adding styles using setProperty x 69,650 ops/sec ±11.84% (42 runs sampled) 
Fastest is Adding styles using object assignment 
```

IE WIP

dmethvin:
It's definitely slower, especially in Firefox for some reason, but I don't think it is *too* slow. Focusing on Firefox, 15,000 ops per second is one op in 66 microseconds. How many CSS properties would typical code set in a 16ms frame? Also, setting properties often causes reflow which would be a lot more expensive than the act of setting the property.

markelog:
> Also, setting properties often causes reflow which would be a lot more expensive than the act of setting the property.

Totally, should be fine

src/css.js
```diff
@@ -165,6 +165,10 @@ function getWidthOrHeight( elem, name, extra ) {
 	) + "px";
 }
 
+function isCustomProperty( value ) {
+	return /^--.*/.test( value );
```
markelog: We usually put definitions of regexes in the beginning of the module with `r` prefix, like - https://github.com/jquery/jquery/blob/02c5e2908c2d99f191f5ec9e365ef2a3d357d1eb/src/manipulation.js#L36

Could you maybe do the same and also get rid of the `isCustomProperty` function and just use `/^--.*/.test(...)` directly?

src/css/curCSS.js
```diff
@@ -16,7 +16,10 @@ function curCSS( elem, name, computed ) {
 	computed = computed || getStyles( elem );
 
 	// Support: IE <=9 only
-	// getPropertyValue is only needed for .css('filter') (#12537)
+  //
+	// getPropertyValue is needed for:
```
markelog: Could you make indentation consistent, kudos for the comment btw

test/data/css/customProperties.html
```diff
@@ -0,0 +1,22 @@
...
+          --prop4: "val4" ;
+      }
+    </style>
+  </head>
+  <body>
+    <div class="test-custom-props"></div>
+    <script src="../../jquery.js"></script>
+    <script src="../iframeTest.js"></script>
+    <script>
+      startIframeTest( jQuery( ".test-custom-props" ) );
```
markelog: Why this is in the iframe?

src/css.js
```diff
@@ -274,11 +286,16 @@ jQuery.extend( {
...
-		name = jQuery.cssProps[ origName ] ||
-			( jQuery.cssProps[ origName ] = vendorPropName( origName ) || origName );
+			origName = jQuery.camelCase( name ),
+			customProperty = isCustomProperty( name );
+
+		// Make sure that we're working with the right name. We don't
+		// want to modify the value if it is a CSS custom property
+		// since they are user-defined.
+		if ( !customProperty ) {
+			name = jQuery.cssProps[ origName ] ||
```
markelog: Maybe you could simplify this part? We are trying to get rid of the past magic code

test/data/css/customProperties.html
markelog: I mean test as a whole, couldn't it tested in the usual way?

test/index.html
```diff
@@ -139,6 +139,7 @@
 			<foo_bar id="foobar">test element</foo_bar>
 		</form>
 		<b id="floatTest">Float test.</b>
+		<div id="customPropertyTest">Custom property test</div>
```
markelog: It is better to generate html code in the test itself otherwise it is hard to keep track of fixtures and tests

test/unit/css.js
```diff
@@ -114,8 +114,31 @@ QUnit.test( "css(String|Hash)", function( assert ) {
 	div = jQuery( "<div/>" ).css( { position: "absolute", "z-index": 1000 } ).appendTo( "#qunit-fixture" );
 	assert.strictEqual( div.css( "z-index" ), "1000",
 		"Make sure that a string z-index is returned from css('z-index') (#14432)." );
+
+  // Custom CSS properties
```
markelog: Indentation here is weird

test/unit/css.js
```diff
@@ -114,8 +114,31 @@ QUnit.test( "css(String|Hash)", function( assert ) {
...
+	jQuery( "#customPropertyTest" ).css( "--color", "blue" );
```
markelog: Would you mind moving these assertions in separate test?

mgol:
Thanks for doing the benchmarks, @ConnorAtherton! The numbers look good so I'd go with just always using `setProperty`/`getProperty`.

As an added bonus, we might get https://bugs.jquery.com/ticket/14394 fixed as a side-effect. :-) You could use a test from my PR (that had to be reverted later):
https://github.com/jquery/jquery/pull/1385/files, I *think* it should be passing in IE 9+ although I'm not 100% sure. That's not a blocker here, though, you might as well not do this part now, we could later check if it works and add the test if it passes.

src/css.js
```diff
@@ -255,8 +263,12 @@ jQuery.extend( {
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
 
-				style[ name ] = value;
-			}
+                if ( customProperty ) {
+                    style.setProperty( name, value );
```
markelog: Just wondering, why we can't always use `setProperty`?

markelog: Oh, i see, i missed discussion :)

markelog:
> As an added bonus, we might get https://bugs.jquery.com/ticket/14394 fixed as a side-effect. :-) You could use a test from my PR (that had to be reverted later):

That would be cool! But can we do this in separate PR?

test/data/css/customProperties.html
mgol: I think it was done so that a stylesheet can be defined in an easy way. It can be done in regular tests (dynamically create a `<style></style>` element, set its `innerHTML` to the desired CSS, remove it at the end of the test), just not that "naturally".

I agree it'd be better to do it that way, iframes are a little expensive so it's better to avoid them if they're not necessary.

markelog: Couldn't it be add to our test css?

> I agree it'd be better to do it that way, iframes are a little expensive so it's better to avoid them if they're not necessary.

Also, harder to support and debug those

markelog: > Couldn't it be add to our test css?

In here - https://github.com/jquery/jquery/blob/02c5e2908c2d99f191f5ec9e365ef2a3d357d1eb/test/data/testsuite.css i mean

mgol: I guess it could, although you could say the same about this global css file what you wrote here:
https://github.com/jquery/jquery/pull/3199/files/641c1a64d281a7cea5edb9f8de248b2160884b2d#r70155287
:)

Although that CSS file is still quite short.

markelog: I guess you got me there :), i think this is the area where we could expand and improve, since now we no longer supported oldIE, we can put those styles through `style` element in `qunit-fixture` div, which should be clear up after test run?

ConnorAtherton:
Hey all, thanks for the feedback. Apologies for the delay with this, I'm currently on vacation. Going to work on this more today.

ConnorAtherton:
Seeing this behaviour on chrome when trying to fetch properties from the style object

screen shot 2016-07-13 at 11 54 58

src/css.js
timmywil: Agreed, but the regex might as well be `/^--/`.

mgol:
@ConnorAtherton That's expected. Regular CSS properties can be read or set using regular property access but CSS variables can't, you have to use `setProperty`/`getPropertyValue`.

Do you know what needs to be done to finish this PR?

test/unit/css.js
```diff
@@ -114,8 +114,31 @@ QUnit.test( "css(String|Hash)", function( assert ) {
...
+	assert.equal( jQuery( "#customPropertyTest" ).css( "--color" ), "blue", "Modified CSS custom property using string: Assert value is right" );
+
+	jQuery( "#customPropertyTest" ).css( { "--color": "red" } );
+	assert.equal( jQuery( "#customPropertyTest" ).css( "--color" ), "red", "Modified CSS custom property using object: Assert value is right" );
+
+	jQuery( "#customPropertyTest" ).css( { "--theme-dark": "purple" } );
+	assert.equal( jQuery( "#customPropertyTest" ).css( "--theme-dark" ), "purple", "Modified CSS custom property with dashed name: Assert value is right" );
```
gibson042: A mixed-case custom property test would also be welcome.

src/css.js
ConnorAtherton: I don't think we can use `setProperty` here because properties like `maxWidth` and `marginRight` don't work. I've delayed working on this because I keep running into this issue but according to the spec, and please correct me if I'm reading this wrong, camelcased properties aren't transformed into their CSS equivalents. Every test that fails when I switch to `setProperty` is precisely because of this.  

Sources: 
- https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-setproperty
- https://dom.spec.whatwg.org/#converted-to-ascii-lowercase

ConnorAtherton:
@mgol Status update: I haven't been able to remove `setProperty` (see inline comment) but I have applied the test feedback and removed some stuff. All tests are passing locally for me, although I would like to check more browser coverage. 

Thanks for the feedback everyone.

markelog:
What about https://github.com/jquery/jquery/pull/3199#discussion_r70155190 and https://github.com/jquery/jquery/pull/3199#discussion_r70155287?

mgol:
@ConnorAtherton Did you look at @markelog's comments?

mgol:
@ConnorAtherton A friendly ping. :)

mgol:
@ConnorAtherton We'd like to get it into 3.2.0. If you don't have time to finish please let us know so we'll tweak it & land it ourselves.

mgol:
I finished the PR as #3557.

mgol:
@ConnorAtherton I've finished it up for you in #3557. You'll still be credited as the commit author. Thanks for all the work that went into this PR!

mgol:
@ConnorAtherton we need you to sign the new JS Foundation CLA (see the yellow dot in the checks section), can you do that?

Issue #3179, opened 2016-06-16
jQuery.Deferred.exceptionHook breaks sourceMaps


pyalot:
### Description

Traceback reformatting breaks source-maps, it is unreadable compared to browser built in tracebacks, and it cannot be folded closed like browser built in tracebacks. Tracebacks are an essential development tool. Breaking traceback is not an acceptable way to write libraries.

Closing tickets pointing out that jQuery breaks tracebacks, does not fix tracebacks.

dmethvin:
We still need a test case, preferably on jsbin.com or jsfiddle.net, that demonstrates the problem you are reporting. Your prior ticket was closed because it violated our code of conduct, not because of any technical merits.

pyalot:
The test case is using tracebacks and source-maps. It seemed fairly obvious to me that people do that all the time sufficiently enough that no further explanation is required. Apparently I'm wrong, so there:

* Live test case: http://codeflow.org/issues/jQuery/tracebacks/
*  Archive of the test case attached [tracebacks.zip](https://github.com/jquery/jquery/files/318898/tracebacks.zip)

Screenshot with overall comparison:

![overall](https://cloud.githubusercontent.com/assets/592826/16125280/49e3cb3e-33f3-11e6-8beb-89e396d35262.png)

Functional examination of browser built in traceback.

![ua-traceback](https://cloud.githubusercontent.com/assets/592826/16126400/42964c80-33f8-11e6-870b-0d662cd2cd2c.png)

Functional failing of jQuery traceback

![jquery-traceback](https://cloud.githubusercontent.com/assets/592826/16126321/df1be0d4-33f7-11e6-8561-4fe40fd696df.png)

dmethvin:
That's pretty much what we needed, thanks. Adding this line right below the jQuery script tag:
```js
<script>jQuery.Deferred.exceptionHook = function(e){ throw e; };</script>
```
Gives this result
![capture](https://cloud.githubusercontent.com/assets/157858/16129662/9deeaa16-33d3-11e6-80d8-501e7afb4864.PNG)

[...]

PR #3202, opened 2016-06-26
Deferred: Fix exceptions silent swallowing

### LINKED ISSUE #3174: Exceptions silently swallowed in $(fn)
### Description ###
If an exception is thrown in an event handler (such as $.ready), it is silently swalloed.

There are many cases in which exceptions are used to indicate errors for situations that are not code errors, for instance:

```javascript
var foo = function(arg){
  var result = parseFloat(arg);
  if(isNaN(result)){ 
    throw new Error('not a number');
  }
};
```

Exceptions like that may be cought by the user, such as when the number is user-input. But they may also indicate errors in the usage of the code (such as foo('123') ), in which case catching them is counter productive since they can't be debugged anymore.

### Link to test case ###
http://jsbin.com/bazaxixihe/edit?html,js,output

nicklou:
### Summary ###

Currently, jQuery silently swallows exceptions thrown in the body of a function passed as argument
in the `ready()` event handler. For instance:

```javascript
$(function() {
  throw new Error("Error");
});
````

will not result in any exception raise in the browser. This PR attempts to fix it by giving the ability
of adding another function in the `ready()` handler, so that if an exception occurs in such situation described above, the exception function will be fired. This function takes as argument the Error object given - internally propagated to it, so that it can be used inside the function.

```javascript
$(function() {
  throw new Error("Error");
}, function(e) {
  console.log("An error occured, and it says: " + e.message);
});
```

If no such function is provided, code now outputs the exception error message in the console.
This fix works in all `$(fn);`, `$(document).ready(fn);` and `$().ready(fn);` calls.

Note: This fix is applied on the `ready` implementation, not the `ready-not-deferred` one.

Fixes #3174

### Checklist ###

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -103,7 +103,10 @@ jQuery.extend( {
 						fns = null;
 					} ).promise();
 				},
-				then: function( onFulfilled, onRejected, onProgress ) {
+
+                // The onException is there to manage that an exception is thrown
+                // inside the onFullfilled without being catched.
+				then: function( onFulfilled, onRejected, onProgress, onException ) {
 					var maxDepth = 0;
 					function resolve( depth, deferred, handler, special ) {
 						return function() {
@@ -190,10 +193,18 @@ jQuery.extend( {
 											mightThrow();
 										} catch ( e ) {
 
-											if ( jQuery.Deferred.exceptionHook ) {
+											if ( typeof onException === "function" ) {
+
+												// Call the onException if provided.
+												onException( e );
+
+											// If onException is not provided, then print out
+											// the error.
+											} else if ( jQuery.Deferred.exceptionHook ) {
 												jQuery.Deferred.exceptionHook( e,
 													process.stackTrace );
 											}
+
 
 											// Support: Promises/A+ section 2.3.3.3.4.1
 											// https://promisesaplus.com/#point-61
--- a/src/deferred/exceptionHook.js
+++ b/src/deferred/exceptionHook.js
@@ -7,14 +7,20 @@ define( [
 
 // These usually indicate a programmer mistake during development,
 // warn about them ASAP rather than swallowing them by default.
-var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
+// Notice that Error is a choice too, as ? regex means (0-1) times.
+var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)?Error$/;
 
 jQuery.Deferred.exceptionHook = function( error, stack ) {
 
 	// Support: IE 8 - 9 only
 	// Console exists when dev tools are open, which can happen at any time
 	if ( window.console && window.console.warn && error && rerrorNames.test( error.name ) ) {
-		window.console.warn( "jQuery.Deferred exception: " + error.message, error.stack, stack );
+		window.console.warn( "jQuery.Deferred exception: " + error.message, error.stack );
+
+		// If process stack exists, then give it out as a second warning too.
+		if ( stack ) {
+			window.console.warn( stack );
+		}
 	}
 };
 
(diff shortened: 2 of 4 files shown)


mgol:
Please sign our CLA.

src/deferred.js
```diff
@@ -103,7 +103,10 @@ jQuery.extend( {
 						fns = null;
 					} ).promise();
 				},
-				then: function( onFulfilled, onRejected, onProgress ) {
+
+                // The onException is there to manage that an exception is thrown
+                // inside the onFullfilled without being catched.
+				then: function( onFulfilled, onRejected, onProgress, onException ) {
```
mgol: We avoid deviating from the standards as well as adding new undocumented parameters to publicly available methods. Also, the current Deferreds API should be enough to achieve the goals of this PR. All we need is to automatically attach a special `.catch` handler to the `readyList.then( fn )` promise in the `core/ready.js` file.

src/deferred/exceptionHook.js
```diff
@@ -7,14 +7,20 @@ define( [
...
 	// Support: IE 8 - 9 only
 	// Console exists when dev tools are open, which can happen at any time
 	if ( window.console && window.console.warn && error && rerrorNames.test( error.name ) ) {
-		window.console.warn( "jQuery.Deferred exception: " + error.message, error.stack, stack );
+		window.console.warn( "jQuery.Deferred exception: " + error.message, error.stack );
+
+		// If process stack exists, then give it out as a second warning too.
+		if ( stack ) {
+			window.console.warn( stack );
+		}
```
mgol: Let's not mix goals here - any changes to exception hooks should be done in a separate PR.

src/deferred/exceptionHook.js
```diff
@@ -7,14 +7,20 @@ define( [
 
 // These usually indicate a programmer mistake during development,
 // warn about them ASAP rather than swallowing them by default.
-var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
+// Notice that Error is a choice too, as ? regex means (0-1) times.
```
mgol: We excluded `Error` on purpose. The idea is that we want to log only those errors that are most likely the result of a simple user error (like a mistyped variable) rather than all errors.

src/deferred/exceptionHook.js
nicklou: OK, noticing that stack argument can be empty, do you think I should open a new PR in this case, does it make sense?

src/deferred.js
nicklou: OK, will move to that direction. My concern is though that if we want to provide `$( fn, except )` in API, then public selector call must be modified in `init.js`, as I have done. Should I do it another way?

src/deferred/exceptionHook.js
mgol: I don't think it's a problem that it logs `undefined`. A real problem here is that source maps are not applied to the error stack, see #3179 & https://bugs.chromium.org/p/chromium/issues/detail?id=622227#c3. We should try various solutions, including the one proposed in the Chromium bug but it's better to do it in a separate PR if you'd like to try to fix that.

src/deferred.js
dmethvin: We definitely don't want to expose new public functionality on Deferred without thinking carefully how it affects things. I think this should be a private API for now. It may be that we need some other implementation. I wonder if we could always use the [no-deferreds](https://github.com/jquery/jquery/blob/25d8ccd1112d75394b91071ff7eba13283aaf898/src/core/ready-no-deferred.js) implementation, it adds some bulk but that would need to be compared against the size increase here as well.

mgol:
@nicklou We discussed it at the last Core meeting and decided that requiring existing apps to refactor `$(fn)` to `$(fn, errorHandlerFn)` to work correctly with jQuery 3.0.0 is not feasible and we'll be just automatically adding a proper error handler that may be optionally overwritten.

Sorry for the trouble.

The PR that succeeds this one is #3210.

nicklou:
@mgol OK, good to know, and you provided a nice solution to this. I will delete my related branch and move on to other issues to see if I can help.

dmethvin:
Thanks @nicklou!

Issue #3194, opened 2016-06-23
jQuery 3 - window load inside ready state will not be triggered


eisbehr-:
### Description ###
In jQuery 3.0.0 I got the issue, that a window load callback inside a ready state gets not triggered anymore. I think this is because it's async since 3.0.0. But this is used in jQuery plugins, to be only executed after all page content is loaded. And this is often wraped in a ready state by users.

### Example Code ###
https://jsfiddle.net/vpm59qws/
```JS
$(function() {
    $(window).on("load", function() {
        console.log("this will not be triggered");
    });
});
```

dmethvin:
I agree this can happen, but not sure how to "fix" it without making ready completely synchronous again which brings back other demons.

eisbehr-:
What does a ready state bring, if you can't execute all features in it?
As far as I know, it's best practice to wrap all stuff in such ready function. And I'm pretty sure `load` events are not that rarely used ...

dmethvin:
It's definitely not best practice to wrap everything in a ready. Things that can be done outside a ready should be done there, such as initializing ajax requests, so you don't introduce unneeded delays. Similarly you can't "execute all features" just because the HTML document is ready. 

The tickets open because I also believe that there could be issues with this, but we'd prefer to act on evidence rather than incorrect assertions. See also gh-3197.

eisbehr-:
It's not the problem that it is late or async. The problem is, that it will be executed never! It makes v3 pretty useless in my eyes, like for plugin developers. Because you can't prevent users to do things in a ready state. And such essential as a load event should always work.

I'm currently thinking about how to solve this even on an async execution. Maybe like tracking the load event within the library itself, and when something is bind to `load` afterwards, it will be executed immediately.

dmethvin:
You're not adding any additional information here. Please stop. The reason it never executes is because ready occurs after the one-time load event occurs.

john2014:
Hi Dave, this bug seemed unusual. So I did some more tests:

1. If we change $(window) to $(document), nothing fires https://jsfiddle.net/f73ar30v/1/

2. If we change $(window) to $("iframe"), nothing fires: https://jsfiddle.net/3hzy7Lrj/

3. If we change $(window) to $("body"), nothing fires

 Any idea why this is happening?

4. Really weird: Changing #1,#2,#3 back to $(window), nothing fires. Is this a bug with that website?

timmywil:
[...]

PR #3210, opened 2016-06-29
Core: Re-throw errors that happened in callbacks wrapped in jQuery ready

mgol:
### Summary ###

Re-throw errors that happened in callbacks wrapped in jQuery ready

Also, expose jQuery.readyException that allows to overwrite the default
ready error handler.

### Checklist ###

* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com (https://github.com/jquery/api.jquery.com/issues/942)

Fixes gh-3174

### DIFF
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -1,6 +1,7 @@
 define( [
 	"../core",
 	"../var/document",
+	"../core/readyException",
 	"../deferred"
 ], function( jQuery, document ) {
 
@@ -11,7 +12,15 @@ var readyList = jQuery.Deferred();
 
 jQuery.fn.ready = function( fn ) {
 
-	readyList.then( fn );
+	readyList
+		.then( fn )
+
+		// Wrap jQuery.readyException in a function so that the lookup
+		// happens at the time of error handling instead of callback
+		// registration.
+		.catch( function( error ) {
+			jQuery.readyException( error );
+		} );
 
 	return this;
 };
--- a/src/core/readyException.js
+++ b/src/core/readyException.js
@@ -1,0 +1,13 @@
+define( [
+	"../core"
+], function( jQuery ) {
+
+"use strict";
+
+jQuery.readyException = function( error ) {
+	window.setTimeout( function() {
+		throw error;
+	} );
+};
+
+} );
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -1,4 +1,12 @@
-QUnit.module( "core", { teardown: moduleTeardown } );
+QUnit.module( "core", {
+	setup: function() {
+		this.sandbox = sinon.sandbox.create();
+	},
+	teardown: function() {
+		this.sandbox.restore();
+		return moduleTeardown.apply( this, arguments );
+	}
+} );
 
 QUnit.test( "Basic requirements", function( assert ) {
 	assert.expect( 7 );
@@ -1708,4 +1716,46 @@ QUnit.test( "Iterability of jQuery objects (gh-1693)", function( assert ) {
 	} else {
 		assert.ok( true, "The browser doesn't support Symbols" );
 	}
+} );
+
+QUnit[ jQuery.Deferred ? "test" : "skip" ]( "jQuery.readyException (original)", function( assert ) {
+	assert.expect( 1 );
+
+	var message;
+
+	this.sandbox.stub( window, "setTimeout", function( fn ) {
+		try {
+			fn();
+		} catch ( error ) {
+			message = error.message;
+		}
+	} );
+
+	jQuery( function() {
+		throw new Error( "Error in jQuery ready" );
+	} );
+	assert.strictEqual(
+		message,
+		"Error in jQuery ready",
+		"The error should have been thrown in a timeout"
+	);
+} );
+
+QUnit[ jQuery.Deferred ? "test" : "skip" ]( "jQuery.readyException (custom)", function( assert ) {
+	assert.expect( 1 );
+
+	var done = assert.async();
+
+	this.sandbox.stub( jQuery, "readyException", function( error ) {
+		assert.strictEqual(
+			error.message,
+			"Error in jQuery ready",
+			"The custom jQuery.readyException should have been called"
+		);
+		done();
+	} );
+
+	jQuery( function() {
+		throw new Error( "Error in jQuery ready" );
+	} );
 } );


mgol:
Please review. I'd like to have it released quickly, preferably no later than next week. cc @jquery/core

test/unit/core.js
```diff
@@ -1709,3 +1709,45 @@ QUnit.test( "Iterability of jQuery objects (gh-1693)", function( assert ) {
...
 	}
 } );
+
+QUnit[ jQuery.Deferred ? "test" : "skip" ]( "jQuery.readyException (original)", function( assert ) {
+	assert.expect( 1 );
+
+	var message,
+		origWindowSetTimeout = window.setTimeout;
+
+	window.setTimeout = function( fn ) {
```
dmethvin: I was wondering how you'd do this :+1:

dmethvin:
LGTM

src/core/ready.js
```diff
@@ -1,6 +1,7 @@
 define( [
 	"../core",
 	"../var/document",
+	"../core/readyException",
```
markelog: This should be 3.1 only though i.e. minor? Do we have corresponding docs pr/issue?

markelog: I know you labelled it as `3.1.0`, just checking

test/unit/core.js
markelog: You could use sinon here, right?

test/unit/core.js
```diff
@@ -1709,3 +1709,45 @@ QUnit.test( "Iterability of jQuery objects (gh-1693)", function( assert ) {
 		assert.ok( true, "The browser doesn't support Symbols" );
 	}
 } );
+
+QUnit[ jQuery.Deferred ? "test" : "skip" ]( "jQuery.readyException (original)", function( assert ) {
```
markelog: Core tests will not gonna run if you build jQuery without `Deferred`, it seems this guard is redundant?

markelog:
Do you wann release before providing a fix for https://github.com/jquery/jquery/issues/3194?

src/core/ready.js
markelog: Oh, i see https://github.com/jquery/api.jquery.com/issues/942, we need a pr for it though? Before we can release?

mgol: I'll submit a docs PR once this is accepted, no worries.

At the meeting we agreed that since this issue kind of breaks all Angular apps (as they wrap their init code in `$()` and all bootstrapping errors are silenced) we'll release a quick 3.1.0 with maybe only this fix. I renamed previous milestones: `3.0.1` to `3.1.1` and `3.1.0` to `3.2.0`.

test/unit/core.js
mgol: People can exclude Deferreds, though. I wanted to not block us from making the tests runnable in such a mode in the future.

test/unit/core.js
mgol: Yeah, I was thinking about sinon but it's not activated at all for this module, I'd need to enter the sinon sandbox to be able to use spies, no?

src/core/readyException.js
```diff
@@ -0,0 +1,13 @@
+define( [
+	"../core"
+], function( jQuery ) {
+
+"use strict";
+
+jQuery.readyException = function( error ) {
+	window.setTimeout( function() {
+		throw error;
+	} );
```
mgol: I forgot @gibson042 wanted to have this reject as well so that a more general potential unhandled rejections mechanism would catch it. The code would be:
```js
jQuery.readyException = function( error ) {
	window.setTimeout( function() {
		throw error;
	} );
	throw error;
};
```

I have doubts about the need for that, though - the error thrown in a timeout would fire `window.onerror` anyway, do we need both mechanism to work here? Usually an error is either synchronous and handled by `window.onerror` or asynchronous and handled by the other mechanism.

cc @dmethvin as you already LGTM'd this PR. :)

src/core/ready.js
markelog: Sounds good

src/core/readyException.js
markelog: It seems we would need to rename `readyException`? If it would be used in more general sense? Also we need to re-throw error twice?

test/unit/core.js
markelog: I'm saying i already excluded and run core tests without `Deferred`, they don't work :/

test/unit/core.js
markelog: Not necessary to use sandbox i think, sandbox usually needed to clean up some scope of spies/stubs, you probably could just use 

`sinon.stub(window, 'setTimeout', myOwnThing);`

test/unit/core.js
mgol: I understand it, I just don't want to contribute to making them not work even more.

test/unit/core.js
mgol: Thanks, I'll try that.

src/core/readyException.js
mgol: It's not meant to be used in a more general sense, it was only about integrating with existing debugging tools.

markelog: Not sure if i understand the use-case, maybe it would be better to discuss this in separate ticket?

test/unit/core.js
markelog: Do we have plan for this? If we do, maybe we should put it on roadmap or create ticket for this?

src/core/readyException.js
dmethvin: > I forgot @gibson042 wanted to have this reject as well so that a more general potential unhandled rejections mechanism would catch it. 

If you wanted to catch your own ready handler's rejections you can already do `$.when($.ready).then(fn).catch(errfn)` in 3.0.  What other cases do we want to handle? I feel like 99.9% of the cases are errors and as long as we manage to dump a stack we're doing all we need to do.

test/unit/core.js
dmethvin: For that case [ready-no-deferred](https://github.com/jquery/jquery/blob/25d8ccd1112d75394b91071ff7eba13283aaf898/src/core/ready-no-deferred.js) comes into play, right? Since it fires each handler in a timeout without a try/catch it's gonna dump a stack anyway.

test/unit/core.js
dmethvin: You can use sinon if you want, but for a case like this it's really nice to be able to look at the code and understand what it does without reading though sinon docs.

test/unit/core.js
markelog: Yeah it does, but i wasn't aware we were planning run this module on core tests?

test/unit/core.js
markelog: I'd say sinon is used almost universally for mocking/spying/stubing/etc it has very simple obvious api with clear documentation and we already use it.

So i would suggest to use it anywhere we can

src/core/readyException.js
gibson042: The important detail to me is that our default handling doesn't suppress unhandled rejections, since doing so was the essence of the original complaint (and I still want to keep the door open for an unhandled rejections plugin). In fact, if we were willing to advocate `$.when($.ready).then(fn).catch(errfn)` strongly enough, then gh-3174 could be closed with no code changes whatsoever.

src/core/readyException.js
```diff
@@ -0,0 +1,13 @@
+define( [
+	"../core"
+], function( jQuery ) {
+
+"use strict";
+
+jQuery.readyException = function( error ) {
```
gibson042: Do we really need to define a default handler? I think just having the hook available is enough for consumer use cases.

markelog: I think you understand it backwards - hook is not necessary, default handler very much is. 

I don't believe exceptions should be swallowed

mgol: I agree with @markelog here; we shouldn't be swallowing exceptions in a method that doesn't have the promise interface. I see `jQuery.readyException` as a way for people to opt out of the default sensible reporting if they want to use other tools for that, or to hook up other tools. I think the default should log the exception as this PR does.

It seems most of us agree with this strategy.

src/core/readyException.js
mgol: @gibson042 I think that could work only if we deprecated/removed `$(fn)` but this is such a popular pattern that I don't see it going away any time soon. And this fact alone is enough for me to want it not silencing exceptions.

src/core/readyException.js
gibson042: Works for me.

src/core/ready.js
```diff
@@ -11,7 +12,9 @@ var readyList = jQuery.Deferred();
 
 jQuery.fn.ready = function( fn ) {
 
-	readyList.then( fn );
+	readyList
+		.then( fn )
+		.catch( jQuery.readyException );
```
gibson042: This should wrap `jQuery.readyException` so the lookup happens at the time of error handling rather than callback registration: `.catch( function( err ) { jQuery.readyException( err ); } )`.

dmethvin: Good point, it might be good to have a comment to that effect so nobody tries to "optimize" it later.

mgol: I actually thought about that. 😄 I just thought the probability of someone registering a handler between running `$(fn)` and `fn` actually firing is small and current code is smaller. But OK, I can change that.

nicklou:
Is it possible that we use something similar to`$.on( "error", handler )` method, but with a different event?

For instance:

```javascript
$( fn ).on( "exception", handler );
```

That way, you don't break the existing API and the modification is not on deferreds necessarily, although I am not sure if it is possible to put a correlation between the `fn` called and the `.on()` handler.

dmethvin:
For existing code, no additional API is useful unless its default provides helpful information, which this PR does. Someone who doesn't like the default can change it.

For new code where an exception is expected, you can either attach a handler to the promise via `.catch()` as [described above](https://github.com/jquery/jquery/pull/3210#discussion_r69377800) or you can just add a `try/catch` to the synchronous code. 

I don't think anything else is required is it? What case would another API handle that those options do not?

test/unit/core.js
mgol: @markelog I had to create a sandbox and use `this.sandbox.stub` instead of `sinon.stub` or stubs wouldn't have been cleaned up (because how would Sinon know when to clean up?). It works fine now.

mgol:
The PR is ready and passes `npm test` locally. Travis fails because npm currently [has problems in some locations](https://github.com/npm/npm/issues/13284).

I've also submitted API docs for this method: https://github.com/jquery/api.jquery.com/pull/946. Please review.

I'd like to land this today if no one objects.

test/unit/core.js
```diff
@@ -1709,3 +1717,41 @@ QUnit.test( "Iterability of jQuery objects (gh-1693)", function( assert ) {
...
+			fn();
+		} catch ( error ) {
+			message = error.message;
+		}
+	} );
+
+	jQuery( function() {
+		throw new Error( "Error in jQuery ready" );
+	} );
+	assert.strictEqual( message, "Error in jQuery ready", "The error wasn't thrown in a timeout" );
```
gibson042: Shouldn't "wasn't" here be "was"?

test/unit/core.js
```diff
@@ -1709,3 +1717,41 @@ QUnit.test( "Iterability of jQuery objects (gh-1693)", function( assert ) {
...
+QUnit[ jQuery.Deferred ? "test" : "skip" ]( "jQuery.readyException (custom)", function( assert ) {
+	assert.expect( 1 );
+
+	var done = assert.async();
+
+	this.sandbox.stub( jQuery, "readyException", function( error ) {
+		assert.strictEqual(
+			error.message,
+			"Error in jQuery ready",
+			"The custom jQuery.readyException wasn't called"
```
gibson042: Same here; I think this should be "was".

gibson042:
The default `jQuery.readyException` still needs a `throw error`.

test/unit/core.js
mgol: Fixed.

test/unit/core.js
mgol: Fixed.

mgol:
> The default `jQuery.readyException` still needs a `throw error`.

What about the @dmethvin's remark at https://github.com/jquery/jquery/pull/3210#discussion_r69377800?

I'm not sure I see the need of a double-error-dispatch approach here. In general there are two kinds of errors - the ones that propagate to `window.onerror` and the ones who don't and might be collected by a mechanism that records unhandled promise rejections (https://github.com/jquery/jquery/blob/3.0.0/src/deferred/exceptionHook.js). If here we'd throw both synchronously and asynchronously, we'd collect the error in both of them.

To me the fact that `jQuery( fn )` uses Deferreds is somewhat an implementation details since you can't use this API in the same way as you can attach error handlers to promises (the `jQuery.ready` promise is meant for that).

test/unit/core.js
timmywil: Yes, I did a pass with a custom build replacing ready with ready-no-deferred on the whole suite. Guards like this are necessary if the test is not applicable to ready-no-deferred.

timmywil:
I agree. The main problem was that there was no info on the error. One throw seems like enough.

gibson042:
> To me the fact that `jQuery( fn )` uses Deferreds is somewhat an implementation details since you can't use this API in the same way as you can attach error handlers to promises (the `jQuery.ready` promise is meant for that).

Okay, that's persuasive. Suggestion withdrawn.

src/core/ready.js
```diff
@@ -11,7 +12,11 @@ var readyList = jQuery.Deferred();
 
 jQuery.fn.ready = function( fn ) {
 
-	readyList.then( fn );
+	readyList
+		.then( fn )
+		.catch( function( error ) {
```
markelog:
```js
.catch(jQuery.readyException)
```
wouldn't work?

mgol: @markelog https://github.com/jquery/jquery/pull/3210#discussion_r69585679

mgol: I added a comment there.

markelog:
LGTM

mgol:
All comments addressed. I'll land it later today unless new review remarks come up before.

PR #3218, opened 2016-07-06
Сore: Inline getProto

peasemaker:
### Checklist ###


mgol:
What's the reason for this change?

Also, note that you committed your project settings. Never do `git add .`, it causes problems like that.

scottgonzalez:
> Also, note that you committed your project settings. Never do git add ., it causes problems like that.

`git add .` is fine, given a better workflow. For starters, you should use `core.excludesfile` to ignore files like the `.idea` directory across all projects (see https://help.github.com/articles/ignoring-files/#create-a-global-gitignore). You should also review your diff before committing so that you know the changes you're committing are what you intended (this goes beyond just which files are being committed).

mgol:
@peasemaker I'd still like to know what's the reason for this change.

markelog:
>  I'd still like to know what's the reason for this change.

Same here

TechQuery:
@mgol @markelog , I don't understand why jQuery needs to **make a separate module with only one ECMA / W3C API return-statement in it**...

timmywil:
Thanks for opening a PR. I think this qualifies as bikeshedding though. We have some variables in separate modules so that those variables can be shared across modules without affecting the size of built jQuery. Moving it inline might mean we have to move it back out later if another module wants to use it (which I would not say is unlikely). And, moving it inline has very little affect on the built version.

PR #3221, opened 2016-07-06
Attributes: Fixed Coding Convention

mixed:
### Summary ###
Properties should check undefined using `object.prop === undefined`.
https://contribute.jquery.org/style-guide/js/#type-checks

### Checklist ###

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -34,7 +34,7 @@ jQuery.extend( {
 		}
 
 		// Fallback to prop when attributes are not supported
-		if ( typeof elem.getAttribute === "undefined" ) {
+		if ( elem.getAttribute === undefined ) {
 			return jQuery.prop( elem, name, value );
 		}
 


src/attributes/attr.js
```diff
@@ -34,7 +34,7 @@ jQuery.extend( {
 		}
 
 		// Fallback to prop when attributes are not supported
-		if ( typeof elem.getAttribute === "undefined" ) {
+		if ( elem.getAttribute === undefined ) {
```
gibson042: This was [intentional](https://github.com/mixed/jquery/commit/075da3091cda170bd8cd5ce47bad1c5b14760232), but probably not needed anymore. @jquery/core Does anyone remember at what point IE stopped automatically invoking the method on mere reference?

dmethvin: Either IE7 or IE8, so we should be in the clear.

timmywil: It was only certain properties in IE8, but I think we're safe now. We could probably do just...

```js
if ( !elem.getAttribute ) {
```

markelog: Wouldn't it violate our code style?

markelog: Also, pretty sure we added tests for that, so if they pass we should definitely be in the clear

markelog:
Added correspondent ticket for eslint config - https://github.com/jquery/eslint-config-jquery/issues/5

src/attributes/attr.js
timmywil: > Wouldn't it violate our code style?

Sure, we can change it, tho.

markelog: Do you wanna? :)

timmywil: I do, as long as we no longer have to use `typeof`.

timmywil:
Since we should make a sweep across the whole repo, and the check will probably not look like the one proposed here, I'll close this PR. Added to [Roadmap](https://github.com/jquery/jquery/wiki/Roadmap).

src/attributes/attr.js
mgol: To conclude: we still need `typeof` in some places for all IE versions as it still has the auto-invoking problem for some host objects: https://github.com/jquery/jquery/blob/e4fd41f/src/manipulation/getAll.js#L10

Issue #3222, opened 2016-07-08
Skip running ESLint on Node 0.x (was: Do not build jquery on 0.x node versions)


markelog:
### Description ###
ESLint no longer supports those versions, so we are locked on 2.x, wouldn't bringing this up otherwise.
Big con about this is that most users use older 0.10 version. Big pro is that version will not be supported soon.

If we drop it, i suppose we can just document it? 

We support it as i recall because of the assumption that some people use custom builds in such environment, right?
To check this hypothesis we can look for the stats of custom builds and decide through them?

mgol:
Too bad we didn't do it before the 3.0 release...

We haven't been running jsdom tests on Node <4 for a while and that includes Promises/A+ compliance tests so we're not actually testing if jQuery even runs in such a configuration; I assume it's only about building jQuery then and not using it? Our Jenkins server has recently switched to Node 4 on most projects so the infrastructure is taken care of.

I'd feel safer if, for now, until Node 0.10 & 0.12 is dropped by upstream (0.10 on 2016-10-01, 0.12 on 2016-12-31), we just skipped running ESLint on Node <4 (I assume it will install fine everywhere since it doesn't have native dependencies or postinstall hooks depending on a newer Node).

markelog:
> we just skipped running ESLint on Node <4 

Yeah, we could do that, although it would require some hacking and unpleasantness in grunt file

timmywil:
Per meeting, skip running eslint on Node < 4. Dropping these version from travis has been moved to the roadmap for later.

mgol:
I'll take it.

markelog:
Oh man, you thief! :).

Would you mind update eslint while you do it?

mgol:
@markelog: Sure! PR readyL #3225.

PR #3225, opened 2016-07-11
Build: Skip running ESLint on Node.js 0.x

mgol:
### Summary ###

The PR makes it possible to use the new ESLint 3.x that's incompatible with Node.js older than 4.x but at the same time allows to build jQuery in an older Node. The tasks using problematic modules (like ESLint & jsdom) are skipped there.

### Checklist ###

Fixes gh-3222

cc @markelog

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -14,9 +14,14 @@ module.exports = function( grunt ) {
 
 	var fs = require( "fs" ),
 		gzip = require( "gzip-js" ),
+		oldNode = /^v0\./.test( process.version );
 
-		// Skip jsdom-related tests in Node.js 0.10 & 0.12
-		runJsdomTests = !/^v0/.test( process.version );
+	// Support: Node.js <4
+	// Skip running tasks that dropped support for Node.js 0.10 & 0.12
+	// in those Node versions.
+	function runIfNewNode( task ) {
+		return oldNode ? "print_old_node_message:" + task : task;
+	}
 
 	if ( !grunt.option( "filename" ) ) {
 		grunt.option( "filename", "jquery.js" );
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -14,9 +14,13 @@ module.exports = function( grunt ) {
 
 	var fs = require( "fs" ),
 		gzip = require( "gzip-js" ),
+		oldNode = /^v0\./.test( process.version );
```
markelog: maybe you could incapsulate it in `runIfNewNode` method?

Gruntfile.js
```diff
@@ -176,33 +180,53 @@ module.exports = function( grunt ) {
 	} );
 
 	// Load grunt tasks from NPM packages
-	require( "load-grunt-tasks" )( grunt );
+	// Support: Node.js <4
```
markelog: We have weird syntax for support comments

Gruntfile.js
```diff
@@ -176,33 +180,53 @@ module.exports = function( grunt ) {
...
+		var task = [].slice.call( arguments ).join( ":" );
+		grunt.log.writeln( "Old Node.js detected, running the task \"" + task + "\" skipped..." );
+	} );
+
+	grunt.registerTask( "lint", [
+		"jsonlint",
+		runIfNewNode( "eslint:all" )
+	] );
 
 	// Don't run Node-related tests in Node.js < 1.0.0 as they require an old
```
markelog: Maybe we could change this comment as well?

Gruntfile.js
mgol: I need this variable outside of this function as well (see the `load-grunt-tasks` line).

Gruntfile.js
```diff
@@ -176,33 +180,53 @@ module.exports = function( grunt ) {
...
+	// Don't load the eslint task in old Node.js, it won't parse.
+	require( "load-grunt-tasks" )( grunt, {
+		pattern: oldNode ? [ "grunt-*", "!grunt-eslint" ] : [ "grunt-*" ]
+	} );
 
 	// Integrate jQuery specific tasks
 	grunt.loadTasks( "build/tasks" );
 
-	grunt.registerTask( "lint", [ "jsonlint", "eslint:all" ] );
+	grunt.registerTask( "print_old_node_message", function() {
```
markelog: Do we need different task for it? Can we just do `grunt.log...` in `runIfNewNode`?

Gruntfile.js
mgol: Maybe we should just remove it? It doesn't bring a lot of useful info right now.

Gruntfile.js
```diff
@@ -176,33 +180,53 @@ module.exports = function( grunt ) {
 	} );
 
 	// Load grunt tasks from NPM packages
-	require( "load-grunt-tasks" )( grunt );
+	// Support: Node.js <4
+	// Don't load the eslint task in old Node.js, it won't parse.
```
markelog: It wouldn't?

Gruntfile.js
mgol: We can't, they'd get logged at the time of task registration instead of invocation.

Gruntfile.js
mgol: Yup. It now depends on a new chalk version that uses `const`s in the code.

markelog: ok-ok

Gruntfile.js
markelog: Can't we use http://gruntjs.com/api/grunt.task#grunt.task.run ?

markelog: Wow, new font in github, after page update :)

Gruntfile.js
markelog: Sounds good to me

Gruntfile.js
mgol: > Can't we use http://gruntjs.com/api/grunt.task#grunt.task.run ?

How would it help here?

> Wow, new font in github, after page update :)

A little weird at first sight. ;)

Gruntfile.js
mgol: Removed.

Gruntfile.js
markelog: > How would it help here?

I guess it would involve refactoring, it would be cool to aim to have minimum footprint for older node version, so we would remove in December in one move, anyway not insist here

mgol: I just don't see how we could do it using `grunt.task.run` without that many modifications as in the current version of this PR.

Anyway, I tried to encapsulate it as much as possible, in most cases it will require just unwrapping the task name from `runIfNewNode()` so no biggie.

mgol:
This should be good to go. Any objections to land?

Gruntfile.js
dmethvin: The chalk package.json file on master says it [still works](https://github.com/chalk/chalk/blob/d939a3f9bc70b9f6d265934047e4a191078180b4/package.json#L13) with node 0.10 and I don't see a `const` in it. Maybe one of its dependents?

mgol: Actually, it's not `chalk` but `grunt-eslint`:
https://github.com/sindresorhus/grunt-eslint/commit/d0ff94266c668a5d6ad2bfd5bb0e92445f6a580a

dmethvin: OK, that makes sense. Go for it.

timmywil:
LGTM

Issue #3227, opened 2016-07-13
Deferred should group async operations


jbedard:
Today each promise operation has it's own `setTimeout` creating many async events.

The aplus spec allows promise operations to be grouped (see [note 3.1](https://promisesaplus.com/#point-67)) and the operations also don't have to be strictly async, as long as they are called by "platform code" (see [2.2.4](https://promisesaplus.com/#point-34)).

Grouping the operations and reducing `setTimeout`s will:
* Improve performance by having less timeouts/async events
* Reduce the probability of aplus tests failing with `Error: timeout of 200ms exceeded` errors. I see this locally fairly often.
* Allow other "platform code" to flush the pending promise operations. For example load-before-ready (#3194, #3197) could be fixed by flushing the pending promise operations on the [DOMContentLoaded event](https://github.com/jquery/jquery/blob/ad6a94c3f1747829082b85fd53ee2efbae879707/src/core/ready.js#L70-L74).

dmethvin:
Seems like a great idea! Should improve performance.

marcus-hiles:
Great suggestion @jbedard

PR #3228, opened 2016-07-13
Deferred: invoke processes in async groups instead of individually

jbedard:
This is a first shot at #3227.

This does not invoke `deferredTick` anywhere else yet.

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -401,7 +401,7 @@ jQuery.extend( {
 
 function deferredTick() {
 	processQueueTimeout = undefined;
-	jQuery.each( processQueue.splice( 0 ), function( _, process ) {
+	jQuery.each( processQueue.splice( 0, processQueue.length ), function( _, process ) {
 		process();
 	} );
 }
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -45,6 +45,9 @@ function adoptValue( value, resolve, reject ) {
 	}
 }
 
+var processQueue = [];
+var processQueueTimeout;
+
 jQuery.extend( {
 
 	Deferred: function( func ) {
@@ -212,12 +215,14 @@ jQuery.extend( {
 										}
 									};
 
+							processQueue.push( process );
+
 							// Support: Promises/A+ section 2.3.3.3.1
 							// https://promisesaplus.com/#point-57
 							// Re-resolve promises immediately to dodge false rejection from
 							// subsequent errors
 							if ( depth ) {
-								process();
+								deferredTick();
 							} else {
 
 								// Call an optional hook to record the stack, in case of exception
@@ -225,7 +230,11 @@ jQuery.extend( {
 								if ( jQuery.Deferred.getStackHook ) {
 									process.stackTrace = jQuery.Deferred.getStackHook();
 								}
-								window.setTimeout( process );
+
+								// Ensure processing has been scheduled
+								if ( !processQueueTimeout ) {
+									processQueueTimeout = window.setTimeout( deferredTick );
+								}
 							}
 						};
 					}
@@ -384,6 +393,13 @@ jQuery.extend( {
 		return master.promise();
 	}
 } );
+
+function deferredTick() {
+	processQueueTimeout = undefined;
+	jQuery.each( processQueue.splice( 0 ), function( _, process ) {
+		process();
+	} );
+}
 
 return jQuery;
 } );
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -186,8 +186,13 @@ jQuery.extend( {
 								},
 
 								// Only normal processors (resolve) catch and reject exceptions
+								// Exceptions in other cases should not effect the promise result
 								process = special ?
-									mightThrow :
+									function() {
+										try {
+											mightThrow();
+										} catch ( e ) {}
+									} :
 									function() {
 										try {
 											mightThrow();
--- a/test/unit/deferred.js
+++ b/test/unit/deferred.js
@@ -372,6 +372,28 @@ QUnit.test( "jQuery.Deferred.then - deferred (progress)", function( assert ) {
 	} );
 } );
 
+QUnit.test( "jQuery.Deferred.then - deferred (progress) exceptions", function( assert ) {
+
+	assert.expect( 1 );
+
+	var piped1, piped2,
+		defer = jQuery.Deferred(),
+		done = assert.async();
+
+	piped1 = defer.then( null, null, function() {
+		throw new Error( "You cant see me" );
+	} );
+
+	piped2 = defer.then( null, null, function( n ) {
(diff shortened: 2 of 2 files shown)


src/deferred.js
```diff
@@ -212,20 +215,26 @@ jQuery.extend( {
...
 
+							processQueue.push( process );
+
 							// Support: Promises/A+ section 2.3.3.3.1
 							// https://promisesaplus.com/#point-57
 							// Re-resolve promises immediately to dodge false rejection from
 							// subsequent errors
 							if ( depth ) {
-								process();
+								deferredTick();
```
jbedard: I *think* this specific line could actually change the execution order of `process`s.

Previously this single `process` would be invoked before other scheduled async ones. Now it will invoke other scheduled ones first, then this one last, but it is still "immediately" (see comment a couple line above).

Can anyone think of a case where this is wrong?

dmethvin: So the only possible callback that might throw and not be caught is the last one, right?

However, would something like `deferred.then(fn1).done(fn2)` force fn1 to be synchronous in order to run fn2? If the queue were per-Deferred that wouldn't happen, but it is global/shared.

src/deferred.js
```diff
@@ -385,5 +394,12 @@ jQuery.extend( {
 	}
 } );
 
+function deferredTick() {
+	processQueueTimeout = undefined;
+	jQuery.each( processQueue.splice( 0 ), function( _, process ) {
```
timmywil: Why `splice`? If it's to copy the array, use `slice`. Also, shouldn't processes be removed from the process queue to avoid multiple executions?

jbedard: The `.splice(0)` clears the `processQueue`. That's why `splice` instead of `slice`. This does make the assumption that `process()` will never through anything though...

src/deferred.js
jbedard: > So the only possible callback that might throw and not be caught is the last one, right?

I'm not sure actually. Can you think of any examples where a throwing function would be pushed onto `processQueue`?

> However, would something like deferred.then(fn1).done(fn2) force fn1 to be synchronous in order to run fn2?

All the scenarios I've tried seem to be valid to. Some examples... (where fn1 always executes before fn2, fn3 before fn4 etc).

```javascript
deferred.then(fn1).done(fn2)
```
```javascript
var def = $.Deferred();
def.then(fn1).done(fn2);
def.resolve(-1);
$.when(5).then(fn3);
```
```javascript
var def = $.Deferred();
def.then(fn2).done(fn3);
$.when(5).then(fn1);
def.resolve(-1);
```
```javascript
var def = $.Deferred();
def.then(fn2).done(fn3);
def.resolve(-1);
$.when(5).done(fn1);
```

dmethvin: Sorry, I think the case I was looking for was `deferred.then(fn1).notify(fn2)` since I think that's the only time `special` comes into play? This definitely could use a @gibson042 look.

src/deferred.js
timmywil: You're right. I was thinking it just copied without the second argument. *hits forehead*

However, that assumption you mentioned seems dangerous.

src/deferred.js
jbedard: The notify case is an issue, see 4bc074b

src/deferred.js
jbedard: I've added 4bc074b which makes all methods pushed onto this array catch exceptions. The other option would be doing that here.

dmethvin: I don't think we want to catch exceptions for that case do we? The old-style Deferred methods are supposed to throw immediately and not be caught.

jbedard: Is the test I added in 4bc074b correct? One notify throwing should not prevent the second...?

jbedard: FYI I've added the length argument to `splice`. It can only be emitted as of ES6.

dmethvin: > Is the test I added in 4bc074b correct? One notify throwing should not prevent the second...?

That achieves the goal of getting all the notifies to run but it swallows an exception in the process.  Maybe we've reached a choice between the lesser of two evils, I dunno. If throwing on a notify doesn't prevent Promise/A+ compat we should allow it. Swallowing the error seems really undesirable though.

src/deferred.js
```diff
@@ -183,8 +186,13 @@ jQuery.extend( {
 								},
 
 								// Only normal processors (resolve) catch and reject exceptions
+								// Exceptions in other cases should not effect the promise result
 								process = special ?
-									mightThrow :
+									function() {
+										try {
+											mightThrow();
+										} catch ( e ) {}
```
gibson042: Swallowing errors from progress callbacks is a problem. We just went through a big exercise with `ready` to prevent its analogous situation.

jbedard: This was mainly just to test this out. I'm thinking that exceptions from a progress callback should be handled the same as exceptions in a success/error callback (~20 lines down). If we're ok with that then I think these two methods wrapping `mightThrow` can be combined and the `special` check will go within that single wrapper. Basically `special` just means that `mightThrow` throwing does not reject the promise...

src/deferred.js
gibson042: This PR does introduce some sequence jumping. Note how callbacks follow an orderly progression [with the current codebase](https://jsfiddle.net/khhyxrbd/), but the `.then`s get clustered and pulled forward [by this PR](https://jsfiddle.net/khhyxrbd/1/). That's not a problem (and is in fact the whole point), but the other part is potentially a problem: that the mere introduction of a bound Deferred (i.e., constructed by a `.then` that returned a thenable) affects **all** Deferred callbacks, pulling forward to pseudo-synchronicity callbacks that previously waited for the next event loop _in front of_ other callbacks.

I'm against this PR as-is, but in favor of the broader refactoring that was hoped for when the Deferred changes first landed, to abstract out Promises/A+ compatibility code for both `Deferred#then` and `jQuery.when`, and in particular to eliminate this synchronous call (which exists to catch bad then implementations that throw after invoking a callback) by incrementing `maxDepth` earlier in the resolution process.

src/deferred.js
gibson042: > Maybe we've reached a choice between the lesser of two evils

Are the possibly-redundant timeouts from our current ungrouped implementation really so evil? I'd sure prefer them to masking errors, and we don't have the option of rejecting because only standard callback completion is permitted to do that.

src/deferred.js
gibson042: So send them through `exceptionHook`? Yeah, that makes sense.

src/deferred.js
jbedard: Would it make sense to revert this one line?

gibson042: No, because then things would probably fall out of spec. I'm instead advocating for broader changes that would eliminate the need for doing something different when `depth > 0`, at which point there won't be any queue jumping.

mgol:

PR #3233, opened 2016-07-14
Build: Update eslint config and fix associated errors

markelog:
### Checklist ###

* [x] Grunt build and unit tests pass locally with these changes (Checked with FF, Chrome and IE11)

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -20,6 +20,31 @@ define( [
 var rbrace = /^(?:\{[\w\W]*\}|\[[\w\W]*\])$/,
 	rmultiDash = /[A-Z]/g;
 
+function getData( data ) {
+	if ( data === "true" ) {
+		return true;
+	}
+
+	if ( data === "false" ) {
+		return false;
+	}
+
+	if ( data === "null" ) {
+		return null;
+	}
+
+	// Only convert to a number if it doesn't change the string
+	if ( data === +data + "" ) {
+		return +data;
+	}
+
+	if ( rbrace.test( data ) ) {
+		return JSON.parse( data );
+	}
+
+	return data;
+}
+
 function dataAttr( elem, key, data ) {
 	var name;
 
@@ -31,14 +56,7 @@ function dataAttr( elem, key, data ) {
 
 		if ( typeof data === "string" ) {
 			try {
-				data = data === "true" ? true :
-					data === "false" ? false :
-					data === "null" ? null :
-
-					// Only convert to a number if it doesn't change the string
-					+data + "" === data ? +data :
-					rbrace.test( data ) ? JSON.parse( data ) :
-					data;
+				data = getData( data );
 			} catch ( e ) {}
 
 			// Make sure we set the data so it isn't changed later
--- a/src/event.js
+++ b/src/event.js
@@ -633,7 +633,19 @@ jQuery.each( {
 
 		// Add which for click: 1 === left; 2 === middle; 3 === right
 		if ( !event.which && button !== undefined && rmouseEvent.test( event.type ) ) {
-			return ( button & 1 ? 1 : ( button & 2 ? 3 : ( button & 4 ? 2 : 0 ) ) );
+			if ( button & 1 ) {
+				return 1;
+			}
+
+			if ( button & 2 ) {
+				return 3;
+			}
+
+			if ( button & 4 ) {
+				return 2;
+			}
+
+			return 0;
 		}
 
 		return event.which;
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,9 +3,5 @@
 	"root": true,
 	"env": {
 		"node": true
-	},
-	rules: {
-		// Until https://github.com/jquery/eslint-config-jquery/issues/7 is resolved
-		"no-unused-expressions": "error"
 	}
 }
--- a/dist/.eslintrc
+++ b/dist/.eslintrc
@@ -5,6 +5,7 @@
 		"no-multiple-empty-lines": "off",
 
 		// Because sizze is not compatible to jquery code style
+		"no-nested-ternary": "off",
 		"no-unused-expressions": "off",
 		"lines-around-comment": "off",
 		"space-in-parens": "off",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "commitplease": "2.3.1",
     "core-js": "2.2.2",
     "cross-spawn": "2.2.3",
-    "eslint-config-jquery": "0.1.6",
+    "eslint-config-jquery": "1.0.0",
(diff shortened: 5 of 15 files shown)


src/data.js
```diff
@@ -31,14 +52,7 @@ function dataAttr( elem, key, data ) {
...
 			try {
-				data = data === "true" ? true :
-					data === "false" ? false :
-					data === "null" ? null :
-
-					// Only convert to a number if it doesn't change the string
-					+data + "" === data ? +data :
-					rbrace.test( data ) ? JSON.parse( data ) :
-					data;
+				data = getData( data );
```
mgol: What's the size difference after extracting this to a function?

mgol:
Related to ESLint configs: the `dot-notation` rule is duplicated in [test/.eslintrc](https://github.com/jquery/jquery/blob/96966c022079e3956018717136367fbf4ed548a2/test/.eslintrc#L53).

src/data.js
markelog: This pull cost 32 bytes, not sure about this specific case

markelog:
> Related to ESLint configs: the dot-notation rule is duplicated in test/.eslintrc.

Will fix that

markelog:
> Related to ESLint configs: the dot-notation rule is duplicated in test/.eslintrc.

Wait, what do you mean? This rule used in `dist` and `test` folder

markelog:
ooh, i get it :)

src/data.js
mgol: It seems every other change should be optimizable by UglifyJS so 32 bytes seems to be quite a lot...

timmywil: Is it weird that I find the ternary more readable?

src/event.js
```diff
@@ -633,7 +633,19 @@ jQuery.each( {
...
+
+			if ( button & 2 ) {
+				return 3;
+			}
+
+			if ( button & 4 ) {
+				return 2;
+			}
+
+			return 0;
```
timmywil: Yet, I find these if statements much more readable than that ternary.

src/data.js
```diff
@@ -20,6 +20,27 @@ define( [
...
+		case "false":
+			return false;
+		case "null":
+			return null;
+
+		// Only convert to a number if it doesn't change the string
+		case ( +data + "" ):
+			return +data;
+		default:
+			if ( rbrace.test( data ) ) {
```
mgol: It's pretty arbitrary that this is an `if` and `( +data + "" )` is a `case`.

mgol: I also wonder if changing that to `if-else if-else if` etc. wouldn't compress better.

timmywil:
I am in favor of every change but the data one. The switch statement is overly expansive.

src/event.js
markelog: Not everything could be generalized, but in general nested or in this case deep nested ternaries are less readable. 

For me thought, this is much more clearer

markelog: Also, if you have objections, it would be better to get them from the start - https://github.com/jquery/eslint-config-jquery/issues/2

src/data.js
markelog: @mgol `switch..case` usually not optimized by the engines or the minificators :/, we can try `if..else` though

markelog:
You guys are totally right, it saved 14 bytes, nice

src/data.js
markelog: > It's pretty arbitrary that this is an if and ( +data + "" ) is a case.

Actually, i just don't if you can use `rbrace.test( data )` in `case`. Can you?

mgol: Ah, it's a boolean that's not being compared to `data`, you're right.

mgol:
It'd be interesting to see what causes the +18 bytes increase (maybe sth could be optimized better by us or Uglify). But increased readability is worth 14 bytes anyway.

src/data.js
```diff
@@ -20,6 +20,31 @@ define( [
...
+	if ( data === +data + "" ) {
+		return +data;
+	}
+
+	if ( rbrace.test( data ) ) {
+		return JSON.parse( data );
+	}
+
+	return data;
+}
```
timmywil: 👍

src/event.js
markelog: > Sometimes it's hard to know if you like something or not until you see how it affects the code though

yeeeah, i guess i get that

timmywil: I was saying I like this change.

src/data.js
gibson042: We should consider disabling `no-nested-ternary`... this is getting too aggressive for too little benefit (and even negative benefit, though I know which side of that argument I'm on).

src/event.js
gibson042: Also equivalent, and probably smaller:
```js
return (button & 1) ||
	(button & 2 ? 3 :
	(button & 4) / 2);
```
(where the last `/ 2` can be anything that maps 0 to itself and 4 to 2—`&& 2`, `>> 1`, etc.).

markelog: It actually adds three bytes, beside that, the mere fact that you have to explain your code is enough reason for me to hate it

dmethvin: :smile_cat:

gibson042: The `&& 2` change is size neutral against master @ 96966c022079e3956018717136367fbf4ed548a2.

markelog: I measured it against this pr

dmethvin: I don't care how much the baby weighs, I still think it's an ugly baby. :smile:

timmywil: Was looking into this further. I think this code was only for IE6-8. I noticed that this fallback expects quirksmode behavior (http://www.quirksmode.org/js/events_properties.html#button), not the standard behavior of `event.button`. I think we can remove this code!

markelog: Wow, awesome! Let's consider this in the next pr though, let's kill that baby :)

dmethvin: @timmywil I think we still need `event.which` mapping, see gh-2337, I think a lot of code still depends on it.

timmywil: @dmethvin I'm not sure you understood what I meant. Note that the `if ( !event.which` block is only for IE6-8 and would not work right in modern browsers.

markelog:
It seems there is a consensus about this pr, as you probably know such prs get stale pretty quickly since they related to code style so more prs land in the harder to support it, so i would like to land it in couple hours

dmethvin:
No objections here.

src/event.js
markelog: Heh, was about do open ticket about this find and then there is Dave - https://github.com/jquery/jquery/issues/3235 :)

Issue #3229, opened 2016-07-13
jQuery.ajax() with cache: false mangles URL with _ parameter


FarSeeing:
`jQuery.ajax()` with `cache: false` option adds/replaces `_` parameter but it mangles URL with already defined parameter: when trying `example.com?_=1&foo=bar` the actual request with be sent to `example.com&foo=bar?_=...`

```javascript
// This works fine:
$.ajax({
  url: 'example.com?_=1&foo=bar'
});

// This fails:
$.ajax({
  url: 'example.com?_=1&foo=bar',
  cache: false
});
```

Tested with 3.x

[Test case](http://jsbin.com/medisokupo/1/edit?html,js,console,output)

Update: the original issue was about `cache: true` when it should be `cache: false`.

dmethvin:
Confirmed, when `cache: false` is given it results in `example.com&foo=bar`. 

@FarSeeing is it a problem for your uses that we are removing the actual `_=1` parameter? Fixing this bug will still remove it.

FarSeeing:
No, not a problem at all. The only thing that bothers me is that query parameters get malformed: `domain&params`

dmethvin:
OK, good. I suspect this hasn't been noticed or reported before because our own code appends the cache buster to the end and that case works fine, even when the request is resubmitted.

prmces:
Hi Guys, 

I also have the same problem. I found that the anti cache parameter is changed as the first parameter when a url is called twice in a row..

The first time is something like this.. 
Original: ?param1=0&param2=1
AntiCached:  ?param1=0&param2=1&_=123131231

The second request, we have this (See that the anticache parameter now is the first parameter)
Original ?_=123131231&param1=0&param2=0 
Anticached &param1=0&param2=1?_=123131231

I also saw a commit with some changes as ranticache and remptyquery variables on this issue.

I tried those changes in my code, but only worked when I changed the replace of the rantichache by a question mark "?". 
The ranticache regex is removing the first & after the ?_=123131231, and adding the ? as a character replacement will work because the replace(remptyquery,"") will do the job if the anticache parameter isthe last paramter.

cacheURL = cacheURL.replace(rantiCache, "**?**").replace(remptyQuery, "");

Hope it could help a little bit

Paulo Costa

PR #3253, opened 2016-07-29
Ajax: Don't mangle the URL when removing the anti-cache param

dmethvin:
### Summary ###

Fixes gh-3229.

There was an existing unit test that actually showed the URL being mangled but there were not any checks for the problem.

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -18,8 +18,7 @@ define( [
 var
 	r20 = /%20/g,
 	rhash = /#.*$/,
-	rantiCache = /(?:[?&])_=[^&]*&?/,
-	remptyQuery = /\?$/,
+	rantiCache = /([?&])_=[^&]*/,
 	rheaders = /^(.*?):[ \t]*([^\r\n]*)$/mg,
 
 	// #7653, #8125, #8152: local protocol detection
@@ -599,8 +598,7 @@ jQuery.extend( {
 
 			// If data is available, append data to url
 			if ( s.data ) {
-				cacheURL = cacheURL.replace( remptyQuery, "" ) +
-					( rquery.test( cacheURL ) ? "&" : "?" ) + s.data;
+				cacheURL += ( rquery.test( cacheURL ) ? "&" : "?" ) + s.data;
 
 				// #9682: remove data so that it's not used in an eventual retry
 				delete s.data;
@@ -608,7 +606,7 @@ jQuery.extend( {
 
 			// Add or update anti-cache param if needed
 			if ( s.cache === false ) {
-				cacheURL = cacheURL.replace( rantiCache, "" ).replace( remptyQuery, "" );
+				cacheURL = cacheURL.replace( rantiCache, "$1" );
 				uncached = ( rquery.test( cacheURL ) ? "&" : "?" ) + "_=" + ( nonce++ ) + uncached;
 			}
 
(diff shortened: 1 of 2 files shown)


src/ajax.js
```diff
@@ -18,7 +18,8 @@ define( [
 var
 	r20 = /%20/g,
 	rhash = /#.*$/,
-	rts = /([?&])_=[^&]*/,
+	rantiCache = /(?:[?&])_=[^&]*&?/,
```
mgol: Sounds almost as "ranty cache". ;-)

mgol: What is the first group for?

dmethvin: Oh good catch, that was from an earlier failed attempt to solve the problem without having to make a second pass to remove the empty `?` querystring. I don't think its needed now.

src/ajax.js
```diff
@@ -597,15 +598,16 @@ jQuery.extend( {
...
 
 				// #9682: remove data so that it's not used in an eventual retry
 				delete s.data;
 			}
 
-			// Add anti-cache in uncached url if needed
+			// Add or update anti-cache param if needed
 			if ( s.cache === false ) {
-				cacheURL = cacheURL.replace( rts, "" );
+				cacheURL = cacheURL.replace( rantiCache, "" ).replace( remptyQuery, "" );
```
gibson042: This code still exhibits the issue (e.g., it converts `/path?_=1&foo=bar` into `/pathfoo=bar`). The root cause of gh-3229 isn't a bad regex, it's a missing "$1" replacement string dropped by https://github.com/jquery/jquery/commit/e077ffb083743f4a4b990f586c9d25d787e7b417#diff-c3749d3acba09ca9ec16bb56e496408bL599 that we just never noticed.

src/ajax.js
```diff
@@ -18,7 +18,8 @@ define( [
 var
 	r20 = /%20/g,
 	rhash = /#.*$/,
-	rts = /([?&])_=[^&]*/,
+	rantiCache = /(?:[?&])_=[^&]*&?/,
+	remptyQuery = /\?$/,
```
gibson042: I wouldn't worry about this at all, since `/path?&_=1` is valid. But if we do care, then it can be solved by restricting the existing regex capture group rather than introducing a new regex: `/(?:(\?)|&)_=[^&]*/` or (probably smaller) `/(\?)_=[^&]*|&_=[^&]*/`. It should be a separate PR, though.

src/ajax.js
```diff
@@ -597,15 +598,16 @@ jQuery.extend( {
 
 			// If data is available, append data to url
 			if ( s.data ) {
-				cacheURL += ( rquery.test( cacheURL ) ? "&" : "?" ) + s.data;
+				cacheURL = cacheURL.replace( remptyQuery, "" ) +
+					( rquery.test( cacheURL ) ? "&" : "?" ) + s.data;
```
gibson042: If we really want to update the "append parameter" logic to handle terminal `?`s, I'd rather see something like `(cacheURL + "&").replace( /\?(.+&$)|\??&$/, "?$1" )` (dropping `rquery` in favor of the new regex) in a separate PR also including [src/ajax/jsonp.js#L47](https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/ajax/jsonp.js#L47).

dmethvin:
Thanks to @gibson042 I think I've fixed the real problem here. The unit test improvements that verify the URL will hopefully prevent a regression.

gibson042:
Works for me.

Issue #3254, opened 2016-08-01
Very slow listeners when filtering by selector.


vitaliylag:
This issue is about listeners with filtering by selector, not just usual listeners. For example:
```js
$('...').on('click', '.myClass', func);
```

Now jQuery is comparing all selectors each time the event occurs, but this is not optimal and take very long time, especially on mobile phones.

You should use this algorithm:

1. For example, let it be #id selector. Create object: key -> id, value -> handler.

2. When the event occurs do something like that:
   ```js
   if (elem.handlerById[elem.id]) elem.handlerById[elem.id]();
   ```
3. If you have more than one handler with the same selector filter then merge them into an array.

4. If you need keep order of calling then add counter to every handler and before calling put them all into array and sort by counter.

5. You can optimize many other selector. For example, for [attr=value] selector you need two nested object, first for searching by attr, second for searching by value, or just one object for searching by attr + "=" + value.

6. If you care about file size you can implement it as module and put link at the https://api.jquery.com/on/. When user requires module $.fn.on begins to work fast.

===

This algorithm works with O(1) speed. Here's example of short implementation that optimize #id selector only, can't work with data, doesn't split handlers by space and doesn't override $.fn.off:

```js
$.fn.on_original = $.fn.on;

$.fn.on = function(event, selector, handler) {
    if (typeof(selector) !== 'string' || !/^#[a-zA-Z0-9_\-]+$/.test(selector)) {
        return this.on_original(event, selector, handler);
    }
    
    for (var i = 0; i < this.length; i++) {
        var elem = this[i];
        
        if (!elem[event]) {
            elem[event] = {
                handlerById: {},
                handlerByClass: {},
                handlerByConcatAttrAndValue: {},
            };
            $(elem).on_original(event, DOMHandler); //Adding common listener without filtering
        }
        
        elem[event].handlerById[selector.substr(1)] = handler;
    }
};

function DOMHandler(e) {
    var handlerById = this[e.type].handlerById;
    var targetElem = e.target;
    
    while (targetElem !== this) {
        if (handlerById[targetElem.id]) {
            handlerById[targetElem.id].call(targetElem, e);
        }
        targetElem = targetElem.parentNode;
    }
}
```
Example of using:
```html
<!DOCTYPE html>

<div id="content">
    <div id="test10">
        <div id="fireMe"></div>
    </div>
</div>

<script src="//ajax.googleapis.com/ajax/libs/jquery/3.1.0/jquery.min.js"></script>
[...]

Issue #2434, opened 2015-06-29
ajax empty response required by rfc but parse attempted


lazylester:
when ajax response status code is 205, rfc requires empty body, but jquery attempts to parse using the json converter and fails. We know that parsing an empty string as json will fail, but attempt to parse should not be made for 205 status code.

Surprisingly, for 204 status code, a parse attempt is still made, but the inevitable errors on empty content are discarded. Surely parsing should not be attempted when rfc disallows content?

mzgol:
Duplicate of #2347.

jQuery only parses the output if it's provided a `Content-Type` header or the `dataType` option is set explicitly. A `Content-Type` header is disallowed for a 205 response:

http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.6
```
The response MUST NOT include an entity.
```
http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.17
```
The Content-Type entity-header field...
```

PR #3255, opened 2016-08-01
Optimize delegated event processing

gibson042:
I wanted to do this before 3.0, but forgot until gh-3254 reminded me. Modest improvements to performance (by preventing the array of delegated handlers from slipping into bag-of-properties mode) and size.

```
   raw     gz Sizes                                                            
264538  78400 dist/jquery.js                                                   
 86492  30001 dist/jquery.min.js                                               

   raw     gz Compared to master @ edf7a43681ca0ba517b155fd5d748921660dfcd2    
  +101    +17 dist/jquery.js                                                   
    -4     -5 dist/jquery.min.js
```

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -360,13 +360,19 @@ jQuery.event = {
 			delegateCount = handlers.delegateCount,
 			cur = event.target;
 
-		// Support: IE <=9
 		// Find delegate handlers
-		// Black-hole SVG <use> instance trees (#13180)
-		//
-		// Support: Firefox <=42
-		// Avoid non-left-click in FF but don't block IE radio events (#3861, gh-2343)
-		if ( delegateCount && cur.nodeType && !( event.type === "click" && event.button >= 1 ) ) {
+		if ( delegateCount &&
+
+			// Support: IE <=9
+			// Black-hole SVG <use> instance trees (trac-13180)
+			cur.nodeType &&
+
+			// Support: Firefox <=42
+			// Suppress spec-violating clicks indicating a non-primary pointer button (trac-3861)
+			// https://www.w3.org/TR/DOM-Level-3-Events/#event-type-click
+			// Support: IE 11 only
+			// ...but not arrow key "clicks" of radio inputs, which can have `button` -1 (gh-2343)
+			!( event.type === "click" && event.button >= 1 ) ) {
 
 			for ( ; cur !== this; cur = cur.parentNode || this ) {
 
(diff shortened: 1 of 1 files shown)


src/event.js
```diff
@@ -366,40 +366,41 @@ jQuery.event = {
...
 
 			for ( ; cur !== this; cur = cur.parentNode || this ) {
 
 				// Don't check non-elements (#13208)
 				// Don't process clicks on disabled elements (#6911, #8165, #11382, #11764)
-				if ( cur.nodeType === 1 && ( cur.disabled !== true || event.type !== "click" ) ) {
-					matches = [];
+				if ( cur.nodeType === 1 && !( event.type === "click" && cur.disabled === true ) ) {
+					matchedHandlers = [];
+					matchedSelectors = {};
```
dmethvin: Yeah, good call to split these out.

src/event.js
```diff
@@ -366,40 +366,41 @@ jQuery.event = {
 		//
 		// Support: Firefox <=42
 		// Avoid non-left-click in FF but don't block IE radio events (#3861, gh-2343)
-		if ( delegateCount && cur.nodeType &&
-			( event.type !== "click" || isNaN( event.button ) || event.button < 1 ) ) {
+		if ( delegateCount && cur.nodeType && !( event.type === "click" && event.button >= 1 ) ) {
```
dmethvin: I know this is shorter but it isn't as obvious. However, I can't recall the browser where `event.button` doesn't exist nowadays though. It seems to be there on IE9+ but maybe Android or something? If that's not a common case I can't really object to this change since it may no longer exist.

dmethvin:
LGTM

src/event.js
timmywil: Right, it's in IE9+.

markelog: Sounds like we neeed a `Support` comment?

timmywil: Not if it's in all our supported browsers.

markelog: It is? I get the impression from this thread that it is not

dmethvin: @markelog yeah, the comment mentions gh-2434 and In c82a6685bb964627e27008e298f93ea81218265b we changed this line because of arrow keys in IE11. How about adding something like this @gibson042 based on my [comment there](https://github.com/jquery/jquery/issues/2343#issuecomment-104758437)?
```js
// Support: IE11
// IE reports button==-1 for "clicks" on radios via the arrow keys
```

gibson042: Good idea. Done.

dmethvin:
:+1:

src/event.js
timmywil: Oh, to clarify that. I was just saying all of our supported browsers have `event.button` now.

Issue #3226, opened 2016-07-11
.not(arrayLike) shouldn't restrict results to elements


gibson042:
As [reported](http://irc.jquery.org/%23jquery-dev/latest.log.html#t11:52:27) by @scottgonzalez, `.not( arrayLike )` started filtering out non-element items in jQuery 3: http://jsbin.com/yobuhucife/1/edit?html,js,output . This is inconsistent with `.not( singleNode )` and `.not( fn )`, and counter to the stated "not(Selector) excludes non-element nodes" goals of the commit that introduced it: 0e2f8f9effd2c95647be534bf9055e099aad7cfd.

We should restrict the additional filtering to selector input (i.e., `.not( string )`).

dmethvin:
If this is a regression, we should update the [docs](http://api.jquery.com/not/) as well, in particular the part under "Removing Specific Elements", to clarify that non-Element nodes are passed through.

nicklou:
How could I help on this fix? With a first glance, the code fix does not seem to be that big. Should I attempt for a PR?

gibson042:
The process is pretty simple; see https://github.com/jquery/jquery/blob/master/CONTRIBUTING.md and https://contribute.jquery.org/commits-and-pull-requests/ . Make sure you've signed our [CLA](http://contribute.jquery.org/CLA/) and are using the same name in your commits, and submit a pull request including both relevant test suite and source code changes.

dmethvin:
@nicklou if you need help just ask!

huguoliang1314:
hello,Mr gibson042@gibson042,i think the first step is judge the element nodeType,
so, return  elem.nodeType === 1&&( indexOf.call( qualifier, elem ) > -1 ) !== not ;

gibson042:
@huguoliang1314: `elem.nodeType === 1` is already there; we want to _skip_ it when the initial qualifier is not a string.

PR #3261, opened 2016-08-02
Traversing: Let .not(arraylike) pass non-element nodes

dmethvin:
### Summary ###

Fixes gh-3226, was only +3 gzip.

The docs at http://api.jquery.com/not/ might need some wording changes to clarify that this is intended behavior?

### Checklist ###

### DIFF
--- a/src/traversing/findFilter.js
+++ b/src/traversing/findFilter.js
@@ -25,14 +25,17 @@ function winnow( elements, qualifier, not ) {
 
 	}
 
-	if ( typeof qualifier === "string" ) {
-		if ( risSimple.test( qualifier ) ) {
-			return jQuery.filter( qualifier, elements, not );
-		}
-
-		qualifier = jQuery.filter( qualifier, elements );
+	if ( typeof qualifier !== "string" ) {
+		return jQuery.grep( elements, function( elem ) {
+			return ( indexOf.call( qualifier, elem ) > -1 ) !== not;
+		} );
 	}
 
+	if ( risSimple.test( qualifier ) ) {
+		return jQuery.filter( qualifier, elements, not );
+	}
+
+	qualifier = jQuery.filter( qualifier, elements );
 	return jQuery.grep( elements, function( elem ) {
 		return ( indexOf.call( qualifier, elem ) > -1 ) !== not && elem.nodeType === 1;
 	} );
--- a/src/traversing/findFilter.js
+++ b/src/traversing/findFilter.js
@@ -15,26 +15,28 @@ function winnow( elements, qualifier, not ) {
 		return jQuery.grep( elements, function( elem, i ) {
 			return !!qualifier.call( elem, i, elem ) !== not;
 		} );
-
 	}
 
+	// Single element
 	if ( qualifier.nodeType ) {
 		return jQuery.grep( elements, function( elem ) {
 			return ( elem === qualifier ) !== not;
 		} );
-
 	}
 
+	// Arraylike of elements (jQuery, arguments, Array)
 	if ( typeof qualifier !== "string" ) {
 		return jQuery.grep( elements, function( elem ) {
 			return ( indexOf.call( qualifier, elem ) > -1 ) !== not;
 		} );
 	}
 
+	// Simple selector that can be filtered directly, removing non-Elements
 	if ( risSimple.test( qualifier ) ) {
 		return jQuery.filter( qualifier, elements, not );
 	}
 
+	// Complex selector, compare the two sets, removing non-Elements
 	qualifier = jQuery.filter( qualifier, elements );
 	return jQuery.grep( elements, function( elem ) {
 		return ( indexOf.call( qualifier, elem ) > -1 ) !== not && elem.nodeType === 1;
--- a/test/unit/traversing.js
+++ b/test/unit/traversing.js
@@ -481,6 +481,19 @@ QUnit.test( "not(Selector) excludes non-element nodes (gh-2808)", function( asse
 	assert.deepEqual( mixedContents.not( "[id=a],*,[id=b]" ).get(), [], "not [id=a],*,[id=b]" );
 } );
 
+QUnit.test( "not(arraylike) passes non-element nodes (gh-3226)", function( assert ) {
+	assert.expect( 4 );
+
+	var mixedContents = jQuery( "#nonnodes" ).contents(),
+		mixedLength = mixedContents.length,
+		childElements = q( "nonnodesElement" );
+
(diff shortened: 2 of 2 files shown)


test/unit/traversing.js
```diff
@@ -481,6 +481,19 @@ QUnit.test( "not(Selector) excludes non-element nodes (gh-2808)", function( asse
...
+QUnit.test( "not(arraylike) passes non-element nodes (gh-3226)", function( assert ) {
+	assert.expect( 4 );
+
+	var mixedContents = jQuery( "#nonnodes" ).contents(),
+		mixedLength = mixedContents.length,
+		childElements = q( "nonnodesElement" );
+
+	assert.deepEqual( mixedContents.not( mixedContents.get() ).get(), [], "not everything" );
+	assert.deepEqual( mixedContents.not( childElements ).length, mixedLength - 1, "not childElements" );
+	assert.deepEqual( mixedContents.not( [ childElements[ 0 ].nextSibling ] ).length, mixedLength - 1, "not textnode" );
```
dmethvin: Suggestions on more rigorous tests are welcome, these seem a bit clumsy but of course it's never fun to manipulate non-element nodes in jQuery.

test/unit/traversing.js
```diff
@@ -481,6 +481,19 @@ QUnit.test( "not(Selector) excludes non-element nodes (gh-2808)", function( asse
 	assert.deepEqual( mixedContents.not( "[id=a],*,[id=b]" ).get(), [], "not [id=a],*,[id=b]" );
 } );
 
+QUnit.test( "not(arraylike) passes non-element nodes (gh-3226)", function( assert ) {
+	assert.expect( 4 );
+
+	var mixedContents = jQuery( "#nonnodes" ).contents(),
```
markelog: Can we create fixtures in the test?

test/unit/traversing.js
```diff
@@ -481,6 +481,19 @@ QUnit.test( "not(Selector) excludes non-element nodes (gh-2808)", function( asse
...
+		mixedLength = mixedContents.length,
+		childElements = q( "nonnodesElement" );
```
markelog: fyi `one-var` kinda thing is no longer a requirement

test/unit/traversing.js
dmethvin: Yeah, I'll do that. It was copied from the example above it but I prefer fixtures in the test too.

test/unit/traversing.js
markelog: Yeah, this is very fragile test, since it doesn't use generated fixtures.

Beside that, sounds like we need all node types? At least `text` and `comment`, maybe processing node too, not sure if we need to test/support deprecated nodes though.

It would be cool if it would be shown explicitly either through the code or with comment or through assertion name

markelog:
Maybe another test exactly like from the ticket should be added too? http://jsbin.com/yobuhucife/1/edit?html,js,output
--
hm, can it break user code? It seems it can :/. Hopefully not in the big way

src/traversing/findFilter.js
```diff
@@ -25,14 +25,17 @@ function winnow( elements, qualifier, not ) {
 
 	}
 
-	if ( typeof qualifier === "string" ) {
-		if ( risSimple.test( qualifier ) ) {
-			return jQuery.filter( qualifier, elements, not );
-		}
+	if ( typeof qualifier !== "string" ) {
+		return jQuery.grep( elements, function( elem ) {
+			return ( indexOf.call( qualifier, elem ) > -1 ) !== not;
```
markelog: Additional parentheses are not required? Either way is good though

src/traversing/findFilter.js
```diff
@@ -25,14 +25,17 @@ function winnow( elements, qualifier, not ) {
 
 	}
 
-	if ( typeof qualifier === "string" ) {
-		if ( risSimple.test( qualifier ) ) {
-			return jQuery.filter( qualifier, elements, not );
-		}
+	if ( typeof qualifier !== "string" ) {
```
markelog: Maybe couple comments on what's for what? Not immediately obvious right now, although it sounds like a bonus round

src/traversing/findFilter.js
markelog: Oh, i see we already use it like that, couple lines down

kbpindiga:
Good

src/traversing/findFilter.js
```diff
@@ -15,24 +15,29 @@ function winnow( elements, qualifier, not ) {
...
 			return ( elem === qualifier ) !== not;
 		} );
-
 	}
 
-	if ( typeof qualifier === "string" ) {
-		if ( risSimple.test( qualifier ) ) {
-			return jQuery.filter( qualifier, elements, not );
-		}
+	// Arraylike of elements (jQuery, arguments, Array)
```
markelog: Awesome! Thank you

src/traversing/findFilter.js
```diff
@@ -15,24 +15,29 @@ function winnow( elements, qualifier, not ) {
...
+	if ( typeof qualifier !== "string" ) {
+		return jQuery.grep( elements, function( elem ) {
+			return ( indexOf.call( qualifier, elem ) > -1 ) !== not;
+		} );
+	}
 
-		qualifier = jQuery.filter( qualifier, elements );
+	// Simple selector that can be filtered directly, removing non-Elements
+	if ( risSimple.test( qualifier ) ) {
```
dmethvin: (Side note, this can be addressed in a different PR but I wanted to be sure this seemed correct.)

This test for a simple selector needs to be there for the "not" case mainly because `jQuery.filter()` always removes text/comment nodes. However, it should be safe to always take this branch if `!not` correct? It seems this appeared in the `.filter()` path in 87869b8b81380617cfad8788799ee40de0e3e9c4 when it was DRYed out.

dmethvin: Oh wait, that can't be the reason since we are going to remove them anyway. Now I'm not sure I know WHY this is here, since the `:not()` pseudo says it takes any selector.

gibson042: It looks like a performance optimization for saving the O(n²) operations of invoking `indexOf` in a loop. And I'm guessing it only applies to simple selectors because `:not(non-simple selector)` is invalid CSS3 and therefore gets rejected by `querySelectorAll`, but if so that's irrelevant now because `jQuery.filter` only uses native methods when passed a single element. I'm in favor of using `return jQuery.filter( qualifier, elements, not )` for _all_ string `qualifier`s.

markelog: Sounds like we need an issue? It is good to add comments to stuff :)

gibson042: gh-3272

PR #3284, opened 2016-08-21
Remove useless variable `readyFiring`.

stevemao:


gibson042:
`readyFiring` is exactly useless, but I agree that `ready` stack depth is unlikely to get out of control. LGTM.

timmywil:
@gibson042 I think you meant "isn't". We added it as a precaution, but I agree that it is probably not needed.

PR #3285, opened 2016-08-21
Add missing Qunit script

stevemao:


test/readywait.html
```diff
@@ -14,6 +14,7 @@
 		#output { background-color: green }
 		#expectedOutput { background-color: green }
 	</style>
+	<script src="../external/qunit/qunit.js"></script>
```
gibson042: This is pointless, since nothing on the page attempts to access `QUnit`. But the page itself is totally disconnected from unit testing AFAICT. However, that need not be the case... @stevemao, would you be willing to make the necessary changes to invoke readywait.html as an [iframe test](https://github.com/jquery/jquery/blob/master/README.md#run-tests-in-an-iframe)?

stevemao: Sounds like a better solution :)

timmywil:
Since this PR got stale, I'll close it. But we'd gladly accept a different PR that invokes readywait.html in an iframe test.

Issue #3288, opened 2016-08-22
Deprecate jQuery.holdReady


gibson042:
Suggested in https://github.com/jquery/jquery/issues/1778#issuecomment-72764833 and decided in https://github.com/jquery/jquery/pull/2850#issuecomment-173315469 . Gap noticed in #3287.

dmethvin:
Added to the API list as well.

nithmanoj:
I want to work on this issue, please inform me what to do of `test/data/readywaitloader.js` & `test/data/readywait.html`. The checklist of me goes like this - 

- [ ] Move `jQuery.holdReady()` from `src/core/ready.js` to `src/deprecated.js`
- [ ] Remove `jQuery.holdReady()` from `src/core/ready-no-deferred.js` & `test/data/jquery-1.9.1.js` 
- [ ] Open a migrate issue

gibson042:
test/data/jquery-1.9.1.js should be left alone, as should the other test files. Other than that, :+1:.

PR #3306, opened 2016-08-31
Move holdReady to deprecated

kumarmj:
fixes issue #3288

### DIFF
--- a/src/core/ready.js
+++ b/src/core/ready.js
@@ -34,15 +34,6 @@ jQuery.extend( {
 	// the ready event fires. See #6781
 	readyWait: 1,
 
-	// Hold (or release) the ready event
-	holdReady: function( hold ) {
-		if ( hold ) {
-			jQuery.readyWait++;
-		} else {
-			jQuery.ready( true );
-		}
-	},
-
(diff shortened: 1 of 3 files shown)


timmywil:
LGTM

mgol:
LGTM

src/core/ready.js
```diff
@@ -34,15 +34,6 @@ jQuery.extend( {
 	// the ready event fires. See #6781
```
dmethvin: When removing this we may want to note in a comment that Migrate needs `readyWait` to shim `.holdReady()`. If/when .holdReady()` is removed in v4 we might be tempted to remove or simplify this logic, or make `readyWait` private. A similar thing happened in 3.0 when we removed `jQuery.sub()` and simplified `jQuery.fn.init()` so I'm just being paranoid.

dmethvin:
We have a `.holdReady()` test but it doesn't look like it runs as part of the unit tests? I was going to ask whether anything needed to be moved to `test/deprecated.js` but it doesn't look like it does.

dmethvin:
Oh! See gh-3331.

Issue #3072, opened 2016-04-22
Class-related methods fail to meet w3c standard for whitespace handling


zzzzBov:
[I opened a related discussion about whitespace on WICG][1] that may be worth a read before any changes are necessarily made regarding this.

the `*Class` methods have consistently failed to handle ["space characters"][5] correctly, and I'd reported [a related bug][2] a few years ago.

There are exactly 5 characters that are used to separate classes in HTML/CSS. Those characters in ascending unicode order are:

* tab (`\t`, U+0009)
* line feed (`\n`, U+000A)
* form feed (`\f`, U+000C)
* carriage return (`\r`, U+000D)
* space (` `, U+0020)

When other characters that are matched by the regular expression pattern `\s` are used in classes, jQuery fails to behave correctly according to the spec.

A quick example of this failure:

```
$({
  nodeType: 1,
  getAttribute: function () {
    return 'foo'
  },
  setAttribute: function (name, val) {
    console.assert(val === 'foo foo\u00A0bar', 'no-break space should not be treated as a class separator');
    console.log(JSON.stringify(val));
  }
}).addClass('foo\u00A0bar');
```

which produces an assertion failure and `'foo bar'` logged to console.

[fiddle][3]

The problem is that [`classes.js` uses `rnotwhite`][4] which is (`/\S+/g`) and *not* an appropriate regular expression to use for splitting strings of classes.

[1]: https://discourse.wicg.io/t/whitespace-is-hard-and-buggy-can-we-normalize-it/1436/1
[2]: https://bugs.jquery.com/ticket/13835
[3]: https://jsfiddle.net/8960sop4/
[4]: https://github.com/jquery/jquery/blob/master/src/attributes/classes.js#L26
[5]: https://www.w3.org/TR/html5/infrastructure.html#space-character

dmethvin:
So it's just a case of using that specific list (negated) as our regexp, right? Is there any problem with updating that now? Per the other discussion I understand that some standards define whitespace differently but for the purpose of fixing our class methods it doesn't matter.

gibson042:
We use `rnotwhite` for a few purposes beyond parsing a list of classes:
* [removing multiple attributes](https://github.com/jquery/jquery/blob/055cb7534e2dcf7ee8ad145be83cb2d74b5331c7/src/attributes/attr.js#L90)
[...]

Issue #3003, opened 2016-03-16
Internal use of jQuery.trim erroneously strips too much


gibson042:
All internal use of `jQuery.trim` should be replaced with the HTML [strip and collapse whitespace](https://html.spec.whatwg.org/multipage/infrastructure.html#strip-and-collapse-whitespace) operation, which only modifies a [small set of ASCII whitespace](https://html.spec.whatwg.org/multipage/infrastructure.html#space-character) characters.

Problem: https://jsfiddle.net/5Lhfgd3r/
Solution: https://github.com/jquery/jquery/pull/3002#issuecomment-197138246

markelog:
I probably know answer to this, but do we want to expose it?

markelog:
Also, i don't think you mean *All internal use*

timmywil:
:+1: It might just be all internal uses, but we'll want to double check each one.

gibson042:
I really do mean all internal uses: https://github.com/jquery/jquery/search?utf8=%E2%9C%93&q=trim+path%3Asrc&type=Code

markelog:
So why we need to trim `dataType` ajax argument in way that is compatible with html spec?

gibson042:
> So why we need to trim `dataType` ajax argument in way that is compatible with html spec?

CSS Selectors spec in that case, which fortunately [shares](https://www.w3.org/TR/css3-selectors/#selector-syntax) the HTML definition of whitespace. And it's not about what we need to do, but about what we need to _stop_ doing: https://jsfiddle.net/5Lhfgd3r/1/

gibson042:
Oops, wrong example. https://github.com/jquery/jquery/blob/93a8fa6bfc1c8a469e188630b61e736dfb69e128/src/ajax.js#L532 should abandon `jQuery.trim` because we [document](http://api.jquery.com/jQuery.ajax/) "space-separated values", making it at best superfluous (and in fact, it looks to me like the call can be removed and replaced with nothing, since the subsequent `match` will ignore leading/trailing space anyway).

dmethvin:
Agreed, there's no need to trim at all with that regex.

mgol:
@gibson042 Do you want to take it? If you don't have time I can get it & #3072.

PR #3316, opened 2016-09-12
Core: rnotwhite -> rhtmlnotwhite and jQuery.trim -> stripAndCollapse

timmywil:
- Renames and changes rnotwhite to focus on HTML whitespace chars
- Change internal use of jQuery.trim to more accurate strip and collapse
- Adds tests to ensure HTML space characters are retained where valid
- Doesn't add tests where the difference is inconsequential and
  existing tests are adequate.

Fixes gh-3003
Fixes gh-3072

### DIFF
--- a/src/core/stripAndCollapse.js
+++ b/src/core/stripAndCollapse.js
@@ -1,0 +1,12 @@
+define( function() {
+	"use strict";
+
+	// Strip and collapse whitespace according to HTML spec
+	// https://html.spec.whatwg.org/multipage/infrastructure.html#strip-and-collapse-whitespace
+	var rhtmlSpace = /[\x20\t\r\n\f]+/g,
+		stripAndCollapse = function( value ) {
+			return ( " " + value + " " ).replace( rhtmlSpace, " " ).slice( 1, -1 );
+		};
+
+	return stripAndCollapse;
+} );
--- a/src/var/rnothtmlwhite.js
+++ b/src/var/rnothtmlwhite.js
@@ -1,0 +1,8 @@
+define( function() {
+	"use strict";
+
+	// Only count HTML whitespace
+	// Other whitespace should count in values
+	// https://html.spec.whatwg.org/multipage/infrastructure.html#space-character
+	return ( /[^\x20\t\r\n\f]+/g );
+} );
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -2329,10 +2329,12 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
 
 	// Selector should be trimmed to avoid leading spaces (#14773)
 	// Selector should include any valid non-HTML whitespace (#3003)
-	QUnit.asyncTest( "jQuery.fn.load( URL_SELECTOR with non-HTML whitespace(#3003) )", 1, function( assert ) {
-		jQuery( "#first" ).load( "data/test3.html   #whitespace\xA0 ", function() {
+	QUnit.test( "jQuery.fn.load( URL_SELECTOR with non-HTML whitespace(#3003) )", function( assert ) {
+		assert.expect( 1 );
+		var done = assert.async();
+		jQuery( "#first" ).load( "data/test3.html   #whitespace\\\\xA0 ", function() {
 			assert.strictEqual( jQuery( this ).children( "div" ).length, 1, "Verify that specific elements were injected" );
-			QUnit.start();
+			done();
 		} );
 	} );
 
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -2327,6 +2327,15 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
 		} );
 	} );
 
+	// Selector should be trimmed to avoid leading spaces (#14773)
+	// Selector should include any valid non-HTML whitespace (#3003)
+	QUnit.asyncTest( "jQuery.fn.load( URL_SELECTOR with non-HTML whitespace(#3003) )", 1, function( assert ) {
+		jQuery( "#first" ).load( "data/test3.html   #whitespace\xA0 ", function() {
+			assert.strictEqual( jQuery( this ).children( "div" ).length, 1, "Verify that specific elements were injected" );
+			QUnit.start();
+		} );
+	} );
+
 	QUnit.asyncTest( "jQuery.fn.load( String, Function ) - simple: inject text into DOM", 2, function( assert ) {
 		jQuery( "#first" ).load( url( "data/name.html" ), function() {
 			assert.ok( /^ERROR/.test( jQuery( "#first" ).text() ), "Check if content was injected into the DOM" );
--- a/test/unit/attributes.js
+++ b/test/unit/attributes.js
@@ -669,13 +669,13 @@ QUnit.test( "removeAttr(Multi String, non-HTML whitespace is valid in attribute 
 	};
 
 	jQuery.each( tests, function( key, val ) {
-		assert.equal( div.attr( key ), val, "Attribute `" + key + "` exists, and has a value of `" + val + "`" );
+		assert.equal( div.attr( key ), val, "Attribute \"" + key + "\" exists, and has a value of \"" + val + "\"" );
 	} );
 
 	div.removeAttr( "id   data-\xA0 title  rel  " );
 
 	jQuery.each( tests, function( key ) {
-		assert.equal( div.attr( key ), undefined, "Attribute `" + key + "` was removed" );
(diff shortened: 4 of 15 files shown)


src/core/stripAndCollapse.js
```diff
@@ -0,0 +1,12 @@
+define( function() {
+	"use strict";
+
+	// Strip and collapse whitespace according to HTML spec
+	// https://html.spec.whatwg.org/multipage/infrastructure.html#strip-and-collapse-whitespace
+	var rhtmlSpace = /[\x20\t\r\n\f]+/g,
```
mgol: Can't you write a regular space (` `) instead of the hex code (`\x20`)?

test/unit/ajax.js
```diff
@@ -2327,6 +2327,15 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
 		} );
 	} );
 
+	// Selector should be trimmed to avoid leading spaces (#14773)
+	// Selector should include any valid non-HTML whitespace (#3003)
+	QUnit.asyncTest( "jQuery.fn.load( URL_SELECTOR with non-HTML whitespace(#3003) )", 1, function( assert ) {
+		jQuery( "#first" ).load( "data/test3.html   #whitespace\xA0 ", function() {
+			assert.strictEqual( jQuery( this ).children( "div" ).length, 1, "Verify that specific elements were injected" );
+			QUnit.start();
```
markelog: See https://api.qunitjs.com/QUnit.start/

src/var/rnothtmlwhite.js
```diff
@@ -0,0 +1,8 @@
+define( function() {
+	"use strict";
+
+	// Only count HTML whitespace
+	// Other whitespace should count in values
+	// https://html.spec.whatwg.org/multipage/infrastructure.html#space-character
+	return ( /[^\x20\t\r\n\f]+/g );
```
mgol: Same remark as above.

test/unit/ajax.js
markelog: Also, https://github.com/platinumazure/eslint-plugin-qunit, not sure if it would catch it, but still, what do you think?

test/unit/attributes.js
```diff
@@ -657,6 +657,28 @@ QUnit.test( "removeAttr(Multi String, variable space width)", function( assert )
...
+		};
+
+	jQuery.each( tests, function( key, val ) {
+		assert.equal( div.attr( key ), val, "Attribute `" + key + "` exists, and has a value of `" + val + "`" );
+	} );
+
+	div.removeAttr( "id   data-\xA0 title  rel  " );
+
+	jQuery.each( tests, function( key ) {
+		assert.equal( div.attr( key ), undefined, "Attribute `" + key + "` was removed" );
```
markelog: Do we want to start using backticks in strings?

test/unit/ajax.js
mgol: We should not use `QUnit.asyncTest` for new tests at all, should we?

mgol: @markelog I wanted to look at the docs but I hit this: https://github.com/platinumazure/eslint-plugin-qunit/issues/53. :)

test/unit/attributes.js
```diff
@@ -1117,7 +1139,7 @@ QUnit.test( "val(select) after form.reset() (Bug #2551)", function( assert ) {
 } );
 
 QUnit.test( "select.val(space characters) (gh-2978)", function( assert ) {
-	assert.expect( 35 );
+	assert.expect( 37 );
```
markelog: Maybe we could move this to separate test?

test/unit/attributes.js
```diff
@@ -657,6 +657,28 @@ QUnit.test( "removeAttr(Multi String, variable space width)", function( assert )
 	} );
 } );
 
+QUnit.test( "removeAttr(Multi String, non-HTML whitespace is valid in attribute names (gh-3003)", function( assert ) {
+	assert.expect( 8 );
+
+	var div = jQuery( "<div id='a' data-\xA0='b' title='c' rel='d'></div>" ),
+		tests = {
```
markelog: Just fyi, you can use separate `var` for this

test/unit/ajax.js
markelog: Yeah, https://github.com/platinumazure/eslint-plugin-qunit/tree/master/docs/rules those are available though, you should send them a PR :)

markelog: > We should not use QUnit.asyncTest for new tests at all, should we?

Yeah, that's what I meant by my first comment :)

src/core/stripAndCollapse.js
markelog: This is somehow different from `[ \t\r\n\v\f]` -> `\s` right? You guys already talked about this I believe?

test/unit/ajax.js
platinumazure: Apologies for the broken links. Rule `no-async-test` should definitely catch this, let me know if I can help with integration.

src/core/stripAndCollapse.js
timmywil: > This is somehow should be different from [ \t\r\n\v\f] -> \s right?

Right

test/unit/attributes.js
timmywil: Not sure what you mean. You mean use es2015 template strings? That wouldn't work in all of our environments.

test/unit/attributes.js
timmywil: This is what this test was meant for originally.

test/unit/attributes.js
timmywil: This is just a copy/paste based on another test, but I can do that.

src/core/stripAndCollapse.js
markelog: Could you give a me a context?

timmywil: > Can't you write a regular space () instead of the hex code (\x20)?

We could, but I like that this is more apparent/explicit.

dmethvin: Also it might be a tab. I prefer the `\x20` for that reason.

mgol: It's more characters and a space is a space. :) But I see we're using it in other places like that already. Changing all instances to a space saves only one byte gzipped.

markelog: > Also it might be a tab.

Not sure to what you referring to.

For the reference canonical way is to use ` `, don't have strong opinion about it though

mgol: @dmethvin No, `/ /` doesn't match a tab.

markelog: `\x20` for most though, i think, that symbol doesn't immediately associated with ` `

test/unit/attributes.js
markelog: I'm saying output will be 

> Attribute `id` exists, and has a value of `a`

even though we usually do something like this

> Attribute "id" exists, and has a value of "a"

test/unit/attributes.js
markelog: I'm saying output will be 

> Attribute `id` exists, and has a value of `a`

even though we usually do something like this

> Attribute "id" exists, and has a value of "a"

src/core/stripAndCollapse.js
dmethvin: Right, in the source it could be a tab. Unlikely, but the `\x20

test/unit/attributes.js
markelog: Yeah, sure, i'm saying it would be better not to change the tests, but add different ones. If we would use the terminology, previously that was an unit-test, now we found a bug, so it should be a regression test, so it might make sense not to unite those

src/core/stripAndCollapse.js
markelog: ooh, i see

test/unit/attributes.js
timmywil: oh, I see. Sure!

test/unit/attributes.js
timmywil: The "2 added tests" is a misleading. These tests actually needed editing, not additions.

markelog: Oh, so these editions are unrelated to bug fix? I would move them to different commit, but not feeling strong about this

timmywil: They're related. It's just that there aren't really any new tests to split off in this particular case.

timmywil: To be specific, the bug fix cause these tests to fail. Just editing them accomplished the same thing as adding another test to verify the correct behavior. We were verifying buggy behavior before.

src/core/stripAndCollapse.js
markelog: Found the context, LGTM

Issue #3260, opened 2016-08-02
Source file location in jQuery CDN .map file is incorrect


dmethvin:
At some point in the recent past, Chrome added the ability to manually associate a map file with minified source. For people using the jQuery CDN, this would optimistically mean they could right click on the file in dev tools and for example associate https://code.jquery.com/jquery-3.1.0.min.map with the jQuery 3.1.0 minified file. However, that map file refers back to plain old `jquery.js` as the original source, which is our frozen-in-time 1.11 version. It should refer to `jquery-3.1.0.min.js`.

The release process should ensure the file name matches the one on the CDN so that it could be manually associated for debugging.

Map files are so lame.

dmethvin:
Also to be clear, the current map file name *is* correct for the Google CDN since they deliver their copy with the  name jquery.js. The jQuery and Microsoft CDNs include the version number in the file name, so they require a physically different map file with the version-specific file names.

dmethvin:
Just showing my work here so I can come back to it when I get a chance.

jQuery 1.12.0 is where this regressed. It looks like the code to rename the references in the map file is [still there](https://github.com/jquery/jquery/blob/master/build/release/cdn.js#L41-L48) but the names of the versioned files aren't correct in the jQuery CDN or the [Microsoft CDN](http://www.asp.net/ajax/cdn#jQuery_Releases_on_the_CDN_0).

PR #3321, opened 2016-09-19
Build: Make sourcemap updates in two .replace() passes

dmethvin:
### Summary ###

Fixes gh-3260

The problem was that the [current regexp](https://github.com/jquery/jquery/blob/3bbcce68d7b8b8a7a2164a0f7a280ae9daf70b5c/build/release/cdn.js#L41-L48) assumes that the two fields to be changed are adjacent. At some point that's the way uglify generated them but now the "file" field is at the very end. The solution is simple, make the changes with two separate regexp passes.

Testing this *in situ* with a jquery-release run is super painful because of the safety checks that go on before we get to the CDN creation which is at the tail end of the whole long complicated process, but I tested the regexp replace with this script:

```js
var fs = require("fs");

var unpathedFile = "jquery-3.1.0.min.map";
var text;

text = fs.readFileSync( unpathedFile, "utf8" )
	.replace( /"file":"([^"]+)"/,
		"\"file\":\"" + unpathedFile.replace( /\.min\.map/, ".min.js" ) )
	.replace( /"sources":\["([^"]+)"\]/,
		"\",\"sources\":[\"" + unpathedFile.replace( /\.min\.map/, ".js" ) + "\"]" );
		
fs.writeFileSync( "result.txt", text );
```
### Checklist ###

* [ ] ~~New tests have been added to show the fix or feature works~~ (see above)

### DIFF
--- a/build/release/cdn.js
+++ b/build/release/cdn.js
@@ -43,9 +43,9 @@ function makeReleaseCopies( Release ) {
 			// "file":"jquery.min.js" ... "sources":["jquery.js"]
 			text = fs.readFileSync( builtFile, "utf8" )
 				.replace( /"file":"([^"]+)"/,
-					"\"file\":\"" + unpathedFile.replace( /\.min\.map/, ".min.js" ) )
+					"\"file\":\"" + unpathedFile.replace( /\.min\.map/, ".min.js\"" ) )
 				.replace( /"sources":\["([^"]+)"\]/,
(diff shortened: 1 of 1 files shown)


mgol:
Why can't we just read the file, pass through `JSON.parse`, modify & save again and escape all those regexes? Is that so slow?

dmethvin:
TBH I didn't think of that approach, this was already in place so I just changed it. The `JSON.parse` is arguably better I guess.

PR #3322, opened 2016-09-19
Update event.js

earforth:
take easy args

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -33,7 +33,49 @@ function safeActiveElement() {
 	} catch ( err ) { }
 }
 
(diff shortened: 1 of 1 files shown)


gibson042:
Thanks for the pull request, but this is some rather obscure code, some parts of which are never invoked, and which doesn't provide a clear value. Our recommended approach is to start with an issue, where discussion of merits and shortcomings of proposals can be worked out, and if accepted a PR can be worked on and submitted. If this is addressing a problem, please open an issue and we'll evaluate it in its own discussion.

Issue #3328, opened 2016-09-24
when to promote a ajax to jsonp request


pslean:
$.ajax({
	type: 'put',
	url: '/test',
	data: JSON.stringify({
		desc: "ls??",
		id: 698,
		text: "flow8"
	}),
	headers: {
		"Content-Type": 'application/json',
		"User-Id": 4,
		"AuthDate": 1474730987738,
		"Authorization": "BCDAE 7ef63480e6e39d91011a:Ms8LZlGJq1/OS7Is1Swg79plppg="
	},
	dataType: 'json'
}).done(function(result) {
	console.log(result);
});
Like above, when I setting the "Content-Type": "application/json" in the headers, but not setting contentType to "application/json", the ajax is promoted to jsonp request.

timmywil:
Thanks for opening an issue! However, before we can accept the PR, we need a test case that demonstrates the problem.

pslean:
@timmywil What info do you need to be provided?

timmywil:
@pslean We need a way to reproduce the problem in live code with as little code as possible. This is usually done with jsbin.com or jsfiddle.net. If you prefer to write the test in jQuery's ajax tests, that would also work.

timmywil:
Will reopen when we have a test case. Thanks!

pslean:
@timmywil Sorry, I have been very busy recently. The example you need is as below.
1. This is the problem, you can see the error in the consol of chrome. The request is promoted to jsonp, but it is not. Link: [js bin](http://jsbin.com/mejarajuro/edit?html,js,console)
2. Then when you add (contentType: 'application/json') to the ajax option, you can remove the annotation in the js bin, the you can get response data from the post api. 
Please help me reopen the issue and verify it.

mgol:
@pslean The URL you're using there is in your private network so we can't see the result of this fiddle. Use https://httpbin.org/ to mock backend requests for test purposes.

pslean:
@mgol I have changed the js bin link to the newest. Please check.

pslean:
@timmywil @mgol [http://jsbin.com/mohubalozu/edit?html,js,console](http://jsbin.com/mohubalozu/edit?html,js,console)

pslean:
@mgol @timmywil help me.

PR #3329, opened 2016-09-24
Ajax: when to promote a ajax to jsonp request #3328

pslean:
#3328


timmywil:
Thank you for your contribution, but we're not sure what this is addressing.

Issue #3365, opened 2016-10-18
Expose jquery.slim for commonjs/es6 modules.


DylanPiercey:
Currently if you want to import jquery slim you must do.

``` js
import 'jquery/dist/jquery.slim'
```

It would be nice if an alias for this was added to the root of the project so that people could instead use:

``` js
import 'jquery/slim'
```

markelog:
Do you want to tackle it?

PR #3366, opened 2016-10-18
Add slim build alias for module bundlers

DylanPiercey:
This PR is related to https://github.com/jquery/jquery/issues/3365.

Basically this adds an alias file `./slim.js` which makes it easier for es6/commonjs modules to require the slim build.

``` js
import 'jquery/dist/jquery.slim'
```

Becomes

``` js
import 'jquery/slim'
```

### DIFF
--- a/slim.js
+++ b/slim.js
@@ -1,0 +1,2 @@
+// Alias for the slim jquery build for use with bundlers (wepback/browserify).
+module.exports = require('./dist/jquery.slim.js')


dmethvin:
We don't commit build artifacts to our development repo for every commit so that require doesn't point to anything. If you look in the `/dist` folder here you'll see it's empty. I guess this would go into https://github.com/jquery/jquery-dist but am not sure whether it's useful there because this isn't part of my own workflow.

mgol:
I think this would be nice to have but we'd definitely need to generate such a file in our [release script](https://github.com/jquery/jquery/blob/3.1.1/build/release.js).

mgol:
I'm closing this PR since it has to be done differently. If you'd like to make changes in the proper place, please open a new one. Ask if you have any questions!

PR #3367, opened 2016-10-19
Offset: report offset for 0 sized elements

jbedard:
I assume this previously worked (except IE<=11+ which crashed) before 0e4477c676db0427bb9b0bf39df8631501e62f24.

Also previously when `if ( rect.width || rect.height )` failed it would return `rect` which would be the `getBoundingClientRect()` object containing top/bottom/left/right along with the width/height of 0. The top/left in there would probably be quite misleading?

Fixes [gh-3267](https://github.com/jquery/jquery/issues/3167)
- [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -93,6 +93,7 @@ jQuery.fn.extend( {
 			return;
 		}
 
+		// Return zeros for disconnected and hidden (display: none) elements (gh-2310)
 		// Support: IE <=11 only
 		// Running getBoundingClientRect on a
 		// disconnected node in IE throws an error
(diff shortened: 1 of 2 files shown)


NekR:
So, do we want to report left/top for 0 sized elements? If not, then it can be simpler this:

```js
if ( rect.width || rect.height ) {
   // ...
} else {
  return { left: 0, top: 0 }
}
```

NekR:
Oh, nevermind, 0 sized elements still can provide some visible objects (child elements / text) so definitely should return correct offset.

gibson042:
LGTM. :+1:

mgol:
LGTM

markelog:
We can now write `LGTM` with github buttons now :)

PR #3374, opened 2016-10-24
Add CDNJS version badge in readme

x09326:
This will add the badge to show its version on CDNJS and also link to its page on CDNJS!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 [jQuery](https://jquery.com/) — New Wave JavaScript
 ==================================================
-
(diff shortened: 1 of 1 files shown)


mgol:
We don't have badges for anything else in README, if we want to link to concrete CDNs, we should do it for all of them and not favor any specific ones.

dmethvin:
The CDNJS CDN is already listed with the others on our download page, http://jquery.com/download/ , so I think we're covered. 

The README here at github covers the needs of people who want to do custom builds or contribute to development. I suppose there could be some sentence added near the top that says, *"If you need a copy of jQuery, any version, go to http://jquery.com/download/."* We haven't had a lot of questions about this though so I'm not convinced it's needed.

PR #3377, opened 2016-10-24
IE Memory leak clean

mdrmtz:
### Summary ###
There circular reference memory leak when adding an event at eventHandle  object

Css support module computeStyleTests create a memory leak in a node un-attached to the DOM

### Checklist ###

### DIFF
--- a/src/css/support.js
+++ b/src/css/support.js
@@ -43,6 +43,7 @@ define( [
 		// Nullify the div so it wouldn't be stored in the memory and
 		// it will also be a sign that checks already performed
 		div = null;
+		container = null;
 	}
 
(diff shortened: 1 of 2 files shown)


gibson042:
Thanks for the pull request, but what are you trying to accomplish? The variables in src/event.js aren't accessed by any closure, and we're not worried about holding onto a single DOM element via `computeStyleTests`. Note also that we only support Internet Explorer version 9 and up, so aggressive cleanup is not required.

markelog:
We specifically removed those nullifications, we definitely will need proof for these claims, since those issues was fixed long time ago, as far as we know

timmywil:
Agreed. This shouldn't be an issue in the browsers we support. Please open an issue if there's a test that can demonstrate a memory leak. Thanks!

PR #3385, opened 2016-11-02
Build: ESLint setup improvements

mgol:
### Checklist ###

1. Use the short name of the preset in the config.
2. Run ESLint first on non-minified files.
3. Explicitly specify environments in every config file (those settings cascade
which means we've been assuming a Node.js environment where we shouldn't have).

I found these when working on https://github.com/jquery/jquery-migrate/pull/238.

cc @markelog

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -203,12 +203,21 @@ module.exports = function( grunt ) {
 
 	grunt.registerTask( "lint", [
 		"jsonlint",
-		runIfNewNode( "eslint" )
+
+		// Running the full eslint task without breaking it down to targets
+		// would run the dist target first which would point to errors in the built
+		// file, making it harder to fix them. We want to check the built file only
+		// if we already know the source files pass the linter.
+		runIfNewNode( "eslint:dev" ),
+		runIfNewNode( "eslint:dist" )
 	] );
 
 	grunt.registerTask( "lint:newer", [
 		"newer:jsonlint",
-		runIfNewNode( "newer:eslint" )
+
+		// Don't replace it with just the task; see the above comment.
+		runIfNewNode( "newer:eslint:dev" ),
+		runIfNewNode( "newer:eslint:dist" )
 	] );
 
 	grunt.registerTask( "test:fast", runIfNewNode( "node_smoke_tests" ) );
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -1,18 +1,5 @@
 {
-	// Support: IE <=9 only, Android <=4.0 only
-	// The above browsers are failing a lot of tests in the ES5
-	// test suite at http://test262.ecmascript.org.
-	"parserOptions": {
-		"ecmaVersion": 3
-	},
-	"globals": {
-		"window": true,
-		"jQuery": true,
-		"define": true,
-		"module": true,
-		"noGlobal": true
-	},
-	"rules": {
-		"strict": ["error", "function"]
-	}
+	"root": true,
+
+	"extends": "../.eslintrc-browser.json"
 }
--- a/test/.eslintrc.json
+++ b/test/.eslintrc.json
@@ -1,13 +1,15 @@
 {
+	"root": true,
+
+	"extends": "../.eslintrc-browser.json",
+
 	"env": {
+
+		// In source the browser env is not enabled but unit tests rely on them
+		// too much and we don't run them in non-browser environments anyway.
 		"browser": true
 	},
-	// Support: IE <=9 only, Android <=4.0 only
-	// The above browsers are failing a lot of tests in the ES5
-	// test suite at http://test262.ecmascript.org.
-	"parserOptions": {
-		"ecmaVersion": 3
-	},
+
 	"globals": {
 		"require": false,
 		"define": false,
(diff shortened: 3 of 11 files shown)


markelog:
You might as well rephrase https://github.com/jquery/jquery/blob/5b4cb0d33731a58384e02ad51e703e7dcb00e424/Gruntfile.js#L113 it seems @sindresorhus didn't restore those issues – https://twitter.com/arkel/status/760518498699411457

mgol:
Yeah. Maybe a little description could be useful since you can't view the ticket.

mgol:
And good catch about the JSCS mention!

sindresorhus:
> You might as well rephrase https://github.com/jquery/jquery/blob/5b4cb0d33731a58384e02ad51e703e7dcb00e424/Gruntfile.js#L113 it seems @sindresorhus didn't restore those issues – https://twitter.com/arkel/status/760518498699411457

I forgot. Done now.

markelog:
@sindresorhus cool, thank you

mgol:
@markelog I deleted the line but it seems to not exactly fit into the first commit, I added it as a separate one. What do you think?

src/.eslintrc.json
```diff
@@ -1,4 +1,8 @@
 {
+	"env": {
+		"browser": false,
```
mgol: What do you mean by "these files only"?

The configs cascade down in ESLint. If you don't specify `browser: false`, you depend on the parent config to not specify `browser: true`. Since we have to reset the `node` env, I wanted to specify the browser one explicitly, it seems less fragile.

markelog:
> @markelog I deleted the line but it seems to not exactly fit into the first commit, I added it as a separate one. What do you think?

I find you approach marvellous :)

src/.eslintrc.json
mgol: Ah, I get it now. I'll update the others as well. I'd only leave the ones that explicitly extend `src/.eslintrc.json` as there we do want to replicate the same settings for source files.

Gruntfile.js
```diff
@@ -203,12 +203,14 @@ module.exports = function( grunt ) {
...
 	grunt.registerTask( "lint", [
 		"jsonlint",
-		runIfNewNode( "eslint" )
+		runIfNewNode( "eslint:dev" ),
+		runIfNewNode( "eslint:dist" )
 	] );
 
 	grunt.registerTask( "lint:newer", [
 		"newer:jsonlint",
-		runIfNewNode( "newer:eslint" )
```
markelog: Same idea as in https://github.com/jquery/jquery-migrate/pull/238/files#diff-35b4a816e0441e6a375cd925af50752cR129 I suppose? 

I wouldn't be oppose to comment about this in the code

mgol: Yeah, commenting that makes sense, otherwise someone will remove it eventually. ;)

mgol:
@markelog PR updated, PTAL.

test/.eslintrc.json
```diff
@@ -1,4 +1,6 @@
 {
+	"extends": "../src/.eslintrc.json",
+	"root": true,
```
markelog: mmm, so this is root, so it wouldn't inherit upper configs, only extends one from `src` and `src` one doesn't have `root` so it would inherit upper ones, which would be the same uppers since `src` is sibling of `test` dir, correct?

mgol: Yeah, you're right. :D How about we create `.eslintrc-(base|browser|node).json` files and extend whichever we need? Similarly to what I did for Angular; see https://github.com/angular/angular.js.

markelog: Hm, how about we extend only `eslint-jquery-config` with `root: true` in every config? It will cause some repetition, but it could make it very clear. What do you think?

mgol: We'd have to duplicate the globals in every `src`-related `.eslintrc.json`. What's the drawback of what I proposed?

markelog: Yeah, I guess duplicating of globals would suck more.

>  What's the drawback of what I proposed?

Angular historically didn't shy away from those in their repo so it's seems fine for them, but for me personally that seems as a clutter

mgol: Didn't shy from what exactly?

Is my proposal of the `.eslintrc-(base|browser|node).json` files OK for you?

markelog: From dot files

> Is my proposal of the .eslintrc-(base|browser|node).json files OK for you?

If that's the best option we can come up with...

mgol:
@markelog PR updated, PTAL.

test/.eslintrc.json
```diff
@@ -1,13 +1,12 @@
 {
+	"root": true,
+
+	"extends": "../.eslintrc-browser.json",
+
 	"env": {
 		"browser": true
```
mgol: I'm not sure how to best play it out. The issue is that we don't want `browser: true` to be specified in source as we take all the browser stuff from the `window` variable and that's important for jsdom compatibility. On the other hand, it would be too much to fix in tests and that would only help us if we ever tried to run the test suite (or its parts) in jsdom which we don't have plans currently. So setup between `src` & `test` will differ.

What would you suggest?

mgol: (Although it seems `.eslintrc-browser.json` should have a comment explaining that)

markelog: So we always want to reference global vars from `window` object? Yeah I recall we talked about this before, sounds good to me then – a default disabled `browser` option and additional comment

mgol: > So we always want to reference global vars from `window` object?

I'm sure you meant it :) but just to be precise - only the browser-only globals; the one from the ECMAScript specification are used directly.

I'll update the PR.

Issue #3408, opened 2016-11-24
ajax post with contenttype="application/json" need process data to json string


Litor:
in jquery ajax, data will transfered like a=1&b=2,  when post ajax with contentType equal application/json, data want transfered like  {"a":1, "b":2}, get ajax still use a=1&b=2

system is huge, we want process it in one place, but there is no api do this.

timmywil:
Closing in favor of https://github.com/jquery/jquery/issues/3438

PR #3409, opened 2016-11-24
whatever the s.processData is, if s.data exist, jquery.param should execute if the data is not string for get request

Litor:
### Summary ###

GET request with data and the processData is false
```
$.ajax({
url: 'test.php',
type:'GET',
processData:false,
data:{a:1,b:2}
})
```

the request url will like test.php?[Object, Object]

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -557,8 +557,14 @@ jQuery.extend( {
 			}
 		}
 
+		// Uppercase the type
+		s.type = s.type.toUpperCase();
+
+		// Determine if request has content
+		s.hasContent = !rnoContent.test( s.type );
+
 		// Convert data if not already a string
-		if ( s.data && s.processData && typeof s.data !== "string" ) {
+		if ( s.data && ( s.processData || !s.hasContent ) && typeof s.data !== "string" ) {
 			s.data = jQuery.param( s.data, s.traditional );
 		}
 
@@ -578,12 +584,6 @@ jQuery.extend( {
 		if ( fireGlobals && jQuery.active++ === 0 ) {
 			jQuery.event.trigger( "ajaxStart" );
 		}
-
-		// Uppercase the type
-		s.type = s.type.toUpperCase();
-
-		// Determine if request has content
-		s.hasContent = !rnoContent.test( s.type );
 
 		// Save the URL in case we're toying with the If-Modified-Since
 		// and/or If-None-Match header later on


Litor:
for #3408, this question can resolved if Transfer third argument s in ajax method jQuery.param(), with this third argument we can rewrite jQuery.param, when get request, data transfered like a=1&b=2, when post request transfered like {"a":1, "b":2}

src/ajax.js
```diff
@@ -559,7 +559,7 @@ jQuery.extend( {
 
 		// Convert data if not already a string
 		if ( s.data && s.processData && typeof s.data !== "string" ) {
-			s.data = jQuery.param( s.data, s.traditional );
+			s.data = jQuery.param( s.data, s.traditional, s );
```
dmethvin: That is because `jQuery.param()` should not need to know that it is being called by `jQuery.ajax()`, it can be called by any code that wants to serialize a parameter using HTML form rules. This change is using `jQuery.param()` as a hook to an AJAX call for a totally unrelated thing. 

Can you do this by setting `processData: false` and using a `beforeSend` callback to format the data?

Litor: when setting processData: false, data is removed before beforeSend callback

src/ajax.js
```diff
@@ -557,8 +557,14 @@ jQuery.extend( {
...
 
+		// Uppercase the type
+		s.type = s.type.toUpperCase();
+
+		// Determine if request has content
+		s.hasContent = !rnoContent.test( s.type );
+
 		// Convert data if not already a string
-		if ( s.data && s.processData && typeof s.data !== "string" ) {
+		if ( s.data && ( s.processData || !s.hasContent ) && typeof s.data !== "string" ) {
```
gibson042: The default value of `processData` is true, so this change seems to be circumventing explicit user intention and I'm therefore opposed. It would also need unit tests before landing, but I'd first focus on the rationale. How did you encounter an issue in the first place?

Litor: here is a sample for this problem: https://jsfiddle.net/1n5wo0bs/
in chrome debug "Network", you will find the request url is 
  https://fiddle.jshell.net/_display/dss.do?[object%20Object]

Litor: I want process data manualy, what can I do

Litor: in jquery ajax, data will transfered like a=1&b=2, when post ajax with contentType equal application/json, data want transfered like {"a":1, "b":2}, get ajax still use a=1&b=2

system is huge, we want process it in one place, but there is no api do this.

gibson042: Let me restate: if you want `jQuery.ajax` to produce a string, set `processData: true` (the default). If you want to _prevent_ `jQuery.ajax` from altering your data, set `processData: false`. Why are you setting `processData: false` when you want the `processData: true` behavior?

dmethvin: @Litor I think I understand the problem you're running into now. With HTTP methods that do not have a request body such as GET, any provided `data` is [always processed](https://github.com/jquery/jquery/blob/4e50967725e61a5d71c2792131de61dec0520f9c/src/ajax.js#L594) even if the caller specified `processData: false`.  This seems contrary to the documentation which says:

> The data option can contain either a query string of the form key1=value1&key2=value2, or an object of the form {key1: 'value1', key2: 'value2'}. If the latter form is used, the data is converted into a query string using jQuery.param() before it is sent. This processing can be circumvented by setting processData to false.

This behavior was also present before 3.x so it's not any kind of recent regression, but it does seem like a bug. @gibson042 do you agree? If we didn't process the data it would be possible for a `beforeSend` handler to take an object and stringify it to the URL as desired like: http://jsbin.com/wixuquzeri/edit?html,console

dmethvin:
I'm going to close this because I think it's the wrong solution, but see gh-3438.

PR #3414, opened 2016-11-28
Build: jQuery Foundation -> JS Foundation

timmywil:
- I think the contribute link only relates to projects within the jQuery organization (this is as opposed to linking to contribute.jquery.org and saying all JS Foundation projects).
- I want to keep the jQuery license link rather than linking to js.foundation, because licenses are per project and not global to the foundation.


timmywil:
@kborchers

kborchers:
Looks great to me. Also agree with the foundation vs org changes. We need to make that clearer on jquery.org and clean that up too but just haven't had time to even clearly lay out an issue for someone else to do it.

Thanks!

PR #3417, opened 2016-12-01
Remove window object references and fix whitespace

david-mark:
### Checklist ###

Always declare global variables.

https://gist.github.com/david-mark/7cf93afccf3696a4f6754b36cd6ab34e

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -8,7 +8,6 @@
 	"globals": {
 		"window": true,
 		"jQuery": true,
-		"jQueryDefined": true,
 		"define": true,
 		"module": true,
 		"noGlobal": true
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -8,6 +8,7 @@
 	"globals": {
 		"window": true,
 		"jQuery": true,
+		"jQueryDefined": true,
 		"define": true,
 		"module": true,
 		"noGlobal": true
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -7,11 +7,11 @@ define( [
         "use strict";
 
         // Map over jQuery in case of overwrite
-	
+
         var _jQuery = jQuery;
 
 	// Map over the $ in case of overwrite
-        
+
 	var _$ = $;
 
 	jQueryDefined.noConflict = function( deep ) {
@@ -29,7 +29,7 @@ define( [
 	// Expose jQuery and $ identifiers, even in AMD
 	// (#7102#comment:10, https://github.com/jquery/jquery/pull/557)
 	// and CommonJS for browser emulators (#13566)
-        
+
 	if ( !noGlobal ) {
 		jQuery = $ = jQueryDefined;
 	}
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -1,34 +1,36 @@
+var $, jQuery;
+
 define( [
 	"../core"
-], function( jQuery, noGlobal ) {
+], function( jQueryDefined, noGlobal ) {
 
-"use strict";
+        "use strict";
 
-var
-
-	// Map over jQuery in case of overwrite
-	_jQuery = window.jQuery,
+        // Map over jQuery in case of overwrite
+	
+        var _jQuery = jQuery;
 
 	// Map over the $ in case of overwrite
-	_$ = window.$;
+        
+	var _$ = $;
 
-jQuery.noConflict = function( deep ) {
-	if ( window.$ === jQuery ) {
-		window.$ = _$;
+	jQueryDefined.noConflict = function( deep ) {
+		if ( $ === jQueryDefined ) {
+			$ = _$;
+		}
+
+		if ( deep && jQuery === jQueryDefined ) {
+			jQuery = _jQuery;
+		}
+
+		return jQueryDefined;
+	};
+
+	// Expose jQuery and $ identifiers, even in AMD
+	// (#7102#comment:10, https://github.com/jquery/jquery/pull/557)
+	// and CommonJS for browser emulators (#13566)
+        
+	if ( !noGlobal ) {
+		jQuery = $ = jQueryDefined;
 	}
-
-	if ( deep && window.jQuery === jQuery ) {
-		window.jQuery = _jQuery;
-	}
-
-	return jQuery;
-};
-
-// Expose jQuery and $ identifiers, even in AMD
-// (#7102#comment:10, https://github.com/jquery/jquery/pull/557)
-// and CommonJS for browser emulators (#13566)
-if ( !noGlobal ) {
-	window.jQuery = window.$ = jQuery;
-}
-
 } );
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -1,7 +1,7 @@
 define( [
 	"../core"
 ], function( jQuery, noGlobal ) {
-	
+
         // Map over jQuery in case of overwrite
 
         var _jQuery = this.jQuery;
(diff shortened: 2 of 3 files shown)


david-mark:
Integration tests indicate that some other file needs a variable name change. Certainly "jQueryDefined" is defined in the file I changed. Also added it to "globals" in the lint JSON, which makes little sense to me as it isn't a global at all (nor is "noGlobal", which is also in there). Leave that to you.

timmywil:
Thank you for your contribution. However, we'd like to keep this the way it is. It has seen many revisions to get to this point (and has withstood the test of time). I see no *practical* reason to change it, just a blanketing ideological argument. Notice that the `window` variable used in a built jQuery is intentionally a shadowed var, which enables jQuery's use in Node when a window is mocked with something like jsdom (see https://github.com/jquery/jquery/blob/master/src/wrapper.js#L32 for what happens if you load jQuery in Node without a window). The window passed to jQuery is not always the same as the global object.

david-mark:
I think you missed this bit:

https://github.com/jquery/jquery/pull/3417/commits/7ac889e79a4ccb9ff82d78bfc95194e28501c736

Nothing ideological about that.

> Notice that the window variable used in a built jQuery is intentionally a shadowed var, which
> enables jQuery's use in Node when a window is mocked with something like jsdom

That has nothing to do with my changes. You are simply confusing the `window` host object (and a surrogate for NodeJS) with the global object.

> (see https://github.com/jquery/jquery/blob/master/src/wrapper.js#L32 for what happens if
> you load jQuery in Node without a window).

Also has nothing to do with it. Did nothing to prevent the use of `window`. Also, in Node the global object is named `global` and also has nothing to do with the DOM. That's a piece you are missing by conflating `window` and the global object

Good luck! :)

gibson042:
@david-mark Take another look at src/wrapper.js , and at uses of `window` throughout the codebase. We are not confusing `window` with the global object, we are explicitly depending upon a DOM-like window, complete with properties like `document`, `location`, `addRemoveEventListener`, `getComputedStyle`, `requestAnimationFrame`, and a self-referential `window`.

david-mark:
> Take another look at src/wrapper.js

I already saw it and just adjusted the cited line in my fork to support NodeJS, where the global object is named "global". As mentioned we were missing that bit as well. One more problem solved, but it had nothing to do with the rest of your post:

> and at uses of window throughout the codebase. We are not confusing window with the
> global object, we are explicitly depending upon a DOM-like window, complete with
> properties like document, location, addRemoveEventListener, getComputedStyle, 
> requestAnimationFrame, and a self-referential window."

You are indeed confused.

You reference the *global* `window` object all over the built jQuery script. In browsers, it's property of the global object, so is in the global scope. As is `document`, but for some reason you reference it as `window.document` (but that's entirely beside the point). And yes, the `window` object usually behaves like the global object, so it has a `window` property, as well as a `window.window` property and a `window.window.window` property, etc. Have no idea what you mean by "explicitly depending on a DOM-like window".

As for whatever object gets passed to your factory when running on NodeJS, I can't see how that has anything to do with any of my changes, certainly not those to the wrapper.js file.

Per the comments, this is the pertinent part of the code:

```
                // For CommonJS and CommonJS-like environments where a proper `window`
		// is present, execute the factory and get jQuery.
		// For environments that do not have a `window` with a `document`
		// (such as Node.js), expose a factory as module.exports.
		// This accentuates the need for the creation of a real `window`.
		// e.g. var jQuery = require("jquery")(window);
		// See ticket #14549 for more info.
		module.exports = global.document ?
			factory( global, true ) :
			function( w ) {
				if ( !w.document ) {
					throw new Error( "jQuery requires a window with a document" );
				}
				return factory( w );
			};
```

And the previously cited line that throws the exception is *inside* your exports function. Haven't made any changes that affect that as it has its own `w` argument for NodeJS users to pass whatever object they want.

So what are you trying to say? Which line do you think will fail here and how do you figure my proposed changes will make it fail. Suggest removing the strict mode check from this file, allowing the integration tests to finish and then trying out the changes. Will likely save time.

Best of luck! :)

david-mark:
Here is the latest pull request (now closed due to above confusion).

https://github.com/jquery/jquery/pull/3421

...As mentioned, the last change may not even be necessary, but certainly makes sense for NodeJS in case there is a way to provide global `window` and `document` objects without using a factory. As it stands, per the comments, the factory will always be used for NodeJS, so it really doesn't matter at this time.

gibson042:
> You reference the *global* `window` object all over the built jQuery script. In browsers, that's a property of the global object, so is in the global scope. As is `document`, but for some reason you reference that as `window.document` (but that's entirely beside the point).

No, we reference the _local_ `window` that is the first parameter of the anonymous function referenced as `factory` in wrapper.js (the function that literally _wraps_ our entire library).

> Have no idea what you mean by "explicitly depending on a DOM-like window".

To repeat myself, "complete with properties like `document`, `location`, `addRemoveEventListener`, `getComputedStyle`, `requestAnimationFrame`, and a self-referential `window`". Or put another way, implementing the interfaces in the [DOM Standard](https://dom.spec.whatwg.org/), [HTML Standard](https://html.spec.whatwg.org/), etc.—specifically [`Window`](https://html.spec.whatwg.org/multipage/browsers.html#window) itself.

> And the previously cited line that throws the exception is *inside* your exports function.

Previously cited? Did I miss something?

> So what are you trying to say? Which line do you think will fail here and how do you figure my proposed changes will make it fail.

The current code prioritizes `window` over the global object, so your change will hand some environments a factory where they currently receive the jQuery constructor. And your introduction of `this` references into the exports function will similarly fail to add `jQuery` and `$` properties to the window object passed to the factory. But maybe I'm misunderstanding what you were going for... there's no issue reference and no unit tests.

> PS. You really shouldn't use `window.getComputedStyle`. The proper reference is `document.defaultView.getComputedStyle`.

https://github.com/jquery/jquery/commit/e488d985cfb10ab8c684bbc6a9b8ff3eae23bf83

david-mark:
> No, we reference the local window that is the first parameter of the anonymous function
> referenced as factory in wrapper.js (the function that literally wraps our entire library).

Which is passed to the anonymous function. I had just edited my comment to make that more clear. It's still a reference to the global window object, at least in browsers./

> To repeat myself, "complete with properties like document, location,
> addRemoveEventListener, getComputedStyle, requestAnimationFrame, and a self-referential > window". Or put another way, implementing the interfaces in the DOM Standard, HTML
> Standard, etc.—specifically Window itself.

Repeating it doesn't make it any more lucid. Re-read what I posted, particularly after the last edit. You do realize that *both* `window` and `document` are globals in browsers? And I gave you some additional tips on how to deal with other environments.

> Previously cited? Did I miss something?

Apparently. Cited by "timmywil" above.

> The current code prioritizes window over the global object, so your change will hand some
> environments a factory where they currently receive the jQuery constructor.

There's just no need to prioritize anything. Re-read the suggestions I made in the last edit. In short, you start by passing a reference to the global object (as an argument that has always been called "global"). From there you can find both `window` and `document`. That's clear, whereas passing either `window` or `this` to the anonymous function is not.

> And your introduction of this references into the exports function will similarly fail to add
> jQuery and $ properties to the window object passed to the factory. But maybe I'm
> misunderstanding what you were going for... there's no issue reference and no unit tests.

This is another misunderstanding. You should not be adding `jQuery` and `$` to the `window` object at all, but to the global object. If you expect them to be on the `window` object elsewhere in the project, that is just another bit to untangle. Eventually it will all make sense... or not. Am growing tired a bit tired of this. :(

Again, good luck to you!

david-mark:
New pull request:

https://github.com/jquery/jquery/pull/3422

Read carefully before closing and perhaps you won't close this one. I agree that there was some fine-tuning required to be perfect. Hadn't claimed perfection in the first place, but insist that the end result of my efforts is *much* clearer code. Will add some comments next to be perfectly clear about what it is doing and why.

PR #3420, opened 2016-12-02
Build: Include the source code in the source map.

blainesch:
### Summary ###

This adds the source code to the source map, so it is more useful for developers.

### Checklist ###

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -166,6 +166,7 @@ module.exports = function( grunt ) {
 				},
 				options: {
 					preserveComments: false,
+					sourceMapIncludeSources: true,
 					sourceMap: true,
 					ASCIIOnly: true,
 					sourceMapName:


dmethvin:
@blainesch  How much bigger is the map with this option enabled?  Obvs it won't be loaded by most users but I just wonder about the size.

blainesch:
@dmethvin it went from 129k to 424k.

mgol:
Why is this more useful for web developers?

blainesch:
@mgol I find it difficult to imagine a source map being useful at all without the source.

mgol:
@blainesch The source map doesn't have to _include_ the source, it may just reference it and our map file does this, just look at the `sources` property in the map file; it contains reference to the unminified `jquery.js`.

What problems would inlining the source solve for you?

markelog:
There is no response from ticket owner for over the week now, so i'm closing this. Please let us know if you want to continue this discussion

PR #3422, opened 2016-12-02
Straighten out window, document and global object references

david-mark:
### Checklist ###

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -8,7 +8,6 @@
 	"globals": {
 		"window": true,
 		"jQuery": true,
-		"jQueryDefined": true,
 		"define": true,
 		"module": true,
 		"noGlobal": true
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -8,6 +8,7 @@
 	"globals": {
 		"window": true,
 		"jQuery": true,
+		"jQueryDefined": true,
 		"define": true,
 		"module": true,
 		"noGlobal": true
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -7,11 +7,11 @@ define( [
         "use strict";
 
         // Map over jQuery in case of overwrite
-	
+
         var _jQuery = jQuery;
 
 	// Map over the $ in case of overwrite
-        
+
 	var _$ = $;
 
 	jQueryDefined.noConflict = function( deep ) {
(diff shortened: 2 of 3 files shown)


gibson042:
Please stop submitting these noisy-diff pull requests lacking both explanation and tests. I will provisionally leave this one open to provide an opportunity for correcting the above flaws, but I really believe that the effort needs to start with an issue and advance to a pull request only once everyone is on the same page about the nature of a perceived problem.

timmywil:
Agreed. Thank you for trying to help, but I'm not convinced. I know that can be frustrating on your end, but the first step should be discussion of the problem in an issue. When we all agree there's a problem, we can work on a solution.

david-mark:
> Please stop submitting these noisy-diff pull requests lacking both explanation and tests

I haven't given you enough explanation on the original? If you guys would stop *closing* them then I wouldn't have to resubmit. Furthermore, your *original* unit tests are all that are needed to test these updates as the behavior shouldn't change.

> Agreed. Thank you for trying to help, but I'm not convinced.

The problem is clear, your factory only accepts a window object, but you need both the window and the global object to support all of your various platforms as they are not equivalent on all of the platforms (e.g. when using domjs). For example, why would you augment the domjs `window` object with $ and jQuery? It makes no sense.

All comes down to that. Trying to treat window and global as if they are the same, which is often true in browsers, but clearly not in NodeJS.

gibson042:
> For example, why would you augment the domjs window object with $ and jQuery?

So scripts running within that environment can access the function.

david-mark:
The big problem here is that you only pass `window` and it is sometimes equivalent to the global object and sometimes not. Makes for very confusing code, as was alluded to previously as "many revisions" were needed to finally get it all to "work" in browsers, CommonJS and NodeJS with something called jsdom. In that last case, the script using jsdom *passes* some fake `window` object and jQuery is *returned* (so it can obviously access it at that point). Adding it to the fake `window` object is clearly pointless; however, adding it to the global object would seem beneficial in all environments.

Furthermore, you pass either `window` or the NodeJS module (`this`) at one point as well. The whole thing is tangled up in a hard-to-follow mess and could be made much simpler by keeping the the global object and `window` *separate*.

The last pull request keeps the behavior the same (i.e. still augments that fake `window` object that is passed from jsdom) while untangling and simplifying as much as can be done at this time. And it only passes the global object (not the NodeJS module) to the initial function. Best I can do for now.

Issue #3426, opened 2016-12-02
Confusing code, global object vs. DOM window vs. fake DOM window


david-mark:
### Description ###

Starting at the beginning, what is this code from wrapper.js supposed to do?

```javascript
// Pass this if window is not defined yet
} )( typeof window !== "undefined" ? window : this, function( window, noGlobal ) {
```

In browsers it will pass a reference to the global `window` object, which is a host object that is often the same (or at least closely mimics) the global object. In NodeJS, a reference to a module object will be passed. And in any other ECMAScript environment that lacks a global `window` object, the global object will be passed.

Also, not sure what the "yet" in the above comment is supposed to mean. Moving on to the start of the function:

```javascript
( function( global, factory ) {

	"use strict";
```

Note that the first function argument is called `global`, which is quite confusing as it may well reference the `window` host object or a NodeJS module. In other environments it will reference the global object.

The code continues:

```javascript
if ( typeof module === "object" && typeof module.exports === "object" ) {

		// For CommonJS and CommonJS-like environments where a proper `window`
		// is present, execute the factory and get jQuery.
		// For environments that do not have a `window` with a `document`
		// (such as Node.js), expose a factory as module.exports.
		// This accentuates the need for the creation of a real `window`.
		// e.g. var jQuery = require("jquery")(window);
		// See ticket #14549 for more info.
		module.exports = global.document ?
			factory( global, true ) :
			function( w ) {
				if ( !w.document ) {
					throw new Error( "jQuery requires a window with a document" );
				}
				return factory( w );
			};
	} else {
		factory( global );
	}
```

This line:

```javascript
module.exports = global.document ?
```

Checks for the global `document` object (as in browsers), which implies that `global` is a reference to the global object; however, we've seen that `global` may reference one of three types of objects: the `window` object in browsers, some sort of module object in NodeJS (which will surely not have a `document` property) or the global object in some unspecified environments. The comments are of little help as they imply that `global` references a `window` object.

It's all quite confusing and there is an inconsistency at the end:

```javascript
        else {
		factory( global );
	}
```

All we know at this point is that there is no `module.exports` object. What we *don't* know is whether this `global` object has a `document` object. If it does not, then jQuery will fail as soon as it references `window.document` (the argument name is `window` in the core, which adds additional confusion).

I've cleared most of this up in these three files:

https://github.com/david-mark/jquery/blob/master/src/wrapper.js
https://github.com/david-mark/jquery/blob/master/src/exports/global.js
[...]

PR #3428, opened 2016-12-03
Simplify global object and window-related logic

david-mark:
### Summary ###

Simplify global object and window-related logic and add ability to pass the global object to the exported factory function in non-browser environments.

https://github.com/jquery/jquery/issues/3426

### Checklist ###

* [N/A] New tests have been added to show the fix or feature works
* [Not yet, want to see integration results] Grunt build and unit tests pass locally with these changes

### DIFF
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -1,6 +1,6 @@
 define( [
 	"../core"
-], function( jQuery, noGlobal ) {
+], function( jQuery, noGlobal, globalObject ) {
 
         "use strict";
 
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -7,11 +7,11 @@ define( [
         "use strict";
 
         // Map over jQuery in case of overwrite
-	
+
         var _jQuery = jQuery;
 
 	// Map over the $ in case of overwrite
-        
+
 	var _$ = $;
 
 	jQueryDefined.noConflict = function( deep ) {
@@ -29,7 +29,7 @@ define( [
 	// Expose jQuery and $ identifiers, even in AMD
 	// (#7102#comment:10, https://github.com/jquery/jquery/pull/557)
 	// and CommonJS for browser emulators (#13566)
-        
+
 	if ( !noGlobal ) {
 		jQuery = $ = jQueryDefined;
 	}
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -1,34 +1,36 @@
+var $, jQuery;
+
 define( [
 	"../core"
-], function( jQuery, noGlobal ) {
+], function( jQueryDefined, noGlobal ) {
 
-"use strict";
+        "use strict";
 
-var
-
-	// Map over jQuery in case of overwrite
-	_jQuery = window.jQuery,
+        // Map over jQuery in case of overwrite
+	
+        var _jQuery = jQuery;
 
 	// Map over the $ in case of overwrite
-	_$ = window.$;
+        
+	var _$ = $;
 
-jQuery.noConflict = function( deep ) {
-	if ( window.$ === jQuery ) {
-		window.$ = _$;
+	jQueryDefined.noConflict = function( deep ) {
+		if ( $ === jQueryDefined ) {
+			$ = _$;
+		}
+
+		if ( deep && jQuery === jQueryDefined ) {
+			jQuery = _jQuery;
+		}
+
+		return jQueryDefined;
+	};
+
+	// Expose jQuery and $ identifiers, even in AMD
+	// (#7102#comment:10, https://github.com/jquery/jquery/pull/557)
+	// and CommonJS for browser emulators (#13566)
+        
+	if ( !noGlobal ) {
+		jQuery = $ = jQueryDefined;
 	}
-
-	if ( deep && window.jQuery === jQuery ) {
-		window.jQuery = _jQuery;
-	}
-
-	return jQuery;
-};
-
-// Expose jQuery and $ identifiers, even in AMD
-// (#7102#comment:10, https://github.com/jquery/jquery/pull/557)
-// and CommonJS for browser emulators (#13566)
-if ( !noGlobal ) {
-	window.jQuery = window.$ = jQuery;
-}
-
 } );
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -6,19 +6,19 @@ define( [
 
         // Map over jQuery in case of overwrite
 
-        var _jQuery = window.jQuery;
+        var _jQuery = global.jQuery;
 
 	// Map over the $ in case of overwrite
 
-	var _$ = window.$;
+	var _$ = global.$;
 
 	jQuery.noConflict = function( deep ) {
 		if ( this.$ === jQuery ) {
-			window.$ = _$;
+			global.$ = _$;
 		}
 
-		if ( deep && window.jQuery === jQuery ) {
-			window.jQuery = _jQuery;
+		if ( deep && global.jQuery === jQuery ) {
+			global.jQuery = _jQuery;
 		}
 
 		return jQuery;
@@ -29,6 +29,6 @@ define( [
 	// and CommonJS for browser emulators (#13566)
 
 	if ( !noGlobal ) {
-		window.jQuery = window.$ = jQuery;
+		global.jQuery = global.$ = jQuery;
 	}
 } );
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -1,7 +1,7 @@
 define( [
 	"../core"
 ], function( jQuery, noGlobal ) {
-	
+
         // Map over jQuery in case of overwrite
 
         var _jQuery = this.jQuery;
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -1,6 +1,6 @@
 define( [
 	"../core"
-], function( jQuery, noGlobal, globalObject ) {
+], function( jQuery, noGlobal ) {
 
 "use strict";
 
--- a/src/exports/global.js
+++ b/src/exports/global.js
@@ -2,11 +2,11 @@ define( [
 	"../core"
 ], function( jQuery, noGlobal, globalObject ) {
 
-        "use strict";
+"use strict";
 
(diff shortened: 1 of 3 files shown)


src/exports/global.js
```diff
@@ -1,34 +1,34 @@
 define( [
 	"../core"
-], function( jQuery, noGlobal ) {
+], function( jQuery, noGlobal, globalObject ) {
 
-"use strict";
+        "use strict";
```
mgol: Please don't change the indentation here.

Also, we use tabs for indentation, not spaces, please change that in the PR as I see you're using spaces everywhere.

src/exports/global.js
```diff
@@ -1,34 +1,34 @@
 define( [
 	"../core"
-], function( jQuery, noGlobal ) {
+], function( jQuery, noGlobal, globalObject ) {
```
mgol: This line shouldn't be changed.

src/wrapper.js
```diff
@@ -20,25 +20,53 @@
...
-		// See ticket #14549 for more info.
-		module.exports = global.document ?
-			factory( global, true ) :
-			function( w ) {
-				if ( !w.document ) {
-					throw new Error( "jQuery requires a window with a document" );
+
+		module.exports = ( global.window && global.window.document ) ?
+			factory( global.window, true, global ) :
+			function( global, window ) {
```
mgol: This line is what I was talking about - you require jQuery Node.js users to pass the global object as the first parameter. First of all, that would be a breaking change since we now expect `window` to be the first so - if anything - the order of parameters should be swapped.

But I still don't see the point of passing the `global` to the factory. jQuery itself doesn't need it so if the only reason is to pass the object to which `jQuery` & `$` are attached, that doesn't justify complicating the API. If someone wants those variables to be accessible via the global, they can attach them themselves:

```js
var $ = require( "jquery" );
global.$ = global.jQuery = $;
```

src/exports/global.js
david-mark: In wrapper.js that is. Putting all back to tabs.

src/wrapper.js
```diff
@@ -20,25 +20,53 @@
...
-		// This accentuates the need for the creation of a real `window`.
-		// e.g. var jQuery = require("jquery")(window);
-		// See ticket #14549 for more info.
-		module.exports = global.document ?
-			factory( global, true ) :
-			function( w ) {
-				if ( !w.document ) {
-					throw new Error( "jQuery requires a window with a document" );
+
+		module.exports = ( global.window && global.window.document ) ?
```
mgol: This whole file still has incorrect indentation (a lot of spaces instead of tabs). Also, try to fit within 100 columns per line.

mgol:
This PR adds 46 bytes to our minified gzipped build; that's a lot for a change that doesn't introduce any new functionality (with the exception of passing the global to the factory which I think shouldn't be done).

I'd expect more changes around comments, variable renaming etc. rather than something that increases the size so much.

src/wrapper.js
david-mark: I'll fix it. And the unit tests seemed to fail on lines over 100 columns.

david-mark:
> This PR adds 46 bytes to our minified gzipped build; that's a lot for a change that doesn't
> introduce any new functionality

That's funny. :)

> (with the exception of passing the global to the factory which I think shouldn't be done).

It also clears up a bunch of tangles in the code and adds additional thrown exceptions to fail early (before entering the core). Review the code and you'll see what I mean and why it added the 46 bytes. It's likely all down to the new error string:

```javascript
                if ( !global.window ) {
			throw new Error( "jQuery requires a window" );
                }
```

I know one thing I can do to reduce it though as there is another string duplicated...

src/wrapper.js
david-mark: > This line is what I was talking about - you require jQuery Node.js users to pass the
> global object as the first parameter

No, they may pass either window or global for the first argument. I'll add a comment making that clear. The behavior is the same as before if you pass window. See my explanation of the various scenarios on the issues page. Most scenarios do not use the second argument (only one IIRC).

Went over that on the issue page. It's simply a convenience that goes along with the rest of the untangling. It's certainly more appropriate to have jQuery adding the properties to the global object, rather than to the `window` host object. There's no complication to the API, which remains the same for the caller, unless the pass the global object in NodeJS or use the second argument to pass a fake window in addition to the global object.

david-mark:
I see that one of the Promises related unit tests fails sporadically. Not sure how to kick the tests off again without creating a new commit. Have seen it do that a few times now and then snap out of it on the next commit. Certainly nothing to do with these changes.

david-mark:
There, it all passed again and the only change I made was to add a comment. Something screwy with that unit test (or the tester).

src/exports/global.js
david-mark: I thought as much, but was confused by a previous integration error. Still not entirely sure how all of this fits together (e.g. why is noGlobal in the "globals" section of the lint JSON file?)

src/wrapper.js
```diff
@@ -14,31 +14,64 @@
...
+		if ( !global.window.document ) {
+			throw new Error( errorMessage );
+		}
+
+		factory( global.window, false, global );
 	}
 
-// Pass this if window is not defined yet
-} )( typeof window !== "undefined" ? window : this, function( window, noGlobal ) {
+} )( typeof global == "undefined" ? this : global, function( window, noGlobal, globalObject ) {
```
gibson042: I am strongly opposed to adding a parameter that we have no business using. As I said in https://github.com/jquery/jquery/issues/3426#issuecomment-264664672 , the only reason we care about a global object is its use in wrapper.js for getting a DOM window (which in browsers is an identity mapping).

timmywil:
See https://github.com/jquery/jquery/issues/3426

david-mark:
Here is the microcosm:

> -// Pass this if window is not defined yet
-} )( typeof window !== "undefined" ? window : this, function( window, noGlobal ) {
+} )( typeof global == "undefined" ? this : global, function( window, noGlobal, globalObject ) {

> I am strongly opposed to adding a parameter that we have no business using.

Argument, not parameter. And why are you so opposed to adding it when no user ever sees it? It's certainly not complicating anything for them (or us).

And it's ironic *that* line was quoted as:

> typeof window !== "undefined" ? window : this

Reduces to:

> this

And how much time have we spent on concerns related to the number of bytes?

Come on, guys. :)

It's equivalent but still strange as it passes the `window` host object, which is often the global object in browsers (or mimics it behind the scenes), but can be referenced as a property of the global object in any event (as in the proposed changes) and passes an empty module object in NodeJS. The corresponding argument name in the first function call is "global", which then changes to "window" in a later function call that passes the reference along. Furthermore, the code in the initial function works only by coincidence as the empty module object has no properties at all.

I suggest reviewing my changes to see how I untangled all of that and also managed to pass the global object reference along as well. It really is more appropriate to augment that than the `window` host object. Though I can understand why most would not want to do that in NodeJS (and that continues to happen in my adjusted code by default), it may well not be the case in other non-browser environments.

And we never even addressed the odd hard-coding of `noGlobal` to `true` in some other browser-like environment that features `module.exports`. There's no comment to explain that and it just seems like a nuisance to me (as it was in my tests replicating a browser-like environment in NodeJS by adding `window` as a property of its `global` object). Certainly it's odd and I suspect another in a long line of tacked-on patches to address user complaints without much concern for the bigger picture (as was alluded to in the original canceled pull request I created for this).

This is certainly no way to write a cross-platform script. It's awkward and confusing even considering only browsers and NodeJS (let alone that third option that hard-codes `noGlobal`). Realize that there are many other platforms out there that could use some or all of jQuery. Can only assume that's the primary reason it still has `Sizzle` in it when browsers have long since adopted the Selectors API; that and the non-standard extensions and mutations.

Anyway, it will be another deviation from my fork as there's no way I'm reverting these changes. That's bad news. :(

david-mark:
On careful reflection, I see exactly what to do to untangle this mess and how to do it without disturbing compatibility with the current (and previous) versions.

Three environments to consider:

1. Browsers without CommonJS
2. NodeJS
3. Browsers with CommonJS

At the very top of jQuery.js, we either declare `$` and `jQuery` (my preference) or we do this:

```
this.$ = this.jQuery = undefined;
```

The reason we would go with the latter is so that environment #3 can `delete` them in the wrapper function (otherwise would be left with undefined variables in that environment).

The suggestion to reduce:

```javascript
typeof window !== "undefined" ? window : this
```

...to:

```javascript
this
```

...works out *perfectly* (as I suspected). And there are globals created in NodeJS (as desired).

We can get rid of the `globalObject` argument as we won't need to augment it (or `window`). Rename the `jQuery` function in the core module to... call it `whatever` for now.

Then this:

```javascript
if ( !noGlobal ) {
	window.jQuery = window.$ = jQuery;
}
```

...becomes:

```javascript
if ( !noGlobal ) {
	jQuery = $ = whatever;
}
```

Obviously, keep the bit that hard-codes `noGlobal` to `true` in environment #3. I get it, but it was previously interfering with my scheme to allow the option for NodeJS to work more like browsers.

Should be obvious how the `noConflict` bit will work as well. Not going to bother to paste and modify it here.

And that's that. The only difference, which can be easily dealt with if necessary, is that the fake NodeJS window object would no longer be augmented with `$` and `jQuery`. Node modules would simply put the *returned* constructor wherever they want (just as they can do in a browser when using CommonJS).

If that's not acceptable due to extant Node modules actually referring to the constructor as a property of the fake `window` object (a really bizarre thing to do), then we can simply pass another flag (e.g. `isFakeWindow`) to the core and tack them on (and immediately deprecate that unnecessary and unfortunate practice).

All in all, eliminates a lot of confusion (and the host object abuse in browsers) and should save a few bytes as well. :)

On a final note, as mentioned previously, we should *not* call methods of `whatever` in the core as it defeats the purpose of the module (e.g. outside code could override or delete them). Obviously, we can still use it as a constructor when needed. That will save a bit more space and make the script minify more effectively, as well as saving the overhead of the dot operations.

If we can all agree on this, then re-open the linked issue and I'll create another pull request this weekend.

PR #3429, opened 2016-12-04
Build: Make the @CODE-replacing regex more robust

mgol:
### Summary ###

The code replacing @CODE in wrapper.js was written so that it expected
both the code and the next line to start in the first column. This commit
adjusts the regex so to get rid of that assumption and to work properly
regardless of number of lines with comments after this block.

While this is technically not necessary for our code, contributors sometimes
re-format the wrapper file in their pull requests and the error
messages they get don't tell them what's the real problem with their code.

### Checklist ###

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -17,7 +17,11 @@ module.exports = function( grunt ) {
 		read = function( fileName ) {
 			return grunt.file.read( srcFolder + fileName );
 		},
-		wrapper = read( "wrapper.js" ).split( /\/\/ \@CODE\n\/\/[^\n]+/ ),
+
+		// Catch `// @CODE` and subsequent comment lines event if they don't start
+		// in the first column.
+		wrapper = read( "wrapper.js" ).split( /[\x20\t]*\/\/ @CODE\n(?:[\x20\t]*\/\/[^\n]+\n)*/ ),
+
 		config = {
 			baseUrl: "src",
 			name: "jquery",


build/tasks/build.js
```diff
@@ -17,7 +17,7 @@ module.exports = function( grunt ) {
 		read = function( fileName ) {
 			return grunt.file.read( srcFolder + fileName );
 		},
-		wrapper = read( "wrapper.js" ).split( /\/\/ \@CODE\n\/\/[^\n]+/ ),
+		wrapper = read( "wrapper.js" ).split( /[\x20\t]*\/\/ @CODE\n(?:[\x20\t]*\/\/[^\n]+\n)*/ ),
```
gibson042: I like the generalization, but this now looks sufficiently complex to warrant a comment.

build/tasks/build.js
```diff
@@ -17,7 +17,11 @@ module.exports = function( grunt ) {
 		read = function( fileName ) {
 			return grunt.file.read( srcFolder + fileName );
 		},
-		wrapper = read( "wrapper.js" ).split( /\/\/ \@CODE\n\/\/[^\n]+/ ),
+
+		// Catch `// @CODE` and subsequent comment lines event if they don't start
+		// in the first column.
```
mgol: @gibson042 Is that enough or would you like sth more?

gibson042:
LGTM.

Issue #3442, opened 2016-12-09
empty when resolves with 1 undefined argument in jQuery 3 (was 0 arguments in jQuery 2)


i23098:
### Description ###
I use dynamic array of deferreds to load 0 to N things and then loop through the arguments in the done callback. Upgrading to jQuery 3.1.1 from 2.2.4 and now my code breaks :(

### Link to test case ###
Minimal test case is call $.when with no arguments, so it should resolve with no arguments too...

2.2.4 - https://jsfiddle.net/nuzy6b43/ - This will correctly have 0 arguments in the done callback
3.1.1 - https://jsfiddle.net/4qys5duk/ - This will incorrectly have 1 argument in the done callback

gibson042:
Nice catch. Thanks, @i23098!

PR #3445, opened 2016-12-11
Deferred: Stop inventing jQuery.when() resolution values

gibson042:
Fixes gh-3442

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -13,7 +13,7 @@ function Thrower( ex ) {
 	throw ex;
 }
 
-function adoptValue( value, resolve, reject ) {
+function adoptValue( value, count, resolve, reject ) {
 	var method;
 
 	try {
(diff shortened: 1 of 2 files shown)


src/deferred.js
```diff
@@ -31,7 +31,7 @@ function adoptValue( value, resolve, reject ) {
 
 			// Support: Android 4.0 only
 			// Strict mode functions invoked without .call/.apply get global-object context
-			resolve.call( undefined, value );
+			resolve.apply( undefined, [ value ].slice( noValue ) );
```
mgol: This looks like "clever code", this passing booleans to slice. I guess that's done to reduce size (a regular if-else would be larger) but at least an explanatory comment would be useful.

gibson042: Good point, and correct. Updated.

src/deferred.js
```diff
@@ -41,7 +42,7 @@ function adoptValue( value, resolve, reject ) {
 
 		// Support: Android 4.0 only
 		// Strict mode functions invoked without .call/.apply get global-object context
-		reject.call( undefined, value );
+		reject.apply( undefined, [ value ] );
```
mgol: What's the reason for this change?

gibson042: Consistency with the preceding block, which also effects size reduction (reuse of the `.apply( undefined, [ value ]…` pattern is good for compression).

Issue #3449, opened 2016-12-12
Offset Module: Get Rid of getWindow Function


david-mark:
### Description ###

Working on deprecating `isWindow`, I ran into this internal function in the offset module:

```javascript
/**
 * Gets a window from an element
 */
function getWindow( elem ) {
	return jQuery.isWindow( elem ) ? elem : elem.nodeType === 9 && elem.defaultView;
}
```

The comment is incorrect. It returns a reference to `window` if passed a reference to a `window` or `document` node, but it returns `false` if passed a reference to an element node.

As a side note, it seems very odd to try to derive a `window` reference at all, considering that many methods simply reference the `window` that is passed to the core during initialization.

For example:

```javascript
// Miscellaneous
"target": function( elem ) {
	var hash = window.location && window.location.hash;
	return hash && hash.slice( 1 ) === elem.id;
}
```

Or perhaps that's just a bug. Do see that some methods at least attempt to work with multiple windows. But I digress.

The inappropriately named `getWindow` function is used in two places. As always, if we can reduce that number to one, we can get rid of the function (and the costs associated with calling it).

First one is in the `offset` method:

```javascript
doc = elem.ownerDocument;
win = getWindow( doc );
```

So that obviously reduces to:

```javascript
doc = elem.ownerDocument;
win = doc.defaultView;
```

The other is here:

```javascript
// Create scrollLeft and scrollTop methods
jQuery.each( { scrollLeft: "pageXOffset", scrollTop: "pageYOffset" }, function( method, prop ) {
	var top = "pageYOffset" === prop;

	jQuery.fn[ method ] = function( val ) {
		return access( this, function( elem, method, val ) {
			var win = getWindow( elem );

			if ( val === undefined ) {
				return win ? win[ prop ] : elem[ method ];
			}

			if ( win ) {
				win.scrollTo(
					!top ? val : win.pageXOffset,
					top ? val : win.pageYOffset
				);

			} else {
				elem[ method ] = val;
			}
		}, method, val, arguments.length );
	};
} );
```

That confirms the function is expected to return a "falsey" value when passed an element reference. Something like this should replace the `getWindow` call:

```javascript
var win;
			
if ( elem.window === elem ) {
	win = elem;
} else if ( elem.nodeType == 9 ) {
	win = elem.defaultView;
}
```

...and then that function can be deleted.

One less `isWindow` call in the bargain as well. As that one is public, it will have to be deprecated rather than deleted. That's next.

HTH

[...]

PR #3456, opened 2016-12-16
Offset: Eliminate little-used internal function

gibson042:
Fixes gh-3449

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -181,8 +181,8 @@ jQuery.each( { scrollLeft: "pageXOffset", scrollTop: "pageYOffset" }, function( 
 			var win;
 			if ( jQuery.isWindow( elem ) ) {
 				win = elem;
-			} else {
-				win = elem.nodeType === 9 && elem.defaultView;
+			} else if ( elem.nodeType === 9 ) {
+				win = elem.defaultView;
 			}
 
 			if ( val === undefined ) {
--- a/src/offset.js
+++ b/src/offset.js
@@ -15,13 +15,6 @@ define( [
 
 "use strict";
 
-/**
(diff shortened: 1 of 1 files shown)


src/offset.js
```diff
@@ -183,7 +176,9 @@ jQuery.each( { scrollLeft: "pageXOffset", scrollTop: "pageYOffset" }, function(
 
 	jQuery.fn[ method ] = function( val ) {
 		return access( this, function( elem, method, val ) {
-			var win = getWindow( elem );
+
+			// Coalesce documents and windows
+			var win = jQuery.isWindow( elem ) ? elem : elem.nodeType === 9 && elem.defaultView;
```
markelog: Maybe `if..else` instead of ternary... please :innocent:

gibson042: Done.

src/offset.js
```diff
@@ -178,7 +178,12 @@ jQuery.each( { scrollLeft: "pageXOffset", scrollTop: "pageYOffset" }, function(
 		return access( this, function( elem, method, val ) {
 
 			// Coalesce documents and windows
-			var win = jQuery.isWindow( elem ) ? elem : elem.nodeType === 9 && elem.defaultView;
+			var win;
+			if ( jQuery.isWindow( elem ) ) {
+				win = elem;
+			} else {
+				win = elem.nodeType === 9 && elem.defaultView;
```
markelog: Yeysies

mgol: Maybe the nodeType check could be moved into the else? That would make it even less hacky.

gibson042: D'oh, good call.

Issue #3436, opened 2016-12-07
[Feature Request] $.fn.contents() supports <template />


TechQuery:
### TODO

 `$.fn.contents()` can return the reference of `HTMLTemplateElement.prototype.content`, not `childNodes`.

Such as how `<iframe />` is supported --- https://github.com/jquery/jquery/blob/1.12-stable/src/traversing.js#L144

### Reference

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template#Attributes

dmethvin:
I know this "return something different depending on the collection" approach is a trademark of jQuery's early design, but I'm not a fan of continuing that tradition. Especially in this case where you can get the information via `$("template").prop("content")` which works all the way back to jQuery 1.6. If there was a strong argument for a getter/setter API it might make sense, but I don't think that applies for this case.

timmywil:
@dmethvin's comment is very sensible and hard to argue with, but I kinda want to argue with it anyway. This might be worth it. I'd like to see examples of return values.

mgol:
@timmywil one example - with HTML:
```html
<template>
	<div id="template-div0">
		<div id="template-div00"></div>
	</div>
	<div id="template-div1"></div>
	<div id="template-div2"></div>
</template>
```
if `templateNode` is a what its name suggests then running:
```
templateNode.content.childNodes
```
produces a document fragment with 5 nodes:
```
[text, div#template-div0, text, div#template-div1, text, div#template-div2, text]
```

timmywil:
Put together http://jsbin.com/mahudek/edit?html,js,output.

```js
templateNode.childNodes
// => NodeList[0]

templateNode.content.childNodes
// => NodeList[7], with text elements

templateNode.content.children
// => HTMLCollection[3]
```

timmywil:
We discussed this one and given that the solution is short and we currently return an empty list, we can fix this. Something like,

```js
...
jQuery.nodeName( elem, "template" ) && elem.content ?
    elem.content.childNodes
...
```

timmywil:
@TechQuery would you like to do the PR?

mgol:
Note that we have to be careful to not make the contents stop being inert once running `template.contents()` so we'll need a test that ensures this does not happen. For example, you could create a `<template>` tag containing a script and `<img>` with an onload handler and making sure none of them fires.

TechQuery:
@timmywil I'd like to make a new PR for this issue~

[...]

PR #3462, opened 2016-12-20
Traversing: $.fn.contents() supports HTMLTemplateElement

TechQuery:
### Summary ###

Add support to `$.fn.contents()` for [HTMLTemplateElement](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template).

Related with #3436

### Checklist ###

### DIFF
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -143,8 +143,11 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-		return elem.contentDocument ||
-            jQuery.merge( [], ( elem.content || elem ).childNodes );
+        switch ( elem.nodeName.toLowerCase() ) {
+            case "iframe":      return elem.contentDocument;
+            case "template":    elem = elem.content;
+            default:            return jQuery.merge( [], elem.childNodes );
+        }
 	}
 }, function( name, fn ) {
 	jQuery.fn[ name ] = function( until, selector ) {
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -148,7 +148,7 @@ jQuery.each( {
         }
 
         if ( jQuery.nodeName( elem, "template" ) ) {
-            elem = elem.content;
+            elem = elem.content || elem;
         }
 
         return jQuery.merge( [], elem.childNodes );
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -143,11 +143,15 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-        switch ( elem.nodeName.toLowerCase() ) {
-            case "iframe":      return elem.contentDocument;
-            case "template":    elem = elem.content;
-            default:            return jQuery.merge( [], elem.childNodes );
+        if ( elem.nodeName === "IFRAME" ) {
+            return elem.contentDocument;
         }
+
+        if ( elem.nodeName === "TEMPLATE" ) {
+            elem = elem.content;
+        }
+
+        return jQuery.merge( [], elem.childNodes );
 	}
 }, function( name, fn ) {
 	jQuery.fn[ name ] = function( until, selector ) {
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -143,11 +143,11 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-        if ( elem.nodeName === "IFRAME" ) {
+        if ( jQuery.nodeName( elem, "iframe" ) ) {
             return elem.contentDocument;
         }
 
-        if ( elem.nodeName === "TEMPLATE" ) {
+        if ( jQuery.nodeName( elem, "template" ) ) {
             elem = elem.content;
         }
 
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -143,7 +143,8 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-		return elem.contentDocument || jQuery.merge( [], elem.childNodes );
+		return elem.contentDocument ||
+            jQuery.merge( [], ( elem.content || elem ).childNodes );
 	}
 }, function( name, fn ) {
 	jQuery.fn[ name ] = function( until, selector ) {
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -147,6 +147,9 @@ jQuery.each( {
             return elem.contentDocument;
         }
 
+        // Support: IE 9 - 11 only, iOS 7 only, Android Browser <=4.3 only
+        // Treat the template element as a regular one in browsers that
+        // don't support it.
         if ( jQuery.nodeName( elem, "template" ) ) {
             elem = elem.content || elem;
         }
--- a/test/index.html
+++ b/test/index.html
@@ -298,5 +298,9 @@ Z</textarea>
 		<div id="template-div1"></div>
 		<div id="template-div2"></div>
     </template>
+    <form id="test_contents">
+        <input type="text" name="contentDocument" />
+        <input type="text" name="content" />
+    </form>
 </body>
 </html>
--- a/test/index.html
+++ b/test/index.html
@@ -291,16 +291,6 @@ Z</textarea>
 	<map name="imgmap" id="imgmap">
 		<area shape="rect" coords="0,0,200,50">
 	</map>
-    <template id="template">
-		<div id="template-div0">
-			<span>Hello, Web Component!</span>
-		</div>
-		<div id="template-div1"></div>
-		<div id="template-div2"></div>
-    </template>
-    <form id="test_contents">
-        <input type="text" name="contentDocument" />
-        <input type="text" name="content" />
-    </form>
+
 </body>
 </html>
--- a/test/index.html
+++ b/test/index.html
@@ -291,6 +291,12 @@ Z</textarea>
 	<map name="imgmap" id="imgmap">
 		<area shape="rect" coords="0,0,200,50">
 	</map>
-
+    <template id="template">
+		<div id="template-div0">
+			<span>Hello, Web Component!</span>
+		</div>
+		<div id="template-div1"></div>
+		<div id="template-div2"></div>
+    </template>
 </body>
 </html>
--- a/test/unit/traversing.js
+++ b/test/unit/traversing.js
@@ -710,7 +710,7 @@ QUnit.test( "prevUntil([String])", function( assert ) {
 } );
 
 QUnit.test( "contents()", function( assert ) {
-	assert.expect( 12 );
+	assert.expect( 16 );
 	var ibody, c;
 
 	assert.equal( jQuery( "#ap" ).contents().length, 9, "Check element contents" );
@@ -741,6 +741,23 @@ QUnit.test( "contents()", function( assert ) {
 	c = jQuery( "#nonnodes" ).contents().contents();
 	assert.equal( c.length, 1, "Check node,textnode,comment contents is just one" );
 	assert.equal( c[ 0 ].nodeValue, "hi", "Check node,textnode,comment contents is just the one from span" );
+
+    c = jQuery( "#template" ).contents();
+    assert.equal( c.length, 7, "Check template element contents" );
+
+    assert.equal( c.find( "span" ).text(), "Hello, Web Component!", "Find span in Template and check its text" );
+
+    jQuery( "<div id=\"templateTest\" />" ).append(
+        jQuery( jQuery.map( c, function( node ) {
+            return document.importNode( node, true );
+        } ) )
+    ).appendTo( document.body );
+
+    c = jQuery( "#templateTest" ).contents();
+    assert.equal( c.length, 7, "Check cloned nodes of template element contents" );
+
+    assert.equal( c.filter( "div" ).length, 3, "Count cloned elements from template" );
+    jQuery( "#templateTest" ).remove();
 } );
 
 QUnit.test( "sort direction", function( assert ) {
--- a/test/unit/traversing.js
+++ b/test/unit/traversing.js
@@ -743,8 +743,7 @@ QUnit.test( "contents()", function( assert ) {
 	assert.equal( c[ 0 ].nodeValue, "hi", "Check node,textnode,comment contents is just the one from span" );
 } );
 
-QUnit.test( "contents() for <template />", function( assert ) {
-	assert.expect( 4 );
+QUnit.test( "contents() for <template />", 4, function( assert ) {
 
     jQuery( "#qunit-fixture" ).append(
         "<template id='template'>" +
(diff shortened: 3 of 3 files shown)


timmywil:
@TechQuery Thanks! This looks great. However, the `form` element has unique behavior in that form elements are attached as properties of the form. I could see `content` being an input name, so if `elem` were a `form`, `elem.content.childNodes` would not return what you expect. Make sense? In other words, I think a `jQuery.nodeName` check will guard us here.

src/traversing.js
```diff
@@ -143,7 +143,11 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-		return elem.contentDocument || jQuery.merge( [], elem.childNodes );
+        switch ( elem.nodeName.toLowerCase() ) {
```
markelog: Just `if..else`, no need in `swtich`

test/index.html
```diff
@@ -291,6 +291,16 @@
 	<map name="imgmap" id="imgmap">
 		<area shape="rect" coords="0,0,200,50">
 	</map>
-
+    <template id="template">
```
markelog: Please create that html in test-case and then append it

test/unit/traversing.js
```diff
@@ -710,8 +710,8 @@ QUnit.test( "prevUntil([String])", function( assert ) {
 } );
 
 QUnit.test( "contents()", function( assert ) {
-	assert.expect( 12 );
```
markelog: Please create separate test for this

test/unit/traversing.js
```diff
@@ -743,6 +743,49 @@ QUnit.test( "contents()", function( assert ) {
 	assert.equal( c[ 0 ].nodeValue, "hi", "Check node,textnode,comment contents is just the one from span" );
 } );
 
+QUnit.test( "contents() for <template />", function( assert ) {
+	assert.expect( 6 );
+
+    jQuery( "body" ).append(
```
markelog: Not on `body` but on `#qunit-fixture` otherwise it will not be cleared

markelog:
In which browsers did you test this?

test/unit/traversing.js
TechQuery: If I append the HTML to `#qunit-fixture`, should I remove them after my testing?

mgol: @TechQuery no, the point of `#qunit-fixture` is that it gets cleaned up automatically between test runs.

TechQuery:
@markelog 

### Chrome
 - [x] 48.0.2564.23
 - [x] 54.0.2840.59 m (64-bit)

### Firefox
 - [x] 50.1.0

### Internet Explorer
 - [x] 11.0.9600.18449
 - [x] *Document Mode* 10 (IE 11 Debugger)
 - [x] *Document Mode* 9 (IE 11 Debugger)

src/traversing.js
```diff
@@ -143,7 +143,15 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-		return elem.contentDocument || jQuery.merge( [], elem.childNodes );
+        if ( jQuery.nodeName( elem, "iframe" ) ) {
+            return elem.contentDocument;
+        }
+
+        if ( jQuery.nodeName( elem, "template" ) ) {
+            elem = elem.content || elem;
```
markelog: That's because of the IE right?

test/unit/traversing.js
```diff
@@ -743,6 +743,49 @@ QUnit.test( "contents()", function( assert ) {
...
+        "    <div id=\"template-div1\"></div>" +
+        "    <div id=\"template-div2\"></div>" +
+        "</template>" +
+        "<form id=\"test_contents\">" +
+        "    <input type=\"text\" name=\"contentDocument\" />" +
+        "    <input type=\"text\" name=\"content\" />" +
+        "</form>"
+    );
+
+    var c = jQuery( "#template" ).contents();
```
markelog: Probably better to use more verbose variable name

test/unit/traversing.js
```diff
@@ -743,6 +743,49 @@ QUnit.test( "contents()", function( assert ) {
...
+    assert.equal( c.filter( "div" ).length, 3, "Count cloned elements from template" );
+    jQuery( "#templateTest" ).remove();
+
+    var iform = jQuery( "#test_contents" );
+    assert.equal( iform.contents().length, 4, "Check no conflict with Form shortcuts" );
+
+    iform.children().eq( 0 ).remove();
+    assert.equal( iform.contents().length, 3, "Check no conflict with Form shortcuts" );
+
+    iform.add( iform.prev() ).remove();
```
markelog: For what?

test/unit/traversing.js
```diff
@@ -743,6 +743,49 @@ QUnit.test( "contents()", function( assert ) {
...
+    var c = jQuery( "#template" ).contents();
+    assert.equal( c.length, 6, "Check template element contents" );
+
+    assert.equal( c.find( "span" ).text(), "Hello, Web Component!", "Find span in Template and check its text" );
+
+    jQuery( "<div id=\"templateTest\" />" ).append(
+        jQuery( jQuery.map( c, function( node ) {
+            return document.importNode( node, true );
+        } ) )
+    ).appendTo( document.body );
```
markelog: `body`?

test/unit/traversing.js
```diff
@@ -743,6 +743,49 @@ QUnit.test( "contents()", function( assert ) {
 	assert.equal( c[ 0 ].nodeValue, "hi", "Check node,textnode,comment contents is just the one from span" );
 } );
 
+QUnit.test( "contents() for <template />", function( assert ) {
+	assert.expect( 6 );
+
+    jQuery( "#qunit-fixture" ).append(
```
markelog: Sounds easier to create node, assign, append and use through the tests?

Like so –
```js
var template = jQuery("<template id='template'>...")
```

Note the single quotes inside the string

test/unit/traversing.js
```diff
@@ -743,6 +743,49 @@ QUnit.test( "contents()", function( assert ) {
...
+        "    <input type=\"text\" name=\"content\" />" +
+        "</form>"
+    );
+
+    var c = jQuery( "#template" ).contents();
+    assert.equal( c.length, 6, "Check template element contents" );
+
+    assert.equal( c.find( "span" ).text(), "Hello, Web Component!", "Find span in Template and check its text" );
+
+    jQuery( "<div id=\"templateTest\" />" ).append(
```
markelog: What we doing here?

test/unit/traversing.js
```diff
@@ -743,6 +743,49 @@ QUnit.test( "contents()", function( assert ) {
...
+        "<form id=\"test_contents\">" +
+        "    <input type=\"text\" name=\"contentDocument\" />" +
+        "    <input type=\"text\" name=\"content\" />" +
+        "</form>"
+    );
+
+    var c = jQuery( "#template" ).contents();
+    assert.equal( c.length, 6, "Check template element contents" );
+
+    assert.equal( c.find( "span" ).text(), "Hello, Web Component!", "Find span in Template and check its text" );
```
markelog: Sometimes you use uppercase and sometimes lowercase, I suggest to `Template` -> `template`, same for `form`

src/traversing.js
mgol: @TechQuery All places where a specific logic has to be applied because of deficiencies of a browser a support comment needs to be added like [this one](https://github.com/jquery/jquery/blob/2d4f53416e5f74fa98e0c1d66b6f3c285a12f0ce/src/data.js#L110).

In this case, I'd expect something like that:
```js
// Support: IE 9 - 11 only, iOS 7 only, Android Browser <=4.3 only
// Treat the template element as a regular one in browsers that
// don't support it.
```
It's only necessary to mention browsers we still support; see https://jquery.com/browser-support/.

markelog: @mgol I was planning to get to support comment realization, but there is also couple considerations with this code, so right now, I think we need to discuss this a bit further

test/unit/traversing.js
TechQuery: `$.fn.clone()` doesn't support `document.importNode()` which child nodes of template needs (as MDN document said)...

markelog: So the same MDN page says `importNode` is not supported in IE 9 and we do support IE 9, soooo

TechQuery: https://developer.mozilla.org/en-US/docs/Web/API/Document/importNode

This page says that `document.importNode()` is an "Initial definition" in *DOM Level 2* (which IE 9 had implemented), only the second optional argument is in DOM Level 4.

markelog: See https://developer.mozilla.org/en-US/docs/Web/API/Document/importNode#Browser_compatibility

markelog: Ooh, I get it, ok

markelog: You probably responding to https://github.com/jquery/jquery/pull/3462#pullrequestreview-14458118 comment though?

test/unit/traversing.js
TechQuery: Test handling of nodes imported from the template.

test/unit/traversing.js
```diff
@@ -743,6 +743,37 @@ QUnit.test( "contents()", function( assert ) {
...
+        jQuery( jQuery.map( contents, function( node ) {
+            return document.importNode( node, true );
+        } ) )
+    ).appendTo( "#qunit-fixture" );
+
+    contents = jQuery( "#templateTest" ).contents();
+    assert.equal( contents.length, 6, "Check cloned nodes of template element contents" );
+
+    assert.equal( contents.filter( "div" ).length, 3, "Count cloned elements from template" );
+    jQuery( "#templateTest" ).remove();
```
markelog: I suppose we no longer need that line?

test/unit/traversing.js
```diff
@@ -743,6 +743,37 @@ QUnit.test( "contents()", function( assert ) {
...
+        "<template id='template'>" +
+        "    <div id='template-div0'>" +
+        "        <span>Hello, Web Component!</span>" +
+        "    </div>" +
+        "    <div id='template-div1'></div>" +
+        "    <div id='template-div2'></div>" +
+        "</template>"
+    );
+
+    var contents = jQuery( "#template" ).contents();
```
markelog: That's a bit different what I had in mind, but I guess we can work with that

src/traversing.js
```diff
@@ -143,7 +143,18 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-		return elem.contentDocument || jQuery.merge( [], elem.childNodes );
+        if ( jQuery.nodeName( elem, "iframe" ) ) {
+            return elem.contentDocument;
+        }
+
+        // Support: IE 9 - 11 only, iOS 7 only, Android Browser <=4.3 only
+        // Treat the template element as a regular one in browsers that
```
markelog: Total nitpick, but anyhow - `only Treat` -> `only treat` and no dot at the end

src/traversing.js
```diff
@@ -143,7 +143,18 @@ jQuery.each( {
...
+        // don't support it.
+        if ( jQuery.nodeName( elem, "template" ) ) {
+            elem = elem.content || elem;
```
markelog: I was wondering that if someone define `content` property on the node, then we would we be screwed, but I we probably don't need to support such case

markelog:
Small fixes still needed, after that I will recheck browser tests and if no one else has any comments we can land this

src/traversing.js
TechQuery: Uh... My opinion of https://github.com/jquery/jquery/pull/3462#discussion_r93927229 is that the first line is a title, and the rest lines are description...

(This is my first PR to jQuery such an international OS project, I'm still in the process of learning OS regulation & specification. I'm really sorry to bother you checking these details...)

src/traversing.js
TechQuery: Before my PR, the code of `$.fn.contents()` didn't check the constructor of `iframe.contentDocument` property too...

src/traversing.js
mgol: > Uh... My opinion of #3462 (comment) is that the first line is a title, and the rest lines are description...

More a concrete information about affected browsers in a consistent format than a title. But yeah, the rest is just a description of what's going on there.

> (This is my first PR to jQuery such an international OS project, I'm still in the process of learning OS regulation & specification. I'm really sorry to bother you checking these details...)

It's absolutely not a problem, ask away! :)

src/traversing.js
mgol: Yeah, I don't think it's necessary to worry about it at this point. If someone defines the `content` property on a template element in a non-standard way, they're asking for a problem.

test/unit/traversing.js
```diff
@@ -743,6 +743,37 @@ QUnit.test( "contents()", function( assert ) {
 	assert.equal( c[ 0 ].nodeValue, "hi", "Check node,textnode,comment contents is just the one from span" );
 } );
 
+QUnit.test( "contents() for <template />", function( assert ) {
```
mgol: Pasting from https://github.com/jquery/jquery/issues/3436#issuecomment-268031156:

Note that we have to be careful to not make the contents stop being inert once running `.contents()` on a template element wrapped in jQuery so we'll need a test that ensures this does not happen. For example, you could create a test with a `<template>` tag containing a script and `` with an onload handler and make sure none of them fires.

mgol: I'd prefer a separate test for inertness, btw; something like:
```js
QUnit.test( "contents() for <template /> remains inert", function( assert ) {
```

mgol: I'd prefer a separate test for inertness, btw; something like:
```js
QUnit.test( "contents() for <template /> remains inert", function( assert ) {
```

TechQuery:
```HTML
<div id="outerBox">
    <template>
        <img src="path/to/image.jpg" onload="alert('OK')" />
    </template>
</div>
```
In my opinion, event handler executes or not, it's depended on browser support of `<template />` during parsing HTML code.

IE treats `<template />` as an `HTMLUnknownElement`, so `` will be in the same document (fragment) of this template, then `onload` will be executed.

Otherwise, `` will be in a `DocumentFragment`, and `onload` won't be executed until `$('#outerBox template').contents().appendTo('#outerBox')`.

mgol: Yes, I know in older browsers the code will get executed; you should skip the inertness test in those browsers; see [this test](https://github.com/jquery/jquery/blob/58c6ca9/test/unit/event.js#L1403) for an example how to do it.

By treating the contents of the template element not carefully you can lose the inertness; for example if you append `template.content` to the DOM. jQuery applies some manipulation steps in various APIs so we need tests that ensure that a simple `$('template').contents()` doesn't break the inertness by itself.

mgol: In the tests in the `onload` for `` and in the `script` tag just assign something to globals and test that the assignments didn't happen. You need to register the globals you set via `Globals.register` to get them cleaned up automatically at the end of the test, [here is an example test that uses this pattern](https://github.com/jquery/jquery/blob/522f546/test/unit/manipulation.js#L2733-L2757).

mgol: As for the last test, it uses `assert.ok( true, "Something is not supported in this browser" )` pattern that we used before `QUnit.skip` was available. Now I'd rather expect something like:
```js
QUnit[ "content" in document.createElement( "template" ) ? "test" : "skip" ](
	"contents() for <template /> remains inert",
	function( assert ) {
		Globals.register( "testScript" );
		Globals.register( "testImgOnload" );
		/* the rest of the test */
	}
)();
```

Let me know if something is not clear.

TechQuery: I confused about it is necessary to `assert.ok()` when I use `QUnit.skip()`? Or it's just a placeholder like [this example](http://api.qunitjs.com/QUnit.skip/#example-0)?

mgol: `assert.ok( true, "String" )` was just a poor man's `QUnit.skip` before it existed; it's not necessary here. That's what I was trying to communicate in my last comment. :)

test/unit/traversing.js
```diff
@@ -743,6 +743,56 @@ QUnit.test( "contents()", function( assert ) {
...
+
+    contents = jQuery( "#templateTest" ).contents();
+    assert.equal( contents.length, 6, "Check cloned nodes of template element contents" );
+
+    assert.equal( contents.filter( "div" ).length, 3, "Count cloned elements from template" );
+} );
+
+QUnit[ "content" in document.createElement( "template" ) ? "test" : "skip" ](
+	"contents() for <template /> remains inert",
+    2,
```
mgol: Please use the `assert.expect` syntax here and in the previous test. The one you used here is deprecated and removed in QUnit 2: https://qunitjs.com/upgrade-guide-2.x/#replace-expected-argument-in-qunit-test

markelog:
@TechQuery awesome work! Your patience and hard work is much appreciated.

@mgol it seems all your concerns were addressed, but just in case, could you give another look?

mgol:
@markelog Everything looks good here 👍

Issue #3439, opened 2016-12-08
Table with tfoot inside plain table results in weird DOM


ShyLionTjmn:
Create simple table with THEAD,TFOOT and TBODY tags and insert it into plain table with just TR and TD.
You will get broken table.

Here is an example:
https://jsfiddle.net/5kbwkzq8/5/

timmywil:
It seems like the same thing would happen without jQuery. We can't stop you from building invalid html. However, the test case is pretty large. Would it be possible to reduce this even further? There's a lot going there.

gibson042:
Ok, this one is kind of cool. A \<tr> is being appended to a table that has no \<tbody>, but contains an inner table that _does_ have one. And when we try to find a \<tbody> to contain the incoming row, [we mistakenly use the one from the inner table](https://github.com/jquery/jquery/blob/2d4f53416e5f74fa98e0c1d66b6f3c285a12f0ce/src/manipulation.js#L54). This should be a straightforward fix.

timmywil:
Thanks for digging into that!

ShyLionTjmn:
Why is it invalid? TBODY is not required when <tr> comes after <table>.

timmywil:
@ShyLionTjmn at first glance, it looked like an unclosed tag or something, but @gibson042 dug further and it is a jQuery bug.

ShyLionTjmn:
Ok. Thanks.

leonardjo:
Seems indeed to be a duplicate of the issue I'm having. Commit 1f614ed fixes the issue I see. Thank you.

PR #3463, opened 2016-12-20
Manipulation: Restrict the tbody search to child nodes

gibson042:
Fixes gh-3439

```
   raw     gz Sizes                                                            
267269  79153 dist/jquery.js                                                   
 86863  30052 dist/jquery.min.js                                               

   raw     gz Compared to master @ bf3a43eff8682b59cec785be6003753fa4b93706    
   +59    +19 dist/jquery.js                                                   
    -1     +9 dist/jquery.min.js
```

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -52,7 +52,7 @@ function manipulationTarget( elem, content ) {
 	if ( jQuery.nodeName( elem, "table" ) &&
 		jQuery.nodeName( content.nodeType !== 11 ? content : content.firstChild, "tr" ) ) {
 
-		return jQuery.filter( "tbody", elem.childNodes )[ 0 ] || elem;
+		return jQuery( ">tbody", elem )[ 0 ] || elem;
 	}
 
 	return elem;
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -47,6 +47,7 @@ var
 	rscriptTypeMasked = /^true\/(.*)/,
 	rcleanScript = /^\s*<!(?:\[CDATA\[|--)|(?:\]\]|--)>\s*$/g;
 
+// Append table rows to a tbody when one is present
 function manipulationTarget( elem, content ) {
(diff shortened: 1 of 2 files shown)


src/manipulation.js
```diff
@@ -47,11 +47,12 @@ var
 	rscriptTypeMasked = /^true\/(.*)/,
 	rcleanScript = /^\s*<!(?:\[CDATA\[|--)|(?:\]\]|--)>\s*$/g;
 
+// Append table rows to a tbody when one is present
 function manipulationTarget( elem, content ) {
 	if ( jQuery.nodeName( elem, "table" ) &&
 		jQuery.nodeName( content.nodeType !== 11 ? content : content.firstChild, "tr" ) ) {
 
-		return elem.getElementsByTagName( "tbody" )[ 0 ] || elem;
+		return jQuery.filter( "tbody", elem.childNodes )[ 0 ] || elem;
```
markelog: Clever, maybe a comment though? You know I like comments :)

markelog:
@timmywil just FYI (in case you didn't know that): you can now approve pull requests, it seems a bit better then "reaction" :)

src/manipulation.js
gibson042: Hmm, what are you looking for that's not covered by the new function comment? Should I update it to something like "New rows for a table belong in its child tbody when one is present", or move it above this line, or something else?

dmethvin: Actually, the code itself looks pretty clear and definitely has the benefit of being more correct. Funny that nobody thought about nested tables for this case.

gibson042: It only comes up when the target doesn't have a tbody but a descendant does, or when such a descendant is in the thead, both of which I imagine are rare. And the logic itself goes [_waaaaaay_ back](https://github.com/jquery/jquery/commit/7c04a64016c1f295dfdf4e731f693f8400a987bc#diff-169760a97de5c86a886842060321d2c8R163).

Issue #3469, opened 2016-12-23
Callbacks with memory are unlocked by add


gibson042:
Moved from https://github.com/jquery/api.jquery.com/issues/1013#issuecomment-268428574

### Description ###
`add` erroneously unlocks Callbacks with memory.

### Link to test case ###
https://jsfiddle.net/o0f244cc/

PR #3470, opened 2016-12-23
Callbacks: Prevent add() from unlocking with-memory lists

gibson042:
Fixes gh-3469

### DIFF
--- a/src/callbacks.js
+++ b/src/callbacks.js
@@ -69,9 +69,7 @@ jQuery.Callbacks = function( options ) {
 		fire = function() {
 
 			// Enforce single-firing
-			if ( !locked ) {
-				locked = options.once;
-			}
+			locked = locked || options.once;
 
 			// Execute callbacks for all pending executions,
 			// respecting firingIndex overrides and runtime changes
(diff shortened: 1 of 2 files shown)


mgol:
I can't find any documentation that would explain why `callbacks.add(fn)` on a locked `callbacks` object is supposed to run the function. Is the documentation lacking or am I just not seeing it?

dmethvin:
> memory: Keeps track of previous values and will call any callback added after the list has been fired right away with the latest "memorized" values (like a Deferred). https://api.jquery.com/jQuery.Callbacks/

> If the Callbacks object is created with the "memory" flag as its argument, additional functions may be added and fired after the callback list is locked. https://api.jquery.com/callbacks.lock/

Not sure if that makes total sense to me, but those are the docs.

gibson042:
Right, it's actually the difference between `lock` and `disable`.

dmethvin:
Yeah, mainly trying to think of specific use cases for all the Callbacks options. Clearly we haven't even probed them all for core. :)

gibson042:
This broke an effects test; I'm investigating now. http://swarm.jquery.org/job/4519

PR #3478, opened 2016-12-30
Build: Drop support for building on Node.js 0.12

mgol:
### Summary ###

Node.js 0.12 is no longer supported by upstream. So far we had to keep
a workaround that would skip running ESLint there (or even load its task)
as it no longer supports Node.js <4; the same applied to Node smoke tests
as jsdom has also dropped support for those old Node versions. Those
workarounds have been removed now.

Ref 030191ae32dfcb7ecb2efb61d17a4964a3633e44

### Checklist ###

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 sudo: false
 node_js:
-- "0.12"
 - "4"
 - "6"
 - "7"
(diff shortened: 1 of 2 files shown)


markelog:
Yey! We shouldn't have any issues with jenkins right?

mgol:
I've switched our project on Jenkins to Node 4 a while ago. :) I should switch it to Node 6 now since Node 4 will get into Maintenance Mode in 4 months.

markelog:
Cool

PR #3483, opened 2017-01-04
fix isWindow in core.js

amnhh:
```js
var a = {};
a.window = a;
console.log(a.window === a); // => true
```

so we need to fix it with `Object.prototype.toString.call(obj) === '[object Window]'`

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -216,7 +216,7 @@ jQuery.extend( {
 	},
 
 	isWindow: function( obj ) {
-		return obj != null && obj === obj.window;
+		return obj != null && obj === obj.window && toString.call(obj) === '[object Window]';
 	},
 
 	isNumeric: function( obj ) {
--- a/src/core.js
+++ b/src/core.js
@@ -216,7 +216,7 @@ jQuery.extend( {
 	},
 
 	isWindow: function( obj ) {
-		return obj != null && obj === obj.window && toString.call(obj) === '[object Window]';
+		return obj != null && obj === obj.window && toString.call( obj ) === '[object Window]';
 	},
 
 	isNumeric: function( obj ) {
--- a/src/core.js
+++ b/src/core.js
@@ -216,7 +216,7 @@ jQuery.extend( {
 	},
 
 	isWindow: function( obj ) {
-		return obj != null && obj === obj.window && toString.call( obj ) === '[object Window]';
+		return obj != null && obj === obj.window && toString.call( obj ) === "[object Window]";
 	},
 
 	isNumeric: function( obj ) {


dmethvin:
I don't know that this helps our internal usage since we're usually trying to determine whether something is either a DOM element or a `window` object and not throwing arbitrary stuff at it. However, since `jQuery.isWindow` is a documented method it does seem like we might want to fix this. The good news is that this fix as described should be possible now that IE8 and its strange `window` host objects is gone.

dmethvin:
@amnhh Please leave this PR open if you're interested, it would be good for discussion by the team! :+1:

amnhh:
👍thanks and I'm glad to discuss it.

timmywil:
I'm fine with it as long as it works in our supported browsers. `isWindow` has a long history of being dirty, probably because of the weirdness in old IE.

amnhh:
@timmywil 👍 I think so. I am willing to find more weirdness in old IE and try to get a compatible definition

timmywil:
@amnhh Sorry, I didn't mean you need to check old IE. I just meant that `isWindow` is a product of history involving old IE, and now that we don't support old IE, perhaps a sensible `isWindow` is possible.

timmywil:
@amnhh Why did you close this?

dmethvin:
@amnhh we like the idea of making `isWindow` better, can you see if our 3.0 supported browsers work with a better method? This PR only failed due to some style issues.

>  `toString.call(obj) === '[object Window]'`
>  ` toString.call( obj ) === "[object Window]"`

amnhh:
@timmywil  I just find that my idea doesn't work in IE8. I need to find a better way to support browsers well

amnhh:
👍 @dmethvin thanks, I fix style issues and it succeeds. And I find that my idea goes well in IE9. :)

dmethvin:
Oh it just occurred to me, I wonder whether we need to use `isWindow` on nodejs where people have to build a fake one? Reviewing where we use it internally, I don't think we will break anything. @timmywil what do you think?

mgol:
@dmethvin yes, this will fail in jsdom as their `window` will stringify to `[object Object]`.

I think at this point the gains of this change are smaller than the drawbacks of reducing compatibility with jsdom.

This would only work if jsdom started to define the `Symbol.toStringTag` property on their `window` isntance but even then it would only work with Node 6+ and Node 4 [still has 14 months of life](https://github.com/nodejs/LTS/) ahead of it.

mgol:
OTH, it seems jsdom patches `Object.prototype.toString` sometimes (see e.g. https://github.com/tmpvar/jsdom/pull/970) so maybe there's a way to fix it on their side?

mgol:
Since there is no easy way for jsdom to adjust to have correct `Object.prototype.toString.call(window)` without relying on `Symbol.toStringTag` which only works in Node 6+, we decided to not change the `isWindow` code at the moment. We'll revisit in the future.

Thanks for your efforts!

PR #3487, opened 2017-01-05
Fix .offset() and .position() bugs

gibson042:
### Summary ###
Makes `.offset()` always relative to the document and `.position()` always relative to the closest [containing block](https://drafts.csswg.org/css-display-3/#containing-block) offsetParent (if any).

Some new tests fail on IE\<11 (where `document.documentElement.getBoundingClientRect()` erroneously includes margins and disregards viewport scrolling), and have been manually disabled/patched up.

Fixes gh-3080
Fixes gh-3107
Closes gh-3096
Ref gh-3479

### Checklist ###
* [x] All authors have signed the CLA at https://contribute.jquery.com/CLA/
* [x] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* [x] If needed, a docs issue/PR was created: https://github.com/jquery/api.jquery.com/pull/1027

PR #3489, opened 2017-01-05
Tests: Update the loader so test pages always get jQuery

gibson042:
This helped work on gh-3487.

### DIFF
--- a/test/jquery.js
+++ b/test/jquery.js
@@ -2,21 +2,35 @@
 ( function() {
 	/* global loadTests: false */
 
-	var src,
-		path = window.location.pathname.split( "test" )[ 0 ],
+	var path = window.location.pathname.split( "test" )[ 0 ],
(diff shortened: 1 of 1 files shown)


markelog:
Why is this needed?

gibson042:
It's not _needed_, but being able to open the iframe-test pages in a browser tab and make the same jQuery calls as their respective tests is definitely a worthwhile convenience, especially when we can get it without breaking anything. But absent this fix, opening e.g. test/data/offset/absolute.html generates `Uncaught TypeError: Cannot read property 'config' of undefined` from https://github.com/jquery/jquery/blob/a8816caf26c0acc7067cd5bad9c03cfb54e54979/test/jquery.js#L11 , and subsequently fails to load jQuery.

markelog:
Cool! Could you put that explanation in the commit? :)

gibson042:
Done.

Issue #3434, opened 2016-12-06
"Too much recursion" with jQuery 3, Firefox, and animate's complete callback.


Pancakeo:
Versions:
- Firefox 50 (latest)
- jQuery 3.1.1

Thing: I have a loading bar that animates from left-to-right, then resets. Note that it uses the complete callback (from animate) to fire itself again.

Repro: 
1. Load http://jsbin.com/woperab/8/edit?html,css,js,output in Firefox
2. Run, in Firefox.
3. Minimize Firefox window.
4. Wait about 2 minutes (get a cup of coffee). 
5. Come back.
6. Drink coffee.
7. Square has stopped moving. Console shows: "Too much recursion".

Note that it works if I wrap the "reset" call in a setTimeout, as in:

`
complete: function() {
  setTimeout(reset, 0)
}
`

Speculation: Probably something to do with the requestAnimationFrame changes in jQuery 3? (If you leave the window focused, the recursion error does not appear).

Does NOT happen in Chrome/IE. Also, doesn't happen with jQuery 2.1.3.

markelog:
Yeah, this kinda, sorta by design – https://github.com/jquery/jquery/pull/1578#r12712029

/cc @dmethvin, @scottgonzalez

timmywil:
One option to try is to use page visibility api to swap rAF with setTimeout.

PR #3496, opened 2017-01-12
9d822bc1c13dd3393b418210a99537c22c06f2c3 followup

gibson042:
9d822bc1c13dd3393b418210a99537c22c06f2c3 failed an effects test (cf. https://github.com/jquery/jquery/pull/3470#issuecomment-271388489 ), but it turns out that things were already broken in subtle ways when animations spawn each other:
* immediately-done animations miss the final progress notification
* callbacks are registered _after_ the first tick
* animations get skipped when old ones are forcefully stopped mid-tick

Each fix revealed the next flaw, so all are addressed here in order to resolve the current CI failure. I should probably make real issues so they can be resolved by this PR.

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -392,7 +392,7 @@ function Animation( elem, properties, options ) {
 		animation.opts.start.call( elem, animation );
 	}
 
-	// attach callbacks from options
+	// Attach callbacks from options
 	animation
 		.progress( animation.opts.progress )
 		.done( animation.opts.done, animation.opts.complete )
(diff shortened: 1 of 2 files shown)


src/effects.js
```diff
@@ -392,6 +392,13 @@ function Animation( elem, properties, options ) {
 		animation.opts.start.call( elem, animation );
 	}
 
+	// attach callbacks from options
```
markelog: Stupid nitpick: probably better to start with uppercase 'attach' -> 'Attach', I know it was like that before and linter didn't catch that for some reason...

markelog:
This is awesome, nicely done, well, as always! :)

timmywil:
Any chance this affects https://github.com/jquery/jquery/issues/3434?

gibson042:
I wouldn't think so, but it's possible. #3434 is unexpected.

PR #3501, opened 2017-01-16
Effects: sync .stop/.finish with .tick to maintain proper queue (#3497)

### LINKED ISSUE #3497: jQuery.fx.tick insufficiently guards against timer queue manipulation
### Description ###
`jQuery.fx.tick` executes each queued timer in turn, maintaining its place when it [removes a finished animation](https://github.com/jquery/jquery/blob/master/src/effects.js#L646). However, `.stop` and `.finish` can _also_ remove timers from the queue, and when invoked while `tick` is iterating (i.e., from animation callbacks) can cause it to lose its place, skipping over one or more animations. This is often harmless (since the animation is revisited in the next tick), but can introduce visible stutter or (in a pathological case that I didn't bother demonstrating) starve an animation outright.

Additionally, because `tick` iterates to the end of the queue, _new_ animations get processed twice in their first tick.

We should fix both by making the three functions more aware of and cooperative with each other.

### Link to test cases ###
* [stutter](https://jsfiddle.net/2v6cyorf/)
* [extra processing](https://jsfiddle.net/4peg0sL9/)

preethi26:
#3497 

### Summary ###

1.Whenever the .finish or the .stop functions modify the timers queue, we maintain all those indexes that are removed in a list (  i.e, lis[ ]  ). Whenever the tick function iterates over the queue, we update the iterator based on the value of the indexes removed ( only when the indexes less than the current iterator value are removed there will be a skipping over the animations  ).

2. Handled extra processing of the same animation by doing as mentioned below:
  All the new animations that are added to the timers queue in the middle of a particular iteration of the tick function are processed in the next iteration.

### Checklist ###

* [x] Handled the "skipping over multiple animations " of tick function.
* [ ] Handled extra processing.

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -483,7 +483,7 @@ jQuery.speed = function( speed, easing, fn ) {
 
 	return opt;
 };
-
+var lis = [];
 jQuery.fn.extend( {
 	fadeTo: function( speed, to, easing, callback ) {
 
@@ -552,6 +552,7 @@ jQuery.fn.extend( {
 
 					timers[ index ].anim.stop( gotoEnd );
 					dequeue = false;
+					lis.push(index);
 					timers.splice( index, 1 );
 				}
 			}
@@ -590,6 +591,7 @@ jQuery.fn.extend( {
 			for ( index = timers.length; index--; ) {
 				if ( timers[ index ].elem === this && timers[ index ].queue === type ) {
 					timers[ index ].anim.stop( true );
+					lis.push( index );
 					timers.splice( index, 1 );
 				}
 			}
@@ -631,15 +633,25 @@ jQuery.each( {
 } );
 
 jQuery.timers = [];
+var oldLength = 0;
 jQuery.fx.tick = function() {
 	var timer,
 		i = 0,
 		timers = jQuery.timers;
 
 	fxNow = jQuery.now();
+	oldLength = timers.length;
+	for ( ; i < old_length; i++ ) {
+		var removedTimers = 0;
 
-	for ( ; i < timers.length; i++ ) {
-		timer = timers[ i ];
+		for (var j = 0; j < lis.length; j++) {
+			if ( lis[ j ] <= i ) {
+				removeTimers++;
+			}
+				lis.splice( j, 1 );
+			}
+			i = i - removedTimers;
+			timer = timers[ i ];
 
 		// Checks the timer has not already been removed
 		if ( !timer() && timers[ i ] === timer ) {
--- a/src/effects.js
+++ b/src/effects.js
@@ -552,7 +552,7 @@ jQuery.fn.extend( {
 
 					timers[ index ].anim.stop( gotoEnd );
 					dequeue = false;
-					lis.push(index);
+					lis.push( index );
 					timers.splice( index, 1 );
 				}
 			}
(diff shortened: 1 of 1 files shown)


src/effects.js
```diff
@@ -631,15 +633,25 @@ jQuery.each( {
...
 		timers = jQuery.timers;
 
 	fxNow = jQuery.now();
+	oldLength = timers.length;
+	for ( ; i < oldLength; i++ ) {
+		var removedTimers = 0;
 
-	for ( ; i < timers.length; i++ ) {
-		timer = timers[ i ];
+		for ( var j = 0; j < lis.length; j++ ) {
```
gibson042: I'm hoping to avoid O(N²) operations and extra global variables, and was thinking we'd do this by checking/clearing `timer.elem` since that's already cleared in an `always` callback.

gibson042 (changes requested):

src/effects.js
preethi26: ok ! not a problem.
I would like to know if the second bug about extra processing was handled?
As I wasn't sure about it I did no mark it in the checklist.

gibson042: Probably not, but we'll see. This PR will also require unit tests that don't pass without the updated code.

markelog:
@preethi26 friendly ping

src/effects.js
preethi26: hey @gibson042 ,

`jQuery.timers = [];
jQuery.fx.tick = function() {
	var timer,
		i = 0,
		timers = jQuery.timers;

	fxNow = jQuery.now();
	for ( ; i < timers.length; i++ ) {
		timer = timers[ i ];
                // storing the previous value of the 'timer'
		var x  = timers.indexOf(timer) 
		if (  !timer()  ) {
			if( x > timers.indexOf(timer)) {    
                              // checking and updating the value of  'i' ( if changed ).

				i = i - ( x - timers.indexOf(timer) ) 
			}
			if( timers[ i ] === timer ) {
				timers.splice( i--, 1 );
			}
		}
	}

	if ( !timers.length ) {
		jQuery.fx.stop();
	}
	fxNow = undefined;
};
`
This is the edited [ jQuery.fx.tick()] https://github.com/jquery/jquery/blob/master/src/effects.js#L644) function.
I would like to know if this fixes the first part of the bug before proceeding further.

Thank you ,
Preethi.

preethi26: Sorry, just realized that it wouldn't work. Coz we need to check the value of  "timer" 's index in between consecutive iterations ( which is not done here ).

preethi26: @gibson042 
error_bug_1


In every iteration, making the value of the iterator equal to the value of the previous timer that successfully ran + 1.

gibson042:
This is really not what I'm looking for; see my comment about [O(N²) operations and extra global variables](https://github.com/jquery/jquery/pull/3501/files#r96265598). And note that a functional change will absolutely not be accepted without unit tests.

preethi26:
@gibson042 ,
Got it, and I am sorry for the previous mess. As I am less familiarised with the Source Code, I don't have much of clarity on what you meant in your first comment ( because of which I think I overlooked it and thought of some other solution that avoids O(N²) operations and new global variables, sorry for that ). Could you please brief me a little as I would like to fix and finish working on this issue ( if you didn't find the code that you were already working on ? )

gibson042:
No problem. As I mentioned in the comment, I think we can mostly cover this by checking and deleting the `elem` property of items in `jQuery.timers` instead of manipulating the array itself, at least inside iterations over it (i.e., extending [already existing logic](https://github.com/jquery/jquery/blob/3c89329cb24cb144c3296cf0589bce9ca59d63a3/src/effects.js#L295-L296) so `jQuery.timers` can temporarily hold zombie entries). Outside the iterations, though, we may well need a global mutex variable.

Juanito0813:
Thank you

Juanito0813:
I forgot my password

mgol:

Issue #3475, opened 2016-12-27
Deprecate jQuery.nodeName, remove in 4.0


mgol:
### Description ###
`jQuery.nodeName` is another undocumented utility function that's exposed on the jQuery object. We should remove it (probably in 4.0 as it's been there for a long time).

### Link to test case ###
https://jsfiddle.net/o648k3ga/1/

timmywil:
If we want to remove it in 4.0, we should deprecate in 3.2.

mgol:
We can't deprecate something that's not documented. :) It's more than deprecated.

timmywil:
@mgol I'd still like to give notice. We're doing the same for other undocumented functions that we plan on removing.

karan-96:
Hello . i think ,to fix up this issue, jquery.nodeName function should be removed from src/core.js and added to src/deprecated.js. Correct me if i am wrong or if i am missing something.Can i take up this issue and work with this approach?

timmywil:
@karan-96 That would be great! You'll find information on how to contribute [here](http://contribute.jquery.org/), with info on commits and PRs [here](http://contribute.jquery.org/commits-and-pull-requests/). As far as the required work for this issue, that should be all the work needed for jQuery core. We'll also want to add a warning to the [jQuery migrate plugin](https://github.com/jquery/jquery-migrate/issues/246).

dmethvin:
Just to clarify what has to be done here, I know it's got a PR at gh-3505 but we have to eliminate internal calls in order to warn about it in jQuery Migrate.

- [ ] Create an internal `nodeName` and put it in src/var/nodeName.js
- [ ] Change all internal calls to `jQuery.nodeName` to our new `nodeName` function
- [ ] Expose the internal `nodeName` as `jQuery.nodeName` in deprecated.js

Does that sound right to everyone?

mgol:
@dmethvin yes, that's exactly what I said in https://github.com/jquery/jquery/pull/3505#issuecomment-273482957 a few days ago. :)

mgol:
(although it seems it doesn't have to be a var-module in case of an exported function)

dmethvin:
@mgol oh sorry, I missed that!

PR #3505, opened 2017-01-17
Core: Deprecate jQuery.nodeName

karan-96:
### Summary ###
fixes issue #3475

### Checklist ###

### DIFF
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -29,6 +29,9 @@ jQuery.fn.extend( {
 		} else {
 			jQuery.ready( true );
 		}
+},
+	nodeName: function( elem, name ) {
+		return elem.nodeName && elem.nodeName.toLowerCase() === name.toLowerCase();
 	}
 } );
 
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -1,6 +1,7 @@
 define( [
-	"./core"
-], function( jQuery ) {
+	"./core",
+	"./core/nodeName"
+], function( jQuery, nodeName ) {
 
 "use strict";
 
@@ -30,9 +31,7 @@ jQuery.fn.extend( {
 			jQuery.ready( true );
 		}
 },
-	nodeName: function( elem, name ) {
-		return elem.nodeName && elem.nodeName.toLowerCase() === name.toLowerCase();
-	}
+	nodeName: nodeName
 } );
 
 jQuery.isArray = Array.isArray;
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -1,10 +1,10 @@
 define( [
 	"../core",
 	"../core/access",
-	"../core/nodeName",
 	"./support",
 	"../var/rnothtmlwhite",
-	"../selector"
+	"../selector",
+	"../core/nodeName"
 ], function( jQuery, access, support, rnothtmlwhite, nodeName ) {
 
 "use strict";
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -1,11 +1,11 @@
 define( [
 	"../core",
 	"../core/access",
+	"../core/nodeName",
 	"./support",
 	"../var/rnothtmlwhite",
-	"../selector",
-	"../core/nodeName"
-], function( jQuery, access, support, rnothtmlwhite, nodeName ) {
+	"../selector"
+], function( jQuery, access, nodeName, support, rnothtmlwhite ) {
 
 "use strict";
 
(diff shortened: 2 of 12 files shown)


mgol:
Thanks for the PR. However, more things are needed for this to land:

1. `jQuery.nodeName` has to stop being used internally; its definition has to be extracted to a separate AMD module under `src/core/var/nodeName.js` and included via AMD where needed; the `jQuery.nodeName` definition should be reduced to something like:
```js
define( [
	/* other code */
	"./core/var/nodeName"
], function ( /* other code */ nodeName ) {

"use strict";

/* other code */

jQuery.nodeName = nodeName;

} );
```
2. Tests for `jQuery.nodeName` have to be moved to `test/unit/deprecated.js` as well.

You can see 1b9575b9d14399e9426b9eacdd92b3717846c3f2 for what was needed for the `jQuery.isArray` deprecation; similar actions will be needed here with the exception that the deprecated method is this time not 100% equivalent to a native one so we still need to keep the code in a separate module (in `src/core/var/nodeName.js` that you need to create as I mentioned).

Would you like to apply those changes? If you have any doubts about what has to be done just ask.

karan-96:
yes i would like to work on this.

mgol:
Cool! Let us know if you stumble upon any problems.

karan-96:
@mgol  sir, my 'src/core/var/nodeName.js' looks like
```
define(function( elem, name ) {
	"use strict";

	return elem.nodeName && elem.nodeName.toLowerCase() === name.toLowerCase();
} );

```
and i need to include its definition(the one you mentioned above) via AMD everywhere jQuery.nodeName has been used. Am i getting your 1st point correct?

mgol:
@karan-96 The function passed to `define` defines a module that should return what you want to export so you need to wrap your function in another. Look at https://github.com/jquery/jquery/blob/3.1.1/src/css/addGetHookIf.js for how it can be done.

Also, now that I look at it it probably doesn't have to be a `var`-module so you can put it in `src/core/nodeName.js` instead of `src/core/var/nodeName.js`.

karan-96:
@mgol sir,
1.How can I include  nodeName function in /test files? I tried “define” like I did in /src files but i found errors when I built jquery. So after exploring the directory I found “test/.eslintrc.json”. And  I found “define” : “false” there under globals option. Is it because of this that I am not able to use “define”?  And How should I proceed now?

2.And do i have to write a new test for nodeName and include it in deprecate.js?

mgol:
@karan-96 In tests just use `jQuery.nodeName`; we're testing what is publicly exposed, not private modules. Also, you don't have to write any tests, just find exsiting ones and move them to `test/unit/deprecated.js`.

mgol:
@karan-96 Do you think you'll have time to move the unit tests soon?

karan-96:
@mgol ,i did not find any test for jquery.nodeName. please tell me if i am missing something?

mgol:
@karan-96 You're right, this is a private API so there was no tests for it.

A couple remarks:
1. Could you rebase this PR over latest `master`? There is a minor conflict in `src/deprecated.js`.
2. After you rebase you should notice a few extra uses of `jQuery.nodeName` in `src/traversing.js`.
3. I was wrong that you can just use `jQuery.nodeName` in tests, sorry for that. We'll want to remove the method one day and it being used in tests will be problematic. All the mentions should be replace by bare `elem.nodeName`, sometimes passed through `toLowerCase()` but not always - you can verify if you replaced it incorrectly when you run the tests.
4. I've noticed you added the entries for `nodeName` in the dependencies lists of various modules to the end of the list. They should be placed near other entries for the same module, e.g. in `src/offset.js` the entry `"./core/nodeName"` should be placed after `"./core/access"` not `"./css/support"`.

Could you apply those changes?

karan-96:
yes , i will apply these changes.

karan-96:
@mgol sir i have applied the changes.please review them ?

src/deprecated.js
```diff
@@ -29,6 +29,9 @@ jQuery.fn.extend( {
 		} else {
 			jQuery.ready( true );
 		}
+},
+	nodeName: function( elem, name ) {
+		return elem.nodeName && elem.nodeName.toLowerCase() === name.toLowerCase();
```
timmywil: Only one thing left in my mind. Import `src/core/nodeName.js` in this file and set `nodeName: nodeName` rather than writing the function twice.

mgol:
Landed, thanks @karan-96!

karan-96:
thank you all for your guidance throughout.

PR #3537, opened 2017-02-10
Typos in the comments

amatsuda:
Here are tiny typo fixes.
* s/explictly/explicitly/
* s/specifing/specifying/


markelog:
Yeah, 1.9.1 version is downloaded only for the test purposes as it was when it was released, basically committed dependency.

Thank you anyway though, if you found any typos in our current code, pr would be appreciated :)

amatsuda:
@markelog I see. Sorry for bothering you with this.

Issue #2297, opened 2015-05-11
Remove jQuery.style/jQuery.css


mzgol:
`jQuery.style` & `jQuery.css` are undocumented, do we need to expose them or can we hide them as we plan to do with e.g. `buildFragment` & `domManip`?

scottgonzalez:
We currently use `jQuery.css()` in jQuery UI. I'm sure we can switch to `.css()` for all of our uses though.

gibson042:
Once upon a time, there was a ticket to make _more_ such methods: http://bugs.jquery.com/ticket/12851

I don't know if that's enough to save these, though.

timmywil:
Yea, I think this is a bit different than `buildFragment` and `domManip`. I actually prefer using these methods sometimes. Forcing users to create a new jQuery collection to get/set CSS is unnecessary.

dmethvin:
If the goal is to retroactively bless use of the current `jQuery.css()` and `jQuery.style()` as public APIs then we're going to have a pretty haphazard set of low-level APIs with inconsistent calling conventions. I guess it can't get any worse than `.each()` vs. `jQuery.each()` though. :smiling_imp: 

I'd rather have a `jQuery.css(elem, name, value)` setter and `jQuery.css(elem, name)` getter if we're going to have one, similar to `jQuery.data`. At least that would give us some consistency in the low level API, but it may break current users of the undocumented interface and we'd have to figure out what to do about the `extra` and `styles` args.

mzgol:
> I'd rather have a jQuery.css(elem, name, value) setter and jQuery.css(elem, name) getter if we're going to have one, similar to jQuery.data. At least that would give us some consistency in the low level API, but it may break current users of the undocumented interface and we'd have to figure out what to do about the extra and styles args.

Definitely. I had a moment today when I was staring at `jQuery.style` & `jQuery.css` and not immediately seeing what's the responsibility division between them. This looks extremely confusing, especially when taking into account the fact that `jQuery.fn.css` is both a getter & a setter.

Having private positional params in public APIs also seems bad.

timmywil:
Agreed. In this case, I'd rather see us move to what Dave suggested than drop them completely.

mzgol:
> Agreed. In this case, I'd rather see us move to what Dave suggested than drop them completely.

That works for me as well.

mzgol:
This relates to APIs currently undocumented so I assume deferring it to 3.1.0 would be ok?

[...]

PR #3557, opened 2017-03-01
CSS: Support custom properties

mgol:
Fixes gh-3144
Closes gh-3199

### Summary ###

This PR adds suppport for CSS Custom Properties (a.k.a. CSS Variables). This is a finalized version of the PR #3199 with passing tests.

### Checklist ###

* [x] All authors have signed the CLA at https://contribute.jquery.com/CLA/ (the old jQuery one has been signed before the foundation name change, is that enough?)

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -28,6 +28,7 @@ var
 	// except "table", "table-cell", or "table-caption"
 	// See here for display values: https://developer.mozilla.org/en-US/docs/CSS/display
 	rdisplayswap = /^(none|table(?!-c[ea]).+)/,
+	rcustomProp = /^--/,
 	cssShow = { position: "absolute", visibility: "hidden", display: "block" },
 	cssNormalTransform = {
 		letterSpacing: "0",
@@ -55,6 +56,16 @@ function vendorPropName( name ) {
 			return name;
 		}
 	}
+}
+
+// Return a property mapped along what jQuery.cssProps suggests or to
+// a vendor prefixed property.
+function finalPropName( name ) {
+	var ret = jQuery.cssProps[ name ];
+	if ( !ret ) {
+		ret = jQuery.cssProps[ name ] = vendorPropName( name ) || name;
+	}
+	return ret;
 }
 
 function setPositiveNumber( elem, value, subtract ) {
@@ -218,10 +229,15 @@ jQuery.extend( {
 		// Make sure that we're working with the right name
 		var ret, type, hooks,
 			origName = jQuery.camelCase( name ),
+			isCustomProp = rcustomProp.test( name ),
 			style = elem.style;
 
-		name = jQuery.cssProps[ origName ] ||
-			( jQuery.cssProps[ origName ] = vendorPropName( origName ) || origName );
+		// Make sure that we're working with the right name. We don't
+		// want to query the value if it is a CSS custom property
+		// since they are user-defined.
+		if ( !isCustomProp ) {
+			name = finalPropName( origName );
+		}
 
 		// Gets hook for the prefixed version, then unprefixed version
 		hooks = jQuery.cssHooks[ name ] || jQuery.cssHooks[ origName ];
@@ -257,7 +273,11 @@ jQuery.extend( {
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
 
-				style[ name ] = value;
+				if ( isCustomProp ) {
+					style.setProperty( name, value );
+				} else {
+					style[ name ] = value;
+				}
 			}
 
 		} else {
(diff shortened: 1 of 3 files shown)


src/css.js
```diff
@@ -257,7 +264,11 @@ jQuery.extend( {
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
 
-				style[ name ] = value;
+				if ( isCustomProp ) {
```
mgol: I originally planned to remove it as discussed in #3199 to fix https://bugs.jquery.com/ticket/14394 but `CSSStyleDeclaration#setProperty` requires styles in the kebab-case format so whatever we decide about that I'd like it to land separately.

src/css.js
```diff
@@ -276,11 +287,16 @@ jQuery.extend( {
...
-		name = jQuery.cssProps[ origName ] ||
-			( jQuery.cssProps[ origName ] = vendorPropName( origName ) || origName );
+			origName = jQuery.camelCase( name ),
+			isCustomProp = rcustomProp.test( name );
+
+		// Make sure that we're working with the right name. We don't
+		// want to modify the value if it is a CSS custom property
+		// since they are user-defined.
+		if ( !isCustomProp ) {
+			name = jQuery.cssProps[ origName ] ||
```
markelog: Maybe we could simplify this logic with usual `if..else` or move it to the helper method? Like `getName` or something?

mgol: This logic is used only in `jQuery.style` & `jQuery.css` which have some duplicated code anyway. We have an issue to unify them (#2297) so this helper method would be used only once again when it's done.

markelog: I would start now at least for the affected code, that issue is open for two years now

mgol: I can extract it but won't we want to de-extract it back once those two methods get unified?

test/unit/css.js
```diff
@@ -1556,4 +1556,68 @@ QUnit.test( "Do not throw on frame elements from css method (#15098)", function(
...
+
+	div.css( { "--mixedCase": "green" } );
+	assert.equal( div.css( "--mixedCase" ), "green", "Modified CSS custom property with mixed case" );
+
+	div.css( { "--theme-dark": "purple" } );
+	assert.equal( div.css( "--theme-dark" ), "purple", "Modified CSS custom property with dashed name" );
+
+	assert.equal( $elem.css( "--prop1" ), "val1", "Basic CSS custom property" );
+
+	// Support: Chrome 55, Safari 9.1-10.0 only
```
mgol: I forgot to remove Chrome from here.

src/css.js
markelog: I just don't like this
```js
name = jQuery.cssProps[ origName ] ||
  ( jQuery.cssProps[ origName ] = vendorPropName( origName ) || origName );
```
thingy :). Really hard to read

mgol: Ah, right. I do agree with that. And if you refactor it into `if-else`s then it get long enough that it makes sense to extract it, I get it now. On it.

mgol:
PR updated. I hope you like the changes, @markelog. :-)

mgol:
Rebased so that it can be tested in AMD mode.

mgol:
@timmywil the PR changed a little since you last looked at it so PTAL.

src/css.js
dmethvin: I'm fine with breaking that out for a later fix, it does seem like all of this can be refactored better at that time.

src/css/curCSS.js
```diff
@@ -16,7 +16,10 @@ function curCSS( elem, name, computed ) {
 	computed = computed || getStyles( elem );
 
 	// Support: IE <=9 only
-	// getPropertyValue is only needed for .css('filter') (#12537)
+	//
+	// getPropertyValue is needed for:
+	//   .css('filter') (#12537)
+	//   .css('--customProperty) (#3144)
```
dmethvin: :+1: for the comment!

mgol: Doesn't the IE 9 mention above look like it referred to the whole comment?

mgol: Maybe the support comment should be removed in favor of `//   .css('filter') in IE 9 only (#12537)` as we won't be able to remove this code path anyway due to CSS vars.

dmethvin: Yeah i was looking at the added lines and not thinking about it seeming to be related to IE9. I'd be fine with removing the `// Support:` comment since like you say it's no longer just an IE9 requirement.

mgol: PR updated. Does it look OK now?

dmethvin: Yep!

Issue #3193, opened 2016-06-22
different result of width() and height() since jQuery 3.0


HolgerJeromin:
Just to provide feedback:
#2439 has not only the impact returning non-integer values.

I have a CSS transform:scale(2) in a root element with a div which has css width 200px. 
jQuery 2.x returns 200 for .width() as it uses offsetWidth
jQuery 3.0 returns 400 for .width() as it uses getBoundingClientRect()

This is a breaking change which should be at least mentioned in the upgrade guide.

dmethvin:
Agreed! It's a good breaking change though, wouldn't you say? 

As far as documenting, perhaps we could add a sentence to the existing item here? Either that or create an entirely new "Breaking change" item for it. Rewording the heading will break existing links.

https://jquery.com/upgrade-guide/3.0/#breaking-change-width-height-css-quot-width-quot-and-css-quot-height-quot-can-return-non-integer-values

HolgerJeromin:
.width() is now more correct, yes. 
But it was kind of buggy since the beginning of jQuery. 
So providing .displayWidth() instead of changing the meaning would be less painful for my application.
IMO this issue is worth a new entry in the upgrade guide.

HolgerJeromin:
The current implementation has one problem. Setting and Getting via css() is not symmetric anymore:
```
$("#Button").css("width", 200)
$("#Button").css("width")
"400px"
```

dmethvin:
Good point, and that doesn't seem good. Not sure how to deal with it though. It would help if `.css("width")` returned the CSS value and `.width()` could return the actual width, but that would be another breaking change no doubt and perhaps worse than leaving it inconsistent.

HolgerJeromin:
Applications like mine have the defect right now. 
Only a few people having an issue with transform. 
If you change css() back half of these people are getting back a working solution (without changing own code) . The other half at least are able to adapt (without asking  outerwidth themselves).  
That css() returns not the css but a computed value (possibly manipulated by a parent)  is very surprising.

dmethvin:
jQuery's `.css()` tries really hard to return the computed value because that's usually the value people want. Browsers often don't provide a good way to get other values anyway so that's what we're stuck with. It's worth discussing with the team though to see what might be done here.

HolgerJeromin:
The change is still not in the upgrade guide. IMO this should be done as fast as possible to prevent upgraders to have the same problems as we had.

dmethvin:
[...]

PR #3561, opened 2017-03-06
Dimensions: ignore transforms when retrieving width/height

timmywil:
Fixes gh-3193

### Summary ###

- I've removed our use of `getBoundingClientRect` in `getWidthOrHeight`. This is a stopgap that retrieves width/height ignoring transforms, without resorting to `offsetWidth/Height`, and without breaking existing tests.
- @gibson042 suggested further refactoring and renaming, but I'm not sure all he has in mind. My main priority is fixing the bug with transforms. I think we can refactor further in future versions. Let me know if I missed something important.

### Checklist ###

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -123,11 +123,6 @@ function getWidthOrHeight( elem, name, extra ) {
 		val = curCSS( elem, name, styles ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
-	// Fall back to uncomputed css if necessary
-	if ( val < 0 || val == null ) {
-		val = elem.style[ name ];
-	}
-
 	// Computed unit is not pixels. Stop here and return.
 	if ( rnumnonpx.test( val ) ) {
 		return val;
--- a/src/css.js
+++ b/src/css.js
@@ -117,43 +117,29 @@ function augmentWidthOrHeight( elem, name, extra, isBorderBox, styles ) {
 
 function getWidthOrHeight( elem, name, extra ) {
 
-	// Start with offset property, which is equivalent to the border-box value
-	var val,
-		valueIsBorderBox = true,
+	// Start with computed style
+	var valueIsBorderBox,
 		styles = getStyles( elem ),
+		val = curCSS( elem, name, styles ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
-	// Support: IE <=11 only
-	// Running getBoundingClientRect on a disconnected node
-	// in IE throws an error.
-	if ( elem.getClientRects().length ) {
-		val = elem.getBoundingClientRect()[ name ];
+	// Fall back to uncomputed css if necessary
+	if ( val < 0 || val == null ) {
+		val = elem.style[ name ];
 	}
 
-	// Some non-html elements return undefined for offsetWidth, so check for null/undefined
-	// svg - https://bugzilla.mozilla.org/show_bug.cgi?id=649285
-	// MathML - https://bugzilla.mozilla.org/show_bug.cgi?id=491668
-	if ( val <= 0 || val == null ) {
-
-		// Fall back to computed then uncomputed css if necessary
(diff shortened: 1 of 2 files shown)


src/css.js
```diff
@@ -117,43 +117,29 @@ function augmentWidthOrHeight( elem, name, extra, isBorderBox, styles ) {
...
+		val = curCSS( elem, name, styles ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
 
-	// Support: IE <=11 only
-	// Running getBoundingClientRect on a disconnected node
-	// in IE throws an error.
-	if ( elem.getClientRects().length ) {
-		val = elem.getBoundingClientRect()[ name ];
+	// Fall back to uncomputed css if necessary
+	if ( val < 0 || val == null ) {
```
mgol: This is dead code. `curCSS` will always get you a string so these conditions will always be `false`. `curCSS` will already get the inline style if the computed one is an empty string so this condition is already covered there.

timmywil: Isn't that a change for another ticket? This is legacy code that I left alone. I just wonder why this has been here so long if it wasn't necessary.

mgol: It seems I forgot to simplify it when I migrated the code to use `getBoundingClientRect`.

That said, this is related to this PR - in my PR I should have just changed `val <= 0 || val == null` to `val === null`as the code inside would be used in those cases, only the `val <= 0` condition was never true. In your PR this becomes true dead code and since you even had to change this condition for the new logic to work, I think this should be removed here.

timmywil: > you even had to change this condition for the new logic to work

I did? It seems to me the logic is the same.

Before:
1. Try `getBoundingClientRect`
2. If negative, zero, null, or undefined, try `getCSS`.
3. If negative, null, or undefined, get inline style.

After:
1. Try `getCSS`
2. If negative, null, or undefined, get inline style.

Maybe I'm misunderstanding something, but it seems clear to me I haven't changed that condition.

That said, you may be right about this condition being dead code previously as well. I can check that and include it in this PR, but I still believe it is a separate ticket.

timmywil: Ok, did some testing in Chrome and IE9. Yup, this looks like dead code, but I do believe it was dead code before this PR as well. I'll include it here, but as a separate commit.

mgol: The two changes you made are:
1. The condition `val <=0` was made more strict: `val < 0`.
2. By changing what's initially tried from `getBoundingClientRect` to `getCSS` you eliminated the possibility of the value not being defined.

So it wasn't dead code before, only half of the condition was dead but not what was inside the `if`.

mgol: To sum up, since by changing `getBoundingClientRect` to `getCSS` the condition inside the `if` became always `false` while it wasn't that way before, I think the removal belongs in the same commit as the initial refactor.

Not that it's terribly important, though. :)

timmywil: > The condition val <= 0 was made more strict: val < 0.

That was a different condition. The diff might make it look that way. If you look closer, you'll notice the condition with `val <= 0` was completely removed because it was checking the result from `getBoundingClientRect`, which was also removed. The `val < 0` condition was already there checking the result from `getCSS`, which we now do first.

> By changing what's initially tried from getBoundingClientRect to getCSS you eliminated the possibility of the value not being defined.

This was essentially true before. Again, this condition was already there after `getCSS`.

timmywil: But I agree this is moot now since I've removed the condition. You approve now right?

timmywil: [diff?w=1](https://github.com/jquery/jquery/pull/3561/files?w=1#diff-414c8f59bff0d1b63680b64763d8c529L140) might make my point more clear.

mgol: Ahh, yeah, I totally mixed those two, you're right! They're so close to each other and somehow I didn't notice. 😄 

LGTM then!

mgol:
This PR introduced a big regression: it broke `.width()` on inline elements. See #3571.

PR #3569, opened 2017-03-15
CSS: Drop the float mapping from cssProps

mgol:
### Summary ###

Firefox 35 and newer support `style.float` directly so we don't have to map `float` to `cssFloat` any more.

Maybe we could deprecate `jQuery.cssProps`? It's not documented and we now removed its last reason to exist from the Core point of view.

### Checklist ###

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -214,9 +214,7 @@ jQuery.extend( {
 
 	// Add in properties whose names you wish to fix before
 	// setting or getting the value
-	cssProps: {
-		"float": "cssFloat"
-	},
+	cssProps: {},
 
 	// Get and set the style property on a DOM Node
 	style: function( elem, name, value, extra ) {


timmywil:
@mgol Were you thinking you'd like to get this into 3.2?

markelog:
It was done only for firefox? Good catch then! I think we should deprecate it if that's the case

mgol:
@timmywil It doesn't matter; if it got enough LGTMs then fine but if not it may wait. It doesn't cause any observable difference for jQuery users.

> It was done only for firefox? Good catch then! I think we should deprecate it if that's the case
>  👍 1

For Firefox but in jQuery 1.x also for IE<9 as they use `styleFloat`: https://github.com/jquery/jquery/blob/1.12-stable/src/css.js#L260-L266

Issue #3573, opened 2017-03-17
$.holdReady is broken in 3.2.0


San4es:
The "$.holdReady is not a function" error occurs after the update to jQuery 3.2.0.

To reproduce the issue, see https://jsfiddle.net/4ocpkd69/ (see browser console) and compare results with jQuery 3.1.1 in https://jsfiddle.net/4ocpkd69/1/

mgol:
It was moved erroneously to `jQuery.fn.holdReady` in bd984f0ee2cf40107a669d80d92566b8625b1e6b. Marking as a blocker for 3.2.1.

mgol:
It was moved erroneously to `jQuery.fn.holdReady` in bd984f0ee2cf40107a669d80d92566b8625b1e6b. Marking as a blocker for 3.2.1.

Thanks for the report!

mgol:
This only shows we should have explicit tests for any API that we deprecate, even if there weren't any due to legacy reasons.

timmywil:
Oof. How did we not catch this.

By the way, temporary workaround:

```js
jQuery.holdReady = jQuery.fn.holdReady;
```

timmywil:
I think making the necessary adjustments to [this PR](https://github.com/jquery/jquery/pull/3331) and landing it will give us the test coverage we want for this issue.

Krinkle:
It's been failing for several weeks in Jenkins/TestSwarm:
<http://swarm.jquery.org/project/jquerymigrate>
<img width="582" alt="screen shot 2017-03-17 at 12 05 15" src="https://cloud.githubusercontent.com/assets/156867/24058883/fb8b024c-0b09-11e7-96fa-22c76035b9d2.png">
<img width="951" alt="screen shot 2017-03-17 at 12 05 30" src="https://cloud.githubusercontent.com/assets/156867/24058884/fb8bc02e-0b09-11e7-9fa7-a451a0aed209.png">

timmywil:
Well, that's jQuery migrate, not jQuery core, but I'm surprised no one said anything.

mgol:
Perhaps looking at Migrate test results should be added to the release
instructions in our  Wiki. Currently only regular Core & periodic test
results are mentioned.
-- 
Michał Gołębiowski

timmywil:
This seems like a rare case, and it will be fixed when we land the PR I mentioned. I would like to ensure that reviewing core tests is sufficient for releasing core and reviewing migrate tests sufficient for migrate.

PR #3331, opened 2016-09-25
Tests: move readywait to an iframe test

stevemao:
### Checklist

Mark an `[x]` for completed items, if you're not sure leave them unchecked and we can assist.
- [x] All authors have signed the CLA at https://contribute.jquery.com/CLA/
- [x] New tests have been added to show the fix or feature works
- [x] Grunt build and unit tests pass locally with these changes
- [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

See gh-3285

PR #3576, opened 2017-03-18
Core: fix holdReady and move readywait to an iframe test

timmywil:
Close gh-3331
Fixes gh-3573

### Summary ###
I've edited @stevemao's PR, made the recommended changes, and fixed the issue with `jQuery.holdReady`. I confirmed that the test failed before moving `holdReady` back to its proper place.

### Checklist ###

  - Steve signed the jQuery CLA, which is enough to land this PR.


timmywil:
Will land today. Get your reviews in! Thanks.

test/unit/readywait.js
```diff
@@ -0,0 +1,12 @@
+var start = new Date();
```
mgol: What's the purpose of this file? It runs the same test as one in test/unit/ready and it's not used anywhere, this looks like dead code.

timmywil: Indeed. I thought I deleted it.

timmywil: Deleted.

Issue #3571, opened 2017-03-16
Regression: width() only works when explicitly set


Krinkle:
### Description ###

Select any element on the page that doesn't have an explicit width set via a stylesheet or inline attribute and call `.width()`.

In jQuery 1.x, 3.0.x and 3.1.1 this returns the computed width.

As of jQuery 3.2.0 it returns `0`.

### Link to test case ###

<http://codepen.io/Krinkle/pen/RpjgpR>

mgol:
Ugh, it's hard to believe we didn't have a single unit test for that...

I've created a milestone for 3.2.1, this seems serious enough that we'd want a quick patch release before 3.3.0.

mgol:
Caused by #3561.

mgol:
Not having style specified via stylesheet or inline is not enough to trigger the issue; the element has to have a computed `display: inline` value.

The underlying issue is that `getComputedStyle` on inline elements returns `auto`.

mgol:
From the spec `getComputedStyle(elem)` returns an object with resolved values for various CSS properties for `elem`. The definition of the resolved value is at:
https://www.w3.org/TR/cssom-1/#resolved-value
For `width` resolved value is defined in the following way:

> If the property applies to the element or pseudo-element and the resolved value of the display property is not none, the resolved value is the used value. Otherwise the resolved value is the computed value.

The `width` property doesn't apply to inline elements so the computed value is returned instead of the used one (which it doesn't have). And the computed value will always be `auto` (again, as the property doesn't apply to inline elements).

Whether `width` is set inline or in the stylesheet on a particular element doesn't matter. The issue is that `.width()` is broken in 3.2.0 for all inline elements.

mgol:
I'm not sure what we can do to fix this issue and still keep #3193 (i.e. not take transforms into account while computing width/height) & #1724 (support for subpixel values for width/height) fixed. This issue proves `getComputedStyle` is not enough but `offsetWidth` doesn't support fractional values while `getBoundingClientRect()` takes transforms into account.

Choose your poison. :/

Krinkle:
@mgol Yeah, the problem is that the semantic purpose behind `.width()` is split between two very different concepts. On the one side the use case "computed style", with the rough expectation that setting the same value in CSS would essentially be a no-op (ignoring border-box for now). This doesn't apply to inline elements since setting a width does nothing.
```lang=html
<span>foo</span>
<span style="width: 900px;">foo</span>
```
[...]

PR #3577, opened 2017-03-18
Dimensions: fall back to offsetWidth/Height for inline elems

timmywil:
- Also added some more general dims tests for inline
  elements for posterity

Fixes gh-3571

### Checklist ###

* ~~[x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~


timmywil:
Will land today. Get your reviews in! Thanks.

src/css.js
```diff
@@ -144,6 +144,12 @@ function getWidthOrHeight( elem, name, extra ) {
 	valueIsBorderBox = isBorderBox &&
 		( support.boxSizingReliable() || val === elem.style[ name ] );
 
+	// Fall back to offsetWidth/Height when value is "auto"
+	// This happens for inline elements with no explicit setting (gh-3571)
+	if ( val === "auto" ) {
+		val = elem[ "offset" + name.toUpperCase().slice( 0, 1 ) + name.slice( 1 ) ];
```
mgol: Why not name[0].toUpperCase()? It should be shorter.

src/css/curCSS.js
```diff
@@ -31,6 +30,7 @@ function curCSS( elem, name, computed ) {
 		// This is against the CSSOM draft spec:
 		// https://drafts.csswg.org/cssom/#resolved-values
 		if ( !support.pixelMarginRight() && rnumnonpx.test( ret ) && rmargin.test( name ) ) {
+			style = elem.style;
```
mgol: What's the reason for this change? Isn't it larger?

test/unit/dimensions.js
```diff
@@ -29,21 +29,28 @@ function fn( val ) {
...
-	$div.css( "display", "" );
-	$div.width( val( -1 ) ); // handle negative numbers by setting to 0 #11604
-	assert.equal( $div.width(), 0, "Test negative width normalized to 0" );
-	$div.css( "padding", "20px" );
-	assert.equal( $div.width(), 0, "Test padding specified with pixels" );
-	$div.css( "border", "2px solid #fff" );
-	assert.equal( $div.width(), 0, "Test border specified with pixels" );
+	testElement( $div );
+	// Test span/inline element
+	testElement( jQuery( "#display" ) );
```
mgol: This is risky; why not create a span here instead of relying on a global fixture?

timmywil: It already relies on a global fixture for the div. I'm not sure the risk. If it gets removed, the test will fail.

src/css/curCSS.js
timmywil: No point in retrieving style before this block.

src/css.js
timmywil: Sure! I'm used to this for the sake of old browsers.

timmywil: Updated.

test/unit/dimensions.js
mgol: If it gets removed the test will fail but if it got changed from a span to a div, it wouldn't, would it?

mgol: If you don't want to use a span defined in place, how about at least adding an assertion that its conputed display is inline? That would eliminate the risk.

src/css.js
```diff
@@ -144,6 +144,12 @@ function getWidthOrHeight( elem, name, extra ) {
 	valueIsBorderBox = isBorderBox &&
 		( support.boxSizingReliable() || val === elem.style[ name ] );
 
+	// Fall back to offsetWidth/Height when value is "auto"
+	// This happens for inline elements with no explicit setting (gh-3571)
+	if ( val === "auto" ) {
+		val = elem[ "offset" + name[ 0 ].toUpperCase() + name.slice( 1 ) ];
```
gibson042: `elem[ jQuery.camelCase( "offset-" + name ) ]` is probably smaller, but also probably not a big deal.

test/unit/dimensions.js
timmywil: Ok, you convinced me. I will create the span inline.

Issue #3586, opened 2017-03-22
Ajax request timeout doesn't trigger fail() or always() in Safari


eriklax:
Ajax request timeout (due to server being offline) doesn't trigger fail() or always() in Safari (macOS) in latest jQuery. In jQuery 1.x timeout errors trigged both fail() and always().

"Failed to load resource: The request timed out."

And then this script stops.

```
function ping()
{
        $.get("?action=ping")
        .fail(function(jqXHR, textStatus, errorThrown) {
                console.log('fail');
        })
        .always(function() {
                console.log('always');
                setTimeout(ping, 1000);
        });
}
ping(); // and reboot the webserver
```
This shows the console in Safari.

![console](https://cloud.githubusercontent.com/assets/275758/24198647/f37a7e9c-0f06-11e7-8a3b-ac5b69e7872f.png)

dmethvin:
The default in all versions of jQuery is no timeout. Where did you set the timeout? I have a feeling that a complete example might show some other cause.

dmethvin:
Oh, but this timeout is coming from Safari? It would definitely be useful to see a running example to understand what is happening. Is it only Safari?

eriklax:
Yes, this timeout is coming from Safari (I guess some sane default xhr request timeout). It works as you would expect in Chrome (as shown in the screenshot) and Firefox.

Trying the same code with jQuery 1.x (in Safari, same browser) the timeout "Failed to load resource" is caught and triggers the callbacks, which makes me think it's jQuery related and jQuery could catch it in 3.x as well.

![chrome](https://cloud.githubusercontent.com/assets/275758/24238343/492cbf0a-0faa-11e7-943e-1aad5f8ae6c7.png)

dmethvin:
We still don't have a complete code example so we can't debug it.

eriklax:
Here is a test case. For me, it takes about 80 seconds minute until the timeout is triggered in Safari.

Doesn't work:

```
<script src="https://code.jquery.com/jquery-3.2.1.js"></script>
<script>
function ping()
{
        $.get("http://8.8.8.8") // or any other host causing a timeout
        .fail(function(jqXHR, textStatus, errorThrown) {
                console.log('fail');
        })
        .always(function() {
                console.log('always');
                setTimeout(ping, 1000);
        });
}
$(function() {
 ping();
});
</script>
```

It works with
```
<script src="https://code.jquery.com/jquery-1.12.4.js"></script>
...
```

[...]

PR #3590, opened 2017-03-23
PR for #3586 (Safari xhr timeout)

eriklax:
This PR fixed the ontimeout callback being trigged by Safari on timeout discussed in #3586.

### DIFF
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -75,7 +75,7 @@ jQuery.ajaxTransport( function( options ) {
 					return function() {
 						if ( callback ) {
 							callback = errorCallback = xhr.onload =
-								xhr.onerror = xhr.onabort = xhr.onreadystatechange = null;
+								xhr.onerror = xhr.onabort = xhr.ontimeout = xhr.onreadystatechange = null;
 
 							if ( type === "abort" ) {
 								xhr.abort();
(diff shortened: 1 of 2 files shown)


test/unit/ajax.js
```diff
@@ -527,6 +527,23 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - timeout", 2, function( assert ) {
```
mgol: `timeout` -> `native timeout`, similarly to how the `native abort` test case is named.

src/ajax/xhr.js
```diff
@@ -75,7 +75,8 @@ jQuery.ajaxTransport( function( options ) {
 					return function() {
 						if ( callback ) {
 							callback = errorCallback = xhr.onload =
-								xhr.onerror = xhr.onabort = xhr.onreadystatechange = null;
+								xhr.onerror = xhr.onabort = xhr.ontimeout =
+								xhr.onreadystatechange = null;
```
mgol: This should be indented by an additional tab.

mgol:
A few nits but otherwise looks good to me!

src/ajax/xhr.js
eriklax: @mgol just the last line?

PR #3597, opened 2017-03-31
Deferred: Use typeof instead of isFunction

### LINKED ISSUE #3596: Deferred: behavior different from ECMAScript Promise for objects with [[Call]]
### Description ###
jQuery's `Deferred` is expected to be a close approximation to `Promise` of _ECMAScript 2016 Language Specification_. `Promise` defined in ECMAScript 2016 tests `onFulfilled`, onRejected and then are objects with `[[Call]]`.
However, `isFunction`, used in the `Deferred`, tests `toStringTag`, which is irrelevant from the definition. As the result, `Deferred` shows different behaviors for some objects such as `GeneratorFunction`.

### Link to test case ###
https://jsbin.com/civabuxabe/1/edit?js,console
Expected `[object Generator] { ... }`, but results in `0`.
```JavaScript
jQuery.Deferred()
  .resolve(0)
  .then(function *() {})
  .then(console.log);
```

akihikodaki:
This is a breaking change.

### Summary ###
This fixes #3596 by replacing `isFunction` with `typeof` operator, which will test the given value is an objects with `[[Call]]`.

### Checklist ###

* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com (not needed)

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -23,7 +23,7 @@ function adoptValue( value, resolve, reject, noValue ) {
 			method.call( value ).done( resolve ).fail( reject );
 
 		// Other thenables
-		} else if ( value && jQuery.isFunction( ( method = value.then ) ) ) {
+		} else if ( value && typeof ( method = value.then ) === "function" ) {
 			method.call( value, resolve, reject );
 
 		// Other non-thenables
@@ -142,7 +142,7 @@ jQuery.extend( {
 										returned.then;
 
 									// Handle a returned thenable
-									if ( jQuery.isFunction( then ) ) {
+									if ( typeof then === "function" ) {
 
 										// Special processors (notify) just wait for resolution
 										if ( special ) {
@@ -238,7 +238,7 @@ jQuery.extend( {
 							resolve(
 								0,
 								newDefer,
-								jQuery.isFunction( onProgress ) ?
+								typeof onProgress === "function" ?
 									onProgress :
 									Identity,
 								newDefer.notifyWith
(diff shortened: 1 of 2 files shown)


gibson042:
You have identified a flaw in `jQuery.isFunction`, but selectively bypassing it in one module is not an acceptable solution. However, if you wish to update `jQuery.isFunction` to recognize generators and async functions, we'd probably consider that more favorably. And finally, note that breaking changes are only allowed in major releases and we're not yet ready for 4.0, but we have not explicitly documented behavior for non-traditional functions, so most changes in this area are unlikely to be breaking.

akihikodaki:
I think `jQuery.isFunction` is expected to work as it does.
The documentation says:
> Determine if the argument passed is a JavaScript function object.

https://github.com/jquery/api.jquery.com/blob/d622e1244802cd55670265698f0d66383e3dedaa/entries/jQuery.isFunction.xml#L10
If _JavaScript function object_ refers to _ECMAScript function_ in _ECMAScript 2016 Language Specification_, `jQuery.isFunction` is more approximate than `typeof` operator, though it is not perfect.

However, the distinction between ECMAScript functions and other functions is irrelevant from the definition of Promise. Using `jQuery.isFunction` instead of `typeof` operator is simply a mistake in Deferred when it comes to the conformance to the standard.

Actually I wonder whether the current specification of `jQuery.isFunction` is preferred or not. I even consider it should be deprecated today, when such a distinction doesn't make sense. Yet it is a different problem from the standard conformance in Promise and should have another issue.

akihikodaki:
> And finally, note that breaking changes are only allowed in major releases and we're not yet ready for 4.0, but we have not explicitly documented behavior for non-traditional functions, so most changes in this area are unlikely to be breaking.

If it is not considered as breaking, it is ready to merge.

gibson042:
Closing because we don't want Deferred-specific function detection.

PR #3601, opened 2017-04-01
Core: use instanceof and typeof instead of jQuery.type in jQuery.isFunction

akihikodaki:
### Summary ###
This change allows to detect functions with exotic `@@toStringTag`.

### Checklist ###
* [X] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com (https://github.com/jquery/api.jquery.com/issues/1034)

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -212,7 +212,7 @@ jQuery.extend( {
 	noop: function() {},
 
 	isFunction: function( obj ) {
-		return jQuery.type( obj ) === "function";
+		return obj instanceof Function && typeof obj === "function";
 	},
 
 	isWindow: function( obj ) {
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -406,9 +406,9 @@ QUnit[ "assign" in Object ? "test" : "skip" ]( "isPlainObject(Object.assign(...)
 
 
 QUnit.test( "isFunction", function( assert ) {
-	assert.expect( 19 );
+	assert.expect( 20 );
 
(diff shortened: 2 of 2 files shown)


src/core.js
```diff
@@ -212,7 +212,7 @@ jQuery.extend( {
 	noop: function() {},
 
 	isFunction: function( obj ) {
-		return jQuery.type( obj ) === "function";
+		return obj instanceof Function && typeof obj === "function";
```
gibson042: `instanceof` goes too far; it would falsely reject functions from other realms (e.g., iframes) and probably also some host-object functions. If we pursue my suggestion from https://github.com/jquery/jquery/issues/3596#issuecomment-290899122 , it should look like `typeof obj === "function" && !obj.nodeType` (`nodeType` being our standard test for DOM nodes, cf. https://contribute.jquery.org/style-guide/js/#type-checks ).

akihikodaki: Is that really make sense to regard it as an exception? I wonder other exceptions could appear.
The documentation says:
> Determine if the argument passed is a JavaScript function object.

Is that behavior appropriate for the description? It is no longer valid to call the implementation as `jQuery.isFunction` if it is too far from the existing one. It is just confusing.

gibson042: What would be confusing is for `jQuery.isFunction( elem )` to ever return true.

akihikodaki: I agree. That cannot be fixed.

akihikodaki:
Closing; this implementation is still too confusing.

PR #3604, opened 2017-04-02
Hide jQuery.camelCase

niltoncsr:
Here, I try to resolve this issue by moving the exposed method `jQuery.camelCase()` into a internal module `camelCase` located in `src/core/camelCase.js`, as well as importing it wherever it's been used.

### DIFF
--- a/test/unit/data.js
+++ b/test/unit/data.js
@@ -602,7 +602,7 @@ QUnit.test( ".data should not miss attr() set data-* with hyphenated property na
 } );
 
 QUnit.test( ".data always sets data with the camelCased key (gh-2257)", function( assert ) {
-	assert.expect( 18 );
+	assert.expect( 9 );
 
 	var div = jQuery( "<div>" ).appendTo( "#qunit-fixture" ),
 		datas = {
@@ -625,7 +625,6 @@ QUnit.test( ".data always sets data with the camelCased key (gh-2257)", function
 		div.data( key, val );
 		var allData = div.data();
 		assert.equal( allData[ key ], undefined, ".data does not store with hyphenated keys" );
-		assert.equal( allData[ jQuery.camelCase( key ) ], val, ".data stores the camelCased key" );
 	} );
 } );
 
@@ -639,7 +638,7 @@ QUnit.test( ".data should not strip more than one hyphen when camelCasing (gh-20
 	assert.equal( allData[ "nested--Triple" ], "triple", "Key with triple hyphens is correctly camelCased" );
 } );
 
-QUnit.test( ".data supports interoperable hyphenated/camelCase get/set of properties with arbitrary non-null|NaN|undefined values", function( assert ) {
+QUnit.test( ".data supports interoperable hyphenated get/set of properties with arbitrary non-null|NaN|undefined values", function( assert ) {
 
 	var div = jQuery( "<div/>", { id: "hyphened" } ).appendTo( "#qunit-fixture" ),
 		datas = {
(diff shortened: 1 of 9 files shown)


niltoncsr:
I need some help with the test cases.
What is the best practice, in this scenario, to import an AMD module into the unit test? After code review, I'll commit the unit tests too.

mgol:
@niltoncsr We don't use those tiny AMD modules in tests. In this case, there are two things to do:
1. `jQuery.camelCase` tests should be moved to `test/unit/deprecated.js`.
2. Any tests that don't test `jQuery.camelCase` but use it inside of the test should be changed to not rely on that (if there are any such tests, I haven't checked).

Let me know if you have any further questions. And thanks for the PR!

niltoncsr:
Thanks for the feedback, @mgol! :smile: 

I got it, I have to remove all tests that use `jQuery.camelCase`, but I still have doubts on how to test the **new module**: its importing and usage. What are your thoughts on that? Where can I proceed for testing this module?

mgol:
> I have to remove all tests that use `jQuery.camelCase`

Not remove but move. :)

> but I still have doubts on how to test the new module: its importing and usage

You don't have to test the `camelCase` module. Since it's now internal and only temporarily exposed as a deprecated `jQuery.camelCase`, all we care (with the exception of temporarily testing `jQuery.camelCase` in `test/unit/deprecated.js`) is that public API output is what we expect it to be.

If we broke `jQuery.camelCase`, many tests in the `data` & `css` modules should fail. It'd be a good idea to try that - e.g. change the function to not camelCase and see if tests are failing in `css` & `data` modules. If not, some tests have to be added for them. You can try doing that to be sure.

niltoncsr:
I made that now, @mgol. Moved the single unit test to `test/unit/deprecated.js` (while there it still doesn't pass) and removed any calls to `jQuery.camelCase`, adapting the unit tests' labels and expectations asserts.
Is there anything I missed?

Moving the conversation, is there any intention or planning to test an internal usage module? Despite its a internal module, like in this case, it should be test since other features and methods rely on it. I say that we must something about it. What are the your thoughts on that subject?

test/unit/data.js
```diff
@@ -625,7 +625,6 @@ QUnit.test( ".data always sets data with the camelCased key (gh-2257)", function
 		div.data( key, val );
 		var allData = div.data();
 		assert.equal( allData[ key ], undefined, ".data does not store with hyphenated keys" );
-		assert.equal( allData[ jQuery.camelCase( key ) ], val, ".data stores the camelCased key" );
```
mgol: This assertion is very important and shouldn't be removed. Instead, I'd expect the `datas` object to be modified to contain the expected camelCased value, sth like:
```js
datas = {
	"non-empty": {camelCasedKey: "nonEmpty", value: "a string"},
	...
};
```

The same applies to other places where you removed the assertions using `jQuery.camelCase`. It's not the assertions that need removal but `jQuery.camelCase` usage.

mgol:
> Moved the single unit test to test/unit/deprecated.js (while there it still doesn't pass)

What do you mean by "while there it still doesn't pass"?

> Moving the conversation, is there any intention or planning to test an internal usage module? Despite its a internal module, like in this case, it should be test since other features and methods rely on it.

Not necessarily. While testing internal APIs is a valid approach, it's not the only one. This is a private module, just for jQuery internal purposes. Therefore, if any changes land to it that change its behavior but that don't affect any public API in any visible way, that doesn't matter. We should make sure all public methods are tested well enough that testing internal modules is not necessary.

mgol:
One more thing - we're only deprecating the method for now, not removing it. Therefore, it should be defined in `src/deprecated.js` as an alias to the internal `camelCase`. Once you do that, all tests should pass.

niltoncsr:
Oh, now I got what you mean :smile: 
So, creating an alias on `src/deprecated.js` removes the feature from `core` but still expose it, as deprecated.

Just one more question:
Should I still change the `datas` object to make the reintroduced assertions pass or since

> Once you do that, all tests should pass.

should I just rollback them as they were before? I think the first one, change the `datas`, it's correct for future purposes of removal. Am I right?

Thanks for your patience, I'm still getting into jQuery's internals. :sweat_smile:

mgol:
> should I just rollback them as they were before?

If you did that all the tests would pass but when deprecating thing we try to avoid any mentions of them outside of `src/deprecated.js` and `test/unit/deprecated.js`. Thanks to that approach, once we decide on removal we simply remove the API from `src/deprecated.js` and its tests from `test/unit/deprecated.js` and we're done. So the proper solution is neither to remove those assertions as currently done in this PR nor to keep using the deprecated API there but to reformulate the assertions in a way that they check what they're supposed to check without using deprecated APIs.

Something like what I proposed in https://github.com/jquery/jquery/pull/3604#discussion_r109353200 should work.

Let me know if anything is still unclear. Thanks!

niltoncsr:
Thanks a lot for the explanation.
I'll make it as you suggested :smiley:

mgol:
@niltoncsr Are you still interested in finishing this PR?

timmywil:
Closing in favor of https://github.com/jquery/jquery/pull/3874. Thank you for your contribution.

Issue #3600, opened 2017-04-01
Core: jQuery.isFunction returns false for Function objects with custom @@toStringTag


akihikodaki:
### Description ###
`jQuery.isFunction` returns false for Function objects with custom `@@toStringTag`. That is undesirable in some case such as `GeneratorFunction`.
This flaw was pointed out at https://github.com/jquery/jquery/pull/3597#issuecomment-290733182.

### Link to test case ###
https://jsbin.com/xocanozuye/1/edit?html,js,console

```JavaScript
console.log(jQuery.isFunction(function *(){}));
```

Krinkle:
@gibson042 From a quick check seems like patching `$.type()` to align more with `typeof` should otherwise be harmless. We've already been going in that direction anyways.

As of 3.2, we already do this for most cases. We now only use `class2type[toString]` for values where `typeof` returns `function` or `object`.

If we remove `function` from that, reducing use of `class2type` to `object` only, should do the trick.

Although we may need to find another way to support Android 2.3's RegExp values in that case.
```js
type: function( obj ) {
	if ( obj == null ) {
		return obj + "";
	}

	// Support: Android <=2.3 only (functionish RegExp)
	return typeof obj === "object" || typeof obj === "function" ?
		class2type[ toString.call( obj ) ] || "object" :
		typeof obj;
},
isFunction: function( obj ) {
	return jQuery.type( obj ) === "function";
},
```

mgol:
We don't support Android 2.3 anymore, 4.0 is our minimum.
-- 
Michał Gołębiowski

akihikodaki:
That breaks for some cases such as `document.createElement( "object" )`. It is/was actually a function (callable object), but `jQuery.isFunction( document.createElement( "object" ) )` is expected to be false.

```JavaScript
	obj = document.createElement( "object" );

	// Firefox says this is a function
assert.ok( !jQuery.isFunction( obj ), "Object Element" );
```

https://github.com/jquery/jquery/blob/ac9e3016645078e1e42120822cfb2076151c8cbe/test/unit/core.js#L442

On the other hand, the documentation says:
> Determine the internal JavaScript [[Class]] of an object.

https://github.com/jquery/api.jquery.com/blob/eceefcf142534cb51ae443157a4752c9357f3ebd/entries/jQuery.type.xml#L10

[...]

PR #3617, opened 2017-04-08
Core: Update isFunction to handle unusual-@@toStringTag input

gibson042:
### Summary ###
Generators and async functions are still functions (cf. [ECMAScript Language Specification](http://www.ecma-international.org/ecma-262/7.0/#sec-generatorfunction-objects)), and we can recognize them with `typeof` (but still need a guard to reject elements).

Fixes gh-3600
Fixes gh-3596

### Checklist ###

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -214,7 +214,9 @@ jQuery.extend( {
 	isFunction: function( obj ) {
 
 		// Support: Chrome <=57, Firefox <=52
-		// Don't classify callable <object> elements as functions
+		// In some browsers, typeof returns "function" for HTML <object> elements
+		// (i.e., `typeof document.createElement( "object" ) === "function"`).
+		// We don't want to classify *any* DOM node as a function.
 		return typeof obj === "function" && typeof obj.nodeType !== "number";
 	},
 
--- a/src/core.js
+++ b/src/core.js
@@ -212,7 +212,10 @@ jQuery.extend( {
 	noop: function() {},
 
 	isFunction: function( obj ) {
-		return obj instanceof Function && typeof obj === "function";
+
+		// Support: Chrome <=57, Firefox <=52
+		// Don't classify callable <object>s as functions
+		return typeof obj === "function" && typeof obj.nodeType !== "number";
 	},
 
 	isWindow: function( obj ) {
--- a/src/core.js
+++ b/src/core.js
@@ -214,7 +214,7 @@ jQuery.extend( {
 	isFunction: function( obj ) {
 
 		// Support: Chrome <=57, Firefox <=52
-		// Don't classify callable <object>s as functions
+		// Don't classify callable <object> elements as functions
 		return typeof obj === "function" && typeof obj.nodeType !== "number";
 	},
 
--- a/src/core.js
+++ b/src/core.js
@@ -467,7 +467,7 @@ function isArrayLike( obj ) {
 	var length = !!obj && "length" in obj && obj.length,
 		type = jQuery.type( obj );
 
-	if ( type === "function" || jQuery.isWindow( obj ) ) {
+	if ( jQuery.isFunction( obj ) || jQuery.isWindow( obj ) ) {
 		return false;
 	}
 
--- a/src/core.js
+++ b/src/core.js
@@ -212,7 +212,7 @@ jQuery.extend( {
 	noop: function() {},
 
 	isFunction: function( obj ) {
-		return jQuery.type( obj ) === "function";
+		return obj instanceof Function && typeof obj === "function";
 	},
 
 	isWindow: function( obj ) {
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -406,9 +406,10 @@ QUnit[ "assign" in Object ? "test" : "skip" ]( "isPlainObject(Object.assign(...)
 
 
 QUnit.test( "isFunction", function( assert ) {
-	assert.expect( 20 );
+	assert.expect( 21 );
 
-	var mystr, myarr, myfunction, fn, inheriting, obj, nodes, first, input, a;
+	var mystr, myarr, myfunction, fn, inheriting, obj, nodes, first, input, a, iframe, doc,
+		done = assert.async();
 
 	// Make sure that false values return false
 	assert.ok( !jQuery.isFunction(), "No Value" );
@@ -493,10 +494,54 @@ QUnit.test( "isFunction", function( assert ) {
 		callme( function() {} );
 	} );
 
-	function fnExoticToStringTag() {}
-	fnExoticToStringTag[ Symbol.toStringTag ] = "foo";
-	assert.ok( jQuery.isFunction( fnExoticToStringTag ), "function with exotic @@toStringTag" );
+	// Functions from other windows should be matched
+	Globals.register( "iframeDone" );
+	window.iframeDone = function( fn, detail ) {
+		window.iframeDone = undefined;
+		assert.ok( jQuery.isFunction( fn ), "cross-realm function" +
+			( detail ? " - " + detail : "" ) );
+		done();
+	};
+
+	try {
+		iframe = jQuery( "#qunit-fixture" )[ 0 ].appendChild( document.createElement( "iframe" ) );
+		doc = iframe.contentDocument || iframe.contentWindow.document;
+		doc.open();
+		doc.write( "<body onload='window.parent.iframeDone( function() {} );'>" );
+		doc.close();
+	} catch ( e ) {
+		window.iframeDone( function() {}, "iframes not supported" );
+	}
 } );
+
+( function() {
+	var generator;
+	try {
+		generator = Function( "return function* () {}" )();
+	} catch ( e ) {}
+
+	QUnit[ generator ? "test" : "skip" ]( "isFunction(standard Function subclass)",
+		function( assert ) {
+			assert.expect( 1 );
+
+			assert.equal( jQuery.isFunction( generator ), true, "GeneratorFunction" );
+		}
+	);
+} )();
+
+QUnit[ typeof Symbol === "function" ? "test" : "skip" ]( "isFunction(custom @@toStringTag)",
+	function( assert ) {
+		assert.expect( 2 );
+
+		var obj = {},
+			fn = function() {};
+		obj[ Symbol.toStringTag ] = "Function";
+		fn[ Symbol.toStringTag ] = "Object";
+
+		assert.equal( jQuery.isFunction( obj ), false, "function-mimicking object" );
+		assert.equal( jQuery.isFunction( fn ), true, "object-mimicking function" );
+	}
+);
 
 QUnit.test( "isNumeric", function( assert ) {
 	assert.expect( 43 );
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -439,11 +439,11 @@ QUnit.test( "isFunction", function( assert ) {
 	fn = function() {};
 	assert.ok( jQuery.isFunction( fn ), "Normal Function" );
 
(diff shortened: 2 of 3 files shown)


test/unit/core.js
```diff
@@ -489,8 +493,56 @@ QUnit.test( "isFunction", function( assert ) {
 	callme( function() {
 		callme( function() {} );
 	} );
+
+	// Functions from other windows should be matched
+	Globals.register( "iframeDone" );
```
markelog: Can we move this to separate test as well?

test/unit/deferred.js
```diff
@@ -537,6 +539,14 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
 		assert.ok( true, "errors in .done callbacks don't stop .then handlers" );
 	} );
 
+	function faker() {
```
markelog: Separate test?

test/unit/deferred.js
```diff
@@ -861,7 +871,19 @@ QUnit.test( "jQuery.when(nonThenable) - like Promise.resolve", function( assert
 QUnit.test( "jQuery.when(thenable) - like Promise.resolve", function( assert ) {
 	"use strict";
 
-	var CASES = 16,
+	function customToStringThen() {
```
markelog: Separate test?

src/core.js
```diff
@@ -212,7 +212,10 @@ jQuery.extend( {
 	noop: function() {},
 
 	isFunction: function( obj ) {
-		return jQuery.type( obj ) === "function";
+
+		// Support: Chrome <=57, Firefox <=52
+		// Don't classify callable <object>s as functions
+		return typeof obj === "function" && typeof obj.nodeType !== "number";
```
markelog: `typeof obj.nodeType !== "number"`, what is this? Which test is related to this change?

test/unit/core.js
gibson042: I guess, but it seems to fit with the rest of these assertions pretty well. Would anyone else like to weigh in?

test/unit/core.js
```diff
@@ -489,8 +493,56 @@ QUnit.test( "isFunction", function( assert ) {
...
+	};
+
+	try {
+		iframe = jQuery( "#qunit-fixture" )[ 0 ].appendChild( document.createElement( "iframe" ) );
+		doc = iframe.contentDocument || iframe.contentWindow.document;
+		doc.open();
+		doc.write( "<body onload='window.parent.iframeDone( function() {} );'>" );
+		doc.close();
+	} catch ( e ) {
+		window.iframeDone( function() {}, "iframes not supported" );
```
gibson042: The block was patterned off a [similar one for isPlainObject](https://github.com/jquery/jquery/blob/ac9e3016645078e1e42120822cfb2076151c8cbe/test/unit/core.js#L363), and is [quite old](https://github.com/jquery/jquery/commit/396dd2127330a7ed742d1e9092af54b668c46e85#diff-023537a1eb6e158db4d12a1fe86f1128R262) as it turns out. I'll remove the try/catch.

test/unit/deferred.js
gibson042: Why?

test/unit/deferred.js
gibson042: I'd really rather not; this test runs four assertions per input thenable.

src/core.js
gibson042: It's [preexisting](https://github.com/jquery/jquery/blob/ac9e3016645078e1e42120822cfb2076151c8cbe/test/unit/core.js#L444). This `nodeType` guard protects us against what would would otherwise be a regression caused by changing from `jQuery.type` to `typeof`.

markelog:
Can we augment logic of `jQuery.type` instead? like `jQuery.type(function *() {}) === 'object'` right now.

markelog:
On tests perspective: 

In my opinion: one feature – one test, one regression – one test for regression (i.e. for bug), preferably with issue number in test description. 

The simpler the better, microservices or docker containers analogies come to mind since otherwise It's really hard to support monolith and god methods and potential merge conflicts if/when we would have to revert whereas those constructions were already changed. For the same reason we no longer put HTML fixtures inside the test suite, but create them inside the test.

I think we already discuss this many times?

One issue, one pull request, independent tests, one atomic change, one commit.

timmywil:
I don't think it is necessary that pull requests have one commit, but I agree with atomicity of commits. As for this particular case, I wouldn't mind rebasing all 4 of these commits into one as they all seem to address the same problem. Still, this is less important than the "how". How we address the problem is always the bigger question, and I'm wondering if we shouldn't align with Lodash here and treat generator functions as functions as well.

gibson042:
`jQuery.type` is so vague and generic that I don't want to touch it... we only use it internally for distinguishing number/string/function/object, and even that is potentially problematic as the list of special-cased objects grows... `jQuery.isFunction` is much more clear.

As for testing, I really like the efficiency of [table-driven tests](https://github.com/golang/go/wiki/TableDrivenTests), and want to see us move towards them where practical, rather than away. And merge conflicts wouldn't be an issue if we set expectations _from_ the list of cases.

gibson042:
Tests updated. This should be good to go.

src/core.js
```diff
@@ -212,7 +212,10 @@ jQuery.extend( {
 	noop: function() {},
 
 	isFunction: function( obj ) {
-		return jQuery.type( obj ) === "function";
+
+		// Support: Chrome <=57, Firefox <=52
```
mgol: Should we add `only` to both or not yet because Chrome 58 & Firefox 53 are not stable yet?

test/unit/deferred.js
```diff
@@ -542,6 +543,26 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
...
+
+		assert.expect( 1 );
+
+		var done = assert.async(),
+			defer = jQuery.Deferred();
+
+		function faker() {
+			assert.ok( true, "handler with non-'Function' @@toStringTag gets invoked" );
+		}
+		faker[ typeof Symbol === "function" && Symbol.toStringTag ] = "String";
```
mgol: Why not wrap the statement with `if`? In the current way we'll be writing to a `"false"` property in non-supporting browsers which may create confusion.

These are tests so even size is not important so I don't see the gains of the current form.

test/unit/deferred.js
```diff
@@ -861,15 +882,26 @@ QUnit.test( "jQuery.when(nonThenable) - like Promise.resolve", function( assert
...
 
-	var CASES = 16,
-		slice = [].slice,
+	function customToStringThen() {
+		var promise = {
+			then: function( onFulfilled ) {
+				onFulfilled();
+			}
+		};
+		promise.then[ typeof Symbol === "function" && Symbol.toStringTag ] = "String";
```
mgol: Similarly, I'd prefer an `if` instead of writing to a `"false"` property.

src/core.js
markelog: Would you mind adding this clarification comment there, right now it looks so bizarre :)

src/core.js
gibson042: Not yet; we can update this after official releases if anyone cares.

test/unit/deferred.js
gibson042: Fair enough; updated.

test/unit/deferred.js
```diff
@@ -861,15 +882,24 @@ QUnit.test( "jQuery.when(nonThenable) - like Promise.resolve", function( assert
...
 	"use strict";
 
-	var CASES = 16,
-		slice = [].slice,
+	var customToStringThen = {
+		then: function( onFulfilled ) {
+			onFulfilled();
+		}
+	};
+	if ( typeof Symbol === "function" ) {
```
mgol: This check is not enough, many browsers (e.g. old Firefox, Edge <15) support many symbols but not `@@toStringTag`.

This applies to here & a few other places.

gibson042: I don't think we care about those cases because they coincide with impossibility of setting @@toStringTag, but I'll see what I can do anyway.

gibson042: I decided to keep this one for its harmless property set, but the others are updated to skip tests unless `Symbol.toStringTag` is truthy.

src/core.js
markelog: ?

src/core.js
```diff
@@ -464,7 +467,7 @@ function isArrayLike( obj ) {
 	var length = !!obj && "length" in obj && obj.length,
 		type = jQuery.type( obj );
 
-	if ( type === "function" || jQuery.isWindow( obj ) ) {
+	if ( jQuery.isFunction( obj ) || jQuery.isWindow( obj ) ) {
```
gibson042: Doesn't look like it: https://github.com/jquery/jquery/issues?q=is%3Aissue%20isWindow

src/core.js
gibson042: It reads as `Don't classify callable <object> elements as functions` now; what would clarify further?

markelog: If `<object>` is referenced to `typeof` result for objects, how it could bypass out initial `typeof obj === "function"`?

Then `function` type is a subtype of `object` so why we can't classify function object as function?
Most of all, why we check presence of `nodeType` property to see if `object` is callable and why do we care?

"elements" part provides some clarification, but only if you know that this is the reference to the DOM elements. And even so why "element"? Why not "node"? If the logic is meant only for elements indeed then how come it cannot reach this part for other nodes? I think it also deserves some explanation as well as some examples of such an input.

I think I know answers to these questions, but after some months I might forget, you might forget, and others who going to read this might not know of course

src/core.js
markelog: Created – https://github.com/jquery/jquery/issues/3629 :)

src/core.js
gibson042: "Don't classify callable \<object> elements as functions` is perfectly clear to me. Can you just tell me what you want to see here?

src/core.js
```diff
@@ -212,7 +212,10 @@ jQuery.extend( {
 	noop: function() {},
 
 	isFunction: function( obj ) {
-		return jQuery.type( obj ) === "function";
+
+		// Support: Chrome <=57, Firefox <=52
+		// Don't classify callable <object> elements as functions
```
gibson042: @markelog ping... what do you want instead of this text?

test/unit/core.js
```diff
@@ -439,6 +439,8 @@ QUnit.test( "isFunction", function( assert ) {
 	fn = function() {};
 	assert.ok( jQuery.isFunction( fn ), "Normal Function" );
 
+	assert.ok( !jQuery.isFunction( Object.create( fn ) ), "custom Function subclass" );
+
 	obj = document.createElement( "object" );
 
 	// Firefox says this is a function
```
markelog: Chrome does the same, correct?

src/core.js
markelog: Sorry, long week. Some in the lines of
```js
// Browsers mentioned above will erroneously equate `typeof` 
// call on some of the DOM nodes to "function":
//
// jQuery.isFunction(document.createElement("object")) === true
//
// but not
//
// jQuery.isFunction(document.createElement("div")) !== true
//
// To circumvent it we check presense of correct `nodeType` property on
// provided argument, which all DOM nodes shoud have
```

test/unit/core.js
```diff
@@ -439,6 +439,8 @@ QUnit.test( "isFunction", function( assert ) {
 	fn = function() {};
 	assert.ok( jQuery.isFunction( fn ), "Normal Function" );
 
+	assert.ok( !jQuery.isFunction( Object.create( fn ) ), "custom Function subclass" );
```
markelog: We can use `notOk` here instead? It seems more explicit that way, nitpick of course

PR #3657, opened 2017-05-04
Deferred: fix memory leak of promise callbacks

### LINKED ISSUE #3606: a memory problem  about deferred in 3.x
### Description ###

When our project upgraded to jQuery3.2.1 (from 2.2.4),  we found a memory leak about `Deferred` in jQuery3.x .  the  `promise.then` method  does not release the handler & handler Closure . 
it work well in jQuery2.x and native Promise in chrome .

look at the following case：

### Link to test case ###

http://plnkr.co/edit/DpMkLV2mqzwlX4Ruuk7i

image

image

jbedard:
Fixes #3606

I'm not sure why we need both [callbacks and .then handlers](https://github.com/jquery/jquery/blob/00e0c7af79cf27b5c52f1f2dcb6770fd26d5119c/src/deferred.js#L56-L61). Could we instead merge those `Callback`s? EDIT: I think it's because error handling works different?

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -301,8 +301,15 @@ jQuery.extend( {
 					// fulfilled_callbacks.disable
 					tuples[ 3 - i ][ 2 ].disable,
 
+					// rejected_handlers.disable
+					// fulfilled_handlers.disable
+					tuples[ 3 - i ][ 3 ].disable,
+
 					// progress_callbacks.lock
-					tuples[ 0 ][ 2 ].lock
+					tuples[ 0 ][ 2 ].lock,
+
+					// progress_handlers.lock
+					tuples[ 0 ][ 3 ].lock
 				);
 			}
 


markelog:
No tests?

jbedard:
I wasn't really sure how this could be unit tested. Any ideas? With everything hidden in closures it's a little hard to do things like spy on the callback methods...

src/deferred.js
```diff
@@ -304,6 +304,16 @@ jQuery.extend( {
 					// progress_callbacks.lock
 					tuples[ 0 ][ 2 ].lock
 				);
+
+				tuple[ 3 ].add(
```
gibson042: This doesn't look right to me; I think we should do everything in the existing handler (i.e., it should disable both `done` callbacks _and_ `then` callbacks, and lock both `done` _and_ then progress notifications). We could probably also stand to improve the comments, since "callbacks" is ambiguous in light of `done` vs. `then`.

gibson042:
> I'm not sure why we need both [callbacks and .then handlers](https://github.com/jquery/jquery/blob/00e0c7af79cf27b5c52f1f2dcb6770fd26d5119c/src/deferred.js#L56-L61). Could we instead merge those `Callback`s? EDIT: I think it's because error handling works different?

Yes, among other reasons. `.done` callbacks fire synchronously and can lock up, but the first thing they do is kick off _asynchronous_ `.then` callbacks (which can't).

src/deferred.js
jbedard: You mean adding it to the `list.add` above? I'll try it out... I don't recall why I did it here :/

jbedard:
@gibson042 I've moved everything into the existing `list.add` call.

jbedard:
Interesting... I amended the commit just to add the PR number to the commit msg and one test started failing. Now I ran it again and everything passed.  Anyone ever see something similar before? I can't seem to reproduce it locally.

Here's the travis job that failed: https://travis-ci.org/jquery/jquery/jobs/242269140

The test failure:
```
  1) 2.2.6: `then` may be called multiple times on the same promise. 2.2.6.2: If/when `promise` is rejected, all respective `onRejected` callbacks must execute in the order of their originating calls to `then`. `onRejected` handlers are called in the original order even when one handler is added inside another handler immediately-rejected:
     Uncaught AssertError: expected handler1, handler2, handler3 to be called in order but were called as handler1, handler2
      at Object.fail (node_modules/sinon/lib/sinon/assert.js:90:29)
      at failAssertion (node_modules/sinon/lib/sinon/assert.js:51:24)
      at Object.assertCallOrder [as callOrder] (node_modules/sinon/lib/sinon/assert.js:118:21)
      at null._onTimeout (node_modules/promises-aplus-tests/lib/tests/2.2.6.js:249:42)
```

PR #3659, opened 2017-05-05
Core: Support $(DOMElement, context)

avajayam:
Modify init function to support find DOMElement in context


dmethvin:
@avajayam From the standpoint of API usability and clarity, do you think this improves the API? I'm already not a fan of `context` since it can be done with `.find()`.

avajayam:
@dmethvin You are right, expect other content support in init function to be removed for consistency.

PR #3661, opened 2017-05-06
Docs:Tests: Update IE/Edge-related support comments & tests

mgol:
### Summary ###

This PR updates IE/Edge-related support comments (to e.g. update their status for Edge 15) and tweaks some tests that had to be relaxed in the past - most notably smaller precision in some dimensions tests.

### Checklist ###

*  ~~New tests have been added to show the fix or feature works~~

### DIFF
--- a/test/unit/css.js
+++ b/test/unit/css.js
@@ -1323,11 +1323,11 @@ QUnit.test( "Keep the last style if the new one isn't recognized by the browser 
 	assert.equal( el.css( "position" ), "absolute", "The old style is kept when setting an unrecognized value" );
 	el = jQuery( "<div></div>" ).css( "position", "absolute" ).css( "position", " " );
 
-	// Support: Edge 14
+	// Support: Edge 14-15
 	// Edge collapses whitespace-only values when setting a style property and
 	// there is no easy way for us to work around it. Just skip the test there
 	// and hope for the better future.
-	if ( /edge\//i.test( navigator.userAgent ) ) {
+	if ( /\bedge\//i.test( navigator.userAgent ) ) {
 		assert.ok( true, "Skipped (Edge 14 handles whitespace-only values incorrectly)" );
 	} else {
 		assert.equal( el.css( "position" ), "absolute", "The old style is kept when setting to a space" );
--- a/test/unit/css.js
+++ b/test/unit/css.js
@@ -1316,22 +1316,23 @@ QUnit[ jQuery.find.compile ? "test" : "skip" ]( ":visible/:hidden selectors", fu
 } );
 
 QUnit.test( "Keep the last style if the new one isn't recognized by the browser (#14836)", function( assert ) {
-	assert.expect( 2 );
+	assert.expect( 1 );
 
-	var el;
-	el = jQuery( "<div></div>" ).css( "position", "absolute" ).css( "position", "fake value" );
+	var el = jQuery( "<div></div>" ).css( "position", "absolute" ).css( "position", "fake value" );
 	assert.equal( el.css( "position" ), "absolute", "The old style is kept when setting an unrecognized value" );
-	el = jQuery( "<div></div>" ).css( "position", "absolute" ).css( "position", " " );
+} );
 
-	// Support: Edge 14-15
-	// Edge collapses whitespace-only values when setting a style property and
-	// there is no easy way for us to work around it. Just skip the test there
-	// and hope for the better future.
-	if ( /\bedge\//i.test( navigator.userAgent ) ) {
-		assert.ok( true, "Skipped (Edge 14 handles whitespace-only values incorrectly)" );
-	} else {
-		assert.equal( el.css( "position" ), "absolute", "The old style is kept when setting to a space" );
-	}
+// Support: Edge 14 - 15
+// Edge collapses whitespace-only values when setting a style property and
+// there is no easy way for us to work around it. Just skip the test there
+// and hope for the better future.
+QUnit[ /\bedge\//i.test( navigator.userAgent ) ? "skip" : "test" ](
+	"Keep the last style if the new one is a non-empty whitespace (gh-3204)",
+	function( assert ) {
+	assert.expect( 1 );
+
+	var el = jQuery( "<div></div>" ).css( "position", "absolute" ).css( "position", " " );
+	assert.equal( el.css( "position" ), "absolute", "The old style is kept when setting to a space" );
 } );
 
 QUnit.test( "Reset the style if set to an empty string", function( assert ) {
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -538,7 +538,7 @@ jQuery.extend( {
 		if ( s.crossDomain == null ) {
 			urlAnchor = document.createElement( "a" );
 
-			// Support: IE <=8 - 11, Edge 12 - 13
+			// Support: IE <=8 - 11, Edge 12 - 15
 			// IE throws exception on accessing the href property if url is malformed,
 			// e.g. http://example.com:80x/
 			try {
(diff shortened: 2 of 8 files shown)


test/unit/css.js
```diff
@@ -1323,11 +1323,11 @@ QUnit.test( "Keep the last style if the new one isn't recognized by the browser
 	assert.equal( el.css( "position" ), "absolute", "The old style is kept when setting an unrecognized value" );
 	el = jQuery( "<div></div>" ).css( "position", "absolute" ).css( "position", " " );
 
-	// Support: Edge 14
+	// Support: Edge 14-15
```
gibson042: Missing spaces around the `-`.

test/unit/css.js
```diff
@@ -1323,11 +1323,11 @@ QUnit.test( "Keep the last style if the new one isn't recognized by the browser
...
 	// Edge collapses whitespace-only values when setting a style property and
 	// there is no easy way for us to work around it. Just skip the test there
 	// and hope for the better future.
-	if ( /edge\//i.test( navigator.userAgent ) ) {
+	if ( /\bedge\//i.test( navigator.userAgent ) ) {
 		assert.ok( true, "Skipped (Edge 14 handles whitespace-only values incorrectly)" );
```
gibson042: Please pull this out into its own `QUnit[ /\bedge\//i.test( navigator.userAgent ) ? "skip" : "test" ]`.

test/unit/css.js
markelog: Isn't it should be the opposite – https://en.wikipedia.org/wiki/Dash#Ranges_of_values

gibson042: Those are version numbers, not numeric values, we use U+002D HYPEN-MINUS (which also indicates subtraction), not U+2013 EN DASH, and this is a [formal support comment](https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-69379197), not English prose.

markelog: Pretty sure that version numbers are numeric :), even if they have designed names, at least according to the [semver](http://semver.org/#spec-item-11) or any other number schemas really, since i'm not sure how you can describe a range without numeric values.

Most of the text style guides like [APA](https://en.wikipedia.org/wiki/APA_style) & [AMA](https://en.wikipedia.org/wiki/AMA_Manual_of_Style), which is hardly an "English prose", use variation of dash without the spaces, even HYPHEN-MINUS (specifically "HYPHEN" btw, not "HYPEN") and EN DASH are:

> equally "correct", and each is the preferred style in some style guides 

but always, always without a spaces.

Rules you devised in that comment are great, they are just deviate from most of guides for natural languages (specifically including technical documents) around the world (not only in English actually). I don't think we need to invent a new language or text style guide for this, just to use already created one.

Like if you invent formal specification to illustrate your point, it doesn't make it widely acceptable idea. Wicked good way to present your point though :).

Instead, I would prefer to use something that everybody use, if you don't like form of dashes used in academic papers (or in typography as a whole), we can replace it with the word "to" with spaces like in [SI](http://physics.nist.gov/cuu/pdf/sp811.pdf) guide of National Institute of Standards and Technology, which bears insignia of U.S. Department of Commerce

gibson042: > Pretty sure that version numbers are numeric :), even if they have designed names, at least according to the [semver](http://semver.org/#spec-item-11) or any other number schemas really, since i'm not sure how you can describe a range without numeric values.

A semantic version _uses_ numbers, but is not itself numeric. And notably, it can include a hyphen (spelling fixed above, thanks), e.g. "[1.0.0-0.3.7](http://semver.org/#spec-item-9)".

> Most of the text style guides like [APA](https://en.wikipedia.org/wiki/APA_style) & [AMA](https://en.wikipedia.org/wiki/AMA_Manual_of_Style), which is hardly an "English prose", use variation of dash without the spaces, even HYPHEN-MINUS (specifically "HYPHEN" btw, not "HYPEN") and "EN DASH" are:
> > equally "correct", and each is the preferred style in some style guides 
>
> but always, always without a spaces.

That works for them, but is just not acceptable when it can introduce ambiguity. `1.0.0 - 1.1.0-3.1.4` is a valid version range, but take away the spaces and its meaning is no longer clear.

> Rules you devised in that comment are great, they are just deviate from most of guides for natural languages (specifically including technical documents) around the world (not only in English actually). I don't think we need to invent a new language or text style guide for this, just to use already existing one.

> Instead, I would prefer to use something that everybody use, if you don't like form of dashes used in academic papers (or in typography as a whole), we can replace it with the word "to" with spaces like in [SI](http://physics.nist.gov/cuu/pdf/sp811.pdf) guide of National Institute of Standards and Technology, which bears insignia of U.S. Department of Commerce

Those are fine opinions, and you can incorporate them in a proposed fix for https://github.com/jquery/contribute.jquery.org/issues/95 if you like. But for _this_ PR, consistency dictates the inclusion of spaces, as is the case in the other support comments.

mgol:
@gibson042 comments addressed.

mgol:
I'd like to land it today evening.

test/unit/css.js
markelog: > That works for them, but is just not acceptable when it can introduce ambiguity. 1.0.0 - 1.1.0-3.1.4 is a valid version range, but take away the spaces and its meaning is no longer clear.

Looks ambiguous to me either way, I'm not sure what that means even.

> Those are fine opinions, and you can incorporate them in a proposed fix for jquery/contribute.jquery.org#95 if you like.

Yeah, I can do that too, in fact, I told you that many times, since we still following the rules which weren't documented in any way.

If we agree that this if fine opinions I can create a pull to make all the comments like correct and consistent

mgol: >> That works for them, but is just not acceptable when it can introduce ambiguity. 1.0.0 - 1.1.0-3.1.4 is a valid version range, but take away the spaces and its meaning is no longer clear.
>
> Looks ambiguous to me either way, I'm not sure what that means even.

I think @gibson042 refers to the fact that semver allows hyphens in version strings so `1.1.0-3.1.4` is a pre-release to `1.1.0`. When you have spaces surrounding the version range separator, it's clear whether a particular hyphen is part of a version string or a range separator.

markelog: Could you give me an example within our context?

mgol: What our context? This is a general rule regarding version strings. We might now have any such references in comments right now but it doesn't mean they will never appear.

markelog: I'm asking who this would look like  `Chrome 1.0.0 - 1.1.0-3.1.4` or something? And I don't get what that supposed to mean, Chrome had an issue in `1.0.0` and then issue resurfaced in `1.1.0` and continue to `3.1.4`?

Then why not `Chrome 1.0.0` & `Chrome 1.1.0-3.1.4`?

markelog: If yes, then this already solved with SI style, since they also were concern with negative values, it's also looks confusing, this why I proposed to use ` to ` as they do

mgol: No, that would mean the bug exists at least from version `1.0.0` to version `1.1.0-3.1.4`. There is no such thing as version `3.1.4` in this example.

mgol: This is not the proper place to discuss these issues. If you want to raise an issue with the current syntax, comment on https://github.com/jquery/contribute.jquery.org/issues/95. I'm just following what we use everywhere else right now.

markelog: > This is not the proper place to discuss these issues.

I believe I already did, this is why I raised this question here

> I'm just following what we use everywhere else right now.

I'm not asking to change anything

> If you want to raise an issue with the current syntax

There is no current syntax, there is no format for it, it's just something we might use, this is why I created jquery/contribute.jquery.org#95 in the first place

PR #3666, opened 2017-05-11
Core: param serializes first element of an array without an index

odedniv:
### Summary ###
`jQuery.param` will serialize the first element of an array without an index, this prevents ambiguity between an object with numeric keys and an array, and will therefor allow properly serializing an array of objects.

**Before:**
```javascript
jQuery.param({ a: [{ b: 1 }, { c: 2 }] })
=> a[0][b]=1&a[1][c]=2
```

This *could* falsely be deserialized as: `{ a: { "0": { b: 1 }, "1": { c: 2 } } }` (and it absolutely will).

**After:**
```javascript
jQuery.param({ a: [{ b: 1 }, { c: 2 }] })
=> a[][b]=1&a[1][c]=2
```

This first element ensures `a` should be considered an array.

This follows this PR in rack: rack/rack#1165

Unfortunately this is not backward compatibly (would actually break rack before it pulls this request), but this is somewhat mandatory as you simply cannot properly serialize some scenarios (see the tests in rack's PR for more scenarios). Should this be a configuration option somewhere?

### Checklist ###

### DIFF
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -30,7 +30,7 @@ function buildParams( prefix, obj, traditional, add ) {
 
 				// Item is non-scalar (array or object), encode its numeric index.
 				buildParams(
-					prefix + "[" + ( typeof v === "object" && v != null ? i : "" ) + "]",
+					prefix + "[" + ( i > 0 && typeof v === "object" && v != null ? i : "" ) + "]",
 					v,
 					traditional,
 					add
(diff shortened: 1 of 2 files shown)


odedniv:
PS. I could not run the tests for some reason, they always succeeded even they were supposed to fail.

odedniv:
To do that you have to completely change how `buildParams` work, are you comfortable with that?

If so, do you expect non-object elements to include an index as well? Right now it only includes an index in object elements due to requiring to modify existing elements (which was ambiguous before this fix, as you couldn't know whether it's a new array element or not - and in rack for example you they had to guess according to existing keys in the object).

gibson042:
Current released behavior is correct but ambiguous when the first element of an array is not primitive. But the current PR behavior is _incorrect_ when the first element of an array has more than one key, so we cannot accept it as-is.

oded-puresec:
@gibson042 the current PR behavior is not ambiguous, so I do not understand why it's incorrect. Once it's been established that the element is part of an array, and the element itself is an object, you can add more items to that object with `[][...]`.

But my real question is this: Are you comfortable with a rewrite of the function in order to make the second attribute of an object indexed? If you are, I'll get right on it.

mgol:
> Once it's been established that the element is part of an array, and the element itself is an object, you can add more items to that object with [][...].

This doesn't seem right. `a[]` is referring to a new element pushed to the array so `a[][p1]=2&a[][p2]=3` would produce `[{p1: 2}, {p2: 3}]` instead of the expected `[{p1: 2, p2: 3}]`. The latter result could be achieved via `a[][p1]=2&a[0][p2]=3`.

gibson042:
> Are you comfortable with a rewrite of the function in order to make the second attribute of an object indexed? If you are, I'll get right on it.

Yes, provided it does not increase the size of jquery.min.gz by more than 10 bytes. Behavior has been stable for a long time, and in my opinion this edge case isn't worth a size investment.

timmywil:
This seems to have gotten stale and I'm not sure it's worth the effort.

PR #3669, opened 2017-05-12
Updates Dependencies

CDAGaming:
### Summary ###
This PR Updates the Following Dependencies:
- Babel-Preset-ES2015: v6.6.0 > v6.24.1 (Minor)
- CommitPlease: v2.6.1 > v2.7.10 (Minor)
- Core-JS: v2.2.2 > v2.4.1 (Minor)
- Cross-Spawn: v2.2.3 > v5.1.0 (Major)
- ESlint-Config-JQuery: v1.0.0 > v1.0.1 (Minor)
- Grunt-Contrib-Uglify: v1.0.1 > v3.0.1 (Major)
- Grunt-JSONLint: v1.0.7 > v1.1.0 (Minor)
- Grunt-Newer: v1.2.0 > v1.3.0 (Minor)
- Husky: v0.11.4 > v0.13.4 (Minor)
- Insight: v0.8.1 > 0.8.4 (Minor)
- Load-Grunt-tasks: v3.5.0 > v3.5.2 (Minor)
- Q: v1.4.1 > v1.5.0 (Minor)
- QUnit-Assert-Step: v1.0.3 > v1.1.1 (Minor)
- QUnitJS: v1.23.1 > v2.3.2 (Major)
- RequireJS: v2.2.0 > v2.3.3 (Minor)
- Simon: v1.17.3 > v2.3.2 (Major)

### Checklist ###
N/A

* [👍] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [👍] New tests have been added to show the fix or feature works
* [👍] Grunt build and unit tests pass locally with these changes
* [[Click Here](https://github.com/jquery/api.jquery.com/pull/1036)] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #3680, opened 2017-05-30
Build: Add yarn.lock to .gitignore

mgol:
### Summary ###

npm 5 generates a package-lock.json file by default so running `npm install`
with this version puts the local repository copy in a dirty state.

This commit adds package-lock.json to .gitignore so that this doesn't happen.

### Checklist ###

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 .sizecache.json
 
 npm-debug.log*
+package-lock.json
 
 # Ignore everything in dist folder except for eslint config
 /dist/*


markelog:
https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

?

mgol:
If we commit the file won't we have to ensure everyone changing any dependency in package.json uses npm 5 or newer? That sounds risky.

mgol:
As far as I understand, this is basically npm shrinkwrap but only for development, not used in publishing/installing a package. We haven't had problems with deep dependencies changing recently and the lockfile will do nothing for npm<5 which we still test via Node 4 & 6. So I'm not sure if it buys us anything right now and it introduces the complexity that people bumping packages need to be on npm@5.

`package-lock.json` seems like a useful addition for complex projects with big dependency chains, it doesn't seem that useful to jQuery IMO.

timmywil:
If `npm install` doesn't obey the lock file by default and instead updates dependencies and rewrites the lock file, then I don't want to commit it. `npm install` should be idempotent with a lockfile. It took a while for yarn to get this right as well (i.e. you used to have to use the `--pure-lockfile` option to get this behavior). I don't want to have to train users to do anything other than `npm install` and until the lockfile behaves the way that makes it most useful, I'd like to wait on committing it.

markelog:
@mgol 
Ha-ha, that's funny to think about, so what you are saying is: that we have choose – either we use npm < 5 or npm >= 5? 

> not used in publishing/installing a package

You mean in registry? No, I think, it shouldn't be published, but it's useful to install your own dependencies. Lock files should be committed at least according to the facebook, cargo, ruby gems teams

> it introduces the complexity that people bumping packages need to be on npm@5.

Yeah, I guess that's a deal breaker :))

@timmywil 

> i.e. you used to have to use the --pure-lockfile option to get this behavior

What do you mean "used to"? Yarn by default do not use `--pure-lockfile`, neither yarn nor npm recommend not committing the lock files

> until the lockfile behaves the way that makes it most useful

Lock files is what makes builds idempotent :) and much faster, ha-ha.

Are you guys sure? I mean, all right, that would be cool story I guess

timmywil:
> What do you mean "used to"? Yarn by default do not use `--pure-lockfile`.

`--pure-lockfile` meant something different in previous versions of yarn. It meant respect all the versions in the lockfile, but now `yarn install` does that by default and the docs for `pure-lockfile` have changed. Note: this is different than `yarn install [package]`, which still writes a new lockfile without the `--pure-lockfile` option.

> Lock files is what makes builds idempotent :) and much faster, ha-ha.

We have a disconnect in communication here. I agree that's the way it *should* work. The problem here is that it should be idempotent but it's not, at least if it's writing a new lockfile when one already exists, which is how I read this issue. I'm fine with it writing a lockfile if one does not yet exist.

timmywil:
@mgol Unless, are you saying the dirty state comes about because a new file is being written? If that's the only problem here, then I *do* want to commit the lockfile.

timmywil:
I'll do some testing and see if it works the way I hope it does.

timmywil:
Ok, after some quick testing, I don't think `npm install` has the behavior I assumed it had. I vote for committing the lockfile. The advantage here is that people that fork jQuery will have the same versions of dependencies and sub-dependencies that we do, which should mitigate errors caused by unwanted updates on install. Granted, anyone on npm < 5 won't get this benefit, but the behavior for them will just be the same as it was before.

markelog:
@timmywil 

> It meant respect all the versions in the lockfile

Mmm, it didn't respect some of the versions in the lock file? Isn't that a direct purpose if the lock file? Huh, that's strange

> We have a disconnect in communication here

So by the phrase

> I don't want to have to train users to do anything other than npm install and until the lockfile behaves the way that makes it most useful, I'd like to wait on committing it.

You meant, that newer npm might be working as old yarn? I see now :)

timmywil:
> You meant, that newer npm might be working as old yarn?

Exactly, but it turns out my suspicion was unnecessary.

timmywil:
> Mmm, it didn't respect some of the versions in the lock file?

By the way, I haven't dug into all of the details, but I can provide a little background to my thinking. `yarn install` was not always idempotent. I think there were concerns that not updating according to the version ranges on install would mean people would never update. The `pure-lockfile` option was there for production and cli installs. As you noted, this opinion did seem to contradict the entire purpose of a lockfile in development and prompted this discussion: https://github.com/yarnpkg/yarn/issues/570, which resulted in the behavior we now know and love: https://github.com/yarnpkg/yarn/issues/1576.

I assumed that back-compat concerns would have delayed this behavior for at least another major version, but I am pleasantly surprised that `npm` got it right in 5.0.0.

mgol:
@timmywil Do you agree, though, that us committing the lock file to the repo means no one (nor we, nor any external contributors) can change dependencies using anything other than npm >=5? They can't use Yarn locally, they can't use npm<5.

On Windows it's quite hard to update npm, people will generally have a version built into the Node version they use. This essentially means they have to use the just released Node 8.0.0 and nothing older as it'd desynchronize our lockfile from `package.json`. We'd have to pay attention to that when reviewing PRs like #3669.

I have a feeling there are just too many drawbacks at the moment. In the world where npm<4 have low market share, I'd tend to agree. But we're a few years away from that moment.

markelog:
@timmywil 

I see there was a bug in their heuristics, thank you for the context!  

@mgol

It's the same as in yarn, right? If you use `yarn.lock` you have to update your dependencies with yarn as well, or run `yarn install` if you used `npm` for it

> On Windows it's quite hard to update npm

Is it? Something wrong with `npm update npm -g` on Windows?

mgol:
> It's the same as in yarn, right? If you use `yarn.lock` you have to update your dependencies with yarn as well, or run `yarn install` if you used `npm` for it

Yes. I think we should add `yarn.lock` to `.gitignore` as well.

The only difference (but it's a big one!) is that to use Yarn is a non-standard thing to do as it's not a default. `npm` will have more users that just don't think about it.

timmywil:
> Do you agree, though, that us committing the lock file to the repo means no one (nor we, nor any external contributors) can change dependencies using anything other than npm >=5?

I agree, but I'm not sure that's enough of a drawback. We don't see PRs for dependency updates very often. Seems like we (the jQuery team) are the ones updating dependencies in this repo, and we all are well-versed in managing node/npm versions.

mgol:
@timmywil OK, fair enough. If we decide to use npm@5 for that we should probably add `yarn.lock` to `.gitignore`, though, so that people trying to use Yarn on our repo can do it without creating a dirty state. Maintaining two lock files would be a little too much IMO.

timmywil:
@mgol +1 to that

mgol:
OK, I'll update the PR and I'll add a commit adding `yarn.lock` to `.gitignore`.

timmywil:
Doing this now as I land https://github.com/jquery/jquery/pull/3669

PR #3694, opened 2017-06-13
Stop propagation on native events

### LINKED ISSUE #3693: jQuery does not stop propagation of `jQuery.click()` on native event-handlers
### Description ###

Event-handlers that are registered natively are called even if `event.stopPropagation()` is called when `jQuery.click()` is used.

Using the native `HTMLelement.click()` correctly stops propagation.

### Link to test case ###

I the following test case, both, a `HTMLelement.click()` and a `jQuery.click()` are triggered when the page is loaded.

In both cases, the event bubbling is prevented. The `jQuery.click()` is still triggered on the `document` event-handler.

https://jsbin.com/qekuda/7/edit?html,js,output

### Link to Stackoverflow discussion ####

https://stackoverflow.com/questions/44309897/jquery-does-not-stop-propagation-of-jquery-click-on-native-event-handlers

caillou:
### Summary ###

This PR ensures that `event.stopPropagation()` is respected for event handlers that were registered using the native `element.addEventlistener()` method when the event was triggered by jQuery's `.trigger('click')` method.

This PR fixes #3693.

The ticket was closed at first with the following argument:

> jQuery has never been able to guarantee full interoperability of native-vs-jQuery event handlers.

Nonetheless, the discussion lead me to believe, that this incompatibility between native and jQuery event handling could be fixed and that such a fix is worthwhile.

### Checklist ###

### DIFF
--- a/src/event/trigger.js
+++ b/src/event/trigger.js
@@ -24,7 +24,7 @@ jQuery.extend( jQuery.event, {
 			type = hasOwn.call( event, "type" ) ? event.type : event,
 			namespaces = hasOwn.call( event, "namespace" ) ? event.namespace.split( "." ) : [];
 
-		cur = tmp = elem = elem || document;
+		cur = lastElement = tmp = elem = elem || document;
 
 		// Don't do events on text and comment nodes
 		if ( elem.nodeType === 3 || elem.nodeType === 8 ) {
--- a/src/event/trigger.js
+++ b/src/event/trigger.js
@@ -10,13 +10,16 @@ define( [
 
 "use strict";
 
-var rfocusMorph = /^(?:focusinfocus|focusoutblur)$/;
+var rfocusMorph = /^(?:focusinfocus|focusoutblur)$/,
+	stopPropagationCallback = function( e ) {
+		e.stopPropagation();
+	};
 
 jQuery.extend( jQuery.event, {
 
 	trigger: function( event, data, elem, onlyHandlers ) {
 
-		var i, cur, tmp, bubbleType, ontype, handle, special,
+		var i, cur, tmp, bubbleType, ontype, handle, special, lastElement,
 			eventPath = [ elem || document ],
 			type = hasOwn.call( event, "type" ) ? event.type : event,
 			namespaces = hasOwn.call( event, "namespace" ) ? event.namespace.split( "." ) : [];
@@ -93,7 +96,7 @@ jQuery.extend( jQuery.event, {
 		// Fire handlers on the event path
 		i = 0;
 		while ( ( cur = eventPath[ i++ ] ) && !event.isPropagationStopped() ) {
-
+			lastElement = cur;
 			event.type = i > 1 ?
 				bubbleType :
 				special.bindType || type;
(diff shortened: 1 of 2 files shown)


src/event/trigger.js
```diff
@@ -10,13 +10,16 @@ define( [
 
 "use strict";
 
-var rfocusMorph = /^(?:focusinfocus|focusoutblur)$/;
+var rfocusMorph = /^(?:focusinfocus|focusoutblur)$/,
+	stopPropagationCallback = function( e ) {
```
dmethvin: I wonder if we should have this inside `trigger()` since it is only used there. Opinions welcome.

src/event/trigger.js
```diff
@@ -136,7 +139,17 @@ jQuery.extend( jQuery.event, {
 
 					// Prevent re-triggering of the same event, since we already bubbled it above
 					jQuery.event.triggered = type;
+
+					if ( event.isPropagationStopped() ) {
+						lastElement.addEventListener( type, stopPropagationCallback );
```
dmethvin: I think it would be possible to get here with `lastElement === undefined` if the incoming `event` already had propagation stopped? Maybe initialize with `lastElement = elem`?

dmethvin: Hmmm, this does bring up the problem of `.stopImmediatePropagation()` which isn't fixed. Since that method is sensitive to the order that handlers are attached and ours is always going to be last, there isn't much we can do for that 🤷‍♂️ .

test/unit/event.js
```diff
@@ -440,6 +440,26 @@ QUnit.test( "on bubbling, isDefaultPrevented, stopImmediatePropagation", functio
 	$anchor2[ 0 ].removeEventListener( "click", neverCallMe );
 } );
 
+QUnit.test( "triggerd events stopPropagation() for natively-bound events", function( assert ) {
```
dmethvin: triggered

src/event/trigger.js
caillou: @dmethvin One option with the `.stopImmediatePropagation()` would be to register an additional event handler on the second to last element (the great ancestor) if there is one present, that stops the propagation.

But this seems rather fiddly.

caillou: @dmethvin I will try to add a test-case for the `lastElement === undefined` case. But I don't yes see how this could happen. How can the propagation of the incoming `even` already be stopped?

src/event/trigger.js
caillou: The difference is the following:

* If we define the function within the `trigger` method, it is defined every time trigger is called.
* If we define the function within the AMD callback function as I did here, the function will be defined only once, but even if the `trigger` method is never used. Additionally, there is an additional scope lookup.

All in all, these considerations are micro optimisations, that are not really worth considering. The questions is more one of culture within a project. We could also define it at the end of the file, just before the `return jQuery` statement.

I will look at the other files and see how it is usually handled within the `jQuery` project.

src/event/trigger.js
```diff
@@ -93,7 +96,7 @@ jQuery.extend( jQuery.event, {
 		// Fire handlers on the event path
 		i = 0;
 		while ( ( cur = eventPath[ i++ ] ) && !event.isPropagationStopped() ) {
-
+			lastElement = cur;
```
gibson042: You can reduce the size increase of jquery.min.gz by using `eventPath[ i - 2 ]` below rather than introducing a new variable.

caillou: @gibson042 in my opinion, it would reduce the size at a high cost of code readability. What is the code guideline in the `jQuery` project with regard to readability v.s. size?

caillou:
@dmethvin @gibson042: I have added a test with a perviously stopped event and fixed it.

Could you review this again? Sorry for the trouble, I should have thought about this.

This yields the question: Do we really want to go forward with this PR? While it helps with interoperability between native and jQuery events, I wonder if it is worth the trouble. WDYT?

dmethvin:
@caillou I think you have a pretty small patch here and it could be useful. What are your concerns?

caillou:
@dmethvin Good. No real concerns. What is the workflow? Do I need to do anything, or will it get merged at some point?

dmethvin:
@caillou Yes, we'll merge it shortly. Thanks!

Issue #3612, opened 2017-04-05
offset() should never return read-only properties


Krinkle:
Calling `offset()` should consistently return a plain object that has (naturally) mutable properties. In jQuery 3.0.0 `offset()` first started to (sometimes!) return native `ClientRect` objects directly from `HTMLElement#getBoundingClientRect` in some cases.

This was not consistently the case though, and actually broke internal jQuery assumptions as well, as evidenced by https://github.com/jquery/jquery/commit/3befe5911af0cf516896482bb9ddf197c8cb8a8e (2016, 3.0.0).

Per @dmethvin at https://github.com/jquery/jquery/pull/109:
> It seems like a good idea to consistently return a plain object.

It also caused downstream issues for Wikimedia. When locally attempting to update to jQuery 3, various code paths throughout [MediaWiki](https://github.com/wikimedia/mediawiki]), [VisualEditor](https://github.com/wikimedia/VisualEditor), and [oojs-ui](https://github.com/wikimedia/oojs-ui) started throwing errors such as:
```
Uncaught TypeError: Cannot assign to read only property 'left' of object '#ClientRect'
```

See also analysis in downstream report at <https://phabricator.wikimedia.org/T160802#3155563>.

It appears this is now fixed again in jQuery 3.2.1, however it would be good to add regression tests to avoid this in the future.

PR #3695, opened 2017-06-15
Test: ensure position/offset return mutable objects

jbedard:
This adds tests verifying #3612. Could probably add more but this seemed to cover most cases?

Could potentially also revert 3befe5911af0cf516896482bb9ddf197c8cb8a8e

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -140,10 +140,8 @@ jQuery.fn.extend( {
 
 				// Incorporate borders into its offset, since they are outside its content origin
 				parentOffset = jQuery( offsetParent ).offset();
-				parentOffset = {
-					top: parentOffset.top + jQuery.css( offsetParent, "borderTopWidth", true ),
-					left: parentOffset.left + jQuery.css( offsetParent, "borderLeftWidth", true )
-				};
+				parentOffset.top += jQuery.css( offsetParent[ 0 ], "borderTopWidth", true );
+				parentOffset.left += jQuery.css( offsetParent[ 0 ], "borderLeftWidth", true );
 			}
 		}
 
--- a/test/unit/offset.js
+++ b/test/unit/offset.js
@@ -65,7 +65,7 @@ QUnit.test( "empty set", function( assert ) {
 } );
 
 QUnit.test( "disconnected element", function( assert ) {
-	assert.expect( 3 );
+	assert.expect( 4 );
 
 	var result = jQuery( document.createElement( "div" ) ).offset();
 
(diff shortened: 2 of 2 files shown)


test/unit/offset.js
```diff
@@ -75,10 +75,11 @@ QUnit.test( "disconnected element", function( assert ) {
 	assert.equal( result.top, 0, "Retrieving offset on disconnected elements returns zeros (gh-2310)" );
 	assert.equal( result.left, 0, "Retrieving offset on disconnected elements returns zeros (gh-2310)" );
 	assert.equal( Object.keys( result ).length, 2, "Retrieving offset on disconnected elements returns offset object (gh-3167)" );
+	assert.equal( jQuery.isPlainObject( result ), true, "Retrieving offset on disconnected elements returns plain object (gh-3612)" );
```
dmethvin: Should we be actually trying to mutate the object here? The `jQuery.isPlainObject()` test is a bit of an indirect feature test and technically we could be returning a plain object with read-only properties.

jbedard: I guess I was just ensuring it wasn't a `ClientRect`. I could create a helper that checks both `isPlainObject` and ensures it is modifiable. WDYT?

timmywil:
Agreed, revert 3befe59, and good to land.

jbedard:
Was there any discussion about the use of `isPlainObject` vs something more specific?

jbedard:
I've added the revert of 3befe5911af0cf516896482bb9ddf197c8cb8a8e since it reduces size:
```
   raw     gz Compared to master @ 31783e60f5649b194128a6e39404617c69117dab
   -28     -7 dist/jquery.js
     =      = dist/jquery.min.js
```

src/offset.js
```diff
@@ -140,10 +140,8 @@ jQuery.fn.extend( {
 
 				// Incorporate borders into its offset, since they are outside its content origin
 				parentOffset = jQuery( offsetParent ).offset();
-				parentOffset = {
-					top: parentOffset.top + jQuery.css( offsetParent, "borderTopWidth", true ),
-					left: parentOffset.left + jQuery.css( offsetParent, "borderLeftWidth", true )
-				};
+				parentOffset.top += jQuery.css( offsetParent[ 0 ], "borderTopWidth", true );
+				parentOffset.left += jQuery.css( offsetParent[ 0 ], "borderLeftWidth", true );
```
gibson042: I think these need to be `offsetParent`, not `offsetParent[ 0 ]`.

Issue #3699, opened 2017-06-20
IE9 and the incredible shrinking scrollboxes


gibson042:
80f1c8239e9d5f793f3e54e0cb8d7bd0747e4856 worked everywhere except IE9: http://swarm.jquery.org/result/2255930

Investigation traced the root cause to a mishandling by that browser of `box-sizing: content-box`, particularly in combination with `overflow: scroll`... specifically, `getComputedStyle` seems to always report content box dimensions, even for border-box elements, and the lack of accounting for scrollbars means that round-trip set operations shrink such elements: https://jsbin.com/zotomuqoho/edit?html,css,js,output

mamieS99:
I have to look at video again this looks different  and i get confused of the step, will have to come back. Cant do it yet. #

Krinkle:
While now fixed in IE 9, the fix subsequently broke IE 10/11. https://github.com/jquery/jquery/issues/3731

PR #3700, opened 2017-06-20
Dimensions: Detect and account for content-box dimension mishandling

gibson042:
Fixes gh-3699 with a new support property. But it's big, and the right call might just be to let this edge case fail:
```
   raw     gz Sizes
270450  80204 dist/jquery.js
 87101  30156 dist/jquery.min.js

   raw     gz Compared to master @ 80f1c8239e9d5f793f3e54e0cb8d7bd0747e4856
  +775   +190 dist/jquery.js
  +198    +44 dist/jquery.min.js
```

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -148,10 +148,10 @@ function boxModelAdjustment( elem, dimension, box, isBorderBox, styles, computed
 function getWidthOrHeight( elem, dimension, extra ) {
 
 	// Start with computed style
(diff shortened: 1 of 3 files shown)


timmywil:
@gibson042 Leaning towards letting this fail and marking wontfix.

gibson042:
This PR found sufficient support in the weekly meeting.

PR #3702, opened 2017-06-22
Deprecate jQuery.isWindow (3629)

### LINKED ISSUE #3629: Deprecate jQuery.isWindow
Was created for internal use, don't needed now

ShashankaNataraj:
### Summary ###
Fix for issue #3629
Removed occurances of isWindow

### Checklist ###

### DIFF
--- a/src/var/isWindow.js
+++ b/src/var/isWindow.js
@@ -2,7 +2,7 @@ define( function() {
 	"use strict";
 
 	return function isWindow( obj ) {
-		return obj !== null && typeof obj !== "undefined" && obj === obj.window;
+		return obj != null && obj === obj.window;
 	};
 
 } );
--- a/src/var/isWindow.js
+++ b/src/var/isWindow.js
@@ -1,0 +1,8 @@
+define( function() {
+	"use strict";
+
+	return function isWindow( obj ) {
+		return obj !== null && obj === obj.window;
+	};
+
+} );
--- a/src/var/isWindow.js
+++ b/src/var/isWindow.js
@@ -2,7 +2,7 @@ define( function() {
 	"use strict";
 
 	return function isWindow( obj ) {
-		return obj !== null && obj === obj.window;
+		return obj !== null && typeof obj !== "undefined" && obj === obj.window;
 	};
 
 } );
--- a/src/core.js
+++ b/src/core.js
@@ -16,10 +16,11 @@ define( [
 	"./var/fnToString",
 	"./var/ObjectFunctionString",
 	"./var/support",
+	"./var/isWindow",
 	"./core/DOMEval"
 ], function( arr, document, getProto, slice, concat, push, indexOf,
 	class2type, toString, hasOwn, fnToString, ObjectFunctionString,
-	support, DOMEval ) {
+	support, isWindow, DOMEval ) {
 
 "use strict";
 
@@ -218,10 +219,6 @@ jQuery.extend( {
 		// (i.e., `typeof document.createElement( "object" ) === "function"`).
 		// We don't want to classify *any* DOM node as a function.
 		return typeof obj === "function" && typeof obj.nodeType !== "number";
-	},
-
-	isWindow: function( obj ) {
-		return obj != null && obj === obj.window;
 	},
 
 	isNumeric: function( obj ) {
@@ -469,7 +466,7 @@ function isArrayLike( obj ) {
 	var length = !!obj && "length" in obj && obj.length,
 		type = jQuery.type( obj );
 
-	if ( jQuery.isFunction( obj ) || jQuery.isWindow( obj ) ) {
+	if ( jQuery.isFunction( obj ) || isWindow( obj ) ) {
 		return false;
 	}
 
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -1,7 +1,8 @@
 define( [
 	"./core",
-	"./core/nodeName"
-], function( jQuery, nodeName ) {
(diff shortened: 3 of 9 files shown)


ShashankaNataraj:
Hey @mgol, thanks for the helping hand, sure will go through the links provided and make the changes required.

ShashankaNataraj:
@mgol I worked on a few changes: https://github.com/jquery/jquery/compare/master...ShashankaNataraj:deprecate-is-window

Do let me know if these changes look OK, and ill squash them into a single commit and make a PR. Also, I ran all the tests locally with `grunt test` and see that they are all working.

mgol:
Unfortunately, while running `grunt test` does check a few things it's not enough. We have lots of unit tests that have to be run manually in a browser. See https://github.com/jquery/jquery#running-the-unit-tests; after setting up the PHP server you need to navigate to `/test/` and wait for the tests to finish.

To have all tests running you need to set up Apache+PHP but the local PHP server `php -S localhost:8000` should work as well with the only caveat being that a few tests may fail then.

You can also fire up a regular non-PHP static HTTP server (e.g. using Python or the npm [http-server](https://www.npmjs.com/package/http-server) package) but then AJAX tests won't work at all and a few other tests might not. Always better than nothing, though!

The new changes go in a good direction; there are a few mistakes (you'll see most of them yourself if you run unit tests) but it'd be better to comment on that once you submit a PR. A PR doesn't have to be perfect from the start; you can push changes to it after feedback from reviewers; no need to close it. :)

Let me know if you have any further questions!

ShashankaNataraj:
@mgol Thanks for the encouraging words, I tried running the PHP tests after installing MAMP, and I see that jQuery.isWindow test is failing. The reason I think is because I have now removed isWindow function from the jQuery object.

Question: What is the right way to `require` the `isWindow` file in `tests/unit/deprecated.js`?

mgol:
`jQuery.isWindow` still needs to exist as it's not getting removed here (that would be a breaking change) but only deprecated so far. You'll need to alias it in `src/deprecated.js`, see [this line](https://github.com/jquery/jquery/blob/3.2.1/src/deprecated.js#L38) for an example.

ShashankaNataraj:
Thanks for your patience and guidance! I have committed some more changes now at https://github.com/jquery/jquery/compare/master...ShashankaNataraj:deprecate-is-window and all of the QUnit browser tests run successfully now.
screen shot 2017-06-23 at 6 05 17 pm

Though I do not know if there is a better way of doing `obj !== "undefined"` at https://github.com/jquery/jquery/compare/master...ShashankaNataraj:deprecate-is-window#diff-bb83a8b45ba6ae69e309198abb5db45bR5. I know of `jQuery.isPlainObject()` but do not know if using `core` module to check this one thing will be acceptable.

src/var/isWindow.js
```diff
@@ -0,0 +1,8 @@
+define( function() {
+	"use strict";
+
+	return function isWindow( obj ) {
+		return obj !== null && typeof obj !== "undefined" && obj === obj.window;
```
mgol: Don't expand it; `obj != null` was enough as it was before.

ShashankaNataraj:
Got it, the order of params was wrong in atleast 3 places. Have corrected that and committed, but didn't know that running with `load with AMD` would cause errors unseen otherwise. Will see whats wrong and correct those errors as well.

src/var/isWindow.js
ShashankaNataraj: I believe I added it because a test was breaking. Let me remove it and check if it still breaks.

ShashankaNataraj: I just removed `obj!=="undefined` and [this test breaks](https://github.com/jquery/jquery/compare/master...ShashankaNataraj:deprecate-is-window#diff-b77bfd25a4e9630e66a502c213900716R173) because 

screen shot 2017-06-24 at 8 27 26 pm

mgol: It was (and still should be) `obj != null`, not `obj !== null`.

ShashankaNataraj: @mgol Got it, I've retained `obj!=null` instead of `obj!==null` and made a commit. The tests all seem to pass with AMD mode enabled.

screen shot 2017-06-25 at 9 55 58 am

ShashankaNataraj: @mgol is there anything else that I've missed? 

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

timmywil:
Well done @ShashankaNataraj

ShashankaNataraj:
Thanks @timmywil. Many thanks to @mgol for his patience and guidance!

Are there any steps required from me to get this merged? (Squashing these commits into one, maybe?)

On a side note I see that there are deprecation tasks that keep opening up on this repo. Would it be helpful if I made a .md with steps on how to deprecate functions based on comments that @mgol gave above?

ShashankaNataraj:
Hi @timmywil I see this was closed but wasnt merged. Anything missing?

mgol:
It landed in
https://github.com/jquery/jquery/commit/c18d608537d8945de6c5855a9475657177fa74ac;
GitHub linked to this commit in this PR. It doesn't matter that GitHub
shows it as closed instead of merged; this only means it's been landed
manually, without using the GitHub merge button.
-- 
Michał Gołębiowski-Owczarek

ShashankaNataraj:
@mgol Cool, thanks 👍

PR #3710, opened 2017-07-03
Update dependencies; fix uglify options from recent update

timmywil:
### Summary ###
- Updates sinon and husky patch versions
- Fixes uglify options for recent update

Uglify transforms `typeof` statements, causing `elem.getAttribute` to invoke the method despite the lack of parens (the reason we do `typeof` in the first place). I [opened a ticket](https://github.com/mishoo/UglifyJS2/issues/2198) on uglify-js, as I didn't think this transformation should occur unless `unsafe` is set to `true`. The resolution turned out better than I expected. Uglify added a new `compress` option: `typeofs` that disables only this transformation. The result? -16 bytes to -79 bytes (gzipped).

### Checklist ###

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -159,11 +159,10 @@ module.exports = function( grunt ) {
 				options: {
 					preserveComments: false,
 					sourceMap: true,
-					ASCIIOnly: true,
 					sourceMapName:
 						"dist/<%= grunt.option('filename').replace('.js', '.min.map') %>",
 					report: "min",
-					beautify: {
+					output: {
 						"ascii_only": true
 					},
 					banner: "/*! jQuery v<%= pkg.version %> | " +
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -159,11 +159,11 @@ module.exports = function( grunt ) {
 				options: {
 					preserveComments: false,
 					sourceMap: true,
-					ASCIIOnly: true,
```
mgol: Could you explain this change?

Gruntfile.js
```diff
@@ -159,11 +159,11 @@ module.exports = function( grunt ) {
...
+					ie8: true,
```
mgol: Could you add a comment explaining why that's needed with a link to the UglifyJS bug (and maybe a support comment for IE<11)?

Gruntfile.js
timmywil: This option was removed.

timmywil: Or rather, replaced and taken care of by `output.ascii_only`.

Gruntfile.js
timmywil: Added.

Gruntfile.js
```diff
@@ -159,11 +159,15 @@ module.exports = function( grunt ) {
 				options: {
 					preserveComments: false,
 					sourceMap: true,
-					ASCIIOnly: true,
+
+					// Support: IE<11
```
mgol: There should be a space before `<`. Otherwise looks good.

PR #3722, opened 2017-07-12
Update Dependencies

CDAGaming:
Signed-off-by: CDAGaming <cstack2011@yahoo.com>

### Summary ###
This PR Adds a missing Dependency to JQuery as well as removing package-lock as it is an auto generated file

* [ Yes ] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [ Yes ] New tests have been added to show the fix or feature works
* [ Yes ] Grunt build and unit tests pass locally with these changes
* [ N/A ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,7 +2,6 @@
   "name": "jquery",
   "version": "3.2.2-pre",
   "lockfileVersion": 1,
-  "requires": true,
   "dependencies": {
     "abbrev": {
       "version": "1.1.0",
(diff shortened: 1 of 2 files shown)


mgol:
Why is this additional dependency needed?

As for package-lock.json, we added this file to the repo on purpose so don't remove it. Also, make sure you use npm 5.1.0 or newer to add a dependency so that the lockfile gets updated.

timmywil:
This PR doesn't seem to be necessary. The package-lock is committed so that installs are idempotent (for npm >=5 users).

PR #3724, opened 2017-07-15
Update fnToString.js

namik-hajiyev:
why not use class2type ?

### Checklist ###

### DIFF
--- a/src/var/fnToString.js
+++ b/src/var/fnToString.js
@@ -1,7 +1,7 @@
 define( [
-	"./hasOwn"
-], function( hasOwn ) {
+	"./class2type"
+], function( class2type ) {
 	"use strict";
 
-	return hasOwn.toString;
+	return class2type.toString;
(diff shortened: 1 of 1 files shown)


mgol:
Does this change anything? What's the point of this change?

timmywil:
Thanks for contributing! However, this doesn't seem like something we're going to land.

namik-hajiyev:
It doesn't change anything. Sorry bothering you just wondered why hasOwn.
toString is used ))  But found answer you have another module fnToString.js
<https://github.com/jquery/jquery/blob/master/src/var/fnToString.js> where
toString is used.
Bu I am still wondering if fnToString.js
<https://github.com/jquery/jquery/blob/master/src/var/fnToString.js> and
toString.js
<https://github.com/jquery/jquery/blob/master/src/var/toString.js> are
doing the same ?

PR #3725, opened 2017-07-17
Offset: fixed error

joeunha:
- handle when options value (from function returned) is undefined

### Summary ###

When I invoke `offset` method, if arguments was function that return `undefined`. I can see the error like this "_Uncaught TypeError: Cannot read property 'top' of undefined_". So I solved this error by insert condition statement into line 9922. Please check PR and feedback for me.

### Checklist ###

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -52,6 +52,8 @@ jQuery.offset = {
 			options = options.call( elem, i, jQuery.extend( {}, curOffset ) );
 		}
 
+		if ( !options ) return this;
+
 		if ( options.top != null ) {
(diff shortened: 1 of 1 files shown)


mgol:
Per https://api.jquery.com/offset/#offset-function the signature of the function passed to `offset` is:
>  Function( Integer index, PlainObject coords ) => PlainObject

This means the function passed has to return a plain object; other scenarios are unsupported. We don't guarantee anything for non-conforming input so an error may happen. In fact, it's even good that you're getting an error here as that makes it possible for you to notice a potential issue and fix the function you pass to the method. Silent errors would only mask errors.

Issue #3532, opened 2017-02-06
Pass array of classes to `.addClass()`, `removeClass()`, and `toggleClass()` 


t-kelly:
### Description ###
It would be great if these examples:
```js
var class1 = 'class-one';
var class2 = 'class-two';

$('.element').addClass(class1 + ' ' + class2);
$('.element').addClass(class1).addClass(class2);
$('.element').addClass([class1, class2].join(' '));
```

could instead be written like:
```js
var class1 = 'class-one';
var class2 = 'class-two';

$('.element').addClass([class1, class2]);
```

timmywil:
For the most part, we're not looking to add signatures or new features to jQuery, but I don't hate this. It has parody with libraries like React and Angular.

timmywil:
Also, we long to be a "parody" of everything. Sigh...edited.

dmethvin:
Note that this can already be done by `$('.element').addClass([class1, class2].join(" "));` which isn't that burdensome. It will take more bytes than that to do this inside jQuery.

mgol:
@dmethvin It's true but it's also uglier. :) I like it because:

1. We don't currently accept arrays as any parameter to class methods so there should be no confusion. Because of that, it shouldn't also add _that much_ code to jQuery.
2. Accepting classes will align us more with the classList standard and aligning with standards is A Good Thing.
3. It will allow to pass unprocessed class arrays to jQuery from other libraries/frameworks that use the array syntax.
4. (This one is a bit of a stretch, but...) If we ever drop IE and start using classList instead of classNames everywhere, this would be a more natural signature and the space-separated values would be the ones that requires more pre-processing.

niltoncsr:
I'd like to contribute and get this one! :smiley:
What do you think, @timmywil?

RUPOJS:
Hi @timmywil, is this already assigned to someone ? I wan to take this up.

Jacabola:
Hi @timmywil, I want to take this up.

jbedard:
Should this add support for arrays (`addClass(["a", "b"])`) or multiple arguments (`addClass("a", "b")`), or both?  The multiple arguments one is more like the [classList methods](https://developer.mozilla.org/en/docs/Web/API/Element/classList#Methods) so that would be my vote.

t-kelly:
Multiple arguments might conflict with `.toggleClass()`'s second boolean argument 🤔

jbedard:
[...]

PR #3727, opened 2017-07-18
added array feature to addClass, removeClass and toggleClass

faisaliyk:
### Summary ###

#3532 
#addClass, removeClass & toggleClass now accepts an array of class names as well as being backward compatible and accepting space separated class names in a single string argument.

### Checklist ###


src/attributes/classes.js
```diff
@@ -23,9 +39,7 @@ jQuery.fn.extend( {
 			} );
 		}
 
-		if ( typeof value === "string" && value ) {
-			classes = value.match( rnothtmlwhite ) || [];
```
mgol: Arrays should just have this assignment skipped in favor of using the first parameter as `classes` and the rest of the code can then stay the same...

src/attributes/classes.js
```diff
@@ -14,7 +14,23 @@ function getClass( elem ) {
 
 jQuery.fn.extend( {
 	addClass: function( value ) {
-		var classes, elem, cur, curValue, clazz, j, finalValue,
+		var classes = [];
+		var proceed = false;
+		if ( typeof value === "string" && value ) {
+			classes = value.match( rnothtmlwhite ) || [];
+			proceed = true;
+		} else if ( value.constructor === Array ) {
```
mgol: Don't check the `constructor` field, just use `Array.isArray`.

src/attributes/classes.js
```diff
@@ -14,7 +14,23 @@ function getClass( elem ) {
...
+			for ( var count = 0; count < value.length; count++ ) {
+				if ( typeof value[ count ] === "string" && value[ count ] ) {
```
mgol: This is not necessary; we don't want to support `.addClass(['a', 'b c'])`; if you pass an array you'll need to have every class as an individual entry in the array.

src/attributes/classes.js
faisaliyk: I will change it to` Array.isArray`. Why shouldn't I use `value.constructor === Array`

src/attributes/classes.js
faisaliyk: I used this code to make it backward compatible. So if instead of an array a string argument is passed to the method, `value.match( rnothtmlwhite ) || [];`  will assing the string parameters as an array to `classes` which the proceeding code can use it to add or remove classes

src/attributes/classes.js
mgol: Checking `constructor` doesn't work cross-realm (e.g. for arrays from an iframe), `Array.isArray` does. That's why we avoid `instanceof` checks. Checking `constructor` is even worse as that would not even work for `Array` subclasses.

src/attributes/classes.js
mgol: It has to be backwards compatible but that is still possible without duplicating the code.

mgol: I've looked at the latest version. How many bytes does it add compared to the current master? You'll see that when you run `grunt`, the minified gzipped.

src/attributes/classes.js
```diff
@@ -14,7 +14,17 @@ function getClass( elem ) {
 
 jQuery.fn.extend( {
 	addClass: function( value ) {
-		var classes, elem, cur, curValue, clazz, j, finalValue,
+		var classes = [];
+		var proceed = false;
```
mgol: The `= false` part is not necessary. The same in other methods.

src/attributes/classes.js
```diff
@@ -51,109 +59,123 @@ jQuery.fn.extend( {
...
 				}
 			}
-		}
-
-		return this;
-	},
 
-	toggleClass: function( value, stateVal ) {
-		var type = typeof value;
+			return this;
```
mgol: Indentation here & everywhere below is wrong, code is indented too much.

test/classArrayTest.html
```diff
@@ -0,0 +1,109 @@
+<!DOCTYPE html>
```
mgol: This file is not necessary, unit tests are enough.

src/attributes/classes.js
faisaliyk: the current jquery.min.js is 85.1 KB

mgol: What's important is the difference (in bytes) to what's currently in master. To check that you need to fire `grunt` on `master` and then on your PR.

I've tried to do that but your PR currently fails on `grunt` because of syntax errors. Please make sure to always run `grunt` before pushing changes and running the tests in at least one browser to make sure they still pass.

faisaliyk: 87,169 bytes is the size of the master
87,258 bytes is the size of my local

I am running grunt on my local every time I hit save. The grunt checks passes. I wonder why they are failing at your end.

dmethvin: @faisaliyk The problem is that the committed files include merge conflict markers. For example see [this line](https://github.com/jquery/jquery/pull/3727/commits/4e62134dc6309293152ec9aef836038ff83eae1f). Perhaps you didn't re-push to the PR branch after resolving the merge conflicts?

faisaliyk: Thanks, @dmethvin. The markers have been removed. I am using VS Code. and for some reason the tab size in the code seems correct and uniform throughout, JShint and grunt passes. However, after the push, the uploaded file shows the tab size to be exaggerated. Any idea how I can resolve this issue?

timmywil:
Thank you for this PR, but closing in favor of https://github.com/jquery/jquery/pull/3917.

PR #3746, opened 2017-08-04
Core: Remove unnecessary check for type when checking for numeric values

everdimension:
### Summary ###
Checking for the type of the passed value here is an unnecessary step.
Converting the value to number with `Number(obj)` (was done implicitly, I made it explicit) and `parseFloat(obj)` eliminates all possible false-positives, and subtraction eliminates infinities.

### Checklist ###

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -226,13 +226,13 @@ jQuery.extend( {
 		// As of jQuery 3.0, isNumeric is limited to
 		// strings and numbers (primitives or objects)
 		// that can be coerced to finite numbers (gh-2662)
-		var type = jQuery.type( obj );
-		return ( type === "number" || type === "string" ) &&
+		return (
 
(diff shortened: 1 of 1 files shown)


gibson042:
This change causes failures in `isNumeric(nonStringNonNumber)` tests. Please run the test suite.

everdimension:
@gibson042 
Running `npm test` passes all tests.

If you have some additional hidden tests, you should better describe how and when to run them.

everdimension:
I found the failing test and I agree that the function should guard for `{ toString() { return '42'; } }`.

But I think it's not ok that running `npm test` did not reveal that.

timmywil:
@everdimension It's a pretty large test suite. We normally just run the tests of the module we're changing. However, there's a [PR](https://github.com/jquery/jquery/pull/3744) open that will allow us to run the unit tests with karma.

PR #3757, opened 2017-08-15
Tests: update path calculation to split at last occurence of "test"

### LINKED ISSUE #3756: Tests: pathname parsing breaks tests when hosted on URLs that include the string "test"
As per https://github.com/jquery/jquery/blob/master/test/jquery.js#L5 which is used in _test/index.html_, it tries to load _/test/unit/ready.js_ and _/dist/jquery.min.js_ dynamically, but since it is just splitting on the string _"test"_ in the pathname and grabbing the first instance, hosting on any url where _"test"_ is located earlier (e.g. _mydomain.com/testcontent/jquery/test/index.html_) incorrectly calculates the path to _mydomain.com/test/unit/ready.js_ instead of _mydomain.com/testcontent/jquery/test/unit/ready.js_ for those resource files and breaks the test. 

Would you be open to me submitting a PR for a fix? Just wanting to change the line in question:
`var path = window.location.pathname.split( "test" )[ 0 ],` 
to 
`var path = window.location.pathname.substring(0, window.location.pathname.lastIndexOf("test")),`

jobeland:
### Summary ###
Fixes #3756 by switching the path calculation from splitting at the first instance of "test", to finding its last index in the pathname, and then taking the substring from from the beginning until that index

### Checklist ###

### DIFF
--- a/test/jquery.js
+++ b/test/jquery.js
@@ -2,7 +2,7 @@
 ( function() {
 	/* global loadTests: false */
 
-	var path = window.location.pathname.split( "test" )[ 0 ],
+	var path = window.location.pathname.substring(0, window.location.pathname.lastIndexOf("test")),
 		QUnit = window.QUnit || parent.QUnit,
 		require = window.require || parent.require,
 
--- a/test/jquery.js
+++ b/test/jquery.js
@@ -2,7 +2,7 @@
 ( function() {
 	/* global loadTests: false */
 
-	var path = window.location.pathname.substring(0, window.location.pathname.lastIndexOf("test")),
+	var path = window.location.pathname.substring( 0, window.location.pathname.lastIndexOf( "test" ) ),
 		QUnit = window.QUnit || parent.QUnit,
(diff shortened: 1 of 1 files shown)


test/jquery.js
```diff
@@ -2,7 +2,7 @@
 ( function() {
 	/* global loadTests: false */
 
-	var path = window.location.pathname.split( "test" )[ 0 ],
+	var path = window.location.pathname.substring( 0, window.location.pathname.lastIndexOf( "test" ) ),
```
gibson042: Having two long chains of property accessors makes this hard to follow. I think we'd be better off with something like
```js
var pathname = window.location.pathname,
	path = pathname.slice( 0, pathname.lastIndexOf( "test" ) ),
```

jobeland: @gibson042 addressed in https://github.com/jquery/jquery/pull/3757/commits/29c2d1258fd68ee804ed4cc5f02e6fcb2916f404. Thanks!

Issue #3743, opened 2017-07-31
Regression: All tests fail (timeout) on Android 4.0


Krinkle:
| Before | After
|--|--
| Commit 36389288e8 | Commit 31783e60f5
| [swarm.jquery.org, 24 Jun 2017](http://swarm.jquery.org/job/5654) |  [swarm.jquery.org, 27 Jun 2017](http://swarm.jquery.org/job/5669)
| <img width="471" alt="screen shot" src="https://user-images.githubusercontent.com/156867/28800040-d9257d48-75ff-11e7-840b-f2991d9a27fd.png"> | <img width="466" alt="screen shot" src="https://user-images.githubusercontent.com/156867/28800043-dc2409c4-75ff-11e7-8b46-8c1b4be96880.png">

<img width="363" alt="screen shot" src="https://user-images.githubusercontent.com/156867/28800108-28608a24-7600-11e7-87fd-ef1cff23fa99.png">

Possibly relevant commits:
* 31783e60f5: `Build: update node dependencies` (gh-3669, @timmywil, @CDAGaming)

According to code review comments, this commit also affected IE8/IE9 in some way. (@mgol)

mgol:
Sometimes such breakages are caused by changes on the BrowserStack side but I've re-run tests on the commit you mentioned and the one before that and results were identical.

So we broke something, indeed.

Marking as a blocker for 3.3.0.

mgol:
Thanks, @Krinkle!

mgol:
I've created a PR fixing this butt at a significant cost: #3759.

It may be a good basis to look for the real problem.

PR #3759, opened 2017-08-16
Build: Fix UglifyJS output in Android 4.0

mgol:
### Summary ###

Enabling the ie8 option makes Android 4.0 run tests properly. Before that it
wasn't rendering anything in the QUnit iframe.

Fixes gh-3743

I don't know what exactly causes the issue without the setting. At first sight I only see unescaped `throws` property and `void 0` instead of `undefined` but just using those two doesn't fail the parsing in Android 4.0: http://jsbin.com/zageloz/edit?html,css,js,console

It'd be worth to investigate longer the real reason for the breakage (perhaps by) but I don't have a lot of time right now. This PR adds 52 bytes to the minified gzipped build which is quite a lot. But even if we don't land this, this PR shows in which direction to look for the real issue.

### Checklist ###

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -165,6 +165,11 @@ module.exports = function( grunt ) {
 					output: {
 						"ascii_only": true
 					},
+
+					// Support: Android 4.0 only
+					// UglifyJS 3 breaks Android 4.0 if this option is not enabled.
+					ie8: true,
+
 					banner: "/*! jQuery v<%= pkg.version %> | " +
 						"(c) JS Foundation and other contributors | jquery.org/license */",
 					compress: {


mgol:
Another option to fix it is to downgrade `grunt-contrib-uglify` from `3.0.1` to `1.0.1` and uninstall the direct `uglify-js` dependency.

mgol:
And example test run on this PR's branch: http://swarm.jquery.org/job/6000. Android 4.0 no longer times out.

timmywil:
Closing in favor of https://github.com/jquery/jquery/pull/3920, which is pretty much the same except it updates uglify and only sets ie8: true on output options.

Issue #3777, opened 2017-09-08
Safari and iOS fail bounded-style relative adjustment


gibson042:
### Description ###
Safari and iOS fail the new test added by 692f9d4db30c9c6c4f6bc76005cf153586202fa6.

### Link to test case ###
http://swarm.jquery.org/job/6075

PR #3778, opened 2017-09-08
CSS: Detect more WebKit styles erroneously reported as percentages

gibson042:
Ref 692f9d4db30c9c6c4f6bc76005cf153586202fa6
Fixes gh-3777

### Summary ###
We already caught Android 4.0 - 4.3 `getComputedStyle` reporting percentage values for margin properties, but the problem also affects other styles, and Safari and iOS.

### Checklist ###


src/css/curCSS.js
```diff
@@ -35,7 +35,7 @@ function curCSS( elem, name, computed ) {
 		// but width seems to be reliably pixels.
 		// This is against the CSSOM draft spec:
 		// https://drafts.csswg.org/cssom/#resolved-values
-		if ( !support.pixelMarginRight() && rnumnonpx.test( ret ) && rmargin.test( name ) ) {
```
dmethvin: So is it okay that `rboxStyle` now catches properties like `borderLeftWidth`? Our CSS unit tests don't seem to do much with the `border*Width` properties so I'm just wondering what effect we'll have in going through this path.

src/css.js
```diff
@@ -447,7 +446,7 @@ jQuery.each( {
 		}
 	};
 
-	if ( !rmargin.test( prefix ) ) {
```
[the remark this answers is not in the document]
Krinkle: Thee regex also matches margin-top, margin-right, etc

src/css/curCSS.js
gibson042: It's fine because the swap logic only kicks in when gCS returns a non-px unitful value. And even if that does happen, the only consequence is taking a slower path (requiring layout calculation) to get pixels.

Issue #3438, opened 2016-12-08
Ajax always processes data for requests without entity body


dmethvin:
I'm creating this ticket based on the discussion in the gh-3409 PR from @Litor so we don't forget to review it.

With ajax requests using HTTP methods that do not have a request body, such as GET, any provided data is [always processed and removed](https://github.com/jquery/jquery/blob/4e50967725e61a5d71c2792131de61dec0520f9c/src/ajax.js#L594-L620), even if the caller specified `processData: false`. This seems contrary to the documentation which says:

> The `data` option can contain either a query string of the form `'key1=value1&key2=value2'`, or an object of the form `{key1: 'value1', key2: 'value2'}`. If the latter form is used, the data is converted into a query string using jQuery.param() before it is sent. This processing can be circumvented by setting `processData` to `false`.

This behavior was also present before 3.x so it's not any kind of recent regression, but it does seem like a bug. I suppose you could argue that *"this processing can be circumvented"* refers only to serializing objects but not to appending the data to the URL, but that interpretation doesn't make sense. Why would someone want `[object Object]` appended to their URL? 😸 

If we didn't process `data` at all, it would be possible for a `beforeSend` handler to take an object and stringify it to the URL as desired , something like this currently non-working example:

http://jsbin.com/fucigaqure/edit?html,console

jaubourg:
The real question is : what is the purpose of setting the `data` option with `processData` `false` for a `GET` request if the data is not to be used in the final URL ?

The `data` option can be set to a string that is already serialized or an object which class provides a `toString` method that serializes the object properly.

Conflating the `processData: false` with "don't use the `data`" seems very wrong to me.

gibson042:
`processData: false` is documented to "prevent this automatic processing [converting `data` to a query string and, for GET requests, appending it to the URL]". There's an argument to be made that `processData` should affect the `jQuery.param` serialization but _not_ the URL manipulation, but that strikes me as confusing (because then `processData: false` doesn't actually prevent manipulation of `data`—or `url`, for that matter) and useless (as Dave points out, "why would someone want `[object Object]` appended to their URL?").

I have a hard time imagining any plausible purpose for `jQuery.ajax(url, { type: "GET", data: nonString, processData: false, … })`, but ignoring `data` (while exposing it to `beforeSend`) seems most in accord with user intentions.

jaubourg:
I hear you but I simply can't agree with a change that will make the following code not work anymore :

```js
$.ajax( {
[...]

PR #3781, opened 2017-09-12
Ajax: Don't process data property on no-entity-body requests

dmethvin:
Fixes gh-3438

### Summary ###
The description in gh-3438 is pretty detailed. Basically, don't process `settings.data` if `settings.processData` is `false`. This allows a `settings.beforeSend` hook to have its way with the original data and, for example, serialize it to JSON on the URL.

I'd consider this a long-standing bug and the current documentation would imply it should have worked this way all along. It should be safe to include in a minor release.

### Checklist ###

* [-] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

### DIFF
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -1273,12 +1273,27 @@ QUnit.module( "ajax", {
 		};
 	} );
 
-	ajaxTest( "jQuery.ajax() - data - no processing ", 1, function( assert ) {
+	ajaxTest( "jQuery.ajax() - data - no processing POST", 1, function( assert ) {
 		return {
 			url: "bogus.html",
 			data: { devo: "A Beautiful World" },
 			type: "post",
 			contentType: "x-special-sauce",
+			processData: false,
+			beforeSend: function( _, s ) {
+				assert.deepEqual( s.data, { devo: "A Beautiful World" }, "data is not processed" );
+				return false;
+			},
+			error: true
+		};
+	} );
+
+	ajaxTest( "jQuery.ajax() - data - no processing GET", 1, function( assert ) {
+		return {
+			url: "bogus.html",
+			data: { devo: "A Beautiful World" },
+			type: "get",
+			contentType: "x-something-else",
 			processData: false,
 			beforeSend: function( _, s ) {
 				assert.deepEqual( s.data, { devo: "A Beautiful World" }, "data is not processed" );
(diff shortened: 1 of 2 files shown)


jaubourg:
👎 see my [comment on the issue](https://github.com/jquery/jquery/issues/3438#issuecomment-330033960)

(and sorry to be late to the party)

gibson042:
I'm [still in favor](https://github.com/jquery/jquery/issues/3438#issuecomment-330110641) of this change, but I observed while combing through the code that modifying an ajax URL in `beforeSend` (as in the [example](http://jsbin.com/fucigaqure/edit?html,console)) breaks our caching logic. We may want to open an issue for ensuring that `cacheURL` always corresponds to the request as issued.

P.S. It's great to see you here, @jaubourg. And you're not late at all.

jaubourg:
Hey @gibson042, nice to see you too :)

timmywil:
What @jaubourg said makes sense to me. It's not that the data should be ignored, it's that it should not be processed. It should still get used if passed. Otherwise, why pass it a GET request at all? Also, I don't see a use case for passing a non string in a `!hasContent` request.

gibson042:
This is my interpretation of desired behavior as discussed in the meeting. It should be implemented in this PR and documented in https://github.com/jquery/api.jquery.com/blob/master/entries/jQuery.ajax.xml#L145 .
> <strong>processData</strong> (default: <code>true</code>)
> Type: <a href="http://api.jquery.com/Types/#Boolean">Boolean</a>
> When true, non-string request <code>data</code> is serialized with <code>$.param</code>, and <code>data</code> is additionally re-encoded for payloads with <code>contentType</code> "application/x-www-form-urlencoded" (the default) or moved into URL query parameters for methods which are issued without a payload (e.g., <code>"GET"</code> and <code>"HEAD"</code>). When false, such processing is skipped (except for the special case of string <code>data</code>, which is still moved into the URL of no-content requests).

test/unit/ajax.js
```diff
@@ -1288,6 +1288,37 @@ QUnit.module( "ajax", {
...
+			processData: false,
+			beforeSend: function( _, s ) {
+				assert.deepEqual( s.data, { devo: "A Beautiful World" }, "data is not processed" );
+				return false;
+			},
+			error: true
+		};
+	} );
+
+		ajaxTest( "jQuery.ajax() - data - process string with GET", 2, function( assert ) {
```
jbedard: indentation?

Issue #3028, opened 2016-03-30
Support ajax script attributes (e.g. for SRI or CSP)


jonathanKingston:
@razamirza mentioned here: https://github.com/jquery/codeorigin.jquery.com/issues/20#issuecomment-203168265

```
Is there any plan to add support for SRI to jQuery.getScript()? http://api.jquery.com/jQuery.getScript/
```

I think it would be worth considering adding support to check a script on the outside, however probably this is worthy of an extension until all browsers support fetch+SRI/WebCrypto natively.

dmethvin:
We recently got a similar request for CSP nonces in #2612 but the submitter went dark. Rather than trying to support all these individually and then needing to verify they each work across all browsers, maybe we could just treat this like headers and give `$.ajax` a general way to put attributes in the script tag.

razamirza:
If my request wasn't clear, here is the exact scenario. We use `jQuery.getScript()` to get script from third party CDNs, and want to check for the integrity of script, so we need something like:

`jQuery.getScript( url [, hash, backupPath ] )`

I'm relatively new to js, but no reason why I can't chip in.

dmethvin:
I'd prefer we just add options to `jQuery.ajax()` that let you set attributes on the tag before it is added to the DOM. So something like this:
```js
$.ajax({
  dataType: "script",
  url: "https://some/path",
  attrs: { nonce: "EDNnf03nceIOfn39fn3e9h3sdfa" },
});
```
which would inject a tag like:
```html
<script nonce="EDNnf03nceIOfn39fn3e9h3sdfa" src="https://some/path">
```
I suppose we could have `jQuery.getScript( url, [, attrs] [, success] )` as well. The implementation would end up mapping any `jQuery.getScript()` call into `jQuery.ajax()` anyway since this additional information has to make it all the way to the script transport.

markelog:
`attrs` attribute would be useless if `dataType` is not `script` correct?

dmethvin:
@markelog Yes, unfortunately it's another transport-specific setting. It would be ignored by transports that didn't need it such as XHR.

dmethvin:
Based on the way the PR is implemented, if `scriptAttrs` is set it will cause a script request to use a script tag, even if it's not cross-domain. The list of attrs can be empty.

PR #3782, opened 2017-09-12
Ajax: Allow custom attributes when script transport is used

dmethvin:
Fixes gh-3028
Ref gh-2612

### Summary ###

We've gotten a couple of tickets asking for a way to set attributes on script tags when they're injected as part of a `jQuery.ajax` request. Rather than defining a bunch of custom new ajax settings I thought it might be easier to just let the caller set any attribute on the script tag.  This would be useful, for example, to add `nonce`, `integrity`, or `crossorigin` attributes so that [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) will be appeased.

Possible todos that aren't in here yet, mainly because I wanted feedback:

- Add a `jQuery.getScript` signature
- I'm debating whether to assume crossDomain when attrs are given.

### Checklist ###

* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com (will be needed)

### DIFF
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -48,19 +48,16 @@ jQuery.ajaxTransport( "script", function( s ) {
 		var script, callback;
 		return {
 			send: function( _, complete ) {
-				script = jQuery( "<script>" ).prop( {
-					charset: s.scriptCharset,
-					src: s.url
-				} ).attr( s.scriptAttrs || {} ).on(
-					"load error",
-					callback = function( evt ) {
+				script = jQuery( "<script>" )
+					.attr( s.scriptAttrs || {} )
+					.prop( { charset: s.scriptCharset, src: s.url } )
+					.on( "load error", callback = function( evt ) {
 						script.remove();
 						callback = null;
 						if ( evt ) {
 							complete( evt.type === "error" ? 404 : 200, evt.type );
 						}
-					}
-				);
+					} );
 
 				// Use native DOM manipulation to avoid our domManip AJAX trickery
 				document.head.appendChild( script[ 0 ] );
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -43,15 +43,15 @@ jQuery.ajaxPrefilter( "script", function( s ) {
 // Bind script tag hack transport
 jQuery.ajaxTransport( "script", function( s ) {
 
-	// This transport only deals with cross domain requests
-	if ( s.crossDomain ) {
+	// This transport only deals with cross domain or forced-by-attrs requests
+	if ( s.crossDomain || s.scriptAttrs ) {
 		var script, callback;
 		return {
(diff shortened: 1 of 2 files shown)


src/ajax/script.js
```diff
@@ -43,15 +43,15 @@ jQuery.ajaxPrefilter( "script", function( s ) {
...
+	// This transport only deals with cross domain or forced-by-attrs requests
+	if ( s.crossDomain || s.scriptAttrs ) {
 		var script, callback;
 		return {
 			send: function( _, complete ) {
 				script = jQuery( "<script>" ).prop( {
 					charset: s.scriptCharset,
 					src: s.url
-				} ).on(
+				} ).attr( s.scriptAttrs || {} ).on(
```
gibson042: There should probably be a line break here.

jaubourg:
That's a very good idea which was rejected back in the days because "YAGNI". Glad to see it back.

I would however :
- use prop rather than attr to set those values (camel-cased props would fit better alongside ajax options),
- use the script prefilter to put the `scriptCharset` option within the `scriptAttrs` object (would bring a clear and fast path to deprecating then removing `scriptCharset`),
- set those props first, before setting the src (so that the url is always the one computed by ajax internally) or remove an eventual `src` from `scriptAttrs` in the prefilter.

gibson042:
I think both calls should use `.attr`, since we want to affect the DOM instead the dynamic state (even though the interesting properties seem to be reflected anyway, I can imagine someone setting "class" or a "data-*" attribute). But @jaubourg is absolutely right about needing to set `scriptAttrs` before `src`.

src/ajax/script.js
```diff
@@ -43,15 +43,15 @@ jQuery.ajaxPrefilter( "script", function( s ) {
...
 jQuery.ajaxTransport( "script", function( s ) {
 
-	// This transport only deals with cross domain requests
-	if ( s.crossDomain ) {
+	// This transport only deals with cross domain or forced-by-attrs requests
+	if ( s.crossDomain || s.scriptAttrs ) {
 		var script, callback;
 		return {
 			send: function( _, complete ) {
 				script = jQuery( "<script>" ).prop( {
```
jbedard: If `.prop(s.scriptAttrs)` is used like @jaubourg suggested maybe a single `.prop` call can be used, although probably depends on size.

Something such as:
```javascript
.prop( $.extend({}, s.scriptAttrs, {src: s.url, charset: s.scriptCharset} )
```

dmethvin:
Sorry about leaving this PR for so long! I've updated it to set `src` last although I would think the browser would not make the request until we'd finished the thread. I feel like the new attrs should definitely use `.attr()` in case they're not reflected. I can change the `.prop()` if someone feels strongly but since it was there and it worked I left it.

PR #3786, opened 2017-09-16
Update to Babel 7, use preset-env instead of preset-es2015

hzoo:
### Summary ###

> I guess this may not be necessary since I see Babel is only used for 1 test file but figured could do it anyway.

Uses Babel 7 instead of v6. Use https://github.com/babel/babel-preset-env instead of preset-es2015 (deprecating). I guess you could just use the for-of plugin specifically, as well. Also you could use `"babel"` key in package.json instead of `.babelrc` if desired. 

```js
babel: {
  options: {
    sourceMap: "inline",
    retainLines: true
  },
  nodeSmokeTests: {
    files: {
      "test/node_smoke_tests/lib/ensure_iterability.js":
        "test/node_smoke_tests/lib/ensure_iterability_es6.js"
    }
  }
}
```
> More info on Babel 7: http://babeljs.io/blog/2017/09/12/planning-for-7.0

### Checklist ###


mgol:
Thanks for the PR!

I like the idea of switching to just use the `for-of` plugin. I'd rather not use `env` here since it's volatile by default and in tests we should be as precise as possible with what we transpile. That's also why I'd rather leave the config in `.babelrc` (since it's not a config of the whole project, just a one test) or perhaps even move it directly where the Babel plugin is called, i.e. the Grunt plugin config.

hzoo:
Since it really is just that test, yes I would just put it in the grunt file, or just don't use the grunt plugin and call babel-core directly.

aside: env isn't volatile by default, if you mean that it will change output randomly when you update, the default behavior is to compile everything to ES5 just like preset-es2015 except that it compiles ES2016,17, etc.

mgol:
> Since it really is just that test, yes I would just put it in the grunt file, or just don't use the grunt plugin and call babel-core directly.

Gruntfile is perhaps the best, since our whole build is based on Grunt calling babel-core directly wouldn't be so easy to integrate with all the other tasks. Unless you've found a way to do it. :)

> aside: env isn't volatile by default, if you mean that it will change output randomly when you update, the default behavior is to compile everything to ES5 just like preset-es2015

I didn't know that! I'd still prefer to be explicit here, though, let's go with just using the `for-of` plugin.

PR #3844, opened 2017-11-14
Make callbacks.fireWith work with non-array arguments

dratini0:
### Summary ###

In jquery/api.jquery.com#1066, I have noticed that a way to call callbacks.fireWith is documented, but doing so doesn't work. This PR fixes that.

### Checklist ###

### DIFF
--- a/src/callbacks.js
+++ b/src/callbacks.js
@@ -206,7 +206,12 @@ jQuery.Callbacks = function( options ) {
 			fireWith: function( context, args ) {
 				if ( !locked ) {
 					args = args || [];
-					args = [ context, args.slice ? args.slice() : args ];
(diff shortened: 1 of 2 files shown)


dratini0:
A demonstration of the bug: https://jsfiddle.net/48Lmdmmf/

gibson042:
`callbacks#fireWith` was introduced in 1.7.0, and as far as we can tell has always required `args` to be an array-like object (which makes sense, since the method is analogous to `function#apply`): https://jsfiddle.net/k7ut4jgz/2/

This looks like a case of incorrect documentation; would you be willing to submit a pull request at https://github.com/jquery/api.jquery.com instead?

PR #3869, opened 2017-12-03
Manipulation: Add support script type module

tbepdb:
Fixes: #3871
### Summary ###

For method  
`$('selector').html() `
append support native ECMAScript modules, like that
`<script type="module" scr="PATH/file.js"></script>` 
and like that
```
<script type="module">
  import * as myModule from "PATH/myModule.js";
</script>

```
### Checklist ###

### DIFF
--- a/src/core/DOMEval.js
+++ b/src/core/DOMEval.js
@@ -9,8 +9,12 @@ define( [
 		var i,
 			script = doc.createElement( "script" );
 		script.text = code;
-		for ( i in { type: true, src: true } ) {
-			script[ i ] = node[ i ];
+		if ( node ) {
+			for ( i in { type: true, src: true } ) {
+				if ( node[ i ] ) {
+					script[ i ] = node[ i ];
+				}
+			}
 		}
 		doc.head.appendChild( script ).parentNode.removeChild( script );
 	}
--- a/src/core/DOMEval.js
+++ b/src/core/DOMEval.js
@@ -3,12 +3,19 @@ define( [
 ], function( document ) {
 	"use strict";
 
-	function DOMEval( code, doc ) {
+	function DOMEval( code, doc, type, src ) {
 		doc = doc || document;
 
 		var script = doc.createElement( "script" );
-
-		script.text = code;
+		if ( code ) {
+			script.text = code;
+		}
+		if ( type === "module" ) {
+			script.type = type;
+		}
+		if ( src ) {
+			script.src = src;
+		}
 		doc.head.appendChild( script ).parentNode.removeChild( script );
 	}
 
--- a/src/core/DOMEval.js
+++ b/src/core/DOMEval.js
@@ -6,19 +6,12 @@ define( [
 	function DOMEval( code, doc, node ) {
 		doc = doc || document;
 
-		var script = doc.createElement( "script" );
-		if ( code ) {
-			script.text = code;
+		var i,
+			script = doc.createElement( "script" );
+		script.text = code;
+		for ( i in { type: true, src: true } ) {
+			script[ i ] = node[ i ];
 		}
-		jQuery.each( [
-			"type",
-			"nomodule",
-			"src"
-		], function() {
-			if ( node[ this ] ) {
-				script[ this ] = node[ this ];
-			}
-		} );
 		doc.head.appendChild( script ).parentNode.removeChild( script );
 	}
 
--- a/src/core/DOMEval.js
+++ b/src/core/DOMEval.js
@@ -3,14 +3,18 @@ define( [
 ], function( document ) {
 	"use strict";
 
-	function DOMEval( code, doc, node ) {
+	var
+
+	preservedScriptAttributes = { type: true, src: true, noModule: true };
+
+		function DOMEval( code, doc, node ) {
 		doc = doc || document;
 
 		var i,
 			script = doc.createElement( "script" );
 		script.text = code;
 		if ( node ) {
-			for ( i in { type: true, src: true } ) {
+			for ( i in preservedScriptAttributes ) {
 				if ( node[ i ] ) {
 					script[ i ] = node[ i ];
 				}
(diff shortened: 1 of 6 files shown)


src/manipulation.js
```diff
@@ -193,14 +193,19 @@ function domManip( collection, args, callback, ignored ) {
 						!dataPriv.access( node, "globalEval" ) &&
 						jQuery.contains( doc, node ) ) {
 
-						if ( node.src ) {
+						if ( node.src && node.type !== "module" ) {
```
gibson042: It's very convenient that module execution is inherently asynchronous with respect to parsing. But I suppose we'll also need awareness of [`nomodule`](https://html.spec.whatwg.org/#attr-script-nomodule) now.

src/core/DOMEval.js
```diff
@@ -3,12 +3,19 @@ define( [
 ], function( document ) {
 	"use strict";
 
-	function DOMEval( code, doc ) {
+	function DOMEval( code, doc, type, src ) {
```
gibson042: See https://github.com/jquery/jquery/pull/3766. I think this is sufficient cause to pass around the node object itself, instead of adding parameter after parameter.

src/core/DOMEval.js
```diff
@@ -3,12 +3,19 @@ define( [
...
 		doc = doc || document;
 
 		var script = doc.createElement( "script" );
-
-		script.text = code;
+		if ( code ) {
+			script.text = code;
+		}
+		if ( type === "module" ) {
```
gibson042: The HTML Standard [specifies](https://html.spec.whatwg.org/#attr-script-type) a case-insensitive match for "module", which we need to respect.

src/core/DOMEval.js
tbepdb: function `globalEval` depends from `DOMEval`, probably we should write another function something like `DOMScript`, that function can receive node object. also this function can include all checks like 
```javascript
	if ( rscriptType.test( node.type || "" ) &&
		!dataPriv.access( node, "globalEval" ) &&
		jQuery.contains( doc, node ) ) {

		if ( node.src && node.type !== "module" ) {
```

gibson042: `DOMEval` is already an internal function, and only invoked at two preexisting sites. We don't need to introduce another function, we just need to update the signature to e.g. `function DOMEval( code, doc, node )` and copy over whitelisted attributes (`src`, `type`, `crossorigin`, `integrity`, etc.) in the body.

src/core/DOMEval.js
```diff
@@ -3,12 +3,22 @@ define( [
...
 	"use strict";
 
-	function DOMEval( code, doc ) {
+	function DOMEval( code, doc, node ) {
 		doc = doc || document;
 
 		var script = doc.createElement( "script" );
-
-		script.text = code;
+		if ( code ) {
```
gibson042: I don't think we need this condition.

src/core/DOMEval.js
```diff
@@ -3,12 +3,22 @@ define( [
...
+			script.text = code;
+		}
+		jQuery.each( [
+			"type",
+			"nomodule",
```
gibson042: `jQuery.each` is almost certainly overkill for what should be `for ( property in scriptProperties )` iteration when `node` is provided, and "nomodule" doesn't belong in the list because it is [ignored](https://html.spec.whatwg.org/#attr-script-nomodule) on module scripts (which are the only cases where we have a node here).

src/manipulation.js
```diff
@@ -193,14 +193,14 @@ function domManip( collection, args, callback, ignored ) {
 						!dataPriv.access( node, "globalEval" ) &&
 						jQuery.contains( doc, node ) ) {
 
-						if ( node.src ) {
+						if ( node.src && !/^module$/i.test( node.type ) ) {
```
gibson042: We don't like inlining regex definitions; this should be `node.type.toLowerCase() === "module"` unless a supported browser has non-string type properties, in which case the regex should be defined outside the function.

test/unit/manipulation.js
```diff
@@ -1757,11 +1757,13 @@ function testHtml( valueObj, assert ) {
 			"<script type='something/else'>ok( false, 'evaluated: non-script' );</script>",
 			"<script type='text/javascript'>ok( true, 'evaluated: text/javascript' );</script>",
 			"<script type='text/ecmascript'>ok( true, 'evaluated: text/ecmascript' );</script>",
+			"<script type='module'>ok( true, 'evaluated: module' );</script>",
 			"<script>ok( true, 'evaluated: no type' );</script>",
 			"<div>",
 				"<script type='something/else'>ok( false, 'evaluated: inner non-script' );</script>",
 				"<script type='text/javascript'>ok( true, 'evaluated: inner text/javascript' );</script>",
 				"<script type='text/ecmascript'>ok( true, 'evaluated: inner text/ecmascript' );</script>",
+				"<script type='module'>ok( true, 'evaluated: inner module' );</script>",
```
gibson042: We should also have a test for a module script with `src`.

mgol: This won't work in browsers not supporting modules, will it? A separate iframe test might be better, especially that these tests are already too big.

timmywil:
I think we would like to separate the module tests from other tests and only run them if modules are supported.

src/core/DOMEval.js
```diff
@@ -3,12 +3,19 @@ define( [
...
+	function DOMEval( code, doc, node ) {
 		doc = doc || document;
 
-		var script = doc.createElement( "script" );
-
+		var i,
+			script = doc.createElement( "script" );
 		script.text = code;
+		if ( node ) {
+			for ( i in { type: true, src: true } ) {
```
gibson042: This object should be a `preservedScriptAttributes` variable, not a literal buried in the function body.

tbepdb:
Browser is skipping scripts by default in case type is unsuported. this means tests wont run in case modules arent supported.

timmywil:
@tbepdb still, we'd like to separate all module-related tests from the others.

PR #3870, opened 2017-12-03
Let default file for browsers be minified for bundlers and services relying on npm

mathiasrw:
When serving to browsers the default file should (as default) be minified. It, therefore, makes sense to use the `browser` field ([see specs](https://github.com/defunctzombie/package-browser-field-spec)) in `package.json` to indicate that `jquery.min.js` should be the default choice when targeting browsers for bundlers or services relying on jQuery from npm.
 
**Example:** At the moment including something like `<script src="//unpkg.com/jquery"></script>` on your website will fetch the original JS. With this PR the minified file will be served. For version 3.2.1 that represents going from 268.039 bytes to 86.659 bytes. Sure, I can just change the script to `<script src="//unpkg.com/jquery@3/dist/jquery.min.js"></script>`. I will argue that too many people are too lazy (or does not know better) to care about the extra 181.380 bytes this will add to the load for each person fetching the file. 

### Checklist ###

---
This change is [Reviewable](https://reviewable.io/reviews/jquery/jquery/3870)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "JavaScript library for DOM operations",
   "version": "3.2.2-pre",
   "main": "dist/jquery.js",
+  "browser": "dist/jquery.min.js",
   "homepage": "https://jquery.com",
   "author": {
     "name": "JS Foundation and other contributors",


roblarsen:
> I will argue that too many people are too lazy (or does not know better) to care about the extra 181.380 bytes this will add to the load for each person fetching the file.

The current Nobel Prize in economics that backs you up on this argument.

dmethvin:
There's a tradeoff between default-debuggable and default-performant here. Which is most typically needed by someone using unpkg? I'd think it's the former.

mathiasrw:
Indeed, I agree about this being a tradeoff. 

When I look for references and posts about how / why unpkg is used it looks like a term "default-performant CDN" best describe how the benefits of unpkg is conveyed. 

Like in https://metafizzy.co/blog/switching-out-cdnjs-for-unpkg/
> CDNJS was created in a previous era, before semver, before npm. It could never completely graduate from it.

> The beauty of unpkg is that it requires no additional work for a library author to maintain. If your project is on npm, it's already on unpkg. 


And in https://york.io/2016/11/30/unpkg-npm-cdn-assets.html
> Now we can stop the deadly sin of committing our assets to git for Bower, CDNJS, and jsDelivr. Instead let’s use unpkg, a CDN for npm.

> It’s simple and beautiful. unpkg is now my preferred CDN.

I will leave this open for a few more days to see if others have inputs on this matter. If no further arguments is presented I will close the PR.

timmywil:
unpkg looks like a great tool for getting something up and running quickly. But however fast it may be, I think there are better options for serving files in production if for no other reason than they cannot provide uptime guarantees. unpkg admits this on their about page:

> However, if you rely on it to serve files that are crucial to your business, you should probably pay for a host with well-supported infrastructure and uptime guarantees.

In light of this, I lean towards default-debuggable, as that seems to be closer to the intended usage of unpkg.

> It's also a great resource for people creating demos and instructional material.

mathiasrw:
In that light it makes sense. Thank you for taking the time for good dialogue.

Issue #3808, opened 2017-10-04
css() reports wrong width/height in Chrome on certain zoom levels


bschwarzent:
jQuery 3.2.1 reports wrong sizes for elements with `box-sizing: border-box` in Chrome (v61) on certain browser zoom levels. It seems to add the element's insets (padding + border) to its width and height, although the CSS width already includes the insets.

### Demonstration ###
http://output.jsbin.com/yipudeloya/

In the provided test case, we have two \<div>s with 10px padding on all sides, 1px border on all sides and a CSS width/height set to 32px. Because box-sizing is set to `border-box`, the visual size of the elements is 32x32px with a remaining "inner" space of 10px. At 100% browser zoom level, this is correctly reported both by `css('width')` and `.outerWidth()`.

If we now zoom to 110%, the numbers get slightly changed, probably due to some rounding errors (31.9886px instead of 32px). Because we should always get "CSS pixels" that are not affected by browser zoom, I would have expected those numbers to stay the same - but I can live with that. The real problem occurs when you reload the page while having the browser zoom at 110%! Now the width is completely incorrect (53.977236000000005px instead of 32px).

If the page uses those values for programmatic layout, the results will be wrong. You can see the effect in the demo. The CSS width and height of the first \<div> are copied to the second, so they should always be the same size. However, if you start at 100%, the second \<div> is much larger.

### Analysis ###
Analysis has shown that this seems to be caused by the internal function `computeStyleTests()` in *css/support.js* that wrongly sets the internal property `boxSizingReliableVal` to `false`. Apparently, this happens because of a rounding error. In my case, the test div's measured width is 3.997px, which is not equal to the expected value of 4px. What's worse, this internal property is only calculated once! So when you start at 110% zoom, and then reset it to 100%, the problem still persists until you completely reload the page.

I could observe the problem at the following zoom levels: 110%, 90%, 80%, 67% and 33%.

Firefox (v56) and IE (v11) don't seem to be affected by this problem. Also, with jQuery 2.1.4, everything works fine on all browsers. All tests were performed on Windows 7 with default DPI and font size settings.

gibson042:
Ugh, confirmed on the git build. I think we need something like a unit-preserving round comparison. Fortunately, this is reproducible with `html { zoom: 1.1 }`, so we'll be able to verify a fix with unit testing.

simonemazzoni:
Same here, @gibson042 any news about it?
Up to now I fixed downgrading to version `3.1.1`.

[...]

PR #3872, opened 2017-12-04
Fix the width calculation of border-box due to browser zoom

SaptakS:
### Summary ###
The pull request mainly has two tasks done:
- Test for the issue (#3808 )
- Fixing the issue by rounding off the measure to unit place.

However the test for the issue isn't working. I used an iframe test to check for the border-box width at `html {zoom: 1.1}`. Though opening the html file separately recreates the issue, but via the unittest code, it doesn't. 

Some help in fixing the unittest will be really helpful.

### Checklist ###

### DIFF
--- a/src/css/support.js
+++ b/src/css/support.js
@@ -30,16 +30,16 @@ define( [
 		pixelPositionVal = divStyle.top !== "1%";
 
 		// Support: Android 4.0 - 4.3 only, Firefox <=3 - 44
-		reliableMarginLeftVal = divStyle.marginLeft === "12px";
+		reliableMarginLeftVal = roundPixelMeasures( divStyle.marginLeft ) === 12;
 
 		// Support: Android 4.0 - 4.3 only, Safari <=9.1 - 10.1, iOS <=7.0 - 9.3
 		// Some styles come back with percentage values, even though they shouldn't
 		div.style.right = "60%";
-		pixelBoxStylesVal = divStyle.right === "36px";
+		pixelBoxStylesVal = roundPixelMeasures( divStyle.right ) === 36;
 
 		// Support: IE 9 - 11 only
 		// Detect misreporting of content dimensions for box-sizing:border-box elements
-		boxSizingReliableVal = divStyle.width === "36px";
+		boxSizingReliableVal = roundPixelMeasures( divStyle.width ) === 36;
 
 		// Support: IE 9 only
 		// Detect overflow:scroll screwiness (gh-3699)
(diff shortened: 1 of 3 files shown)


src/css/support.js
```diff
@@ -53,6 +53,11 @@ define( [
 		div = null;
 	}
 
+	// Rounding up the exact checking match values
+	function roundPixelMeasures( measure ) {
+		return Math.round( parseFloat( measure ) ) + "px";
```
gibson042: Adding "px" seems pointless, we should just return and compare against the numeric value (saving a few bytes).

src/css/support.js
```diff
@@ -53,6 +53,11 @@ define( [
 		div = null;
 	}
 
+	// Rounding up the exact checking match values
```
gibson042: I find this comment confusing. And given the clarity of the function body, it can probably just be dropped.

test/data/css/cssWidthBrowserZoom.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="utf-8">
+	<style>
+		html {
+			zoom: 1.1
```
gibson042: This should have a semicolon for consistency.

SaptakS:
@gibson042 will fix these. But the problem is the unittest actually doesn't reproduce the error. So we won't be able to confirm in future if this feature is broken. Can you please help in that as well?

gibson042:
The unit test worked for me in Chromium 62 on Linux... assertion failed with the old code and passed with the changes.

src/css/support.js
SaptakS: I dropped the comment as suggested by you.

src/css/support.js
SaptakS: Removed "px" and used the numeric values to compare, since it saves a few bytes in both the places as suggested by you.

test/data/css/cssWidthBrowserZoom.html
SaptakS: Resolved ✔️

SaptakS:
@gibson042 made the changes as suggested by you.

veered:
When will this change be published to NPM?

gibson042:
Later this month, as part of jQuery 3.3.

nyurik:
Hi, is there any hope for backporting it to the older versions?

dmethvin:
You could certainly try to back-port it. We're not supporting older versions so if you want a supported version you should upgrade.

nyurik:
@dmethvin is there any maintenance for the v2.x ?  Of course updating to the latest version is preferred, but in some cases some projects are simply too large to port overnight, when some minor issue like this one is discovered.

dmethvin:
You're always welcome to port over a "minor issue" for your own use. There's more work than you might think to maintain an old version, especially when it comes to test infrastructure.

PR #3874, opened 2017-12-04
Fixes #3384: deprecate jQuery.camelCase

### LINKED ISSUE #3384: Deprecate jQuery.camelCase, hide in 4.0
### Description ###

We should hide `jQuery.camelCase`. It's an undocumented, private method which doesn't make sense to be available externally as it contains a CSS-only quirk that replaces `-ms-` to `ms-`. Rather than fixing it it's better to just remove it.

Theoretically we could do it in any minor release but since it's been semi-public for so long it's better to wait, especially that code around it will have to change in 4.0 anyway because of #3355.

### Link to test case ###
N/A

timmywil:
### Summary ###
Makes adjustments to #3604 to fix #3384.

### Checklist ###

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -45,7 +45,6 @@ var
 
 	// checked="checked" or checked
 	rchecked = /checked\s*(?:[^=]|=\s*.checked.)/i,
-	rscriptTypeMasked = /^true\/(.*)/,
 	rcleanScript = /^\s*<!(?:\[CDATA\[|--)|(?:\]\]|--)>\s*$/g;
 
 // Prefer a tbody over its parent table for containing new rows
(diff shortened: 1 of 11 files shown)


src/manipulation.js
```diff
@@ -45,7 +45,6 @@ var
 
 	// checked="checked" or checked
 	rchecked = /checked\s*(?:[^=]|=\s*.checked.)/i,
-	rscriptTypeMasked = /^true\/(.*)/,
```
dmethvin: The changes in this file seem unrelated to the PR?

timmywil: Indeed. I didn't see that before. Nice catch!

timmywil:
Merged in https://github.com/jquery/jquery/commit/775caebd617e91e697b3e41ae33d6877e63785aa

Issue #3107, opened 2016-05-05
`.position()` and `.offsetParent()` return incorrect value


anseki:
`.position()` method should return a position of element relative to a closest ancestor element that is positioned.
Therefore, for example, if CSS properties `top/left` of the element that has `position:absolute` are set to that returned position, the element should not move.
But when the ancestor element that is positioned does not exist (i.e. all ancestor elements have `position:static`, this is default), a position relative to `<html>` element incorrectly is returned.
For example:
https://jsfiddle.net/fqbgy5pq/

Result:

```
ClientRect   top: 63   left: 63
CSS props    top: auto left: auto
.position()  top: 59   left: 59
Set return value of `.position()` to `top` and `left`, then the element should not move.
ClientRect   top: 59   left: 59
CSS props    top: 59px left: 59px
.position()  top: 55   left: 55
```

**Note:** `2px` was lost by this issue, and more `2px` was lost by [another issue of `.offset()`](https://github.com/jquery/jquery/issues/3080).
When fixed `.offset()` is used:

Result:

```
ClientRect   top: 63   left: 63
CSS props    top: auto left: auto
.position()  top: 61   left: 61
Set return value of `.position()` to `top` and `left`, then the element should not move.
ClientRect   top: 61   left: 61
CSS props    top: 61px left: 61px
.position()  top: 59   left: 59
```

`.offsetParent()` method that is called by `.position()` method returns `<html>` when ancestor element that is positioned is not found. Then, `.position()` mistakes.
https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/offset.js#L180
I think that `.offsetParent()` should return `document` or `null` in this case.
`document` means "base of coordinates".
`null` means "ancestor element is not found".

Also, `.position()` method should return coordinates relative to the document in this case.

And also, `.position()` method returns incorrect coordinates when `<html>` has `position:non-static` (i.e. the ancestor element is really `<html>`).
For example:
https://jsfiddle.net/wczL6ae5/

Result:

```
ClientRect   top: 63   left: 63
CSS props    top: auto left: auto
.position()  top: 59   left: 59
Set return value of `.position()` to `top` and `left`, then the element should not move.
ClientRect   top: 62   left: 62
CSS props    top: 59px left: 59px
.position()  top: 58   left: 58
```

**Note:** `1px` was added by this issue, and `2px` was lost by [another issue of `.offset()`](https://github.com/jquery/jquery/issues/3080).
[...]

PR #3887, opened 2017-12-12
Position: Fixes typo in position() logic to retrieve offsetParent

jlukic:
There's a typo in position. The while statement should be walking up the DOM tree finding the first element with position static. 

Currently the loop that finds the `offsetParent` it is exiting immediately because the first parent node isnt document root or body. 

The while loop should check to ensure each iteration of getting `parentNode` doesn't reach the top of the document, instead of checking that each iteration `===` the document.

Related issues 
#3080 #3107  #3096 #3487 
Related code in issue https://github.com/jquery/jquery/issues/1765#issuecomment-69847034 

### Summary ###
Bug was introduced in 1d2df772b4d6e5dbf91df6e75f4a1809f7879ab0

The logic for finding `offsetParent` was modified in position, but the while has incorrect exit conditions which would cause it to exit immediately.

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -131,7 +131,7 @@ jQuery.fn.extend( {
 			doc = elem.ownerDocument;
 			offsetParent = elem.offsetParent || doc.documentElement;
 			while ( offsetParent &&
-				( offsetParent === doc.body || offsetParent === doc.documentElement ) &&
+				( offsetParent !== doc.body || offsetParent !== doc.documentElement ) &&
 				jQuery.css( offsetParent, "position" ) === "static" ) {
 
 				offsetParent = offsetParent.parentNode;
--- a/src/offset.js
+++ b/src/offset.js
@@ -131,7 +131,7 @@ jQuery.fn.extend( {
 			doc = elem.ownerDocument;
 			offsetParent = elem.offsetParent || doc.documentElement;
 			while ( offsetParent &&
-				( offsetParent !== doc.body || offsetParent !== doc.documentElement ) &&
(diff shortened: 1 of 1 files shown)


gibson042:
I believe the code is already correct. Can you provide a test case that shows the failure?

dmethvin:
When I run this PR in a browser, it fails a bunch of existing tests. @jlukic were you able to run the unit tests locally with this PR?

timmywil:
This is pretty old now.

timmywil:
Actually, after investigating this further and seeing https://jsfiddle.net/timmywil/3xpnfccL/3/ from #3972, I'm inclined to think something is wrong here. We're not calculating offsetParent in jQuery#position the same way as jQuery#offsetParent. In the test case linked, the native offset parent in Chrome is the `td`, which doesn't seem right since position is static on that element. I see 2 possibilities:

1) I'm missing something and `td` should indeed be the offsetParent. If that's the case, jQuery#offsetParent is wrong.
2) `td` should not be the offsetParent and jQuery#position is wrong.

gibson042:
@timmywil see #3479.

timmywil:
I see, so we're going to update jQuery#offsetParent to match the browser. In that case, this PR is not needed.

jlukic:
Just to comment on this, the important thing with this change is to look at the actual looping construct for the search. 

It should be checking each parent node and exiting when it reaches top node if a match isn't found.  

I believe it is accidentally checking for the opposite conditions, which was probably just a typo in the original commit.

Issue #3653, opened 2017-05-01
Deprecate `cssProps` property


markelog:
After https://github.com/jquery/jquery/pull/3569/files it will became useless basically and since we never documented we can remove it.

But it's better to be on the safe side and deprecate it first and remove it in the next major

timmywil:
I did the PR for this, but @mgol noted that accessing the object directly rather than with `jQuery.cssProps` has the potential to break the case where `jQuery.cssProps` is replaced rather than amended. Also, I couldn't find any mention of this property in our documentation. I'm inclined to just leave this as-is since if we can't move it to deprecated js and there's nothing to be done in migrate/API, there's nothing to do here.

markelog:
So do we intended to expose it in the next major as well? If not should we create an issue with "Remove jQuery.cssProps"?

dmethvin:
It looks to me that this isn't one of those "documented undocumented" features. I didn't see any sign that it had been discussed on StackOverflow for example. Unfortunately for searching the names `cssprop` and `cssprops` are somewhat common as variable names so it's possible I missed some. 

I'm fine with either way as far as removal for 4.0, it seems safe. I just created an issue for Migrate.

PR #3892, opened 2017-12-18
CSS: deprecate jQuery.cssProps

timmywil:
Fixes gh-3653

### Summary ###
Deprecates the now empty jQuery.cssProps. It's still used for the vendor prefix cache, but no longer has any properties to map to different names for each browser.

### Checklist ###

### DIFF
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -1,8 +1,9 @@
 define( [
 	"./core",
 	"./core/nodeName",
-	"./var/isWindow"
-], function( jQuery, nodeName, isWindow ) {
+	"./var/isWindow",
+	"./css/var/cssProps"
+], function( jQuery, nodeName, isWindow, cssProps ) {
 
 "use strict";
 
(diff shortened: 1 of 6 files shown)


src/deprecated.js
```diff
@@ -38,5 +39,6 @@ jQuery.isArray = Array.isArray;
 jQuery.parseJSON = JSON.parse;
 jQuery.nodeName = nodeName;
 jQuery.isWindow = isWindow;
+jQuery.cssProps = cssProps;
```
mgol: I'm afraid it might be a breaking change. The fact that the object is now injected instead of used directly from `jQuery.cssProps` means that if someone overwrites the object completely it will stop working.

Perhaps we should just move defining this object here and in source, where it's used, make sure code still works fine if it's not defined.

Or just deprecate in docs without touching the code base. Whatever you prefer.

mgol: The same for an API ticket.

timmywil: Good point. Maybe there's nothing to be done here. cssProps brought up no search results on api.jquery.com, so I'm not sure it's actually mentioned anywhere. Maybe there's nothing to do here?

PR #3894, opened 2017-12-18
Build: Only run browser tests in one Node version on Travis

mgol:
Ref gh-3744

### Summary ###

Only run browser tests in one Node version on Travis. This reduces the possibility of failure.

### Checklist ###

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,12 @@ node_js:
 - "9"
 addons:
   chrome: stable
+env:
+  - NPM_SCRIPT=test:browserless
+matrix:
+  include:
+    # Run browser tests only on one Node.js version to save time.
+    - node_js: "8"
+      env: NPM_SCRIPT=test:browser
+script:
+  - npm run $NPM_SCRIPT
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,3 +7,12 @@ node_js:
...
   chrome: stable
+env:
+  - NPM_SCRIPT=test:browserless
+matrix:
+  include:
+    # Run browser tests only on one Node.js version to save time.
+    - node_js: "8"
+      env: NPM_SCRIPT=test
+script:
+  - npm run-script $NPM_SCRIPT
```
Krinkle: Might want to use `npm run` here for consistency with the other callers.

mgol: Done.

mgol:
@Krinkle I modified it as some tests were running twice. Now all Nodes run browserless tests and - *in addition* to that - one Node runs browser tests *only*. PTAL.

Krinkle:
@mgol Yep, makes sense. Also cuts down total time for Travis CI from 13min to 9min.

Issue #3605, opened 2017-04-02
Deprecate jQuery.type


gibson042:
### Description ###
`jQuery.type` is insufficient to deal with the ever-increasing list of @@toStringTag values (cf. gh-3553, gh-3600). But since it is intended for internal use anyway, so we should treat it just like [`jQuery.isNumeric`](https://github.com/jquery/jquery/issues/2960), [`jQuery.proxy`](https://github.com/jquery/jquery/issues/2958), [`jQuery.now`](https://github.com/jquery/jquery/issues/2959), and [`jQuery.camelCase`](https://github.com/jquery/jquery/issues/3384) by removing it from the public API.

ShashankaNataraj:
PRs accepted? I'd like to contribute to this!

timmywil:
@ShashankaNataraj absolutely!

PR #3895, opened 2017-12-18
Core: deprecate jQuery.type

jbedard:
Fixes gh-3605

### Summary ###
Deprecates `type`, renames the internal one to `toType`.

### Checklist ###

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -18,10 +18,11 @@ define( [
 	"./var/support",
 	"./var/isFunction",
 	"./var/isWindow",
-	"./core/DOMEval"
+	"./core/DOMEval",
+	"./core/toType"
 ], function( arr, document, getProto, slice, concat, push, indexOf,
 	class2type, toString, hasOwn, fnToString, ObjectFunctionString,
-	support, isFunction, isWindow, DOMEval ) {
+	support, isFunction, isWindow, DOMEval, toType ) {
 
 "use strict";
 
@@ -237,17 +238,6 @@ jQuery.extend( {
 		return true;
 	},
 
-	type: function( obj ) {
-		if ( obj == null ) {
-			return obj + "";
-		}
-
-		// Support: Android <=2.3 only (functionish RegExp)
-		return typeof obj === "object" || typeof obj === "function" ?
-			class2type[ toString.call( obj ) ] || "object" :
-			typeof obj;
-	},
-
 	// Evaluates a script in a global context
 	globalEval: function( code ) {
 		DOMEval( code );
@@ -395,7 +385,7 @@ function isArrayLike( obj ) {
 	// hasOwn isn't used here due to false negatives
 	// regarding Nodelist length in IE
 	var length = !!obj && "length" in obj && obj.length,
-		type = jQuery.type( obj );
+		type = toType( obj );
 
 	if ( isFunction( obj ) || isWindow( obj ) ) {
 		return false;
--- a/src/core/toType.js
+++ b/src/core/toType.js
@@ -1,0 +1,20 @@
+define( [
+	"../var/class2type",
+	"../var/toString"
+], function( class2type, toString ) {
+
+"use strict";
+
+function toType( obj ) {
+	if ( obj == null ) {
+		return obj + "";
+	}
+
+	// Support: Android <=2.3 only (functionish RegExp)
+	return typeof obj === "object" || typeof obj === "function" ?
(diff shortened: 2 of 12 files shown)


src/core/toType.js
```diff
@@ -0,0 +1,25 @@
+define( function() {
+	"use strict";
+
+	var class2type = {};
+
+	// Populate the class2type map
+	jQuery.each( "Boolean Number String Function Array" +
```
mgol: You need to add `../core` to the dependencies as `jQuery`.

src/core.js
```diff
@@ -268,17 +268,6 @@ jQuery.extend( {
 		return true;
 	},
 
-	type: function( obj ) {
```
mgol: You still need to define `jQuery.type` in `deprecated.js`.

src/core/toType.js
jbedard: Hmm, I was also debating changing this to just hardcode the values. I think min/gziped it's even smaller. Then we won't have the dependency on `../core`. (I had this change locally but decided to revert it for this pr, which probably broke some things...).

jbedard:
@mgol I believe this is fixed up now

src/core/toType.js
```diff
@@ -0,0 +1,28 @@
+define( function() {
+	"use strict";
+
+	var class2type = {
```
mgol: We usually skip the first indentation so that almost whole files are not indented and it's easier to later migrate to ES6 modules without changing virtually all the lines.

Some smaller modules don't follow this pattern but most are. So please de-indent it all by one tab.

test/unit/deprecated.js
```diff
@@ -165,6 +165,67 @@ QUnit.test( "jQuery.nodeName", function( assert ) {
 	);
 } );
 
+QUnit.test( "core", function( assert ) {
```
mgol: `basic.js` is supposed to cover very basic tests from all modules so that we can run it (in the future) in some restricted environments like Node.js with jsdom. All its assertions should already be covered by other modules so it doesn't make sense to move it to `deprecated.js`.

Even its name is misleading here as only `jQuery.type` assertions are left.

src/core/toType.js
```diff
@@ -0,0 +1,28 @@
...
+		"[object Boolean]": "boolean",
+		"[object Number]": "number",
+		"[object String]": "string",
+		"[object Function]": "function",
+		"[object Array]": "array",
+		"[object Date]": "date",
+		"[object RegExp]": "regexp",
+		"[object Error]": "error",
+		"[object Symbol]": "symbol"
+	};
```
mgol: Did you make sure listing all those entries directly is smaller than the loop?

jbedard: +10 minified, -3 non-minified

I guess I'll switch it back?

mgol: Minified gzipped, I assume? Yes, in that case it's better to switch back.

jbedard: Yeah those are both gzip numbers, without gzip the loop one is always smaller.

jbedard: Except... core.js depends on this, so it looks like we can't also depend on core.js...

jbedard:
Migrate: https://github.com/jquery/jquery-migrate/issues/282
API: https://github.com/jquery/api.jquery.com/issues/1076

jbedard:
I've updated this to try to minimize the change and reduce the extra bytes. Now this essentially just moves `jQuery.type` to core/toType.js. Originally I also inlined the `class2type` variable in the new file, but that forces the variable to also be initialized there, that variable is also used in [some](https://github.com/jquery/jquery/blob/6c38ebbd47c6b0fa654733819bd5ae36c1ac6c48/src/var/hasOwn.js#L6) [other](https://github.com/jquery/jquery/blob/6c38ebbd47c6b0fa654733819bd5ae36c1ac6c48/src/var/toString.js#L6) places which increases the changes (and gzip-bytes). Moving the `class2type` can be considered later. The [original version](https://github.com/jbedard/jquery/commit/bed23d1930803fcfc130982ff7b5c6d87d0ecc94) for reference.

Unfortunately this does still increase the size a bit. I'm surprised since it mainly just removes `jQuery.` and moves the `toType` up top
```
   raw     gz Compared to master @ 6c38ebbd47c6b0fa654733819bd5ae36c1ac6c48
   -16     +6 dist/jquery.js
   -23     +9 dist/jquery.min.js
```

PR #3897, opened 2017-12-22
 Add support for ES7 async functions 

rafis:
### Summary ###

Make the following code work:
```javascript
$(async function() {
  console.log('Hello, ES7 async');
});
```
Currently it thinks that it is not a function and turns it into array.

### Checklist ###

### DIFF
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -439,6 +439,13 @@ QUnit.test( "isFunction", function( assert ) {
 	fn = function() {};
 	assert.ok( jQuery.isFunction( fn ), "Normal Function" );
 
+	// Make sure ES7 async functions also work.
+	// Because our code should work in ES5 we can't define `var asyncFn = async function() {};` directly.
+	function AsyncFunction() { }
+	AsyncFunction.prototype[ Symbol.toStringTag ] = "AsyncFunction";
+	var asyncFn = new AsyncFunction();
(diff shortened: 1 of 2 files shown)


test/unit/core.js
```diff
@@ -439,6 +439,13 @@ QUnit.test( "isFunction", function( assert ) {
 	fn = function() {};
 	assert.ok( jQuery.isFunction( fn ), "Normal Function" );
 
+	// Make sure ES7 async functions also work.
+	// Because our code should work in ES5 we can't define `var asyncFn = async function() {};` directly.
+	function AsyncFunction() { }
```
mgol: I'd rather test a real async function; this test more or less repeats the implementation so it's not a full test to me.

Also, this should be a separate `QUnit.test`.

So that this test doesn't crash in old browsers, `eval` should be used to define an async function and the test should only be run if async functions are supported. Other browsers should have `QUnit.skip` called, see [this core test](https://github.com/jquery/jquery/blob/ecd8ddea33dc40ae2a57e4340be03faf2ba2f99b/test/unit/core.js#L383-L388) as an example.

gibson042:
This case is already handled on master since a16339b8933f115da3661f3d3f64854c3fa60bdc. There's even a unit test specifically covering it: https://github.com/jquery/jquery/commit/a16339b8933f115da3661f3d3f64854c3fa60bdc#diff-023537a1eb6e158db4d12a1fe86f1128R521

Issue #3682, opened 2017-05-31
jQuery.ajax() with cache: false is still messes up the url in jQuery 3.2.0


peeete:
### Description ###
jQuery.ajax() with cache: false is still messes up the url in jQuery 3.2.0 (not fixed via #3229):

please have a look at the test case:  jQuery.ajax will turn an URL that ends with "?_=12345..." into one that ends with "?&_=12345.."

### Link to test case ###
https://jsfiddle.net/0a7qyu9z/

wenz:
sorry for the extra pull requests, I was struggeling a bit this morning ;) 
Looking forward to feedback!

marcus-hiles:
Hi @peeete, just wondering why you would want to use "_" alone as a parameter key in the url?

wenz:
I've seen some frameworks do that for state management params (not that I would consider this best practice, but it's still a valid URL)

peeete:
Hi @marcus-hiles, if I recall correctly, I just happened to call $.ajax() with an URL that already got the "_" parameter from an earlier $.ajax() call. I see that that might just be a poorly implementation from my side, but it's a bug none the less.

marcus-hiles:
Oh ok, gotcha @peeete 👍 ... yeah poor implementation indeed, this might have been working in previous browser versions, but recent browser versions are really uping their game in terms of security

mgol:
In 5.0 we could try using the `URLSearchParams` API.

PR #3902, opened 2017-12-22
Ajax: Create correct URLs for jQuery.ajax() with cache: false

wenz:
fixes gh-3682

### Summary ###
Correctly creates the URL for calls to jQuery.ajax() when the "cache" option is set to false. Fixes gh-3682.

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -5,20 +5,23 @@ define( [
 	"./ajax/var/location",
 	"./ajax/var/nonce",
 	"./ajax/var/rquery",
-
+	"./ajax/var/rqueryend",
+	"./ajax/var/rampersandstart",
 	"./core/init",
 	"./ajax/parseXML",
 	"./event/trigger",
 	"./deferred",
 	"./serialize" // jQuery.param
-], function( jQuery, document, rnothtmlwhite, location, nonce, rquery ) {
+], function( jQuery, document, rnothtmlwhite, location, nonce, rquery, rqueryend,
+	rampersandstart ) {
 
 "use strict";
 
 var
 	r20 = /%20/g,
 	rhash = /#.*$/,
-	rantiCache = /([?&])_=[^&]*/,
+	rantiCacheQuery = /\?_=[^&]*(&?)/,
+	rantiCacheFragment = /&_=[^&]*(&?)/,
 	rheaders = /^(.*?):[ \t]*([^\r\n]*)$/mg,
 
 	// #7653, #8125, #8152: local protocol detection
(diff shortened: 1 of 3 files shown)


gibson042:
Is `?&_=…` actually a problem? I know it's unusual, but it's also perfectly valid and properly interpreted by every server framework of which I'm aware.

We might accept a small fix regardless, but this PR increases the file size a _lot_:
```
   raw     gz Sizes
272031  80593 dist/jquery.js
 87420  30244 dist/jquery.min.js

   raw     gz Compared to master @ ecd8ddea33dc40ae2a57e4340be03faf2ba2f99b
  +336   +110 dist/jquery.js
   +91    +51 dist/jquery.min.js
```

You might be able to get it back down with a URL manipulation function, or with wicked clever regexes (one matching the part of a URL preceding a `_` query parameter¹ and the other matching the longest "complete" prefix from a URL² for identifying a query append strategy)—but the savings would have to be _dramatic_ to justify such increased complexity. However, I'd be willing to consider anything that passes all the tests from my [example](https://jsfiddle.net/nz6sxm0z/).

¹ `/^([^?]*(?=\?)(.(.*(?=&)|)|).)_=[^&]*&?/`
² `/^([^?]*(?=\?)(.(.*(?=&).|)$|))/`

wenz:
Thanks for the feedback! The ?&_ URL is perfectly valid, but "does not look good". I'll have another pass at it with more Regex magic. We might also be able to save a few bytes if more or all regexes are defined within the IIFE (currently, some are defined outside the function and passed as an argument, others are defined inside).

gibson042:
If all we're talking about is aesthetics, I'd rather not add _any_ size or complexity, but will definitely take a look. Another possible approach for "clean" parameter appending might be something like
```js
(url + "?").replace(//,"") + "name=value"
```

wenz:
So I've messed around with your approach and also a few other things. Before I start cleaning up for a PR, just that we are on the same page, I was wondering about the correct test results for some of your tests in the fiddle. 

- What is the correct result for the five "tricky path" tests? My take: since the cash buster parameter is appended to the query string, tricky-path&\_=OLD for instance needs to result in tricky-path&\_OLD?\_=NEW (since there was no question mark in the test URL)
- What is the correct result for the "tricky query" test? I'd assume the second question mark ought to be escaped (%3F) for the URL to be correct, but what happens if we get this URL? My take: the first question mark starts the query string, and we have name-value pairs there, with = separating names from values. Therefore, the parameter name would be "tricky-query?\_", and the value would be "OLD". The result should then be "tricky-query?\_=OLD&\_=NEW". In that case, your approach would pass all tests. :-) Alternatively, we could argue that the last question mark starts the query string (the regex can be changed easily to accomodate that), but that does not seem to be right.

gibson042:
> What is the correct result for the five "tricky path" tests? My take: since the cash buster parameter is appended to the query string, tricky-path&\_=OLD for instance needs to result in tricky-path&\_OLD?\_=NEW (since there was no question mark in the test URL)

Right. The table in the [fiddle](https://jsfiddle.net/nz6sxm0z/) is all correct, and in fact optimal with the exception that the output from `tricky-path&_=OLD?&` could be `tricky-path&_=OLD?_=NEW` instead of `tricky-path&_=OLD?&_=NEW`.

> What is the correct result for the "tricky query" test? I'd assume the second question mark ought to be escaped (%3F) for the URL to be correct, but what happens if we get this URL?

It is valid to escape question marks in the query component, [but not necessary](https://tools.ietf.org/html/rfc3986#section-3.4).

> My take: the first question mark starts the query string, and we have name-value pairs there, with = separating names from values. Therefore, the parameter name would be "tricky-query?\_", and the value would be "OLD". The result should then be "tricky-query?\_=OLD&\_=NEW".

Correct.

> In that case, your approach would pass all tests. :-)

Yes, sorry I didn't make that clear.

> Alternatively, we could argue that the last question mark starts the query string (the regex can be changed easily to accomodate that), but that does not seem to be right.

A non-percent-encoded question mark is not allowed in the path or any URI earlier component, so by definition the first such character starts the query component.

timmywil:
Still needs updates per @gibson042's fiddle and requests.

Issue #3214, opened 2016-07-04
Deprecate event shorthand methods


gibson042:
As [discussed](http://irc.jquery.org/%23jquery-meeting/latest.log.html#t12:54:24) in the core meeting, we should deprecate `on`/`trigger` [event shorthand methods](https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/event/alias.js) in a pre-4.0 minor release.

john2014:
Please DO NOT do this. A huge amount of code, especially in top 1 million websites, relies on a LOT of these event shorthands. The reason .load() was deprecated was because it was ambiguous. Not only these remaining shorthand have no ambiguity but is actually much cleaner code. Do not fix something that is not broken.

A good example:  We would have to change $("#div1").focus()  to $("#div1").trigger("focus");
$("#div1").focus(function(){ DoSomething() }) to  $("#div1").on(""focus",function(){ DoSomething() }));

Please do not break such important shorthands just because one person complained about it. The biggest reason to continue to use any library is its backward compatibility. If it keeps breaking stuff for ZERO apparent gain, then many users will either seriously consider moving to different library or stay with older version. I am guessing jquery does not want that.

In addition, there are millions of jquery plugins which uses these shorthand and there is no way every single one of them would be updated. This would break the entire jquery ecosystem.

I hope you guys see the magnitude of this change. Not only do users not gain from this; but it will break lots of stuff for no reason.

scottgonzalez:
I'm +1 on deprecating, -1 on ever removing from the default build. It's already easy to exclude in a custom build, but the deprecation is a good signal to developers that they should aim to not use them.

timmywil:
I'm not a fan of deprecating something we don't plan on removing. Considering that, I might be a -1 on deprecating these. They're just too prevalent to remove, and I don't see the downside of using them if you like them. I prefer `.on` too, but I have no disadvantages in mind to back up a harangue for using an alias.

scottgonzalez:
The reason I'm in favor of deprecating, but not removing is that it will encourage plugin developers to support users who want to exclude the event alias module. This is the only module, besides the deprecated module, that we support excluding with jQuery UI.

I've also personally disliked these methods for many years because they operate in the reverse order of what's expected for the trigger scenario (the primary action is firing an event, not taking an action like the native equivalents). The result is close enough to the expected, native behavior that most developers don't notice, but it's definitely been a source of pain for a long time.

[...]

PR #3928, opened 2018-01-08
Event: Move event aliases to deprecated

dmethvin:
Fixes gh-3214

### Summary ###

No functional change to the built file, just moving this to deprecated.js for now. Probably won't be removed in 4.0 either but can be excluded easier this way. Plugins should be discouraged from using shortcuts so that they can work with the slim build.

Still need to review the api docs and create PRs there.

### Checklist ###

* [-] New tests have been added to show the fix or feature works

### DIFF
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -3,7 +3,9 @@ define( [
 	"./core/nodeName",
 	"./core/camelCase",
 	"./var/isWindow",
-	"./var/slice"
+	"./var/slice",
+
+	"./event/alias"
 ], function( jQuery, nodeName, camelCase, isWindow, slice ) {
 
 "use strict";
--- a/src/jquery.js
+++ b/src/jquery.js
@@ -11,7 +11,6 @@ define( [
 	"./queue/delay",
 	"./attributes",
 	"./event",
-	"./event/alias",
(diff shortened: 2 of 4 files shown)


timmywil:
@dmethvin Did you mean to include other file changes besides tests?

dmethvin:
@timmywil Oh that was the question I had on this, which is why I hadn't submitted it. Right now the aliases are in a separate module in `event/alias.js` and I didn't think I wanted to remove this module and move it to `deprecated.js`. If someone is currently doing a custom build that removed the aliases but not the deprecated stuff we'd mess it up. What's your take?

timmywil:
@dmethvin I suppose if they're already in their own file, the whole point of sequestering them is already accomplished. Here's an idea, though: Move [this](https://github.com/jquery/jquery/blob/master/src/jquery.js#L14) to [here](https://github.com/jquery/jquery/blob/master/src/deprecated.js#L6).

dmethvin:
Ah, yes, like the others there. I'll do that, thanks!

dmethvin:
Conflicts resolved and landed at 022b69a44e42684bdd0029dd456bedb3b495cc24

PR #3956, opened 2018-01-27
use toType in isFunction, fixes #3955

### LINKED ISSUE #3955: wrong type check in isFunction
https://github.com/jquery/jquery/commit/a16339b8933f115da3661f3d3f64854c3fa60bdc introduced a bug. The typecheck is not correct. The `toType` method was not used and produces failing unit tests at https://github.com/zurb/foundation-sites

DanielRuf:
This also fixes https://github.com/zurb/foundation-sites/issues/10886

### Checklist ###

### DIFF
--- a/test/unit/deferred.js
+++ b/test/unit/deferred.js
@@ -543,11 +543,12 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
 	} catch ( _ ) {}
 } );
 
+console.log( typeof Symbol === "function" && Symbol.toStringTag ? "asd" : "456" );
 QUnit[ typeof Symbol === "function" && Symbol.toStringTag ? "test" : "skip" ](
 	"jQuery.Deferred.then - IsCallable determination (gh-3596)",
 	function( assert ) {
 
-		assert.expect( 1 );
+		assert.expect( 0 );
 
 		var done = assert.async(),
 			defer = jQuery.Deferred();
@@ -888,7 +889,7 @@ QUnit.test( "jQuery.when(thenable) - like Promise.resolve", function( assert ) {
 		}
 	};
 	if ( typeof Symbol === "function" ) {
-		customToStringThen.then[ Symbol.toStringTag ] = "String";
+		//customToStringThen.then[ Symbol.toStringTag ] = "String";
 	}
 
 	var slice = [].slice,
--- a/test/unit/deprecated.js
+++ b/test/unit/deprecated.js
@@ -406,7 +406,7 @@ supportjQuery.each(
 	function( subclass, source ) {
 		var fn;
 		try {
-			fn = Function( "return " + source )();
+			fn = Function( "return " + source );
 		} catch ( e ) {}
 
 		QUnit[ fn ? "test" : "skip" ]( "isFunction(" + subclass + ")",
@@ -429,8 +429,8 @@ QUnit[ typeof Symbol === "function" && Symbol.toStringTag ? "test" : "skip" ](
 		obj[ Symbol.toStringTag ] = "Function";
 		fn[ Symbol.toStringTag ] = "Object";
 
-		assert.equal( jQuery.isFunction( obj ), false, "function-mimicking object" );
-		assert.equal( jQuery.isFunction( fn ), true, "object-mimicking function" );
+		assert.equal( jQuery.isFunction( obj ), true, "function-mimicking object" );
+		assert.equal( jQuery.isFunction( fn ), false, "object-mimicking function" );
 	}
 );
 
--- a/src/var/isFunction.js
+++ b/src/var/isFunction.js
@@ -1,4 +1,6 @@
-define( function() {
+define( [
+	"../core/toType"
+], function( toType ) {
 	"use strict";
 
 	return function isFunction( obj ) {
@@ -7,7 +9,7 @@ define( function() {
       // In some browsers, typeof returns "function" for HTML <object> elements
       // (i.e., `typeof document.createElement( "object" ) === "function"`).
       // We don't want to classify *any* DOM node as a function.
-      return typeof obj === "function" && typeof obj.nodeType !== "number";
+      return toType( obj ) === "function" && toType( obj.nodeType ) !== "number";
   };
 
 } );


DanielRuf:
I guess the unit tests were initially added but never checked or new unit tests written. Strings are not Promises and a Function should not be self executing.

gibson042:
I don't know what this is, but commenting out and reversing test assertions without explanation is no way to get changes accepted.

DanielRuf:
See the failing Travis tests. Your `isFunction` test and other things are basically wrong.

https://github.com/zurb/foundation-sites/commits/develop

And also the unit tests. An empty string is not equal to a thenable or promise for example, the function should not be self executing, ...

DanielRuf:
It broke working unit tests and so did it break Foundation Sites and possibly more (and the migrations for the deprecations are not documented so far if we need any). =(

DanielRuf:
I did a `git bisect` and the mentioned commit was the cause.

gibson042:
We don't have time to sift through someone else's test suite. The instructions in our new issue template are quite clear:
>  Briefly describe the issue you've encountered
>  *  What do you expect to happen?
>  *  What actually happens?
>  *  Which browsers are affected?
>  Provide a *minimal* test case, see https://webkit.org/test-case-reduction/
>  Use the latest shipping version of jQuery in your test case!
>  We prefer test cases on https://jsbin.com or https://jsfiddle.net

DanielRuf:
@ncoden what do you think? So far jQuery 3.3 breaks 40 unit tests and so does it Foundation Sites.

dmethvin:
@DanielRuf have you verified that this change actually fixes Zurb? Also is the second commit here the correct one? It still has a `console.log` in it so I'm wondering if the changes were intended.

DanielRuf:
Yes, it fixes Zurb Foundation Sites. I cloned jQuery into node_modules/jQuery, did `git bisect` and always created a fresh build.

I have force pushed a new commit but it is not here because the PR is already closed.

test/unit/deprecated.js
```diff
@@ -429,8 +429,8 @@ QUnit[ typeof Symbol === "function" && Symbol.toStringTag ? "test" : "skip" ](
 		obj[ Symbol.toStringTag ] = "Function";
 		fn[ Symbol.toStringTag ] = "Object";
 
-		assert.equal( jQuery.isFunction( obj ), false, "function-mimicking object" );
-		assert.equal( jQuery.isFunction( fn ), true, "object-mimicking function" );
+		assert.equal( jQuery.isFunction( obj ), true, "function-mimicking object" );
+		assert.equal( jQuery.isFunction( fn ), false, "object-mimicking function" );
 	}
```
kswedberg: I'm confused. Shouldn't a function-mimicking object return `false` when tested against `jQuery.isFunction()` (and `true` for object-mimicking function)? Why the change here?

DanielRuf: obj is changed to Function a few lines above and typeof gives function for this. And also the new isFunction check. The literal sets it to another type.

DanielRuf: My PR is just a partly fix as the unit tests sre not correct and the expected behaviour of the new method in general after its deprecation (still do not know what is needed to migrate other things and plugins which rely on these. Personally I hoped there would have been a beta release instead of rushing out with a new one and without testing against big frameworks).

DanielRuf:
At least https://github.com/chaijs/chai-jquery is affected and possibly much more. The change clearly introduced a breaking change regarding the behavior compared against 3.2.1.

test/unit/deprecated.js
dmethvin: @DanielRuf it's always possible to test against master. As you can imagine there are a LOT of major frameworks using jQuery and we can't take on the responsibility of testing every release against them. Does Zurb have its CI set up to test against jQuery betas? If so a beta would have caught this. If not, I don't think it would have made a difference.

dmethvin:
This was supposed to be a bug fix, which is why it was not held to a semver major. The fact that the unit tests have to be reversed tells me that there is some assumption made somewhere (jQuery or Zurb) that isn't correct.  To the point @gibson042 made, we could really use a reduced test case. I looked at the failing Zurb tests and can't quite work through them. I'm out of time for now. I will note, however, that Zurb is using ES classes compiled through Babel which may be introducing some strange behavior.

test/unit/deprecated.js
DanielRuf: See package.json. Just (stable) releases are installed by npm.

DanielRuf:
You can easily verify it by cloning foundation-sites, installing the dependencies, going into the node_modules directory, deleting the jquery directory and recreating it with a git clone. In there do a `git checkout https://github.com/jquery/jquery/commit/a16339b8933f115da3661f3d3f64854c3fa60bdc`, rebuild and rerun the testsuite of foundation-sites.

My first check was to compare the dep trees of a succeeding and a failing commit and the difference was jQuery 3.2.1 vs 3.3.1. I checked all commits with `git bisect` in the jquery directory which clearly gave https://github.com/jquery/jquery/commit/a16339b8933f115da3661f3d3f64854c3fa60bdc as cause.

I'll provide some more commands from the Travis CI build.

DanielRuf:
There was no change in foundation-sites which would cause this.

DanielRuf:
First commit that failed:

https://github.com/zurb/foundation-sites/commit/dd020a7058e3e711c2a1e135e2e33829f99075f4

Build:
https://travis-ci.org/zurb/foundation-sites/builds/331005436

Last successful commit:
https://github.com/zurb/foundation-sites/commit/05697a06c61421aeec9465a138f4bc41aae12751

Build:
https://travis-ci.org/zurb/foundation-sites/builds/330365427

Run the test suite with `npm run test:ci`

test/unit/deferred.js
```diff
@@ -888,7 +889,7 @@ QUnit.test( "jQuery.when(thenable) - like Promise.resolve", function( assert ) {
 		}
 	};
 	if ( typeof Symbol === "function" ) {
-		customToStringThen.then[ Symbol.toStringTag ] = "String";
+		//customToStringThen.then[ Symbol.toStringTag ] = "String";
```
gibson042: This change is incorrect, a `then` method that mimics another type must still be used because that's how native Promises behave.

test/unit/deprecated.js
```diff
@@ -406,7 +406,7 @@ supportjQuery.each(
 	function( subclass, source ) {
 		var fn;
 		try {
-			fn = Function( "return " + source )();
+			fn = Function( "return " + source );
```
gibson042: This change is incorrect, `fn` is supposed to be a generator or async function in browsers that support those constructs. We just use `Function` to avoid syntax errors in the test suite itself for browsers that don't support them.

test/unit/deprecated.js
```diff
@@ -429,8 +429,8 @@ QUnit[ typeof Symbol === "function" && Symbol.toStringTag ? "test" : "skip" ](
 		obj[ Symbol.toStringTag ] = "Function";
 		fn[ Symbol.toStringTag ] = "Object";
 
-		assert.equal( jQuery.isFunction( obj ), false, "function-mimicking object" );
-		assert.equal( jQuery.isFunction( fn ), true, "object-mimicking function" );
+		assert.equal( jQuery.isFunction( obj ), true, "function-mimicking object" );
+		assert.equal( jQuery.isFunction( fn ), false, "object-mimicking function" );
```
gibson042: These changes are incorrect, function-mimicking objects are not callable and object-mimicking functions still are (and jQuery.isFunction is intended to test that characteristic).

dmethvin:
> There was no change in foundation-sites which would cause this.

Yes, which is why I said there is some *assumption* that isn't correct. If there is some assumption in foundation-sites that isn't guaranteed in jQuery's API contract it's quite possible for even a patch release to cause serious breakage. That is why we'd like to have a simple test case to figure out why this broke.

This PR seems to be aimed in the wrong direction but if you can provide a reduced case in gh-3955 we can work with you from there to figure out what is going on and determine a resolution.

test/unit/deferred.js
DanielRuf: Sure, but the unit test awaits a thenable, not an empty string.

test/unit/deprecated.js
DanielRuf: I ran the testsuite locally and this caused also a failing unit test on your side.

test/unit/deprecated.js
DanielRuf: Well, but it didn't and a simple `typeof` told otherwise.

test/unit/deferred.js
```diff
@@ -543,11 +543,12 @@ QUnit.test( "jQuery.Deferred.then - spec compatibility", function( assert ) {
 	} catch ( _ ) {}
 } );
 
+console.log( typeof Symbol === "function" && Symbol.toStringTag ? "asd" : "456" );
 QUnit[ typeof Symbol === "function" && Symbol.toStringTag ? "test" : "skip" ](
 	"jQuery.Deferred.then - IsCallable determination (gh-3596)",
 	function( assert ) {
 
-		assert.expect( 1 );
+		assert.expect( 0 );
```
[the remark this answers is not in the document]
DanielRuf: Well, the test was also failing here as you can see on Travis when using the right definition for the deprecated functions. A deprecation should not change the behavior of its previous method definition.

DanielRuf:
https://github.com/search?utf8=%E2%9C%93&q=+%2388+An+in-range+update+of+jquery+is+breaking+the+build+&type=Issues

dmethvin:
Okay so it sounds like it could be due to a pretty common dependency. Hopefully we can get a test case soon.

DanielRuf:
> Okay so it sounds like it could be due to a pretty common dependency. Hopefully we can get a test case soon.

Imo https://github.com/jquery/jquery/commit/a16339b8933f115da3661f3d3f64854c3fa60bdc is still the cause. Not sure which dependency you mean.

https://github.com/julmot/mark.js/issues/186

test/unit/deferred.js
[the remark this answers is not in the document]
gibson042: The deprecation didn't change anything, the fix for #3596 did.

test/unit/deferred.js
gibson042: Right, it needs to await a thenable _in which the `then` method mimics a non-function_.

test/unit/deprecated.js
gibson042: The "deprecated" module is passing in all browsers: http://swarm.jquery.org/job/6744

test/unit/deprecated.js
gibson042: Right, the failure was reported in #3600 and fixed in a16339b8933f115da3661f3d3f64854c3fa60bdc, and we still don't know why you want to revert that because you haven't provided a reduced test case demonstrating your unfulfilled expectations.

Issue #3479, opened 2016-12-31
.position() and .offsetParent() return incorrect results in FF + recent versions of Chrome


craigkovatch:
The W3 offsetParent algorithm changed on 9/15. It now includes ancestor elements with a CSS transform applied. Chrome and (maybe?) FF are in compliance with the new spec, while IE and Safari are not. As a result .position() is now returning incorrect results in these browsers.

IMO jQuery will probably need to implement its own ancestor traversal rather than relying on DOM-provided .offsetParent, otherwise results will forever depend on the browser+version of the client.

See: w3c/csswg-drafts#409

gibson042:
I just submitted gh-3487, which should correct `.offset()` and `.position()` behavior in browsers implementing https://github.com/w3c/csswg-drafts/commit/180b348a1ac0931043cd195651fc5164463d2bce (although we should still add new tests for _this_ issue), but I don't want to override host environment offsetParent logic. Instead, I think we should deprecate `.offsetParent()`—gh-3487 already eliminates internal use.

craigkovatch:
Thanks for hopping on this!

Looks like your solution is reliant on getBoundingClientRect -- have you accounted for the way Chrome's return value for this API changes when pinch-to-zoomed?

Also, does this compute the correct value when translation or scaling transformations are present? I don't see anything in the tests around that property.

craigkovatch:
Also, my suggestion that these methods don't rely on offsetParent (whether DOM or jQuery) was because the results will differ depending on how recent the browser is. My understanding was always that jQuery tries to hide these details "under the covers" so the author doesn't have to do browser-sniffing. Is that still the case?

gibson042:
> Looks like your solution is reliant on getBoundingClientRect -- have you accounted for the way Chrome's return value for this API changes when pinch-to-zoomed? (See also https://bugs.chromium.org/p/chromium/issues/detail?id=489206)

> Also, does this compute the correct value when translation or scaling transformations are present? I don't see anything in the tests around that property.

Explicitly no to both, and in fact we'll soon be documenting that `.position()` is not reliable for elements with transformed ancestors (cf. https://github.com/jquery/jquery/issues/3193#issuecomment-232419543 ), although that may change in the future (and if so, gh-3487 is a step towards it).

[...]

PR #3972, opened 2018-02-08
Fix native offsetParent difference

psyafter:
### Summary ###
When using native offsetParent method so according to https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent TD and TABLE are returned even non-positioned at all.

### Checklist ###

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -131,6 +131,9 @@ jQuery.fn.extend( {
 			// when a statically positioned element is identified
 			doc = elem.ownerDocument;
 			offsetParent = elem.offsetParent || doc.documentElement;
+			while ((offsetParent.nodeName == "TD" || offsetParent.nodeName == "TABLE") && 
+			       jQuery.css(offsetParent, "position") === "static")
+				offsetParent = offsetParent.offsetParent || doc.documentElement;
 			while ( offsetParent &&
 				( offsetParent === doc.body || offsetParent === doc.documentElement ) &&
 				jQuery.css( offsetParent, "position" ) === "static" ) {


dmethvin:
Hey @psyafter can you create an issue that includes a test case showing when the current code gives the wrong answer? If there is a bug, the PR should include a unit test that ensures it's been fixed.

psyafter:
Hi @dmethvin . I'm not good with it. But you can see the bug in my simple example: https://jsfiddle.net/3xpnfccL/1/

spn2 element get wrong position.top because it in table

psyafter:
Hi @dmethvin . I'm not good with it. But you can see the bug in my simple example: https://jsfiddle.net/3xpnfccL/1/

spn2 element get wrong position.top because it in table

timmywil:
Thank you for contributing, but I think jQuery is going to move away from thinking of `offsetParent` as the closest ancestor with position `static` and match behavior with the spec, which as you noted says the `offsetParent` is the closest ancestor with position `static` *or* is a `table`, `td`, or `th` element. This is a breaking change, however, and we should note this in the docs and fix the `.offsetParent()` method to match `.position()`, which will be fixed by #3479.

timmywil:
I meant without.

psyafter:
@timmywil it will not be fixed? currently is breaking change between 3.2 to 3.3 (in 3.2 and before it was always closest to not static, but is closest to table/td/th as well).
After moving to 3.3 app UI is broken by this change.

timmywil:
@psyafter It's debatable whether this can be considered a bugfix. Our docs don't mention the exception for table elements, but native `offsetParent` has this exception in the spec. Which is expected depends on how familiar you are with native behavior. I can't tell you specifically how to work around this change without seeing your app and I'm sorry for the inconvenience, but we've learned not to fight against native behavior. We end up causing more problems.

Issue #3504, opened 2017-01-16
Properly detect attachment of Shadow DOM elements


gibson042:
There are a few places where we check document attachment with `jQuery.contains( elem.ownerDocument, elem )` (`isHiddenWithinTree` for show/hide, `curCSS` for inline style fallback, `buildFragment`/`domManip`/`remove`/`jQuery.clone` for managing script evaluation during DOM manipulation), all of which [fail inside Shadow DOM](https://github.com/jquery/jquery/issues/2854#issuecomment-272604563):
> `$.contains(document, elem)` will return false when elem is shown in a shadow dom.
change to `elem.getRootNode({composed: true}) === elem.ownerDocument` works.

We should investigate the cost of a remedy (e.g., a Shadow-DOM-compatible `isAttached( elem )` function).

timmywil:
Yes to going forward and investigating, but we may scrap it later depending on how involved this gets.

gibson042:
As noted in the meeting, we'd probably benefit from an internal `isAttached` even if it doesn't handle shadow DOM, because it better communicates intent.

PR #3977, opened 2018-02-14
Core: Use isAttached to check for attachment of element

SaptakS:
### Summary ###
This change replaces the use of `jQuery.contains` to check for attachment of DOM
by `isAttached` function to better communicate what is being done.

Ref gh-3504

### Checklist ###

### DIFF
--- a/src/css/var/isHiddenWithinTree.js
+++ b/src/css/var/isHiddenWithinTree.js
@@ -1,9 +1,9 @@
 define( [
 	"../../core",
-	"../../selector"
+	"../../var/isAttached"
 
 	// css is assumed
-], function( jQuery ) {
+], function( jQuery, isAttached ) {
 	"use strict";
 
 	// isHiddenWithinTree reports if an element has a non-"none" display style (inline and/or
@@ -27,7 +27,7 @@ define( [
 			// Support: Firefox <=43 - 45
 			// Disconnected elements can have computed display: none, so first confirm that elem is
 			// in the document.
-			jQuery.contains( elem.ownerDocument, elem ) &&
+			isAttached( elem ) &&
 
 			jQuery.css( elem, "display" ) === "none";
 	};
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -1,5 +1,6 @@
 define( [
 	"./core",
+	"./var/isAttached",
 	"./var/concat",
 	"./var/isFunction",
 	"./var/push",
@@ -23,7 +24,7 @@ define( [
 	"./traversing",
 	"./selector",
 	"./event"
-], function( jQuery, concat, isFunction, push, access,
+], function( jQuery, isAttached, concat, isFunction, push, access,
 	rcheckableType, rtagName, rscriptType,
 	wrapMap, getAll, setGlobalEval, buildFragment, support,
 	dataPriv, dataUser, acceptData, DOMEval, nodeName ) {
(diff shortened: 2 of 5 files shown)


src/css/var/isHiddenWithinTree.js
```diff
@@ -1,9 +1,10 @@
 define( [
 	"../../core",
-	"../../selector"
+	"../../selector",
+	"../../var/isAttached"
```
mgol: This is incorrect as this is the third parameter but it's used as a second one. You need to put that before `../../selector`. Could you run tests in AMD mode at least in one browser locally? It would catch those issues.

Unfortunately you can't do that via our Node.js setup but most non-AJAX tests should work without a PHP server. Or you can run one via `php -S localhost:7000` from your jQuery repository directory and open `http://localhost:7000`, choosing a module you modified (other than ajax) and enabling the AMD checkbox.

src/manipulation/buildFragment.js
```diff
@@ -77,13 +78,13 @@ function buildFragment( elems, context, scripts, selection, ignored ) {
 			continue;
 		}
 
-		contains = jQuery.contains( elem.ownerDocument, elem );
+		attached = isAttached( elem );
```
mgol: A nice variable name change, this reads better than what we had before 👍.

src/var/isAttached.js
```diff
@@ -0,0 +1,10 @@
+define( function() {
+	"use strict";
+
+	return function isAttached( obj ) {
+
+      // Replace the use of contains function to check attachment (gh-3504)
```
mgol: This comment is not necessary, at least not in this form. We don't need to describe what the code looked like previously or link to the issue/PR that changed it.

src/css/var/isHiddenWithinTree.js
SaptakS: Will check this.

src/var/isAttached.js
SaptakS: Okay. Will remove it for the moment.

src/var/isAttached.js
```diff
@@ -0,0 +1,8 @@
+define( function() {
+	"use strict";
+
+	return function isAttached( obj ) {
+		return jQuery.contains( obj.ownerDocument, obj );
+  };
```
timmywil: I'm surprised this passed lint. Looks like a tabs/spaces mixture here.

SaptakS: Really sorry for this. I am not sure how this mixture happened.

timmywil: No worries. I'm just concerned that travis didn't pick it up.

src/var/isAttached.js
```diff
@@ -0,0 +1,8 @@
+define( function() {
+	"use strict";
+
+	return function isAttached( obj ) {
+		return jQuery.contains( obj.ownerDocument, obj );
```
gibson042: This `jQuery.contains` dependency should be captured in the `define` arguments, as it was in the other source files.

src/css/var/isHiddenWithinTree.js
```diff
@@ -1,9 +1,10 @@
 define( [
 	"../../core",
+	"../../var/isAttached",
 	"../../selector"
```
gibson042: I don't think `selector` is needed as a direct dependency any longer.

src/manipulation.js
```diff
@@ -23,7 +24,7 @@ define( [
 	"./traversing",
 	"./selector",
```
gibson042: :+1: This one stays, though (manipulation functions use `jQuery.isXMLDoc` and `jQuery.text`).

src/var/isAttached.js
```diff
@@ -0,0 +1,11 @@
+define( [
+	"../core",
+	"../selector"
```
mgol: Maybe add a comment that was previously in `curCSS`?
```
	"../selector" // Get jQuery.contains
```

SaptakS: Done.

mgol:
Landed, thanks!

PR #3985, opened 2018-02-23
Update: selector native

arguiot:
### Summary ###

Now using MDN polyfill: https://developer.mozilla.org/en-US/docs/Web/API/Element/matches

### Checklist ###

### DIFF
--- a/src/selector-native.js
+++ b/src/selector-native.js
@@ -36,10 +36,17 @@ define( [
 var hasDuplicate, sortInput,
 	sortStable = jQuery.expando.split( "" ).sort( sortOrder ).join( "" ) === jQuery.expando,
 	matches = documentElement.matches ||
-		documentElement.webkitMatchesSelector ||
-		documentElement.mozMatchesSelector ||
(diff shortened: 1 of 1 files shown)


mgol:
Thanks for the PR.

I'm not sure why we'd need the polyfill, though. Every browser we support supports one of the prefixed versions: https://caniuse.com/#feat=matchesselector

If anything, we could drop the Opera `o` & Firefox `moz` prefixes as those browsers have long supported `Element#matches` and would only leave the `ms` one for IE and `webkit` for Android <5.

arguiot:
@mgol you're totally right. But for older browser, it won't be useless.

And I guess that if it's on MDN, which is a very good source, it would be better to add it, just in case.

mgol:
jQuery only supports browsers listed at https://jquery.com/browser-support/; none of this browsers require a polyfill. If you add to that that the `selector-native` module is optional, I don't think we should add unneeded polyfills there.

timmywil:
Thank you for the contribution, but I agree with @mgol.

PR #3994, opened 2018-03-02
Build: Seasonal update of uglify and its options

markelog:
```
 =     +1 dist/jquery.js
-31    -31 dist/jquery.min.js
```
### Checklist ###

* [] New tests have been added to show the fix or feature works
* [] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -261,7 +261,6 @@ module.exports = function( grunt ) {
 					compress: {
 						"hoist_funs": false,
 						loops: false,
-						unused: false,
 
 						// Support: IE <11
 						// typeofs transformation is unsafe for IE9-10
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "sizzle": "2.3.3",
     "strip-json-comments": "2.0.1",
     "testswarm": "1.1.0",
-    "uglify-js": "3.3.4"
+    "uglify-js": "3.3.12"
   },
   "scripts": {
     "build": "npm install && grunt",


Gruntfile.js
```diff
@@ -259,9 +259,9 @@ module.exports = function( grunt ) {
 					banner: "/*! jQuery v<%= pkg.version %> | " +
 						"(c) JS Foundation and other contributors | jquery.org/license */",
 					compress: {
+						"pure_getters": true,
```
mgol: Hmm, what does it exactly do? The Uglify README mentions:

> UglifyJS will assume that object property access (e.g. `foo.bar` or `foo["bar"]`) doesn't have any side effects

but technically we can't guarantee that as `jQuery.event.addProp` creates a getter utilizing a provided hook.

markelog: My understanding is from going to something like:

```js
foo.bar++;
if (foo.bar) {
  console.log('test');
}
```
to something like:

```js
++foo.bar && console.log("test");
```
In the first case, If you have a getter on "bar" it will be executed twice, in the second case – once i.e. we should be fine

mgol: I see, that meaning of the flag makes sense. On the other hand, I imagine something like the following:
```js
var bar,
    flag = foo.a;
if (flag) {
    bar = flag + 1;
}
```
might be minified to (I'm skipping other optimizations here):
```js
var bar;
if (foo.a) {
    bar = foo.a + 1;
}
```
which would invoke a getter twice which - in turn - may lead to a performance penalty as the getter is invoked twice.

markelog: Well, we don't have getters anyway

mgol: We do, our event object is implemented via getters now.

jbedard: Those event getters are super fast and should be pure functions. It is also normal and expected that they will be called many times.

The potential of calling those getters multiple times for the benefit of size reduction sounds good to me.

mgol: OK, I've checked what's the difference between the output with `pure_getters` on and off and it was huge, 57 bytes gzipped. I prettyfied the code then and compared the two outputs and I saw that one of the differences is `pure_getters` cuts out this:
https://github.com/jquery/jquery/blob/3.3.1/external/sizzle/dist/sizzle.js#L193-L195
```js
	// Support: Android<4.0
	// Detect silently failing push.apply
	arr[ preferredDoc.childNodes.length ].nodeType;
```
and this: https://github.com/jquery/jquery/blob/3.3.1/external/sizzle/dist/sizzle.js#L1521-L1523
```js
			// Accessing this property makes selected-by-default
			// options in Safari work properly
			if ( elem.parentNode ) {
				elem.parentNode.selectedIndex;
			}
```
etc.

In other words, in our code getters are sometimes invoked for side effects only so we cannot enable this flag.

markelog: Nice find @mgol, 'pure_getters' option is removed

markelog:
We probably can play with the options a little bit more, but we can do that in another season :).
Yet again - nice work @mgol :)

PR #3996, opened 2018-03-02
Add Shadow DOM feature to isAttached

SaptakS:
### Summary ###

**This PR needs to be merged only after #3977** 

This change allows `isAttached()` function to address shadowDOMElements even if the browser doesn't support `getRootNode()`. 

Ref gh-3504, gh-3977

### Checklist ###

* [ ] New tests have been added to show the fix or feature works (**Will add once this method is reviewed and confirmed**)


mgol:
Can you rebase this PR now that your other one has landed?

SaptakS:
@mgol done.

src/var/isAttached.js
```diff
@@ -1,11 +1,19 @@
 define( [
 	"../core",
+	"./getShadowRoot",
+	"./document",
```
gibson042: We shouldn't have a dependency on document, see below.

src/var/getShadowRoot.js
```diff
@@ -0,0 +1,15 @@
+define( function() {
+	"use strict";
+
+	return function getShadowRoot( obj ) {
```
gibson042: For clarity, please update both this function and `isAttached` to use `elem` instead of `obj`.

src/var/isAttached.js
```diff
@@ -1,11 +1,19 @@
...
 	"../selector" // Get jQuery.contains
-], function( jQuery ) {
+], function( jQuery, getShadowRoot, document ) {
 	"use strict";
 
+	// Replace the use of contains function to check attachment (gh-3504)
 	return function isAttached( obj ) {
+
+		// Check if browser supports Shadow DOM
+		if ( document.head.createShadowRoot || document.head.attachShadow ) {
```
gibson042: `isAttached` is a potentially hot path and I'd rather not see a support check inside of it. How about refactoring getShadowRoot.js to define the smart function conditionally (falling back on `function( elem ) {}`), and then always using it here:
```js
return jQuery.contains( elem.ownerDocument, elem ) ||
	getShadowRoot( elem ) === elem.ownerDocument;
```

src/var/getShadowRoot.js
```diff
@@ -0,0 +1,15 @@
...
+
+		// check for support of getRootNode function
+		if ( typeof obj.getRootNode === "function" ) {
+			return obj.getRootNode( { composed: true } );
+		} else {
+			var doc = obj.ownerDocument;
+			return doc && doc.host ? getShadowRoot( doc.host ) : doc;
```
gibson042: This will return the outer document for detached elements, which is incorrect. And we should probably have a loop instead of recursive calls anyway, but be mindful to return empty at all levels (e.g., including a shadow DOM inside a detached element).

src/var/getShadowRoot.js
```diff
@@ -0,0 +1,22 @@
+define( [
+	"./document"
+], function( document ) {
+	"use strict";
+
+	return function getShadowRoot( elem ) {
+
+		// Check if browser supports Shadow DOM
+		if ( document.head.createShadowRoot || document.head.attachShadow ) {
```
mgol: Why do we need to check both?

mgol: This `if` is still in the function so it still executes every time `isAttached` is invoked. I think @gibson042 meant to define this function conditionally, to run the Shadow DOM support test outside of the function and assign a proper definition based on that. In this way this condition will fire only once.

SaptakS: So to get it clear, you mean this should be written as a condition wherever isAttached is invoked and send some parameter based on that?

SaptakS: > Why do we need to check both?

So `createShawowRoot` gives a false negative in safari version 10.1 since it uses 0.v1 of shadowDOM element which had `attachShadow`. So I included both.

mgol: I don't know what's @gibson042's opinion on that but personally I wouldn't botter with supporting Shadow DOM v0; it was a preview and the standard is the v1 version. And if we add v0 support we won't be able to remove it so easily later.

mgol: > So to get it clear, you mean this should be written as a condition wherever isAttached is invoked and send some parameter based on that?

Something like:
```js
var myFunction;
if (someCondition) {
    myFunction = () => someValue;
} else {
    myFunction = () => someOtherValue;
```
(but in ES5 etc., this is just to get an idea) Then the condition is only checked once and is never checked when `myFunction` is getting invoked.


We did something in this spirit for some `cssHooks`, see:
https://github.com/jquery/jquery/blob/3.3.1/src/css/addGetHookIf.js
and its uses:
https://github.com/jquery/jquery/blob/3.3.1/src/css.js#L414-L425
https://github.com/jquery/jquery/blob/3.3.1/src/offset.js#L218-L229

SaptakS: @mgol isn't in the above examples, in https://github.com/jquery/jquery/blob/3.3.1/src/css/addGetHookIf.js#L10 the condition is checked everytime the `addGetHookIf` is invoked?

mgol: `addGetHookIf` is only called once per each css hook used.

timmywil:
This has been reviewed, but needs tests.

mgol:
@SaptakS will you have time to finish this PR soon?

SaptakS:
@mgol I will work on it this weekend. Sorry for the delay.

mgol:
@SaptakS We're trying to finish stuff planned for `3.4.0`. Will you have time to finish this today or tomorrow? If not, we may have to take over.

SaptakS:
I will give it a try today. Sorry for the delay.

test/unit/css.js
```diff
@@ -641,6 +641,48 @@ QUnit.test( "show/hide detached nodes", function( assert ) {
...
+	shadowRoot.innerHTML = "<style>.hidden{display: none;}</style>" +
+		"<div class='hidden' id='shadowdiv'><p class='hidden' id='shadowp'>" +
+		"<a href='#' class='hidden' id='shadowa'></a></p>" +
+		"<code class='hidden' id='shadowcode'></code>" +
+		"<pre class='hidden' id='shadowpre'></pre><span class='hidden' id='shadowspan'></span></div>" +
+		"<table class='hidden' id='shadowtable'><thead class='hidden' id='shadowthead'>" +
+		"<tr class='hidden' id='shadowtr'><th class='hidden' id='shadowth'></th></tr></thead>" +
+		"<tbody class='hidden' id='shadowtbody'><tr class='hidden'>" +
+		"<td class='hidden' id='shadowtd'></td></tr></tbody></table>" +
+		"<ul class='hidden' id='shadowul'><li class='hidden' id='shadowli'></li></ul>";
```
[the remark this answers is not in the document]
SaptakS: Yes. I will do that for sure.

test/unit/css.js
```diff
@@ -966,6 +1008,26 @@ QUnit[ jQuery.find.compile && jQuery.fn.toggle ? "test" : "skip" ]( "detached to
 		"cascade-hidden element in detached tree" );
 } );
 
+QUnit[ jQuery.find.compile && jQuery.fn.toggle && document.head.createShadowRoot ? "test" : "skip" ]( "shadow toggle()", function( assert ) {
+	assert.expect( 4 );
+	jQuery( "<div id='shadowHost'></div>" ).appendTo( "#qunit-fixture" );
+	var shadowHost = document.querySelector( "#shadowHost" );
+	var shadowRoot = shadowHost.attachShadow( { mode: "open" } );
+	shadowRoot.innerHTML = "<style>.hidden{display: none;}</style><div id='shadowHiddenChild' class='hidden'></div><div id='shadowChild'></div>";
```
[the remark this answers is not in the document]
SaptakS: Okay. makes sense

test/unit/css.js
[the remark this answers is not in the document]
SaptakS: Done.

test/unit/css.js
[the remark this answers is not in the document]
SaptakS: Done.

test/unit/css.js
```diff
@@ -641,6 +641,62 @@ QUnit.test( "show/hide detached nodes", function( assert ) {
...
+		"tr": "table-row",
+		"th": "table-cell",
+		"td": "table-cell",
+		"ul": "block",
+		"li": "list-item"
+	};
+
+	jQuery.each( test, function( selector, expected ) {
+		var shadowChild = shadowRoot.querySelector( "#shadow" + selector );
+		var x = jQuery( shadowChild );
```
gibson042: Let's use a more meaningful name like `$shadowChild`.

test/unit/css.js
```diff
@@ -966,6 +1022,28 @@ QUnit[ jQuery.find.compile && jQuery.fn.toggle ? "test" : "skip" ]( "detached to
...
+	jQuery( "<div id='shadowHost'></div>" ).appendTo( "#qunit-fixture" );
+	var shadowHost = document.querySelector( "#shadowHost" );
+	var shadowRoot = shadowHost.attachShadow( { mode: "open" } );
+	shadowRoot.innerHTML = "<style>.hidden{display: none;}</style>" +
+		"<div id='shadowHiddenChild' class='hidden'></div>" +
+		"<div id='shadowChild'></div>";
+	var shadowChild = shadowRoot.querySelector( "#shadowChild" );
+	var shadowHiddenChild = shadowRoot.querySelector( "#shadowHiddenChild" );
+
+	var x = jQuery( shadowChild );
```
gibson042: Let's use a more meaningful name like `$shadowChild`.

src/core/isAttached.js
```diff
@@ -0,0 +1,25 @@
...
+	"../var/document",
+	"../selector" // Get jQuery.contains
+], function( jQuery, getShadowRoot, document ) {
+	"use strict";
+
+	// Replace the use of contains function to check attachment (gh-3504)
+
+	var isAttached;
+
+	// Check if browser supports Shadow DOM
```
gibson042: Let's drop the preceding comment and update this to something like "Check attachment across shadow DOM boundaries when possible (gh-3504)".

test/unit/effects.js
```diff
@@ -220,6 +220,38 @@ supportjQuery.each( hideOptions, function( type, setup ) {
...
+
+		var shadowHost = document.querySelector( "#shadowHost" );
+		var shadowRoot = shadowHost.attachShadow( { mode: "open" } );
+		shadowRoot.innerHTML = "<style>.hidden{display: none;}</style>" +
+			"<span id='shadowChild' class='hidden'></span>";
+		var shadowChild = shadowRoot.querySelector( "#shadowChild" );
+
+		var $shadowChild = jQuery( shadowChild );
+		var displayNone = $shadowChild.css( "display" );
+		var display = "block";
```
mgol: @gibson042 Note that this hardcodes `"block"` as opposed to the non-shadow test above this one. But, as @SaptakS made me realize - the check for `$span.css( "display" )` in the non-shadow test requires to show the element and later hide it which changes how the element is hidden so the test doesn't really seem to test what it should. Also, in both cases, the expected value is `"block"` so it doesn't make sense to compute it during the test.

What do you think about changing the non-shadow test to avoid this check? (outside of this PR, of course).

mgol:
There was one test failure which I fixed. This is ready for a final review.

@gibson042 do you want to have a look?

SaptakS:
@gibson042 @mgol updated to put size optimization.

Issue #3978, opened 2018-02-14
Expose KeyboardEvent.code on jQuery event


waffledonkey:
### Background ###
Keyboard events introduce new properties, `.key` and `.code`. The latter being especially useful because it normalizes keyboard layouts (eg. Dvorak) and different locales.

https://www.w3.org/TR/uievents-key/
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
https://developers.google.com/web/updates/2016/04/keyboardevent-keys-codes
 
### Description ###
If you register a keydown event listener using W3 addEventListener and then `console.log` the event, the code property is present.  If, however, you do the same thing using jQuery the code property is absent (presumably removed) -- but the key property is curiously present.

```
document.addEventListener('keydown', function (e) {
    console.log('W3:', e.type, e.code, e.key, e.keyCode);
});

$(document).on('keydown', function (e) {
    console.log('jQuery:', e.type, e.code, e.key, e.keyCode);
});
```
*Results*
<pre>
jQuery: keydown <strong>undefined</strong> ArrowRight 39
W3: keydown ArrowRight ArrowRight 39
</pre>

### Link to test case ###
see: https://jsfiddle.net/waffledonkey/vwgu2u12/

While you're in there, IE implemented an older spec so they use 'Left', 'Right', 'Up' and 'Down' versus 'ArrowLeft', 'ArrowRight', 'ArrowUp' and 'ArrowDown' -- there may be other differences...  If you still consider yourself an abstraction library I would *LOVE* if you could help pull IE up the hill.

dmethvin:
Agreed and I am thinking this is a typo because [we have `char` and `key`[(https://github.com/jquery/jquery/blob/4a2bcc27f9c3ee24b3effac0fbe1285d1ee23cc5/src/event.js#L618-L620)  but not `code`. I don't even see a `char` listed anywhere in the specs? @jbedard do you know? It's been around since jQuery 1.7 so it's not recent. I know the specs have changed several times over the years (which, keyCode/charCode, key/char).

The problem with trying to normalize the names in IE11 is that it is especially bloating to put a comprehensive mapping like that in jQuery. If every person who ever used jQuery needed it that might make sense, but very few people need more than one or two keys.

jbedard:
[...]

PR #3998, opened 2018-03-03
Changed event.js to add "code" property to keyboard event object

tmybr11:
### Summary ###

Added the "code" property to the Event object passed to keyboard-triggered event handlers, as suggested on [#3978](https://github.com/jquery/jquery/issues/3978).

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -616,6 +616,7 @@ jQuery.each( {
 	shiftKey: true,
 	view: true,
 	"char": true,
+	code: true,
 	charCode: true,
 	key: true,
 	keyCode: true,


dmethvin:
Not all properties that we lazy-copy are on all events that the browser can generate, and (at least last I checked) not all the properties are supported by all browsers. We'd need a matrix of event names and valid properties for the event by browser that could cover at least most of them.

To ensure we copy correctly from a native event we would need to dispatch a native event, and the mechanism to do that is browser-dependent as well right now--some support the newer event constructors and some only support the DOM2 `document.createEvent()` interface. I am not sure it's even possible to set all the properties we want to test through either.

mgol:
@dmethvin My main worry is that some future refactor that will move this code to a different place will lose one of the properties and we won't notice. To prevent that we don't really need to test with real event objects, just mocking it would suffice.

Obviously, that would amount to just duplicating the event name list in tests without testing the substance but it still seems a little bit more secure than what we have now.

tmybr11:
I will update the API docs asap and I can also add a new test if you think it will be necessary for this update.

timmywil:
We could just do a super-dumb test and make sure that all of these properties are actually on the jQuery.Event prototype.

mgol:
@timmywil Yeah, something like that would be enough here.

@tmybr11 What we're talking about applies to all properties so your PR is fine, the test we're talking about is a matter for another PR/commit.

Issue #3986, opened 2018-02-23
Don't fill jQuery.cssProps if mapping isn't needed


dmethvin:
While working on https://github.com/jquery/jquery-migrate/pull/291 the tests failed indicating that we are using `jQuery.cssProps` still in a [somewhat insidious way](https://github.com/jquery/jquery/blob/4a2bcc27f9c3ee24b3effac0fbe1285d1ee23cc5/src/css.js#L66). Each time someone accesses a CSS property we're populating it with properties like `{ width: "width" }`. That makes it hard for Migrate to warn since just about anything can end up there through no fault of external code. For now I won't land this warning and leave https://github.com/jquery/jquery-migrate/issues/283 open until we can change the logic to only populate `jQuery.cssProps` when it needs to actually change the property name.

dmethvin:
Fixing this is messier than I'd like because we have a [unit test](https://github.com/jquery/jquery/blob/4a2bcc27f9c3ee24b3effac0fbe1285d1ee23cc5/test/unit/css.js#L1031-L1046) making some pretty crazy assertions. I'm not certain that aliasing one standard CSS property to another was the intended use case but that's what we're up against. Just a thought for 4.0.0, do we even need the mapping to vendor-specific properties at that point?

mgol:
@dmethvin The unit test is just checking the mapping, I wouldn't expect anyone in real life aliasing `top` to `left`.

We need the mapping to cache vendor prefixed values which are still used by browsers, although not really for new APIs. There are lots of prefixed properties in IE 9+ and I think it's useful we handle it for users.

What we could do is to introduce another internal object that would be consulted if there is no match in `jQuery.cssProps`; we'd write vendor-prop resolved values to this separate object. This would influence the size slightly negatively but hopefully not too much?

Another solution: IIRC Migrate used to differentiate between user calls & jQuery internal ones by setting a certain flag at the beginning of a method that uses the deprecated thing and resetting it at the end (I don't remember what it was). Wouldn't something like that work here?

PR #4005, opened 2018-03-08
CSS: Avoid filling jQuery.cssProps

dmethvin:
Fixes gh-3986

This avoids filling jQuery.cssProps by introducing a second internal
prop cache. It allows jQuery Migrate to detect external usage.

### DIFF
--- a/src/css/finalPropName.js
+++ b/src/css/finalPropName.js
@@ -1,0 +1,39 @@
+define( [ "../var/document" ], function( document ) {
+
+"use strict";
+
+var cssPrefixes = [ "Webkit", "Moz", "ms" ],
+	emptyStyle = document.createElement( "div" ).style,
+	vendorProps = {};
+
+// Return a vendor-prefixed property or undefined
+function vendorPropName( name ) {
+
+	// Check for vendor prefixed names
+	var capName = name[ 0 ].toUpperCase() + name.slice( 1 ),
+		i = cssPrefixes.length;
+
+	while ( i-- ) {
+		name = cssPrefixes[ i ] + capName;
+		if ( name in emptyStyle ) {
+			return name;
+		}
+	}
+}
+
+// Return a potentially-mapped jQuery.cssProps or vendor prefixed property
+function finalPropName( name ) {
+	var final = jQuery.cssProps[ name ] || vendorProps[ name ];
+
+	if ( final ) {
+		return final;
+	}
+	if ( name in emptyStyle ) {
+		return name;
+	}
+	return vendorProps[ name ] = vendorPropName( name ) || name;
+}
+
+return finalPropName;
+
+} );
--- a/src/effects/Tween.js
+++ b/src/effects/Tween.js
@@ -1,7 +1,9 @@
 define( [
 	"../core",
+	"../css/finalPropName",
+
 	"../css"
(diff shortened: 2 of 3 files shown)


mgol:
What's the size difference?

src/css/finalPropName.js
```diff
@@ -0,0 +1,38 @@
+define( [ "../var/document" ], function( document ) {
+
+"use strict";
```
mgol: Please add an empty line below this one; I think that's what we do elsewhere.

src/effects/Tween.js
```diff
@@ -14,6 +16,8 @@ Tween.prototype = {
 	constructor: Tween,
 	init: function( elem, options, prop, end, easing, unit ) {
 		this.elem = elem;
+		this.styleElem = elem.nodeType === 1 && ( jQuery.cssHooks[ prop ] ||
```
mgol: What's the reason for this change? We're now unit-testing `jQuery.Tween`, in particular the behavior of its properties so it's kind of a public API now. I'd rather not expose internal properties on Tween objects.

mgol:
Unit tests are failing in Travis.

src/effects/Tween.js
dmethvin: Mainly I was trying to avoid doing this calculation each time and just do it once in the init. It's no problem to move it back though.

dmethvin:
Not sure why I didn't see those failures. Moving the test back out of the init fixed the issue, I guess if all props in the Tween are considered fair game at any time that's not a good optimization.

 I did notice that in the tween.js tests we're mocking `jQuery.style()` which isn't a public interface.

dmethvin:
@mgol was there something else you wanted to review before I landed this? I noticed the tag.

src/css/finalPropName.js
```diff
@@ -0,0 +1,39 @@
...
+		name = cssPrefixes[ i ] + capName;
+		if ( name in emptyStyle ) {
+			return name;
+		}
+	}
+}
+
+// Return a potentially-mapped jQuery.cssProps or vendor prefixed property
+function finalPropName( name ) {
+	var final = jQuery.cssProps[ name ] || vendorProps[ name ];
```
Krinkle: Hm.. I don't think this resolved the concern from https://github.com/jquery/jquery-migrate/pull/291.

The jquery-migrate pull proposes a warning for getting (and theoretically, setting) of `jQuery.cssProps` itself. That may seem odd, but is actually fine, given that detecting the setting of sub-properties is hard, and, in order to set a sub-property, one has to get the cssProps object first, which will trigger the warning. All good.

However, the code here in core/css.js, still gets the `jQuery.cssProps` object the same way as before, so it seems that would trigger the warning?

The only thing that changed is it now gets the object once instead of twice.

dmethvin: In the previous code, we **set** a property on `jQuery.cssProps`. Now we no longer do that, so Migrate can look for situations where external code sets values here.

Krinkle: Yep, but the warning added in https://github.com/jquery/jquery-migrate/pull/291 is not for getting/setting unknown properties on `jQuery.cssProps`, it warns for getting (and setting) of the `cssProps` property of `jQuery`.

PR #4010, opened 2018-03-14
Dimensions: avoid fetching boxSizing when setting width/height

### LINKED ISSUE #3991: width/height css set hook causes forced reflows
### Description ###
https://github.com/jquery/jquery/commit/3fcddd6e72e7e318c0b062e391d60867732318ae#diff-414c8f59bff0d1b63680b64763d8c529R370 started invoking the `isBorderBox` test 100% of the time instead of only when necessary. In my case this causes a series of `.css({width})` calls to cause a "forced reflow" (that's the chrome terminology in the performance tool).

Updating the width/height `set` to avoid computing `isBorderBox` unless necessary fixes it for me locally, but makes the code a bit uglier:
```
6582c6582
< 				isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
---
> 				isBorderBox,
6587c6587
< 					isBorderBox,
---
> 					isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
6593c6593
< 			if ( isBorderBox && support.scrollboxSize() === styles.position ) {
---
> 			if ( support.scrollboxSize() === styles.position && (undefined === isBorderBox ? (isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box") : isBorderBox) ) {
```

However so far I have been unable to reproduce this in jsbin, so there must be another condition that I'm not understanding such as a specific position type on the element or parent element?

Here's an example where a series of `ng-style="{width: X}"` in AngularJS causes this (and jQuery 3.1 does not cause it):
jq33-reflow

### Link to test case ###
TODO

jbedard:
- this avoids forcing a reflow in some cases

Fixes #3991

```
 raw     gz Compared to 3991 @ 625e19cd9be4898939a7c40dbeb2b17e40df9d54
+147    +63 dist/jquery.js
 +12    +12 dist/jquery.min.js
```

Adds a bit of size but I think it's worth it...

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -378,7 +378,12 @@ jQuery.each( [ "height", "width" ], function( i, dimension ) {
 		set: function( elem, value, extra ) {
 			var matches,
 				styles = getStyles( elem ),
-				isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
+				scrollBoxSize = support.scrollboxSize() === styles.position,
+
+				// To avoid forcing a reflow, only fetch boxSizing if we need it (gh-3991)
+				boxSizingNeeded = scrollBoxSize || extra,
+				isBorderBox = boxSizingNeeded &&
+					jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
 				subtract = extra && boxModelAdjustment(
 					elem,
 					dimension,
@@ -389,7 +394,7 @@ jQuery.each( [ "height", "width" ], function( i, dimension ) {
 
 			// Account for unreliable border-box dimensions by comparing offset* to computed and
 			// faking a content-box to get border and padding (gh-3699)
-			if ( isBorderBox && support.scrollboxSize() === styles.position ) {
+			if ( isBorderBox && scrollBoxSize ) {
 				subtract -= Math.ceil(
 					elem[ "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ) ] -
 					parseFloat( styles[ dimension ] ) -


src/css.js
```diff
@@ -378,7 +378,11 @@ jQuery.each( [ "height", "width" ], function( i, dimension ) {
 		set: function( elem, value, extra ) {
 			var matches,
 				styles = getStyles( elem ),
-				isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
+				scrollBoxSize = support.scrollboxSize() === styles.position,
+
+				// Only fetch boxSizing if required to avoid forcing a reflow (gh-3991)
+				isBorderBox = ( extra || scrollBoxSize ) &&
+					jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
```
gibson042: I like what the code is doing, but the reasons could be a little more clear. Suggestion to improve that (and reduce size a little bit as well):
```patch
diff --git i/src/css.js w/src/css.js
index fc000be..d7181f4 100644
--- i/src/css.js
+++ w/src/css.js
@@ -380,8 +380,9 @@ jQuery.each( [ "height", "width" ], function( i, dimension ) {
 				styles = getStyles( elem ),
 				scrollBoxSize = support.scrollboxSize() === styles.position,
 
-				// Only fetch boxSizing if required to avoid forcing a reflow (gh-3991)
-				isBorderBox = ( extra || scrollBoxSize ) &&
+				// To avoid forcing a reflow, only fetch boxSizing if we need it (gh-3991)
+				boxSizingNeeded = scrollBoxSize || extra,
+				isBorderBox = boxSizingNeeded &&
 					jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
 				subtract = extra && boxModelAdjustment(
 					elem,
@@ -393,7 +394,7 @@ jQuery.each( [ "height", "width" ], function( i, dimension ) {
 
 			// Account for unreliable border-box dimensions by comparing offset* to computed and
 			// faking a content-box to get border and padding (gh-3699)
-			if ( scrollBoxSize && isBorderBox ) {
+			if ( isBorderBox && scrollBoxSize ) {
 				subtract -= Math.ceil(
 					elem[ "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ) ] -
 					parseFloat( styles[ dimension ] ) -
```

jbedard: That reduces size? Interesting, I thought an extra var would do the opposite...

I agree that is clearer 👍

jbedard: Adding the var and changing the comment:
```
   +38    +11 dist/jquery.js
     =      = dist/jquery.min.js
```
Then changing the order in the `if`:
```
     =     -1 dist/jquery.js
     =      = dist/jquery.min.js
```
So it didn't really change the size at all (but does make it more readable, so :+1:)... @gibson042 is that the results you were getting?

gibson042: I don't remember exactly, but this looks good to me.

PR #4014, opened 2018-03-19
Remove alias

gabrielrdantas:
Alias is not being used

### DIFF
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -5,9 +5,7 @@ define( [
 	"./core/toType",
 	"./var/isFunction",
 	"./var/isWindow",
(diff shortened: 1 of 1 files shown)


mgol:
It is used, this module doesn't return anything but defines event aliases. It's included purely for its side effects.

dmethvin:
@gabrielrdantas I know it is subtle, but notice that there is a space before that name and the ones above it. That indicates we don't use it directly in the module but do need it to be included so that it adds methods to the common `jQuery` object. In this case the deprecated event aliases are in a separate module but referencing it here ensures that it will be pulled in to the full build.

Issue #4045, opened 2018-04-18
Object tag with svg being ignored at contents()


emibloque:
First of all, thank you very much for this project and for your work!

### Description

A regression could have been introduced in this commit 3e3b09d and, as it is returning `documentContent` only for `<iframe>`, it is no longer possible to access the document content of an `<object>` with an svg provided in the `data` attribute.

For example this would stop working since version `3.2.0`:
```html
<object id="svg-object" data="1x1.svg"></object>
```
```js
var svgObject = $("#svg-object")

svgObject.on("load", function(){
	svgObject.contents().find('svg');
})
```

### Test link

The test to reproduce this bug would need a `svg` in the same domain, as it isn't possible to upload it to any of the platforms, I cannot provide a test link. However, I've prepared a PR with the fix and the test case provided in it is the following:
```js
QUnit.test( "contents() for <object />", function( assert ) {
	assert.expect( 2 );

	var svgObject = jQuery( "<object id='svg-object' data='" + baseURL + "1x1.svg'></object>" );
	var done = assert.async();

	svgObject.on( "load", function() {
		var contents = jQuery( "#svg-object" ).contents();
		assert.equal( contents.length, 1, "Check object contents" );
		assert.equal( contents.find( "svg" ).length, 1, "Find svg within object" );
		done();
	} );

	jQuery( "#qunit-fixture" ).append( svgObject );
} );
```

PR #4046, opened 2018-04-18
Traversing: $.fn.contents() support for object

emibloque:
Fixes gh-4045

### Summary ###

As described in gh-4045, this PR allows `<object>`  document content to be accessed again using `contents()` function.

### Checklist ###

### DIFF
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -145,7 +145,7 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-        if ( nodeName( elem, "iframe" ) ) {
+        if ( nodeName( elem, "iframe" ) || nodeName( elem, "object" ) ) {
             return elem.contentDocument;
         }
 
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -145,7 +145,7 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-        if ( nodeName( elem, "iframe" ) || nodeName( elem, "object" ) ) {
+				if ( typeof elem.contentDocument !== "undefined" ) {
             return elem.contentDocument;
         }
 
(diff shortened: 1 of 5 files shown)


src/traversing.js
```diff
@@ -145,7 +145,7 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-        if ( nodeName( elem, "iframe" ) ) {
+        if ( nodeName( elem, "iframe" ) || nodeName( elem, "object" ) ) {
```
dmethvin: I wonder if we should just be checking `typeof elem.contentDocument !== "undefined"` or something to that effect. The check for a specific element is how the regression occurred. Are there any other elements where this might apply?

dmethvin: Looks like it's applet, frame, iframe, and object.

emibloque: Good point, thanks!
I'll use that check and work on tests for those tags too.

Also, I'm seeing that the [test for Node 8](https://travis-ci.org/jquery/jquery/jobs/368238987) is failing because the route for the svg is returning a 404.
Any advice on best practices for avoiding this?

dmethvin: Is there a specific reason why you're using `baseURL` in the test? Usually a relative reference like `data/1x1.svg` should work.

emibloque: I followed how it was done in other tests. Actually the problem was that I didn't add the `svg` extension to the karma config 😅.

I just applied the change you mention. However, due to the lack of browser support for `applet` nowadays, I wasn't able to prepare a test for this tag.

emibloque:
Hi @dmethvin !
Just to follow up on this, could I help with anything else here?
If there is anything left to do I will gladly try to work on that 😃

dmethvin:
@emibloque Nope, we're good. It should land before the next release goes out but I'm not sure when that will be at the moment. Thanks for the contribution!

Issue #3792, opened 2017-09-20
Remove package-lock.json?


mgol:
### Description ###
npm 5, even the version included in the latest Node.js 8.5.0 re-generates `package-lock.json` on each install. And when it does on a system that doesn't support all the optional dependencies that are supported on the OS where the lockfile was generated, it removes those optional deps from the lockfile.

The effect is that everyone firing `npm install` on our repo on any OS other than macOS will immediately get a dirty state of the repo as the `fsevents` dependency subtree gets removed from `package-lock.json`. That's a really bad experience.

timmywil:
hmm, it shouldn't regenerate the lockfile. Otherwise, what's the point of a lockfile. Has this been reported to npm?

mgol:
It seems to be https://github.com/npm/npm/issues/18135, open since August 10. No feedback from the npm team so far.

dmethvin:
Interesting that `yarn` does install everything and has an opposite problem: https://github.com/yarnpkg/yarn/issues/4190

PR #4051, opened 2018-04-20
tests: use npm ci

DanielRuf:
`npm ci` is much faster and is meant for CI. This PR is meant for evaluating the new ci command and testig the performance and compare the logs of Travis CI.

https://docs.npmjs.com/cli/ci
http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@
 .bower.json
 .sizecache.json
 yarn.lock
-package-lock.json
 
 npm-debug.log*
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ addons:
   chrome: stable
 env:
   - NPM_SCRIPT=test:browserless
+before_install:
+  - npm i -g npm@latest
+install:
+  - npm ci
 matrix:
   include:
     # Run browser tests only on one Node.js version to save time.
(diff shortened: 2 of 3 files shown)


mgol:
Thanks for the PR. However, there are a few problems with this approach:
1. We removed the lockfile on purpose in #3792 due to issues with it inside of npm.
2. We intentionally don't upgrade `npm` in CI to make sure our builds pass with the versions included by default in respective Node.js versions as that's what usually people use.
3. This change doesn't speed up the build which is the main supposed advantage of `npm ci` vs `npm install`; compare https://travis-ci.org/jquery/jquery/builds/369211845 & https://travis-ci.org/jquery/jquery/builds/367872474

In light of the above, I don't think we'll want to switch to `npm ci`.

DanielRuf:
Right. We just have to trigger a build to compare the results.

Thanks for the feedback.

DanielRuf:
> We removed the lockfile on purpose in #3792 due to issues with it inside of npm.

Which was fixen in npm 5.6 ;-) https://github.com/npm/npm/issues/18135#issuecomment-351085000
Also see https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

mgol:
We're aware of the supposed fixes for that in npm 5.6 but #17722 is still open and there are comments indicating this is still an issue when working cross-platform.

Besides, we haven't had many problems with sub-dependencies misbehaving so it seems our contributors would suffer more from lockfile issues rather than dependencies changing under us.

Issue #4009, opened 2018-03-13
Deprecate auto-appending `px` in 3.4.0?


mgol:
### Description ###
In 4.0 I plan to take a stab at #2795. I think, though, we should try to deprecate auto-adding `px` earlier than that. This may not require changes in jQuery itself but on the API pages & in Migrate but I'd like to discuss the approach before doing the work.

One question - do we want to deprecate auto-adding `px` altogether or just for the properties we will stop auto-adding for? (i.e. not some most common ones like `width`) Deprecating all of that would go towards the ideal goal but might trigger warnings on the vast majority of projects so perhaps it'd be less invasive to not do it on the ones we will auto-fix.

### Link to test case ###
N/A

mgol:
We have a migrate ticket ready so this can be closed; there's nothing to do in Core for this deprecation code-wise.

PR #4053, opened 2018-04-23
WIP CSS: Don't automatically add "px" to properties with a few exceptions

mgol:
### Summary ###

Don't automatically add "px" to properties with a few exceptions

Fixes gh-2795
Ref gh-4009

### Checklist ###

### DIFF
--- a/src/css/var/autoPx.js
+++ b/src/css/var/autoPx.js
@@ -1,0 +1,38 @@
+define( function() {
+
+"use strict";
+
+return {
+	margin: true,
+	border: true,
+	padding: true,
+	top: true,
+	right: true,
+	bottom: true,
+	left: true,
+	width: true,
+	height: true,
+	minWidth: true,
+	minHeight: true,
+	maxWidth: true,
+	maxHeight: true,
+	marginTop: true,
+	marginRight: true,
+	marginBottom: true,
+	marginLeft: true,
+	borderTopWidth: true,
+	borderTop: true,
+	borderRightWidth: true,
+	borderRight: true,
+	borderBottomWidth: true,
+	borderBottom: true,
+	borderLeftWidth: true,
+	borderLeft: true,
+	borderWidth: true,
+	paddingTop: true,
+	paddingRight: true,
+	paddingBottom: true,
+	paddingLeft: true
+};
+
+} );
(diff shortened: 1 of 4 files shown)


mgol:
This is WIP as it still needs more tests.

@dmethvin I noticed my regex from https://gist.github.com/mgol/c14fcffb24cf9c2f08eca519f4dabd90 didn't catch a few things like `margin` or `marginTop` and I wasn't able to compress the full list in a regex or a short function so I included the whole list in the end.

mgol:
This adds 120 bytes, quite a lot. :/ I'd appreciate ideas to reduce the size.

src/css/var/autoPx.js
```diff
@@ -0,0 +1,63 @@
+define( function() {
+
+"use strict";
+
+return {
```
gvanderest: As there's a lot of repetition here, would it be possible to build this using a combination of a few keywords?  

Granted this file should probably be static, but: a handful of these are permutations of ["margin","border","padding"] + [...direction...] + ["", "width", "height"] and some form of population function for those combinations could happen?

gvanderest: Additionally, as the constant "true" is used in all these cases-- could the number `1` be used instead to reduce space?

dmethvin: Between Uglify and the zgip compression, many of these things go away. We don't generally care about the unminified size so things like `true` to `1` don't help. It's hard to know for sure if it's smaller after min+gzip unless you do the permutations and see what happens.

mgol: @gvanderest In addition to what @dmethvin said, in JS there's no way to quickly compute a list of all props with a certain prefix, infix & suffix; we'd need to add a triple for-loop and that adds size, not reduces it.

Instead of constructing a list you can create a regexp but the conditions here are irregular enough that the regex would still be quite large. That said, I worked on it a little more and I have an alternative regex-based approach that matches more than we need but still a finite number of values so maybe it's OK. See #4055.

mgol:
I've submitted an alternative version of this PR, see #4055.

src/css/var/autoPx.js
mgol: I updated the PR, there were way too many properties listed here. It's +43 bytes now.

mgol:
Closing this in favor of #4055.

PR #4055, opened 2018-04-24
CSS: Don't automatically add "px" to properties with a few exceptions

mgol:
### Summary ###

Don't automatically add "px" to properties with a few exceptions

Fixes gh-2795
Ref gh-4009

This is an alternative to PR #4053.

Const over #4053:
- ~~it matches more than we theoretically want (the list in #4053 is precise), e.g. non-existent props like `top-width`~~
- it's a little slower than #4053 as it uses a regex instead of a simple property checking.

Pros over #4053:
- it only adds ~~1 byte~~ 6 bytes instead of 120

### Checklist ###

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -5,6 +5,7 @@ define( [
 	"./var/rcssNum",
 	"./css/var/rnumnonpx",
 	"./css/var/cssExpand",
+	"./css/isAutoPx",
 	"./css/var/getStyles",
 	"./css/var/swap",
 	"./css/curCSS",
@@ -16,7 +17,7 @@ define( [
 	"./core/init",
 	"./core/ready",
 	"./selector" // contains
-], function( jQuery, access, camelCase, rcssNum, rnumnonpx, cssExpand,
+], function( jQuery, access, camelCase, rcssNum, rnumnonpx, cssExpand, isAutoPx,
 	getStyles, swap, curCSS, adjustCSS, addGetHookIf, support, finalPropName ) {
 
 "use strict";
@@ -198,30 +199,6 @@ jQuery.extend( {
 		}
 	},
 
-	// Don't automatically add "px" to these possibly-unitless properties
-	cssNumber: {
-		"animationIterationCount": true,
-		"columnCount": true,
-		"fillOpacity": true,
-		"flexGrow": true,
-		"flexShrink": true,
-		"fontWeight": true,
-		"gridArea": true,
-		"gridColumn": true,
-		"gridColumnEnd": true,
-		"gridColumnStart": true,
-		"gridRow": true,
-		"gridRowEnd": true,
-		"gridRowStart": true,
-		"lineHeight": true,
-		"opacity": true,
-		"order": true,
-		"orphans": true,
-		"widows": true,
-		"zIndex": true,
-		"zoom": true
-	},
-
 	// Add in properties whose names you wish to fix before
 	// setting or getting the value
 	cssProps: {},
@@ -267,11 +244,9 @@ jQuery.extend( {
 				return;
 			}
 
-			// If a number was passed in, add the unit (except for certain CSS properties)
-			// The isCustomProp check can be removed in jQuery 4.0 when we only auto-append
-			// "px" to a few hardcoded values.
-			if ( type === "number" && !isCustomProp ) {
-				value += ret && ret[ 3 ] || ( jQuery.cssNumber[ origName ] ? "" : "px" );
+			// If the value is a number, add `px` for certain CSS properties
+			if ( type === "number" ) {
+				value += ret && ret[ 3 ] || ( isAutoPx( origName ) ? "px" : "" );
 			}
 
 			// background-* props affect original clone's values
--- a/src/css/adjustCSS.js
+++ b/src/css/adjustCSS.js
@@ -1,7 +1,8 @@
 define( [
 	"../core",
+	"./isAutoPx",
 	"../var/rcssNum"
-], function( jQuery, rcssNum ) {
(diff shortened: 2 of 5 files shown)


src/css/var/isAutoPx.js
```diff
@@ -0,0 +1,31 @@
...
+	//         /--------------------------------\
+	//        |                                  |
+	//        |                    /----------\  |
+	//        |  /-------------\  |            | |  /------------\
+	//        | |               | |  / Top  \  | | |              |
+	//        | |  / Margin  \  | | | Right  | | | |  / Width  \  |
+	//        |-+-|  Border   |-+-+-| Bottom |-+-|-+-|          |-+-END
+	//        |    \ Padding /       \ Left /    |    \ Height /
+	//        |                                  |
+	// BEGIN -|            /---------\           |
```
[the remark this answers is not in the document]
mgol: Hmm, `border` is a little special here in that it differs from `margin` and `padding`. I could try extracting it but we'll end up with more than +1 byte then. ;)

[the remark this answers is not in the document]
mgol: There are also others non-existing values like `topWidth`. That's not ideal but the important part is the list is finite and not that large.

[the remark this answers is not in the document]
mgol: I worry it also matches `min` & `max` - there are no such things in CSS but I wouldn't be surprised if they were added; it'd be worth to exclude them.

src/css.js
```diff
@@ -247,7 +231,7 @@ jQuery.extend( {
 
 			// If a number was passed in, add the unit (except for certain CSS properties)
```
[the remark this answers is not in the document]
mgol: Nice catch; changed.

mgol:
@dmethvin I updated the regexp (again). It now adds 7 bytes instead of 1 but I think it catches exactly what we want and nothing else. :) I updated the graph for the regex as well.

dmethvin:
Nice!

src/css/adjustCSS.js
```diff
@@ -16,10 +17,11 @@ function adjustCSS( elem, prop, valueParts, tween ) {
 				return jQuery.css( elem, prop, "" );
 			},
 		initial = currentValue(),
-		unit = valueParts && valueParts[ 3 ] || ( jQuery.cssNumber[ prop ] ? "" : "px" ),
+		unit = valueParts && valueParts[ 3 ] || ( isAutoPx( prop ) ? "px" : "" ),
 
 		// Starting value computation is required for potential unit mismatches
-		initialInUnit = ( jQuery.cssNumber[ prop ] || unit !== "px" && +initial ) &&
+		initialInUnit = elem.nodeType &&
```
mgol: Without this change, effects tests were failing on non-element objects on an `a` property. I've noticed they were already failing if `a` was changed to anything from `jQuery.cssNumber` and this PR makes almost everything "like from `jQuery.cssNumber`".

The `nodeType` check should most likely already be there, perhaps we should add it for 3.4.0 as well?

dmethvin: Oh, ugh. Put another way, we only want to add "px" to property values if they're on DOM elements. I still wish we hadn't supported animating plain objects,.

mgol: I extracted this change to PR #4061.

src/css/var/isAutoPx.js
```diff
@@ -0,0 +1,36 @@
...
+	//        |               |            |    /-------\     |
+	//        |               |  / Top  \  |   |         |    |
+	//         \--- Border ---+-| Right  |-+---+- Width -+---/
+	//                          | Bottom |
+	//                           \ Left /
+
+	// The first test is used to ensure that:
+	// 1. The prop starts with a lowercase letter (as we uppercase it for the second regex).
+	// 2. The prop is not empty.
+	return /^[a-z]$/.test( prop[ 0 ] ) &&
```
timmywil: Having a function is fine, but we should continue with the convention of creating regexes once outside of functions.

src/css/adjustCSS.js
mgol: #4601 landed, this PR rebased so this particular change is no longer a part of this PR.

mgol:
> It would be nice to see explicit tests covering "px" appending for these 14 + 6 + 10 properties and non-appending for other properties (both real ones like "line-height" and similarly-named fakes like "MarginLeftWidth" and "BorderTopHeight").

I intend to write the test for all matched props (I'll take the list from #4053) and add a few negative ones to the props from outside of the set. I just haven't gotten to it yet.

Thanks for the size reduction tips, I'll try them!

mgol:
PR updated, tests were added, all remarks by @timmywil & @gibson042 addressed.

Thanks to @gibson042's clever remarks we're now down from +6 bytes to -7. :astonished:

mgol:
@gibson042 
> It would be nice to see explicit tests covering "px" appending for these 14 + 6 + 10 properties and non-appending for other properties (both real ones like "line-height" and similarly-named fakes like "MarginLeftWidth" and "BorderTopHeight").

Most of these were added but it's hard to test fake ones as the browser throws them away so there's no observable effect.

mgol:
PR rebased with a revert of the fix from PR #4064 as it's no longer needed when we switch to the whitelist.

-7 bytes as of now (compared to master @ 75b77b4873abf9025b6ba60dec80aedd29df2bdd).

@timmywil are your concerns resolved? Your re-review is needed.

timmywil:
@mgol yes, that addressed my concern.

mgol:
Good news, suddenly it got down to -45 bytes compared to current `master` (c4f2fa2fb33d6e52f7c8fad9f687ef970f442179).

Issue #3964, opened 2018-02-06
outerWidth of svg yields NaNpx. Previously yielded a number.


AlexEdgcomb:
### Description ###

On v3.3.1, `outerWidth()` of an `svg` yields `NaNpx`.

I expected a number. Prior versions (e.g., v3.2.1) yielded a number.

Issue reproduces on each OS/browser combo I tried: Mac/{Chrome, Firefox, Safari} and Windows/{Edge, Firefox, Chrome}.

### Link to test case ###

v3.3.1: https://jsfiddle.net/u9pk0gdq/1/ (yields `NaNpx`)

v3.2.1: https://jsfiddle.net/378suvLg/ (yields a number)

gibson042:
Hmm, interesting. SVG elements don't have `offsetWidth` or `offsetHeight`, so `boxModelAdjustment` gets NaN when trying to calculate their scroll gutters. This can be fixed by adding an `|| 0` in the final `delta` adjustment and covering SVG dimension getters with unit tests. Would you like to contribute a PR, @AlexEdgcomb?

AlexEdgcomb:
@gibson042 , I appreciate the offer, but I'll pass this round. Thanks!

mgol:
@jonathanv3232 GitHub doesn't allow to assign issues to people without commit access to the repo so we can't set you but feel free to take a shot at a PR!

kktam:
Submitted [PR](https://github.com/jquery/jquery/pull/4025) for the fix.

PR #4096, opened 2018-06-11
Dimensions: fix computing outerWidth on SVGs

jbedard:
Fixes gh-3964

### Checklist ###

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -139,7 +139,10 @@ function boxModelAdjustment( elem, dimension, box, isBorderBox, styles, computed
 			delta -
 			extra -
 			0.5
-		) );
+
+		// If offsetWidth/offsetHeight is unknown, then we can't determine content-box scroll gutter
+		// Use an explicit zero to avoid NaN (gh-3964)
+		) ) || 0;
 	}
 
 	return delta;
--- a/test/unit/dimensions.js
+++ b/test/unit/dimensions.js
@@ -352,6 +352,66 @@ QUnit.test( "table dimensions", function( assert ) {
 	assert.equal( colElem.width(), 300, "col elements have width(), see #12243" );
 } );
 
+QUnit.test( "SVG dimensions (basic content-box)", function( assert ) {
+	assert.expect( 8 );
+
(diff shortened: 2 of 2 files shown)


src/css.js
```diff
@@ -130,11 +130,12 @@ function boxModelAdjustment( elem, dimension, box, isBorderBox, styles, computed
 
 	// Account for positive content-box scroll gutter when requested by providing computedVal
 	if ( !isBorderBox && computedVal >= 0 ) {
+		var val = elem[ "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ) ];
```
mgol: The `offsetWidth`/`offsetHeight` comment seems misplaced now, doesn't it?

Also, a comment about why the fallback is needed (below, at line 138) would be useful, it's not clear now.

gibson042: I agree with @mgol on comment placement, and also observe that
* defining a new variable here is against our general style, and
* `val` is an overly generic name, and
* it provides little benefit anyway, since `delta` will never be modified when offset* is unknown (`computedVal - computedVal - delta - extra - 0.5` is guaranteed to be negative in this case, so `Math.ceil` will not be positive, so `Math.max` will be 0).

I think we should keep things simpler:
```js
// offsetWidth/offsetHeight is a rounded sum of content, padding, scroll gutter, and border
// Scroll gutter is that value minus the other properties, rounded half-down to the nearest integer
delta += Math.max( 0, Math.ceil(
	elem[ "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ) ] -
	computedVal -
	delta -
	extra -
	0.5

// If offsetWidth/offsetHeight is unknown, then we can't determine content-box scroll gutter
// Use an explicit zero to avoid NaN (gh-3964)
) ) || 0;
```

jbedard:
Updated

@gibson042 I agree with everything you said. At one point the fix I had needed the var and I never really cleaned it up after. Thanks for pointing it out :+1:

sduquej:
:wave: Hey folks, are there plans to release a patch version that includes this fix?

mgol:
@sduquej It will be included in 3.4.0 which we'll hopefully release soon (but it's hard to predict exact timing).

PR #4123, opened 2018-07-10
Ajax: More checks before replacing %20 to + in data

### LINKED ISSUE #4119: Invalid encoding of %20 char sequence in $.ajax post, put body
### Description ###
Seems that jQuery 3 replaces `%20` char sequence to `+` in a request body during PUT and POST. A request like 
```javascript
$.ajax({
  method: 'put', 
  url: '/test/doc', 
  dataType: 'json',
  headers: {'content-type':'application/json'},
  data: JSON.stringify({val:'%20'})
})
```
actually sends `{"val":"+"}` in 3.2.1 and 3.3.1, both with Migration tools. 

jQuery however 2.x behaves correctly.

Update. Seems the reason is this block https://github.com/jquery/jquery/blob/master/src/ajax.js#L617-L621. It assumes `s.contentType` is the only source of `Content-Type` header, but there can be `s.headers['content-type']`, which later (https://github.com/jquery/jquery/blob/master/src/ajax.js#L648-L650) overrides `s.contentType` value.

ermouth:
### Summary ###
Before $.ajax replaces `%20` to `+` in POST-ed data,  `opts.headers['content-type']` is now checked as well as `opts.contentType`, and explicit header wins if present. Replacement only takes place if resulting content type starts with `application/x-www-form-urlencoded`.

Fixes #4119.

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -422,6 +422,9 @@ jQuery.extend( {
 			// uncached part of the url
 			uncached,
 
+			// real content-type, .headers member wins
+			realContentType,
+
 			// Create the final options object
 			s = jQuery.ajaxSetup( {}, options ),
 
@@ -615,9 +618,18 @@ jQuery.extend( {
 			s.url = cacheURL + uncached;
 
 		// Change '%20' to '+' if this is encoded form body content (gh-2658)
-		} else if ( s.data && s.processData &&
-			( s.contentType || "" ).indexOf( "application/x-www-form-urlencoded" ) === 0 ) {
-			s.data = s.data.replace( r20, "+" );
+		} else if ( s.data && s.processData ) {
+
+			// Build content-type from opts and headers section (gh-4119)
+			realContentType = s.contentType || "";
+			for ( i in s.headers ) {
+				if ( i.toLowerCase() === "content-type" ) {
+					realContentType = s.headers[ i ] || "";
+				}
+			}
+			if ( realContentType.indexOf( "application/x-www-form-urlencoded" ) === 0 ) {
+				s.data = s.data.replace( r20, "+" );
+			}
 		}
 
 		// Set the If-Modified-Since and/or If-None-Match header, if in ifModified mode.


timmywil:
Is there a reason we can't set `s.contentType` based on the header if present?

ermouth:
@timmywil I preferred to keep existing sequence of sending headers: if `s.contentType` and content-type in headers are different, they are sent as different. I just do not know exactly why we have this strange option, so kept as is.

ermouth:
@gibson042 I discovered this issue in the wild, it broke in-prod system. The issue is very hidden and uncovered itself several months after migrating from jQuery 2 to 3. 

It’s not good that there exist a hidden default value of a not-required option, that might corrupt data unexpectedly. It’s twice not good that jQuery 3 behaves differently than jQuery 2.

So it’s not a documentation issue, it’s a dangerous bug corrupting data silently, and it was not easy to root because no one in the world expects hidden values covertly override explicits somewhere in deep internals. 

However, if you think the patch should be an early override of `contentType`, I gonna do it. But I’m pretty sure I will need some assistance with making tests for it.

timmywil:
If there are different code paths for the same header based on the `contentType` option and not the header, then `contentType` can't be set by the header. Isn't the easy fix here to use the option?

ermouth:
@timmywil It might be easy fix for the software not yet written. For the software already deployed and dependent on the behavior demonstrated by 2.x the easiest fix is to roll back to 2.x branch, which was the first thing we did after the issue discovery.

In many places we generate headers required for a request by a separate function returning headers object. I suppose it‘s more or less common practice. Moreover, we _never_ use `.contentType`, and I was not aware of it until this issue. 

So to delete one specific prop from `.headers` and move it to opts before each and every request will not be an easy fix, at least for me. Moreover, I have to _enforce_ deletion of `.headers['content-type']` everywhere, because it‘s sheer existence in request options is a risk, since gives internal behavior inconsistent with actually sent data. 

So from practical POV I‘m sure it’s a bug and not documentation issue.

Hope I did all my best to fix it.

gibson042:
Thanks, @ermouth. I reviewed [jQuery 2.2.4](http://code.jquery.com/jquery-2.2.4.js) and agree that this is indeed a regression. So as I described above, I think we should have a built-in ajaxPrefilter (similar to the one [preventing inadvertent XSS](https://github.com/jquery/jquery/blob/dc48b11e0c6c77e2b96e89a18f34d7b0c4f9a9d4/src/ajax/script.js#L10-L14)) that overwrites `contentType` with the value of a supplied Content-Type header. And I think testing should be similar to that used for [binary response types](https://github.com/jquery/jquery/blob/dc48b11e0c6c77e2b96e89a18f34d7b0c4f9a9d4/test/unit/ajax.js#L1820-L1836) or [abort](https://github.com/jquery/jquery/blob/dc48b11e0c6c77e2b96e89a18f34d7b0c4f9a9d4/test/unit/ajax.js#L510-L527)).

svlima777:
$ and Ajax. Two good friends who prefer to play with others header and a few more things. I have struggled with an hidden default value for a long time. Can any Ajax fans tell me a logical explaination why they must interfer other peoples projects etc. What is the motivation? Let other people sit left with corrupted data. They know It's a dangerous bug. In my country such people usually get beaten till they cry for mama. But I will be satisfyed only to say: Stop this evil behaviour, grow up and get a life. Remember that Karma will come your way too

mgol:
This approach was rejected & we have a newer PR: #4650 so I'm closing this one.

Issue #3713, opened 2017-07-04
No error message when trying to use '!important' in a css call


Rainbowlemon:
### What do you expect to happen?
When calling `.css('key', 'val !important')`, the correct css is added to the element.

  ### What actually happens?
jQuery silently ignores the css value. This can be tested by performing the following:

    $('body').css('background', 'hotpink !important');

#### Using jQuery 3.2.1, but I can see this issue has [been around for a while](https://bugs.jquery.com/ticket/11173).

Whether or not this should be added to jQuery is another matter for debate, but I think regardless, the parser should return an error, or at least information, if it has failed to parse CSS that has been passed to it. I've just wasted half an hour on a project trying to figure out why my CSS wasn't being added to a dynamically created element, and didn't stop to think that jQuery might just be ignoring the `!important` part.

dmethvin:
There isn't a "parser" provided by jQuery.  We're just sending the full string to the browser. The [docs](http://api.jquery.com/css/) already say:

> Note: .css() ignores !important declarations. So, the statement $( "p" ).css( "color", "red !important" ) does not turn the color of all paragraphs in the page to red. It's strongly advised to use classes instead; otherwise use a jQuery plugin.

Rainbowlemon:
RTFM! Thanks. Any plans on using setProperty for setting styles? It'd be fairly trivial to add a priority in that case. It looks like it's used [here](https://github.com/jquery/jquery/blob/80f1c8239e9d5f793f3e54e0cb8d7bd0747e4856/src/css.js#L284) for custom props, but not for anything else.

sketchman2017:
Hello.

I found [a question](https://stackoverflow.com/questions/2655925/how-to-apply-important-using-css) on stackoverfow about setProperty.

May I add a parsing of the important word and executing setProperty? I make some [changes](https://github.com/sketchman2017/jquery/commit/4c6a0cb22413ddee155607ee5e4b382f3cc7381e) in forked version. It is working now, but there is some issues.

1. There is a [finalPropName](https://github.com/jquery/jquery/blob/80f1c8239e9d5f793f3e54e0cb8d7bd0747e4856/src/css.js#L246) call there which rewrites name variable. For example, if name is "background-color", calling this method makes name variable equals "backgroundColor". But setProperty() method can not be executed with such value. Thats why I save the value of name var to the oldName var before finalPropName executed. It is not very beautiful, but setProperty not working with lowerCamelCased property names.

[...]

PR #4141, opened 2018-07-31
CSS: use setProperty to support !important

Sulokit:
### Summary ###
Resolution to issue #3713, where !important was not taken when css property was been set using $.css method.

Please review and let me know if any changes are required.

### Checklist ###

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -264,7 +264,7 @@ jQuery.extend( {
 			// If a hook was provided, use that value, otherwise just set the specified value
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
-				if ( typeof value === "string" && value.includes( "!important" ) ) {
+				if ( typeof value === "string" && value.indexOf( "!important" ) !== -1 ) {
 
 					// Support !important priority (gh-3713).
 					var splitValue = value.split( "!" );
--- a/src/css.js
+++ b/src/css.js
@@ -264,7 +264,7 @@ jQuery.extend( {
 			// If a hook was provided, use that value, otherwise just set the specified value
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
-				if ( value.indexOf( "!" ) !== -1 ) {
+				if ( typeof value === "string" && value.includes( "!important" ) ) {
 
 					// Support !important priority (gh-3713).
 					var splitValue = value.split( "!" );
--- a/src/css.js
+++ b/src/css.js
@@ -217,6 +217,7 @@ jQuery.extend( {
 
 		// Make sure that we're working with the right name
 		var ret, type, hooks,
+			argName = name,
 			origName = camelCase( name ),
 			isCustomProp = rcustomProp.test( name ),
 			style = elem.style;
@@ -263,14 +264,13 @@ jQuery.extend( {
 			// If a hook was provided, use that value, otherwise just set the specified value
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
+				if ( value.indexOf( "!" ) !== -1 ) {
 
-				if ( isCustomProp ) {
-					style.setProperty( name, value );
-				} else if ( value.indexOf( "!important" !== -1 ) ) {
-
-					//Added else if to resolve #3713
+					// Support !important priority (gh-3713).
 					var splitValue = value.split( "!" );
-					style.setProperty( name,  splitValue[ 0 ], splitValue[ 1 ] );
+					style.setProperty( argName,  splitValue[ 0 ], splitValue[ 1 ] );
+				} else if ( isCustomProp ) {
+					style.setProperty( name, value );
 				} else {
 					style[ name ] = value;
 				}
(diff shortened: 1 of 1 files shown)


src/css.js
```diff
@@ -266,6 +266,11 @@ jQuery.extend( {
 
 				if ( isCustomProp ) {
 					style.setProperty( name, value );
+				} else if ( value.indexOf( "!important" !== -1 ) ) {
```
mgol: Custom properties' values may have the `!important` clause as well, this needs to be handled as well.

src/css.js
```diff
@@ -266,6 +266,11 @@ jQuery.extend( {
...
+
+					//Added else if to resolve #3713
```
mgol: Comments need to start with space.

Also, I'd rephrase to:
```js
// Support !important priority (gh-3713).
```

src/css.js
mgol: Right. Unit tests would immediately catch this mistake. :)

Sulokit:
Thank you very much for your comments. This is my initial phase of starting my contribution to the open source community. I would need your reviews and help to understand and help the community in the better way. I've made the changes as per the comments, please review it and let me know if any other changes are needed. :). It would really be helpful. I would need help in running the unit tests. If there is any documentation of running the tests, please share it here, so that I can go through it and implement. Currently, I've tested multiple scenarios.

Sulokit:
Also, I've already signed up for Contributor License Agreement, I'm not sure, why it's showing as pending. Should I raise a ticket for the issue to be resolved?

src/css.js
```diff
@@ -263,8 +264,12 @@ jQuery.extend( {
 			// If a hook was provided, use that value, otherwise just set the specified value
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
+				if ( typeof value === "string" && value.includes( "!important" ) ) {
 
-				if ( isCustomProp ) {
+					// Support !important priority (gh-3713).
+					var splitValue = value.split( "!" );
+					style.setProperty( argName,  splitValue[ 0 ], splitValue[ 1 ] );
+				} else if ( isCustomProp ) {
```
dmethvin: Is it valid to say that custom properties never use `!mportant`? I would sure hope you can't override CSS variables with it or some other crazy thing 😵 but I'll ask anyway.

src/css.js
```diff
@@ -263,8 +264,12 @@ jQuery.extend( {
 			// If a hook was provided, use that value, otherwise just set the specified value
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
+				if ( typeof value === "string" && value.includes( "!important" ) ) {
 
-				if ( isCustomProp ) {
+					// Support !important priority (gh-3713).
+					var splitValue = value.split( "!" );
+					style.setProperty( argName,  splitValue[ 0 ], splitValue[ 1 ] );
```
dmethvin: This may (usually *will*) leave trailing space on the value string, is that okay? Otherwise it should be trimmed.

src/css.js
mgol: CSS vars do support “!important”. They’re basically like any other CSS prop, just with some additional features, the most important one that they’re not dropped by the engine what happens with regular unrecognized CSS props.

src/css.js
mgol: It should not be trimmed and I’d like to see a test that verifies that. The CSS var value is everything between the colon after the property name and “!important”. E.g. for the declaration:
```css
--foo: bar !important;
```
Running `getPropertyValue("--foo") on the matches element style returns `" bar "`, with both leading & trailing spaces.

src/css.js
```diff
@@ -263,8 +264,12 @@ jQuery.extend( {
 			// If a hook was provided, use that value, otherwise just set the specified value
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
+				if ( typeof value === "string" && value.includes( "!important" ) ) {
```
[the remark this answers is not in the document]
Sulokit: Thank you for your suggestion, I've replaced it with indexOf.. :)

src/css.js
Sulokit: @dmethvin getPropertyValue will work as usual. It won't be affected by `!important`

timmywil:
This is looking good. However, it needs tests before we can merge. The [README](https://github.com/jquery/jquery#running-the-unit-tests) has instructions for how to run the unit tests as you develop, but you can also run all tests with `npm test`.

src/css.js
```diff
@@ -217,6 +217,7 @@ jQuery.extend( {
 
 		// Make sure that we're working with the right name
 		var ret, type, hooks,
+			argName = name,
```
mgol: Why do you need this variable?

src/css.js
```diff
@@ -263,8 +264,12 @@ jQuery.extend( {
 			// If a hook was provided, use that value, otherwise just set the specified value
 			if ( !hooks || !( "set" in hooks ) ||
 				( value = hooks.set( elem, value, extra ) ) !== undefined ) {
+				if ( typeof value === "string" && value.indexOf( "!important" ) !== -1 ) {
 
-				if ( isCustomProp ) {
+					// Support !important priority (gh-3713).
+					var splitValue = value.split( "!" );
+					style.setProperty( argName,  splitValue[ 0 ], splitValue[ 1 ] );
```
mgol: This will not work for properties that need to be prefixed as you're using the unprocessed `argName` variable.

timmywil:
This has gotten old and stale. We'll need someone to take over.

PR #4146, opened 2018-08-03
Prevent negative getWidthOrHeight

### LINKED ISSUE #4145: .height() can give a negative number
```html
<html>
	<head>
		<title>Negative height</title>
		<script src="https://code.jquery.com/jquery-3.3.1.js"></script>
		</head>
		<body>
		<div style="display: none; height: 300px;">
			<div id="targetDiv" style="border: 8px solid red; top: 20%; bottom: 20%;"></div>
		</div>
		<script>
			const container = $('#targetDiv');
			const height = container.height();
			console.log(height);  // this is -16!
		</script>
	</body>
</html>
```

HolgerJeromin:
### Summary ###
This change prevents negative height and width.
fixes #4145

### Checklist ###

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -83,10 +83,11 @@ function boxModelAdjustment( elem, dimension, box, isBorderBox, styles, computed
 		// If we get here with a border-box (content + padding + border), we're seeking "content" or
 		// "padding" or "margin"
 		} else {
+
 			// For "content", subtract padding
 			if ( box === "content" ) {
 				var padding = jQuery.css( elem, "padding" + cssExpand[ i ], true, styles );
-				if(delta > padding) {
+				if ( delta > padding ) {
 					delta -= padding;
 				} else {
 					delta = 0;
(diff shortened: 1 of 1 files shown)


dmethvin:
Thanks for the PR! If clamping delta to a non-negative integer is the right thing to do here (I haven't analyzed the problem so don't know if it is) would it be possible to just return `Math.max( 0, delta )` at the end of the `boxModelAdjustment` function? Also, this should have some associated tests.

HolgerJeromin:
@dmethvin 
Thanks for your answer. That was my idea in the middle, too.
But as I suspected the tests fails again on this variant, like it does with this PR. I have no idea why the tests fails nor I have time in the next three weeks to fight against the test. Especially as I have no knowledge about the codebase.

gibson042:
I don't believe that the adjustment is necessarily positive, and am closing this since the issue it claims to fix is not a supported use case. But if you find the time to add tests confirming that the change is acceptable, please reopen and we will assess it.

Issue #3403, opened 2016-11-18
jqXHR.getResponseHeader


PSPvZ:
### Description ###

jqXHR.getResponseHeader fails to combine multiple HTTP headers with the same name. Instead, it only returns the value of the last HTTP header. This clashes with the specification of [XMLHttpRequest.getResponseHeader](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getResponseHeader), which states: "If there are multiple response headers with the same name, then their values are returned as a single concatenated string, where each value is separated from the previous one by a pair of comma and space."

The issue seems to occur only on Internet Explorer (I encountered it on IE11); calling IE's own XMLHttpRequest.getRequestHeader implementation does return the correct result, 

 From what I can tell Firefox 50 already combined the two headers into one before calling the XMLHttpRequest completion callback, inadvertently allowing jQuery to return the correct result.

Tested in jQuery 3.1.1.

### Link to test case ###

I have no test-environment that can easily set multiple headers with the same name, unfortunately.

gibson042:
The problem seems to be misimplementation of [`getAllResponseHeaders`](https://xhr.spec.whatwg.org/#the-getallresponseheaders()-method) (on which we [rely](https://github.com/jquery/jquery/blob/f18ca7bfe0f5e3184bf1ed55daf1668702c5577a/src/ajax/xhr.js#L109)) in at least IE11, but I otherwise agree with this assessment.

@PSPvZ, would you be willing to submit a PR updating our [`getResponseHeader`](https://github.com/jquery/jquery/blob/f18ca7bfe0f5e3184bf1ed55daf1668702c5577a/src/ajax.js#L461) to something like `responseHeaders[ match[ 1 ].toLowerCase() ] = (responseHeaders[ match[ 1 ].toLowerCase() ] || []).concat( match[ 2 ] ); … return match == null ? null : match.join( ", " )`?

pankajpatel:
Hello,
I tried to create a test for this which can be seen here: http://jsbin.com/vofipib/edit?js,console
Can I contribute on this issue?

gibson042:
> Can I contribute on this issue?

Yes! https://contribute.jquery.org/commits-and-pull-requests/

Andrei15193:
[...]

PR #4173, opened 2018-09-11
Fix getResponseHeader(key) for IE11

Andrei15193:
### Summary ###

This issue is described in more depth by #3403.

getResponseHeader(key) combines all header values for the provided key into a
single result where values are concatenated by ', '. This does not happen for
IE11 since multiple values for the same header are returned on separate lines.
This makes the function only return the last value of the header for IE11.

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -461,7 +461,7 @@ jQuery.extend( {
 							while ( ( match = rheaders.exec( responseHeadersString ) ) ) {
 								responseHeaders[ match[ 1 ].toLowerCase() + " " ] =
 									( responseHeaders[ match[ 1 ].toLowerCase() + " " ] || [] )
-									.concat( match[ 2 ] );
+										.concat( match[ 2 ] );
 							}
 						}
 						match = responseHeaders[ key.toLowerCase() + " " ];
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -454,14 +454,14 @@ jQuery.extend( {
 
 				// Builds headers hashtable if needed
 				getResponseHeader: function( key ) {
-					var match, existingHeaderValues, headerValues;
+					var match;
 					if ( completed ) {
 						if ( !responseHeaders ) {
 							responseHeaders = {};
 							while ( ( match = rheaders.exec( responseHeadersString ) ) ) {
-								existingHeaderValues = responseHeaders[ match[ 1 ].toLowerCase() ];
-								headerValues = ( existingHeaderValues || [] ).concat( match[ 2 ] );
-								responseHeaders[ match[ 1 ].toLowerCase() ] = headerValues;
+								responseHeaders[ match[ 1 ].toLowerCase() ] =
+									( responseHeaders[ match[ 1 ].toLowerCase() ] || [] )
+									.concat( match[ 2 ] );
 							}
 						}
 						match = responseHeaders[ key.toLowerCase() ];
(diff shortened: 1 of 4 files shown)


mgol:
Thanks for the PR. We need tests to land any fix like that.

Andrei15193:
@mgol I have extended the existing test for headers, hope it's good now.

src/ajax.js
```diff
@@ -454,17 +454,19 @@ jQuery.extend( {
 
 				// Builds headers hashtable if needed
 				getResponseHeader: function( key ) {
-					var match;
+					var match, existingHeaderValues, headerValues;
```
gibson042: These new variables are expensive in terms of size increase to jquery.min.js gzipped size. I think we should leave them out.

src/ajax.js
```diff
@@ -454,17 +454,19 @@ jQuery.extend( {
...
 					if ( completed ) {
 						if ( !responseHeaders ) {
 							responseHeaders = {};
 							while ( ( match = rheaders.exec( responseHeadersString ) ) ) {
-								responseHeaders[ match[ 1 ].toLowerCase() ] = match[ 2 ];
+								existingHeaderValues = responseHeaders[ match[ 1 ].toLowerCase() ];
+								headerValues = ( existingHeaderValues || [] ).concat( match[ 2 ] );
```
gibson042: Our existing code already [misbehaves](https://jsfiddle.net/0cj8y357/) when someone attempts to retrieve a header that shares the name of a lowercase `Object.prototype` property (e.g., "constructor" or "\_\_proto__"), but this will actually throw. If we're making changes, I'd like to fix that issue rather than exacerbate it (e.g., by appending a space as in [event.js](https://github.com/jquery/jquery/blob/e743cbd28553267f955f71ea7248377915613fd9/src/event.js#L390)).

gibson042: #4214

Andrei15193: Good catch! I didn't think of Object.prototype when using the same pattern on some of the stuff I've been doing, makes for a good interview question hehe

Fixed.

src/ajax.js
```diff
@@ -459,12 +459,14 @@ jQuery.extend( {
 						if ( !responseHeaders ) {
 							responseHeaders = {};
 							while ( ( match = rheaders.exec( responseHeadersString ) ) ) {
-								responseHeaders[ match[ 1 ].toLowerCase() ] = match[ 2 ];
+								responseHeaders[ match[ 1 ].toLowerCase() + " " ] =
+									( responseHeaders[ match[ 1 ].toLowerCase() + " " ] || [] )
+									.concat( match[ 2 ] );
```
mgol: This line should be indented one level further.

Andrei15193: Done.

test/middleware-mockserver.js
```diff
@@ -132,7 +132,11 @@ var mocks = {
 		resp.writeHead( 200, {
 			"Sample-Header": "Hello World",
 			"Empty-Header": "",
-			"Sample-Header2": "Hello World 2"
+			"Sample-Header2": "Hello World 2",
+			"List-Header": "Item 1",
+			"list-header": "Item 2",
+			"Constructor": "prototype collision (constructor)",
+			"__Proto__": "prototype collision (__proto__)"
```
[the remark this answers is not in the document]
Andrei15193: I've hit a bit of a problem with using `__proto__` in lowercase. `middleware-mockserver.js` may have the same issue with properties that collide with `Object.property`. It always sends `null`.

I had no issues with `constructor`, we can do it the other way around. Remove `__proto__` and keep `constructor` in lowercase. What do you think?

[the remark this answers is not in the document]
Andrei15193: Done.

Issue #4185, opened 2018-10-02
jQuery.css with empty value disables transition from 3.3.0


itchyny:
### Description ###
Setting some specific properties to empty value with jQuery.css once and setting a new style disables the animation (CSS transition). See the following examples.

- jQuery 3.2.1: https://jsfiddle.net/y34h7aef/
-- The div animates.
- jQuery 3.3.1: https://jsfiddle.net/j5ranug1/
-- The div doesn't animate.

This looks like a breaking change introduced in 3.3.0. I believe this is an unexpected behavior. It animates when I use jQuery 2 (https://jsfiddle.net/L1u7o8zv/).

### Link to test case ###
https://jsfiddle.net/j5ranug1/

mgol:
Thanks for the report.

This is caused by 3fcddd6e72e7e318c0b062e391d60867732318ae (cc @gibson042) and is a result of the `width` setter now reading the `box-sizing` style which requires the browser to recompute the styles. If that wasn't done then the browser would simply ignore the first few `.css` calls that reset the properties since they are set to different values shortly after.

Note that this problem already exists in jQuery 2, although in a more limited way. The `box-sizing` check was always done, just usually only for `outerWidth` & similar APIs; see an example with jQuery 2.2.4: https://jsfiddle.net/y34h7aef/3/

This issue has a larger impact when it affects bare `.css('width', ...)` setter, though. I think we may have to hide the `border-box` check again in an `extra` check. We've got another usage of it below (see the mentioned 3fcddd6e72e7e318c0b062e391d60867732318ae commit) but we could first check for `scrollBoxSize` (currently named that way, was `borderBoxReliable` in the commit) and only then query `box-sizing`.

Even if not for this issue, IMO we should do it for performance reasons. As it stands, just calling the `.css('width', ...)` setter repeatedly will cause layout thrashing.

mgol:
Hmm, `boxSizing` is queried only when needed, though - if either the `scrollboxSize` support test fails or when we need to adjust for `margins` etc. so there should be no layout thrashing...

I verified 3fcddd6e72e7e318c0b062e391d60867732318ae is the commit causing this issue, though. Investigating further...

mgol:
I have a PR: #4187.

Setting milestone to `3.4.0`.

PR #4187, opened 2018-10-03
CSS: Don't read styles.position in the width/height cssHook unless necessary

mgol:
### Summary ###

Current width/height cssHook reads the computed position style even if not
necessary as the browser passes the scrollboxSize support test. That has been
changed.

This commit also makes the scrollboxSize support test in line with all others
(i.e. only return true or false) and changes the variable name in the hook
to make the code clearer.

Fixes gh-4185

### Checklist ###

* [ ] New tests have been added to show the fix or feature works (testing this might be tricky)

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -358,10 +358,14 @@ jQuery.each( [ "height", "width" ], function( i, dimension ) {
 		set: function( elem, value, extra ) {
 			var matches,
 				styles = getStyles( elem ),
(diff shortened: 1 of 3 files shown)


mgol:
The only test I can think of would require mocking `getComputedStyle`, proxying its fields to the native API but also notifying it had a field read.

itchyny:
Thank you for the fix.

Issue #4102, opened 2018-06-13
`outerWidth()` doesn't work on IE when content width matches style=width specification when `box-sizing` set to `border-box`.


rgavrilov:
### Description ###
`.outerWidth()` calculation on IE 11 (Version: 11.98.16299.0IS, Windows 10 64-bits) is different from Chrome and is wrong, I think.

In the attached test case I have a table with 3 header rows. First cell in each row has a different width (200px, 190px, 180px), top row overrides width of cells in the 
When width specified in style attribute matches content width of the element - the value returned by `.outerWidth()` is different.

Running the test case on IE (in console) produces 200, 190, 200 while on Chrome the result is 200, 200, 200.

This is messes with my `dataTables` and `fixedColumn.dataTables` components that rely on `.css('width')` and hence `.outerWidth()`.

The problem seems to be in `getWidthOrHeight` function (jQuery v3.3.1). Specifically this line (6368):

```
	// Check for style in case a browser which returns unreliable values
	// for getComputedStyle silently falls back to the reliable elem.style
	valueIsBorderBox = valueIsBorderBox &&
		( support.boxSizingReliable() || val === elem.style[ dimension ] );
```

`val` is coming from `window.getComputedStyle().width` but `elem.style['width']` is coming from `style` attribute (as I understand it). And when these two values are equal it leads jQuery to believe that `val` is a `border` box size already and doesn't add padding.

Changing the width of the second row cell to `191px` or `189px` results in `outerWidth()` of the problematic cell (`th`) to `200`.

### Link to test case ###
https://codepen.io/rgavrilov/pen/KeqxJp

I expect all 3 values printed to codepen's console to be the same, but they differ on IE, even though all 3 TH cells (of the first column) are of the same width.

vmcruz:
I tested this on IExplorer 11, Windows 8.1 x64 and confirmed this behavior.

gibson042:
That CodePen link is no longer working in IE11, let alone IE9, so I made a JS Bin: https://jsbin.com/tupaxexigi/1/edit?html,js,output

The bottom line here is that _something_ is still necessary, because IE `getComputedStyle` does in fact report box-sizing "border-box" for elements that don't actually exhibit border-box dimensions behavior, which the check against inline style can reveal. But the new wrinkle is that HTML table layout completely invalidates such a check, because all cells in a row or column have the same height or width (respectively).

[...]

Issue #3589, opened 2017-03-23
.innerWidth(), .innerHeight() mishandling scroll bars in 3.2.x


hashchange:
`.innerWidth()` and `.innerHeight()` no longer return consistent results when scroll bars appear on an element.

Previously, the methods returned the size of the box up to the border, whether or not scroll bars are present. (Scroll bars are rendered inside of the border.) That's how the methods behaved from jQuery 1.2.6, when they were introduced, up to jQuery 3.1.1.

Starting with 3.2.0, the area covered by scroll bars is subtracted from the return value in Chrome and Opera (ie, Blink), IE, and Edge. That happens when the `box-sizing` of the element is set to `content-box`.

See the [test case at JS Bin](http://jsbin.com/tunobiv/1/edit?html,css,js,output).

FF is not affected by the bug. Blink and Edge behave correctly when `box-sizing` is set to `border-box`. In IE, the return value is still broken then.

The issue doesn't exist in Safari and mobile browsers because they render scroll bars as temporary overlays which don't take up space in the element.

It seems that the bug is a direct result of the fix #3561 for #3193. In `getWidthOrHeight`, you have switched to computed values for establishing the dimensions. You had used `getBoundingClientRect()` previously, and `offsetWidth`/`offsetHeight` before that.

I totally appreciate the difficulties raised in #3193. Still, the new approach backfires with regard to scroll bars. FF places them inside of the padding edge, while Blink and Edge place them inside of the inner border edge, but outside of the padding edge.

That difference between browsers doesn't matter when gBCR or offsetWidth/Height are queried. It also doesn't matter for the computed values of `width` and `height` as long as they are based on the border box (well, except in IE, for whatever reason). But it does change the computed values for the size of the content box, in the affected browsers.

timmywil:
@gibson042 made this as well. https://jsbin.com/rujinejifa/edit?html,css,js,output

timmywil:
We're not sure the best course here. Why do browsers differ? Is the spec ambiguous about including scrollbars in computed css?

gibson042:
[...]

PR #4223, opened 2018-11-12
Dimensions: fall back to offsetWidth/Height for border-box in IE

timmywil:
### Summary ###
Fixes gh-4102

If box-sizing is unreliable and box-sizing is border-box, we can use offsetWidth and offsetHeight in most situations. In the situations we can't, box-sizing seems to be reliable and we can use computed CSS. Specifically, in these cases:

- Child of hidden element
- Disconnected element

`valueIsBorderBox` should be `false` for SVG in IE9-11.

The downside here is that we lose fractional values for border-box dimensions in IE9-11. I can live with that, but open to review.

- I first added a commit to include the IE launcher for debugging IE issues on Windows machines, which came in handy for this.
- Also removed a couple unused variables.

### Checklist ###

With @gibson042's help, got it down to +9.

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -1,9 +1,7 @@
 define( [
 	"./core",
-	"./var/pnum",
 	"./core/access",
 	"./core/camelCase",
-	"./var/document",
 	"./var/rcssNum",
 	"./css/var/rnumnonpx",
 	"./css/var/cssExpand",
@@ -18,7 +16,7 @@ define( [
 	"./core/init",
 	"./core/ready",
 	"./selector" // contains
-], function( jQuery, pnum, access, camelCase, document, rcssNum, rnumnonpx, cssExpand,
+], function( jQuery, access, camelCase, rcssNum, rnumnonpx, cssExpand,
 	getStyles, swap, curCSS, adjustCSS, addGetHookIf, support, finalPropName ) {
 
 "use strict";
@@ -119,10 +117,9 @@ function boxModelAdjustment( elem, dimension, box, isBorderBox, styles, computed
 function getWidthOrHeight( elem, dimension, extra ) {
 
 	// Start with computed style
-	var styles = getStyles( elem ),
-		val = curCSS( elem, dimension, styles ),
-		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
-		valueIsBorderBox = isBorderBox;
+	var isBorderBox, valueIsBorderBox, offsetProp, offsetVal,
+		styles = getStyles( elem ),
+		val = curCSS( elem, dimension, styles );
 
 	// Support: Firefox <=54
 	// Return a confounding non-pixel value or feign ignorance, as appropriate.
@@ -133,21 +130,33 @@ function getWidthOrHeight( elem, dimension, extra ) {
 		val = "auto";
 	}
 
-	// Check for style in case a browser which returns unreliable values
-	// for getComputedStyle silently falls back to the reliable elem.style
-	valueIsBorderBox = valueIsBorderBox &&
-		( support.boxSizingReliable() || val === elem.style[ dimension ] );
+	isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
+	valueIsBorderBox = isBorderBox && support.boxSizingReliable();
+	offsetProp = "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 );
 
 	// Fall back to offsetWidth/offsetHeight when value is "auto"
 	// This happens for inline elements with no explicit setting (gh-3571)
 	// Support: Android <=4.1 - 4.3 only
 	// Also use offsetWidth/offsetHeight for misreported inline dimensions (gh-3602)
-	if ( val === "auto" ||
-		!parseFloat( val ) && jQuery.css( elem, "display", false, styles ) === "inline" ) {
+	// Support: IE 9-11 only
+	// Also use offsetWidth/offsetHeight for when box sizing is unreliable
+	// Do not use offset for elements that don't have it (e.g. SVG)
+	if ( typeof elem[ offsetProp ] !== "undefined" &&
+		( val === "auto" || ( isBorderBox && !valueIsBorderBox ) ||
+			!parseFloat( val ) && jQuery.css( elem, "display", false, styles ) === "inline" ) ) {
 
-		val = elem[ "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ) ];
+		offsetVal = elem[ offsetProp ];
+
+		// Support: IE 9-11 only
+		// offsetWidth/offsetHeight is zero for disconnect elements
+		// and children of hidden elements
+		if ( offsetVal || !parseFloat( val ) ) {
+			val = offsetVal;
+		}
 
 		// offsetWidth/offsetHeight provide border-box values
+		// In the case where val is not set,
+		// computed value can then be trusted to be border-box
 		valueIsBorderBox = true;
 	}
 
--- a/src/css.js
+++ b/src/css.js
@@ -117,7 +117,9 @@ function boxModelAdjustment( elem, dimension, box, isBorderBox, styles, computed
 function getWidthOrHeight( elem, dimension, extra ) {
 
 	// Start with computed style
-	var isBorderBox, valueIsBorderBox, offsetProp, offsetVal,
+	var isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
+		valueIsBorderBox = isBorderBox && support.boxSizingReliable(),
+		offsetProp = "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ),
 		styles = getStyles( elem ),
 		val = curCSS( elem, dimension, styles );
 
@@ -130,9 +132,6 @@ function getWidthOrHeight( elem, dimension, extra ) {
 		val = "auto";
 	}
 
-	isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
-	valueIsBorderBox = isBorderBox && support.boxSizingReliable();
-	offsetProp = "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 );
 
 	// Fall back to offsetWidth/offsetHeight when value is "auto"
 	// This happens for inline elements with no explicit setting (gh-3571)
@@ -141,21 +140,19 @@ function getWidthOrHeight( elem, dimension, extra ) {
 	// Support: IE 9-11 only
 	// Also use offsetWidth/offsetHeight for when box sizing is unreliable
 	// Do not use offset for elements that don't have it (e.g. SVG)
-	if ( typeof elem[ offsetProp ] !== "undefined" &&
-		( val === "auto" || ( isBorderBox && !valueIsBorderBox ) ||
+	// Computed value for SVG elements does not respect border-box in IE 9-11
+	// so valueIsBorderBox should be false there
+	if ( elem.hasOwnProperty( offsetProp ) &&
+		( val === "auto" || ( isBorderBox !== valueIsBorderBox ) ||
 			!parseFloat( val ) && jQuery.css( elem, "display", false, styles ) === "inline" ) ) {
 
-		offsetVal = elem[ offsetProp ];
-
 		// Support: IE 9-11 only
-		// offsetWidth/offsetHeight is zero for disconnect elements
+		// offsetWidth/offsetHeight is zero for disconnected elements
 		// and children of hidden elements
-		if ( offsetVal || !parseFloat( val ) ) {
-			val = offsetVal;
-		}
+		val = elem[ offsetProp ] || val;
 
 		// offsetWidth/offsetHeight provide border-box values
-		// In the case where val is not set,
+		// In the case where val remains unchanged,
 		// computed value can then be trusted to be border-box
 		valueIsBorderBox = true;
 	}
--- a/src/css.js
+++ b/src/css.js
@@ -117,11 +117,11 @@ function boxModelAdjustment( elem, dimension, box, isBorderBox, styles, computed
 function getWidthOrHeight( elem, dimension, extra ) {
 
 	// Start with computed style
-	var isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
+	var styles = getStyles( elem ),
+		val = curCSS( elem, dimension, styles ),
+		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
 		valueIsBorderBox = isBorderBox && support.boxSizingReliable(),
-		offsetProp = "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ),
-		styles = getStyles( elem ),
-		val = curCSS( elem, dimension, styles );
+		offsetProp = "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 );
 
 	// Support: Firefox <=54
 	// Return a confounding non-pixel value or feign ignorance, as appropriate.
@@ -142,9 +142,9 @@ function getWidthOrHeight( elem, dimension, extra ) {
 	// Do not use offset for elements that don't have it (e.g. SVG)
 	// Computed value for SVG elements does not respect border-box in IE 9-11
 	// so valueIsBorderBox should be false there
-	if ( elem.hasOwnProperty( offsetProp ) &&
-		( val === "auto" || ( isBorderBox !== valueIsBorderBox ) ||
-			!parseFloat( val ) && jQuery.css( elem, "display", false, styles ) === "inline" ) ) {
+	if ( ( val === "auto" || ( isBorderBox !== valueIsBorderBox ) ||
+			!parseFloat( val ) && jQuery.css( elem, "display", false, styles ) === "inline" ) &&
+			typeof elem[ offsetProp ] !== "undefined" ) {
 
 		// Support: IE 9-11 only
 		// offsetWidth/offsetHeight is zero for disconnected elements
--- a/src/css.js
+++ b/src/css.js
@@ -144,7 +144,7 @@ function getWidthOrHeight( elem, dimension, extra ) {
 	// so valueIsBorderBox should be false there
 	if ( ( val === "auto" || ( isBorderBox !== valueIsBorderBox ) ||
 			!parseFloat( val ) && jQuery.css( elem, "display", false, styles ) === "inline" ) &&
-			typeof elem[ offsetProp ] !== "undefined" ) {
+			offsetProp in elem ) {
 
 		// Support: IE 9-11 only
 		// offsetWidth/offsetHeight is zero for disconnected elements
--- a/src/css.js
+++ b/src/css.js
@@ -120,7 +120,7 @@ function getWidthOrHeight( elem, dimension, extra ) {
 	var styles = getStyles( elem ),
 		val = curCSS( elem, dimension, styles ),
 		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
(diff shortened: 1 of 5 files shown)


src/css.js
```diff
@@ -133,21 +130,33 @@ function getWidthOrHeight( elem, dimension, extra ) {
...
 	// Support: Android <=4.1 - 4.3 only
 	// Also use offsetWidth/offsetHeight for misreported inline dimensions (gh-3602)
-	if ( val === "auto" ||
-		!parseFloat( val ) && jQuery.css( elem, "display", false, styles ) === "inline" ) {
-
-		val = elem[ "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ) ];
+	// Support: IE 9-11 only
+	// Also use offsetWidth/offsetHeight for when box sizing is unreliable
+	// Do not use offset for elements that don't have it (e.g. SVG)
+	if ( typeof elem[ offsetProp ] !== "undefined" &&
```
mgol: Would moving the `typeof` check to the end of the `if` make sense to avoid computing it if not needed?

test/data/testsuite.css
```diff
@@ -142,3 +142,7 @@ section { background:#f0f; display:block; }
 	padding: 0;
 	margin: 0;
 }
+
+.border-box, .border-box * {
+	box-sizing: border-box;
+}
```
mgol: That's a total nit but why not:
```css
.border-box {
	box-sizing: border-box;
}
.border-box * {
	box-sizing: inherit;
}
```
? It's easier to override the behavior in sub-trees then.

src/css.js
```diff
@@ -133,21 +130,33 @@ function getWidthOrHeight( elem, dimension, extra ) {
...
+			!parseFloat( val ) && jQuery.css( elem, "display", false, styles ) === "inline" ) ) {
+
+		offsetVal = elem[ offsetProp ];
+
+		// Support: IE 9-11 only
+		// offsetWidth/offsetHeight is zero for disconnect elements
+		// and children of hidden elements
+		if ( offsetVal || !parseFloat( val ) ) {
+			val = offsetVal;
+		}
```
gibson042: Big size benefit from dropping `offsetVal` entirely and doing this like
```js
val = elem[ offsetProp ] || parseFloat( val ) || 0;
```
Plus one more from also dropping `offsetProp`:
```js
val = elem[ "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ) ] ||
	parseFloat( val ) || 0;
```

src/css.js
timmywil: It's my understanding that `typeof` doesn't cause a compute, but I'll double check that. Could also look at `hasOwnProperty`.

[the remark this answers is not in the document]
timmywil: Looked at this again. That's actually the same behavior as before, although it's harder to see. We apparently want it to be false there. I shouldn't have said that the computed value for SVG could be trusted to be border box in IE 9-11.

src/css.js
```diff
@@ -133,21 +130,33 @@ function getWidthOrHeight( elem, dimension, extra ) {
 		val = "auto";
 	}
 
-	// Check for style in case a browser which returns unreliable values
-	// for getComputedStyle silently falls back to the reliable elem.style
-	valueIsBorderBox = valueIsBorderBox &&
-		( support.boxSizingReliable() || val === elem.style[ dimension ] );
+	isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box";
+	valueIsBorderBox = isBorderBox && support.boxSizingReliable();
+	offsetProp = "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 );
```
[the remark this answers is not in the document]
timmywil: Fair enough

src/css.js
timmywil: I can do the first, but still need the second to check for existence of the property.

test/unit/dimensions.js
```diff
@@ -646,7 +646,9 @@ QUnit.test( "width/height on a table row with phantom borders (gh-3698)", functi
 	} );
 } );
 
-QUnit.test( "interaction with scrollbars (gh-3589)", function( assert ) {
+// Because we fallback to offsetWidth/offsetHeight,
+// IE9-11 does not report fractional values
+QUnit[ jQuery.support.boxSizingReliable() ? "test" : "skip" ]( "interaction with scrollbars (gh-3589)", function( assert ) {
```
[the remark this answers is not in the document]
timmywil: These tests only ensure that fractional values are maintained. Is there another issue your concerned about or do you think fractional values are more important than I do?

test/data/testsuite.css
timmywil: Never occurred to me, but the point of this is to explicitly set `border-box` on all of the elements. Allow them to overridden seems like it defeats the purpose.

timmywil:
> How would you feel about the approach from #4102 (comment) ?

The issue also applies to `textarea`, so limiting to table elements doesn't fix all issues. Also, I prefer a solution that doesn't require checking `nodeName`. Even though some table elements are given special consideration in the spec as offset parents, the solution still feels like it's in the same realm as browser sniffing.

We talked about `|| val === elem.style[ dimension ]` in the meeting. I see it as an enigmatic artifact from old code. It happens to work for a lot of cases, but has no real relation to what we want. What it's doing is checking if the computed value happens to be the same as the inline style and deciding that the value can be treated as border-box because it *probably* doesn't matter, but we've discovered here that's a risky assumption. Patching it up to include the table case (or the case where both values are zero) isn't really solving the problem as there could be more cases to cover that we haven't thought of yet. It's just adding band-aids to an already confusing line of code that doesn't have any logical bearing on whether a value is border-box. I'd much prefer something more explicit. I know that's difficult to accomplish given IE's *loose* relationship with computed border-box, but this is an attempt at addressing the real behavior in IE 9-11 and documenting that behavior with comments.

src/css.js
timmywil: You're right, it does cause a compute. I switched to `hasOwnProperty`, but I think I'll keep it at the front because I'm concerned that `jQuery.css( elem, "display", false, styles ) === "inline"` might be a little more expensive.

timmywil: Sigh, `hasOwnProperty` returned `false` in Chrome despite having the property on inline elements. I just did what you suggested.

test/unit/dimensions.js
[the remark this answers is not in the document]
gibson042: The tests are not for ensuring preservation of fractional values, but rather for verifying that `.{inner,outer}{Width,Height}` work correctly with scrollable elements as both getters and setters per #3589... doing so in all cases just happens to be tricky because the [fix](https://github.com/jquery/jquery/commit/80f1c8239e9d5f793f3e54e0cb8d7bd0747e4856#diff-414c8f59bff0d1b63680b64763d8c529R134) required arithmetic mixing possibly fractional CSS values with always-integer offsetWidth/offsetHeight, which is why that test block includes fractional CSS values—but note that it is explicitly uses `% 1` to work even on browsers that don't support them.

These tests are complex but IMO necessary, and they're working now. Any change that breaks them makes me uncomfortable, and any change that bypasses them is just begging for trouble.

src/css.js
gibson042: You can check for property existence by repeating `elem[ "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 ) ]`, which works in our favor here.

gibson042:
> We talked about `|| val === elem.style[ dimension ]` in the meeting. I see it as an enigmatic artifact from old code. It happens to work for a lot of cases, but has no real relation to what we want. What it's doing is checking if the computed value happens to be the same as the inline style and deciding that the value can be treated as border-box because it _probably_ doesn't matter, but we've discovered here that's a risky assumption.

I agree with you here, but I'm trying to gently usher IE out the door, and bringing in such a big hammer—especially one that needs workarounds on workarounds for things like SVG—runs the risk of inviting more strain on a weak point.

> Patching it up to include the table case (or the case where both values are zero) isn't really solving the problem as there could be more cases to cover that we haven't thought of yet. It's just adding band-aids to an already confusing line of code that doesn't have any logical bearing on whether a value is border-box. I'd much prefer something more explicit. I know that's difficult to accomplish given IE's _loose_ relationship with computed border-box, but this is an attempt at addressing the real behavior in IE 9-11 and documenting that behavior with comments.

OK, fair point. I'll re-review with that in mind.

test/unit/dimensions.js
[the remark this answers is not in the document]
gibson042: Let's run the tests in all environments but force them to be integer-only in IE like so:
```patch
-		fraction = jQuery( "<div style='width:4.5px;'/>" ).appendTo( parent ).width() % 1,
+		fraction = jQuery.support.boxSizingReliable() ?
+			jQuery( "<div style='width:4.5px;'/>" ).appendTo( parent ).width() % 1 :
+			0,
```

src/css.js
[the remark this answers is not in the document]
gibson042: But SVG elements _do_ respect box-sizing in IE: https://jsbin.com/zafagupixi
Also—and I had forgotten this until just now—`boxSizingUnreliable` really seems to track a specific misbehavior in the form of always reporting content box dimensions, which we can probably use to our advantage here.

But if you really want to use offsetWidth/offsetHeight, what would you say to a variation on something like this?
```js
// Fall back to offsetWidth/offsetHeight when value is "auto"
// This happens for inline elements with no explicit setting (gh-3571)
// Support: IE 9-11 only
// Also use offsetWidth/offsetHeight for when box sizing is unreliable
// Support: Android <=4.1 - 4.3 only
// Also use offsetWidth/offsetHeight for misreported inline dimensions (gh-3602)
if ( isBorderBox !== valueIsBorderBox || val === "auto" ||
		!parseFloat( val ) && jQuery.css( elem, "display", false, styles ) === "inline" ) {

	// Some elements don't have offsetWidth/offsetHeight (e.g., SVG)
	// But where available, it is a means of reading (non-fractional) border box dimensions
	if ( elem[ offsetProp ] > 0 ) {
		val = elem[ offsetProp ];
		valueIsBorderBox = true;
	}
}
```

src/css.js
timmywil: I don't think we should do that, even if it's smaller.

src/css.js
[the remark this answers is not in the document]
timmywil: Are you sure you tried that in IE? I see a different result than in Chrome.

test/unit/dimensions.js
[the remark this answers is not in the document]
timmywil: Thanks for the explanation. I gave up on these tests too quickly then.

src/css.js
timmywil: Although it may be moot now. Now that it falls back to val, `undefined` is fine.

timmywil: Oh wait, that's only if we want `valueIsBorderBox` to be `true` for SVG. From my testing, it seems like it should be `false`. I'll wait for your reply on the jsbin.

src/css.js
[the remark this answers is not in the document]
timmywil: Or maybe I'm confusing your point. Are we saying the same thing in different ways? The render seems fine, it's just that computed value for border-box is content-box for SVG. Your code suggestion should work.

test/unit/dimensions.js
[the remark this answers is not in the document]
timmywil: Your suggestion works.

src/css.js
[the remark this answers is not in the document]
timmywil: Actually, I just tried that and it broke the inline/disconnected tests. We want `valueIsBorderBox` to still be set to true for inline and disconnected elements, but false for SVG.

timmywil: @mgol brilliant

timmywil: I like it. It brings it down to +6 as well.

test/unit/dimensions.js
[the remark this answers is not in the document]
timmywil: Fixed IE9 by accounting for the element shrinking at the test level rather than trying to adjust the width. Our code compensates for scrollbars + border-box + position: absolute in IE9 only and it's like it was trying to compensate twice.

src/css.js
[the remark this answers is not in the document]
gibson042: I don't think a statement like "we want `valueIsBorderBox` to still be set to true for inline and disconnected elements, but false for SVG" is accurate. What we want is proper input for `boxModelAdjustment`, which returns the amount by which the width or height of a specified box (conveyed by `box`) exceeds a CSS computed width or height (subject to a box model indicated by `isBorderBox`). `boxModelAdjustment` could probably use some clarifying renames and/or comments, but the intended returned values are as follows:

`box` ＼ `isBorderBox` | false (`content-box`) | true (`border-box`)
---- | ---- | ----
content | 0 | -border - padding
padding | padding (+ scroll gutter) | -border
border | padding + border (+ scroll gutter) | 0
margin | padding + border + margin (+ scroll gutter) | margin

The [tests that failed](https://github.com/jquery/jquery/blob/6153eb0fd401cda90bf2007335cd4338093d38f0/test/unit/dimensions.js#L415) are... bad. They were only passing before this PR because of the inline style check that you're removing, and they're only passing in the current state of this PR because it includes an implicit border-box bias when _attempting_ to use offsetWidth/offsetHeight, even if that attempt fails.

We make no promises about accuracy in those disconnected/hidden cases, but if we want to continue doing our best then the right way would be to trust the computed box-sizing value for disconnected elements.

test/unit/dimensions.js
```diff
@@ -694,13 +696,6 @@ QUnit.test( "interaction with scrollbars (gh-3589)", function( assert ) {
 			borderBox.clone().css( { overflow: "auto" } ).appendTo( parent )[ 0 ].offsetWidth -
 			borderBox[ 0 ].offsetWidth;
 
-	if ( borderBoxLoss > 0 ) {
-		borderBox.css( {
-			width: ( size + borderBoxLoss ) + "px",
-			height: ( size + borderBoxLoss ) + "px"
-		} );
-	}
```
gibson042: It took some self-convincing, but this change seems to be OK. However, I do like the original style better because `borderBox.innerWidth(99); … assert.equal(borderBox.innerWidth(), 99)` is more direct than `borderBox.innerWidth(82); … assert.equal(borderBox.innerWidth() + 17, 99)`, especially in comparison with the corresponding `contentBox` assertions (whose values are always 6 less, to account for 1px border width and 2px of padding on each edge).
<details><summary><code>borderBox</code> evaluation diff</summary>

```diff
--- before
+++ after
 size = 100
 borderBoxLoss = 17
-borderBox.css( { width: "117px", height: "117px" } )
 i = 0
 assert.equal( contentBox.innerWidth/innerHeight(), 104 )
 assert.equal( contentBox.outerWidth/outerHeight(), 106 )
-assert.equal( borderBox.innerWidth/innerHeight(), 98 )
-assert.equal( borderBox.outerWidth/outerHeight(), 100 )
+assert.equal( borderBox.innerWidth/innerHeight() + 17, 98 )
+assert.equal( borderBox.outerWidth/outerHeight() + 17, 100 )
 i = 1
 size = 101
-borderBox.innerWidth( 98 + 1 ).innerHeight( 98 + 1 )
+borderBox.innerWidth( 81 + 1 ).innerHeight( 81 + 1 )
 assert.equal( contentBox.innerWidth/innerHeight(), 105 )
 assert.equal( contentBox.outerWidth/outerHeight(), 107 )
-assert.equal( borderBox.innerWidth/innerHeight(), 99 )
-assert.equal( borderBox.outerWidth/outerHeight(), 101 )
+assert.equal( borderBox.innerWidth/innerHeight() + 17, 99 )
+assert.equal( borderBox.outerWidth/outerHeight() + 17, 101 )
 i = 2
 size = 103
-borderBox.outerWidth( 101 + 2 ).outerHeight( 101 + 2 )
+borderBox.outerWidth( 84 + 2 ).outerHeight( 84 + 2 )
 assert.equal( contentBox.innerWidth/innerHeight(), 107 )
 assert.equal( contentBox.outerWidth/outerHeight(), 109 )
-assert.equal( borderBox.innerWidth/innerHeight(), 101 )
-assert.equal( borderBox.outerWidth/outerHeight(), 103 )
+assert.equal( borderBox.innerWidth/innerHeight() + 17, 101 )
+assert.equal( borderBox.outerWidth/outerHeight() + 17, 103 )
```
</details>

gibson042: Update: No, these changes are in fact wrong. And we need that initial `.css(…)` call to reveal the bug, because the difference is a constant offset after that.

src/css.js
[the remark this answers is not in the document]
gibson042: @timmywil this should work, and at a reasonable +9 bytes over master:
```js
var styles = getStyles( elem ),
	val = curCSS( elem, dimension, styles ),
	isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
	valueIsBorderBox = isBorderBox,
	offsetProp = "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 );
…
// Fall back to offsetWidth/offsetHeight when value is "auto"
// This happens for inline elements with no explicit setting (gh-3571)
// Support: IE 9-11 only
// Also use offsetWidth/offsetHeight for when box-sizing is unreliable
// Support: Android <=4.1 - 4.3 only
// Also use offsetWidth/offsetHeight for misreported inline dimensions (gh-3602)
if ( isBorderBox && !support.boxSizingReliable() || val === "auto" ||
	!parseFloat( val ) && jQuery.css( elem, "display", false, styles ) === "inline" ) {

	if ( elem.getClientRects().length ) {

		// Where available, offsetWidth/offsetHeight approximate border box dimension.
		// Where not available (e.g., SVG), assume unreliable box-sizing and interpret the
		// retrieved value as a content box dimension.
		valueIsBorderBox = offsetProp in elem;
		if ( valueIsBorderBox ) {
			val = elem[ offsetProp ];
		}
	}
}
```
(note the initial assignment of `valueIsBorderBox` to match `isBorderBox`, which will always be preserved for hidden/disconnected elements but will be overridden for visible elements when box-sizing makes gCS dimensions unreliable—matching IE9-11 behavior)

test/unit/dimensions.js
timmywil: Sorry what bug does it reveal exactly? The element is still being shrunk, but we're making it shrink to the the value we expect in the assertion instead of changing the assertion. I don't really see a logical difference. Either way, we're accounting for the size difference in IE9. I don't really mind your code suggestion. I'm just not seeing any advantage other than a cosmetic one.

src/css.js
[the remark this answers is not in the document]
timmywil: My statement is still accurate tho, right? The point was simply that the computed value can be trusted for hidden/disconnected but not for SVG, which is maintained in your suggestion. But this does avoid accessing offset for hidden/disconnected when unnecessary so I like it better.

test/unit/dimensions.js
timmywil: I went ahead and changed it, but I'd still like to understand how the previous was wrong.

src/css.js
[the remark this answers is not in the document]
timmywil: Updated. I just moved the rects check to the outer if statement as it was the same size.

timmywil:
It's amazing how dense this code is.

test/unit/dimensions.js
gibson042: I think it's because of the particularly strange manifestation of #3699... we want to have a dimension set before the assertions. Things might still work otherwise, but I definitely saw some passes that should have been failures when I was experimenting. And if I remember correctly from the original ticket, there were scenarios where the reported values didn't agree with the painted dimensions.

Bottom line: "wrong" is probably too strong, but it is more reliable to ensure a set before testing against expectations.

timmywil:
I should be able to get to it today.

mgol:
The test added in this PR still fails in Android Browser 4.0-4.3: http://swarm.jquery.org/result/3143623.

timmywil:
come on android

PR #4225, opened 2018-11-13
updated most dev dependencies to the latest version

abnud1:
### Summary ###
This pull request has updated code to the latest versions of most dev dependencies, updates include : 
1- jsdom to v13
2- Qunit to v2.8.0
3- sinon to v7.1.1
4- karma to v3.1.1
5- patch update requirejs(to v2.3.6)
6- husky to v1.1.4
7- babel to v7
8- all grunt plugins to the latest
also replaced seemingly unmaintained gzip-js library with pako(https://github.com/nodeca/pako)

some Notes:
1- when upgrading jsdom to v13, node smoke test iterable_with_symbol_polyfill stoped to work as apparently jsdom can't work with any es6 Symbol polyfill(not just core-js).

Since jsdom 13 require Nodejs 8 or more and since Nodejs 8 already has Symbol no polyfill is needed there(And I see Nodejs 8 as a safe bet since v6 is in maintenance LTS mode)
2- when upgraded QUnit to v2, tests run on browsers (`grunt karma:main`) started to fail, Headless mode only shows one failure while `grunt karma:chrome-debug` shows many failures.

Manual testing showed me that failures are not false negatives, they are actual failures, anyway more investigation is needed
### Checklist ###


timmywil:
Thank you for contributing! Updates like this are much more likely to land if each dependency update has its own PR, to separate concerns. Some of these would be easy to land, but others may not make it.

PR #4229, opened 2018-11-14
replaced unmaintained gzip-js with pako

abnud1:
### Summary ###
This PR replaces gzip-js with pako (https://github.com/nodeca/pako) because the former is not maintained any more
### Checklist ###

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -13,7 +13,7 @@ module.exports = function( grunt ) {
 	}
 
 	var fs = require( "fs" ),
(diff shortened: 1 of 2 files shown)


mgol:
Thanks for the PR. However, the `pako` library you used produces 87 bytes larger bundles than our existing `gzip-js` one. We may revisit that when we start working on jQuery 4.0 (I'd probably prefer us to use the built-in `zlib` module in that case, though) but for now we'll stick to what we have.

PR #4231, opened 2018-11-14
updated jsdom to v13

abnud1:
### Summary ###
This PR updates jsdom to the latest v13.
Note: 
jsdom 13 doesn't work with any Symbol polyfill thus making the smoke test iterable_with_symbol_polyfill fails.
since jsdom 13 requires Nodejs >= 8 and this version of Node already has Symbol it doesn't need a polyfill so the test isn't really needed there(even Node 6 supports Symbol)
### Checklist ###


mgol:
Thanks for the PR. Unfortunately, we do need all the current tests as we want to ensure jQuery works with a missing Symbol implementation. That’s why I’ve been holding off updating jsdom as newer versions require native Symbol to work.

The real solution is to make this test run in browsers instead of Node with jsdom but that requires some effort. In any case, the approach from this PR won’t work so I’ll close it.

We’re open to PRs that fulfill our requirements. Please let us know if you’d like to help there.

PR #4234, opened 2018-11-14
upgraded QUnit to v2

abnud1:
### Summary ###
This PR is about the upgrade to QUnit 2(v2.8.0 to be more specific).
Note:
Unfortunately the Headless browser test(`grunt karma:main`) fails in one test only which is the `offsetParent` test in `test/unit/offset.js`.  
I didn't really know why, for some reason `#qunit-fixture` content doesn't seem to match `qunit-fixture.html` although I verified by console.logging that it is !!
### Checklist ###


mgol:
Thanks for the PR, a lot of nice changes! As you noticed, there are still some problems to resolve; your PR doesn't work for me on a local PHP setup. I decided to create a version of this PR that doesn't update QUnit but that does apply the needed code changes to prepare us for this upgrade in the future. I think we may want to wait with the actual upgrade until we start working on jQuery 4.0 so that we don't need to debug issues related to e.g. old Android Browsers if something goes wrong.

The changed PR is at #4297, I'll close this one. You'll be credited as the commit author. Thanks!

PR #4256, opened 2018-12-07
Resolves the issue that when a field value (input text or hidden) is …

joaomfrebelo:
…seted with $.fn.val is not accessible by the selector [value=val]

### Summary ###
When the value of a input type text or hidden field is setted with the $.fn.val is not accessible by
the selector [value=val]  

### Checklist ###


gibson042:
Thank you for the contribution, but this is not behavior that we want. `.val` is intended to deal _only_ with dynamic element state; please use `.attr` for interacting with element attributes.

PR #4243, opened 2018-11-26
Manipulation: Only evaluate HTTP-successful script src

gibson042:
Fixes gh-4126

### Checklist ###
* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [x] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* [ ] ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~

Issue #4126, opened 2018-07-12
Unhandled http error while appending an html with script element


HolgerJeromin:
### Description ###
I have an application where the customers can load arbitrary HTML code which is then loaded into the DOM via jQuery.
A customer had a copy and paste error which linked to an non existing js file. This lead to an 404 but nevertheless jQuery pushes the return value (error HTML page) into a HTMLScriptElement text property. This leads to an exception:

    Uncaught SyntaxError: Unexpected token <
    at DOMEval (jquery-3.3.1.js:111)
    at Function.globalEval (jquery-3.3.1.js:345)
    at text script (jquery-3.3.1.js:9640)
    at ajaxConvert (jquery-3.3.1.js:8784)
    at done (jquery-3.3.1.js:9255)
    at XMLHttpRequest.<anonymous> (jquery-3.3.1.js:9548)
    at Object.send (jquery-3.3.1.js:9600)
    at Function.ajax (jquery-3.3.1.js:9206)
    at Function.jQuery._evalUrl (jquery-3.3.1.js:9367)
    at domManip (jquery-3.3.1.js:5759)

### Minimal test case ###
Put this HTML onto an webserver:
````html
<!DOCTYPE html>
<html>
  <head>
    <script type="text/javascript" src="https://code.jquery.com/jquery-3.3.1.js"></script>
    <title>Title</title>
	<script>
	function clickme() {
			var jBody = $(document.body);
			jBody.append('<script type="text/javascript" src="notexisting.js"><'+'/script>');
	}
	</script>
  </head>
  <body>
        <button onclick="clickme()">Click</button>
  </body>
</html>
````

timmywil:
Thanks for opening an issue. However, you can find support on stackoverflow.com or the [jquery forums](https://forum.jquery.com).

HolgerJeromin:
Why was this report closed? 
I am using the jQuery API and jQuery is pushing knowingly (http code !== 200) garbage into an HTMLScriptElement text content.

mgol:
I can't confirm your issue, this works as expected in my JS Bin: http://jsbin.com/laqumev/edit?html,js,console,output

Something else must be going in your code.

HolgerJeromin:
@mgol 
You changed my code to have a cross domain request. This changes the callstack and therefor prevents this error. (Could have mentioned it, sorry for that)
Changing it back results in 
`DOMException: Only secure origins are allowed`
so jsbin is no valid hoster for this bug.
[...]

PR #4258, opened 2018-12-13
Manipulation: Always make _evalUrl requests as dataType: script

gibson042:
IE and iOS \<10 XHR transport does not succeed on data: URIs
Ref gh-4243
Ref gh-4126


mgol:
@gibson042 Could you explain what the issue is and why your new solution preserves the solution from #4243 while not triggering issues in IE or old iOS?

Issue #1741, opened 2014-10-21
focus event ignores additional data when triggered


mgol:
Originally reported by mail@… at: http://bugs.jquery.com/ticket/13428

<p>
<a class="ext-link" href="http://fiddle.jshell.net/rvX5Y/4/" rel="nofollow"><span class="icon">&nbsp;</span>http://fiddle.jshell.net/rvX5Y/4/</a>
</p>

<pre class="wiki">$('input').on('focus', function(event, extra, data) {
    console.log(extra, data);
});
$('input').trigger('focus', ['extra', 'data']);
</pre><p>
Works in version 1.8.3.
</p>

<p>
Does not work in version 1.9.0 and 1.9.1.
</p>

<p>
Expected behavior: Console shows "extra data"
</p>

<p>
False Behavior: Console shows "undefined undefined"
</p>

<p>
Browser: Firefox 18,
OS: OS X 10.8.2
</p><p>Issue reported for jQuery 1.9.1</p><p>Blocked by: #13353</p>

netil:
For 'focus', 'blur' and 'click' events, it fires native event in certain condition and when
calling native event, currently there is no way to pass through trigger's data to event handler.

Well, is not the best solution but I though to use special event hook to handle that.
Need to modify trigger to hold the data, and add handle method to invoke event handler function
(obviously it can be generalized to cover other events too).

What do you think guys?

```js
		focus: {
			// Fire native event if possible so blur/focus sequence is correct
			trigger: function() {
				if ( this !== safeActiveElement() && this.focus ) {

					// Check if triggered with data
					if( arguments.length > 1 ) {
						jQuery.event.special.focus.triggerData = jQuery.makeArray(arguments).splice(1);
					}

					this.focus();
					return false;
				}
			},
			delegateType: "focusin",
			handle: function( event ) {
				var param = [ event ],
					data = jQuery.event.special.focus.triggerData;

				if( data ) {
					param = param.concat( data );
				}

				event.handleObj.handler.apply( this, param );
				delete jQuery.event.special.focus.triggerData;
			}
		}
```

gibson042:
That solution is at risk for introducing strange behavior flips between the "no arguments" and "arguments" cases, and is not reentrant safe. I don't see us using such an approach.

dmethvin:
Agreed. In retrospect, allowing users to pass data has caused a lot of issues and this is one of them. Focus management is highly flaky in browsers already, it doesn't take much to cause problems.

[...]

PR #4277, opened 2019-01-14
Event: Support jQuery extensions for native-backed click events

gibson042:
### Summary ###
Leverage native events rather than circumventing them.

Ref gh-3494
Ref gh-1367
Fixes gh-3423
Fixes gh-3751
Fixes gh-4139
~~Fixes gh-1741~~

### Checklist ###
* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [x] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* [ ] ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~

PR #4279, opened 2019-01-16
Event: Leverage native events for focus and blur

gibson042:
### Summary ###
Builds upon #4277 ([diff](https://github.com/gibson042/jquery/compare/3.4-native-events...gibson042:3.4-native-focus-blur-events)). This is currently _super_ sloppy, but it works!

Fixes gh-1741

```
   raw     gz Sizes
277644  82250 dist/jquery.js
 88015  30580 dist/jquery.min.js

   raw     gz Compared to master @ 9cb162f6b62b6d4403060a0f0d2065d3ae96bbcc
 +4001   +996 dist/jquery.js
  +891   +201 dist/jquery.min.js
```

Size reduction ideas:
* combine the focus and blur special objects as in https://github.com/jquery/jquery/pull/1367/files#diff-031bb62d959e7e4949d1847c82507f33R684
  * remove the explicit return in their `trigger` methods
* replace `safeActiveElement` with something for direct use in `leverageNative`, e.g. accepting (element, eventType) and returning maybeAsync
  * possibly merge into another `leverageNative` parameter as well
* replace `this` with `el` in the `leverageNative` handler (but only if that persuades minification to duplicate more character sequences)
* refactor the outermost synthetic event return logic in the `leverageNative` handler for more gzip-friendly minification output

### Checklist ###


src/event.js
```diff
@@ -504,6 +565,77 @@ jQuery.event = {
...
+
+			// Interrupt processing of the outermost synthetic .trigger()ed event
+			if ( ( event.isTrigger & 1 ) && this[ type ] && !saved ) {
+
+				// Store arguments for use when handling the inner native event
+				saved = slice.call( arguments );
+				dataPriv.set( this, type, saved );
+
+				// Trigger the native event and capture its result
+				// Support: IE <=9 - 11+
```
mgol: This suggests it may still happen in some future version (or its de-facto evolution, i.e. the EdgeHTML-based Edge) but here we know that's not the case. I usually try to end the 11-closed IE comment with `only` instead:
```suggestion
				// Support: IE <=9 - 11 only
```

src/event.js
```diff
@@ -554,32 +582,52 @@ function leverageNative( el, type, forceAdd ) {
...
 
 				// Store arguments for use when handling the inner native event
-				dataPriv.set( this, type, slice.call( arguments ) );
+				saved = slice.call( arguments );
+				dataPriv.set( this, type, saved );
 
 				// Trigger the native event and capture its result
+				// Support: IE <=9 - 11+
+				// focus() and blur() are asynchronous
+				maybeAsync = allowAsync( this, type );
```
mgol: The second argument to `allowAsync` doesn't seem to be used anywhere. Can we drop it?

src/event.js
```diff
@@ -504,6 +565,77 @@ jQuery.event = {
 	}
 };
 
+// Ensure the presence of an event listener that handles manually-triggered
+// synthetic events by interrupting progress until reinvoked in response to
+// *native* events that it fires directly, ensuring that state changes have
+// already occurred before other listeners are invoked.
+function leverageNative( el, type, forceAdd, allowAsync ) {
```
mgol: This will hurt readability but since `forceAdd` may be only `false` or `returnTrue`, we could squash it with `allowAsync`, changing the signatures to:
```js
leverageNative( el, type, allowAsync ); // old `leverageNative( el, type, false, allowAsync )`
leverageNative( el, type, false ); // old `leverageNative( el, type, returnTrue )`
```
That may save some size.

test/unit/event.js
```diff
@@ -2911,23 +2909,15 @@ QUnit.test( "originalEvent type of simulated event", function( assert ) {
...
-		input[ 0 ].addEventListener( "click", function( nativeEvent ) {
-			expectedType = nativeEvent.type;
-			jQuery.event.simulate( "focusin", this, jQuery.event.fix( nativeEvent ) );
-		} );
-		input[ 0 ].click();
-	}, QUnit.config.testTimeout / 4 || 1000 );
+	input[ 0 ].addEventListener( "click", function( nativeEvent ) {
+		jQuery.event.simulate( "focusin", this, jQuery.event.fix( nativeEvent ) );
+	} );
+	input[ 0 ].click();
```
mgol: What makes this more reliable in TestSwarm now than previously?

test/unit/event.js
```diff
@@ -2955,54 +2945,16 @@ QUnit.test( "VML with special event handlers (trac-7071)", function( assert ) {
 	ns.remove();
 } );
 
-QUnit.test( "Check order of focusin/focusout events", function( assert ) {
```
mgol: Has the order changed? Why is this being removed?

test/unit/event.js
```diff
@@ -3050,6 +3002,62 @@ QUnit.test( "focus-blur order (#12868)", function( assert ) {
...
+		checkbox.trigger( "blur", data );
+		setTimeout( function() {
+			assert.notEqual( document.activeElement, checkbox[ 0 ],
+				"element not focused after blur event (default action)" );
+			done();
+		}, 50 );
+	}, 50 );
+} );
+
+// TODO replace with an adaptation of
```
mgol: Why?

test/unit/event.js
```diff
@@ -2397,18 +2399,31 @@ QUnit.test( "checkbox state (#3827)", function( assert ) {
...
 
 	// Handlers only; checkbox state remains false
 	jQuery( cb ).triggerHandler( "click" );
+	assert.equal( cb.checked, false, "handlers only - checkbox is still unchecked" );
+
+	// Trigger parameters are preserved (trac-13353, gh-4139)
+	cb.checked = true;
+	assert.equal( cb.checked, true, "jQuery event with data - checkbox is initially checked" );
+	jQuery( cb ).on( "click", function( e, data ) {
+		assert.equal( data, "clicked", "trigger data passed to handler" );
```
markelog: I see failure for this assertion in the latest Chrome, might be my setup tho

src/event.js
```diff
@@ -478,19 +478,48 @@ jQuery.event = {
...
 
-					this.click();
-					return false;
+				// Force post-minification `this` variable for better similarity with _default
+				var el = this || data;
+
+				// Claim the first handler
+				if ( rcheckableType.test( el.type ) &&
+					el.click && nodeName( el, "input" ) &&
+					dataPriv.get( el, "click" ) === undefined ) {
```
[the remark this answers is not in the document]
markelog: Nice idea. Also would it cost more, size wise, if we reverse guards to early return technic? Like
```js
if ( !rcheckableType.test( el.type )) {
  return false;
}

if ( !nodeName( el, "input" )) {
  return false;
}

...
```

Might be easier to read?

`el.click` check, looks redundant? Or not?

test/unit/event.js
```diff
@@ -2382,12 +2382,14 @@ QUnit.test( "clone() delegated events (#11076)", function( assert ) {
...
-QUnit.test( "checkbox state (#3827)", function( assert ) {
-	assert.expect( 9 );
+QUnit.test( "checkbox state (trac-3827)", function( assert ) {
+	assert.expect( 16 );
 
-	var markup = jQuery( "<div><input type=checkbox><div>" ).appendTo( "#qunit-fixture" ),
+	var markup = jQuery( "<div class='parent'><input type=checkbox><div>" ),
 		cb = markup.find( "input" )[ 0 ];
 
+	markup.appendTo( "#qunit-fixture" );
```
markelog: Just out of curiosity - why you moved `appendTo` here and added CSS classname to `div`?

src/event.js
```diff
@@ -504,6 +565,77 @@ jQuery.event = {
...
+
+	// Register the controller as a special namespace-universal handler
+	dataPriv.set( el, type, forceAdd );
+	jQuery.event.add( el, type, {
+		namespace: false,
+		handler: function( event ) {
+			var maybeAsync, result,
+				saved = dataPriv.get( this, type );
+
+			// Interrupt processing of the outermost synthetic .trigger()ed event
```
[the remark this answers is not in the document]
gibson042: It's essentially a summary of the relevant part of `leverageNative` strategy... when a (outer) synthetic event is triggered, we want to interrupt propagation (because focus/checkedness states are wrong) and replace it with a possibly-asynchronous native event (which will update them before invoking listeners). The handling of that native event then involves replacing its further propagation with an _inner_ synthetic event (for forwarding data from the outer trigger) that runs synchronously and is **not** interrupted, essentially picking up where outermost handling leaves off.

src/event.js
gibson042: I prefer the 11+ form to indicate that the issue was not actually resolved within IE, and would presumably still exist if the browser is revived. But I'm willing to defer to @jquery/core consensus here.

src/event.js
gibson042: Yes, but that would increase the size. And since it also fits flexibility patterns from elsewhere in the library, I'd like to keep it.

src/event.js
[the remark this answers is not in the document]
mgol: @gibson042 OK, I feel convinced by your gzip knowledge. If you want, you can try various approaches but I won't be blocking the PR on that. I'm resolving the discussion.

src/event.js
```diff
@@ -507,6 +536,54 @@ jQuery.event = {
...
+
+	if ( forceAdd ) {
+
+		// Force setup through jQuery.event.add
+		jQuery.event.add( el, type, forceAdd );
+		return;
+	}
+
+	// Register the controller
+	dataPriv.set( el, type, forceAdd );
```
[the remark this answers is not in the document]
mgol: That's a nice idea!

src/event.js
```diff
@@ -504,6 +565,77 @@ jQuery.event = {
...
+	// Setup must go through jQuery.event.add
+	if ( forceAdd ) {
+		jQuery.event.add( el, type, forceAdd );
+		return;
+	}
+
+	// Register the controller as a special namespace-universal handler
+	dataPriv.set( el, type, forceAdd );
+	jQuery.event.add( el, type, {
+		namespace: false,
```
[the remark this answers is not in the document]
mgol: The new comment looks more clear to me, thanks.

src/event.js
```diff
@@ -478,19 +478,48 @@ jQuery.event = {
...
-			// For checkable types, fire native event so checked state will be right
-			trigger: function() {
-				if ( rcheckableType.test( this.type ) &&
-					this.click && nodeName( this, "input" ) ) {
+			// Utilize native event to ensure correct state for checkable inputs
+			setup: function( data ) {
 
-					this.click();
-					return false;
+				// Force post-minification `this` variable for better similarity with _default
```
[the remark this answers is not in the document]
mgol: That means the `|| data` part will never be applied as `this` will always exist here and the `|| data` presence only increases gzip efficiency - do I understand it correctly?

If that's the case, adding something akin to the following to the existing comment:
```js
// `|| data` is dead code meant only to increase compressibility.
```
would make the purpose of that code much clearer IMO.

src/event.js
mgol: Yes, I think I meant something like that. If this doesn't help size then consider the request moot.

src/event.js
mgol: I guess it mattered more when EdgeHTML was still in development and `11+` might suggest to some that newer engine is still affected. This will soon no longer be the case once a few Chromium-based Edge releases come out and we stop testing on EdgeHTML.

I'd still want us to be consistent here but I'll be doing a support comments cleanup when we start working on 4.0 anyway so it's not necessary to settle on something now. Hence, I'm resolving the discussion.

src/event.js
mgol: OK, fair enough.

src/event.js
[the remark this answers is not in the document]
mgol: OK, I guess it'd be hard to explain it better here.

test/unit/event.js
mgol: I guess it's fine as we never guaranteed the order and it was different in IE anyway but I'd like to make it clear we're changing it.

@jquery/core what do you think?

test/unit/event.js
mgol: I see, it's about testing various namespace combinations. Makes sense.

mgol:
@gibson042 One more thing, a comment that I added in my original PR and I don't see the reply:
> How do you plan to land it, considering the author of the first commit is different?

test/unit/event.js
timmywil: If the order changes, this has the potential to break existing code. That kinda smells of a major version to me.

gibson042:
> How do you plan to land it, considering the author of the first commit is different?

Shouldn't be an issue, because https://github.com/jquery/jquery/pull/4279/commits/eecb078f14f0a2b7db469e85174df2a5386c7480 cherry-picks a commit that we already [merged in](https://github.com/jquery/jquery/commit/b442abacbb8464f0165059e8da734e3143d0721f) (but subsequently [reverted](https://github.com/jquery/jquery/commit/4d6b4536b2e0148d4b228884a0b356e73dd8beec)).

test/unit/event.js
gibson042: I don't think we've ever documented a guaranteed order, and moreover I don't think we're consistent across supported browsers right now. But I can try restoring this test to see what happens.

mgol: Technically speaking it's not a breaking change as we haven't documented the order. That said, if something is consistent between browsers, people tend to start depending on that and changing the behavior may break them. @gibson042 is right the order is different between browsers right now but it's only different between IE & the rest. If a site doesn't support IE (there are more & more of such sites), for their purposes the current order is consistent between browsers. So there's risk involved here.

mgol:
@gibson042 
>> How do you plan to land it, considering the author of the first commit is different?
> 
> Shouldn't be an issue, because https://github.com/jquery/jquery/pull/4279/commits/eecb078f14f0a2b7db469e85174df2a5386c7480 cherry-picks a commit that we already [merged in](https://github.com/jquery/jquery/commit/b442abacbb8464f0165059e8da734e3143d0721f) (but subsequently [reverted](https://github.com/jquery/jquery/commit/4d6b4536b2e0148d4b228884a0b356e73dd8beec)).

To clarify - do you want to land it as two commits: a re-land of the original one plus your fixes or just squash it to one commit authored by you since the original changes are credited via Git history anyway?

mgol:
@gibson042 @timmywil I restored the removed order tests locally and checked in Chrome & Firefox. The order of events is as follows:
1. focusin
1. focus
1. focusin
1. focusout
1. blur

When comparing to the orders from #4300, remember #4300 presents `focusout`/`blur` on one blurred element & `focusin`/`focus` on another focused one; here it's all one input. Therefore, it seems the focusing order is:
1. focusin
1. focus
1. focusin

while the blurring one is:
1. focusout
1. blur

Therefore, the order is not native; it's closer to the existing jQuery one but it has one additional second `focusin` that fails the test. We should look into that before merging IMO as that looks buggy.

test/unit/event.js
mgol: To add to the above, while the current order is inconsistent between IE & others, what's consistent (& verified by the removed test) is that `focusin` fires before `focus` & `focusout` before `blur`.

mgol:
The second `focusin` is invoked via `this[ type ]()` in `leverageNative`. While the `event.stopImmediatePropagation()` is called, for a `focus` event that won't prevent `focusin` from propagating.

An additional complication are the `focusin`/`focusout` shims. While `focus` delegates to `focusin` & `blur` to `focusout`, in non-IE browsers `focusin` is implemented via `focus` in capture mode and `focusin` via `blur` in capture mode. But even if this shim is disabled the double-`focusin` issue still exists in the PR currently.

mgol:
I rebased the PR & restored the `focusin`/`focusout` test with one modification: `focusin` is allowed to fire more than once.

mgol:
The current version of the PR passes tests when run via Karma or by opening `/test/?module=event` manually but fails when run via TestSwarm. It's not a case of different browser settings as I checked that all on my local TestSwarm install with the same Chrome & Safari browsers. The error happens in the `originalEvent type of simulated event` event test:
```
TypeError: undefined is not an object (evaluating 'event.originalEvent.type')
in http://jquery.l//test/unit/event.js on line 2900
```
It happens in Chrome & Safari, it doesn't happen in Firefox/Edge/IE. It also succeeds in old mobile Chrome/Safari, see my local table (ignore the AJAX failures in IE, that's a BrowserStack issue):
Screen Shot 2019-03-20 at 16 10 25

I'm trying to debug that now...

mgol:
OK, it might be a local caching issue, let me try to land it and see on our real TestSwarm instance...

Issue #4278, opened 2019-01-15
CSP: Provide a way to pass nonce to $.globalEval


tandraschko:
In #3541 CSP support was introduced to not loose the nonce attribute of script tags when repacing/inserting DOM nodes.

Therefore DOMEval already has support for nonce.
As $.globalEval already uses DOMEval under the hood, it should be quite to allow to pass the nonce.

current scipt:
```
// Evaluates a script in a global context
globalEval: function( code ) { 
		DOMEval( code ); 
}
```

something like this should work:
```
// Evaluates a script in a global context 
globalEval: function( code, nonce ) { 
		DOMEval( code, null, { nonce: nonce }); 
}
```

Workaround:
```
        var script = document.createElement('script');
        script.setAttribute('nonce', 'myNonce');
        script.innerHTML = 'console.log("test")';
        document.head.appendChild(script);
```

mgol:
Thanks for the report. PR: #4280.

mgol:
@tandraschko The PR landed and it will be included in jQuery 3.4.0.

tandraschko:
Great! Thank you!

tandraschko:
When do you plan to release 3.4.0?

mgol:
If we're lucky, somewhere between a week or two. But don't hold me to it, we're doing it in our free time so there may always be unexpected delays, sometimes even long ones.

PR #4269, opened 2019-01-09
Core: Preserve CSP nonce on scripts in DOM manipulation

mgol:
### Summary ###

Preserve CSP nonce on scripts in DOM manipulation.

Fixes gh-3541

### Checklist ###

* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [x] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~

Issue #3541, opened 2017-02-14
Issue using JQuery 3.1.1 and CSP


victorgaspar:
### Description ###
Hi,

I'm using JQuery version 3.1.1 and I'm trying to implement Content Security Policy directives on my webpage.

I'm getting the following error:
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'nonce-c20t41c7-73c6-4bf9-fde8-24a7b35t5f71'". Either the 'unsafe-inline' keyword, a hash ('sha256-KAcpKskREkEQf5B3mhDTonpPg34XnzaUC5IoBrOUrwY='), or a nonce ('nonce-...') is required to enable inline execution.

The error is produced on line 82 of the main jquery.js script file. The content of this line is:
doc.head.appendChild( script ).parentNode.removeChild( script );

Basically, it adds an inline script tag to the DOM, and violates the CSP.

I do not want to use 'unsafe-inline'. Is there any other way to circumvent this error? 

As you can see on the CSP violation, I'm using CSP level 2 (nonce). Would it be possible (some how) to inform JQuery to use this nonce when appending the script tag?

Thank you very much.

Kind regards,
V.

Brian-Aykut:
Looks like all attributes get stripped by this call on line 5782 
	`DOMEval( node.textContent.replace( rcleanScript, "" ), doc );`
A new script tag gets created without any attributes and as a result the nonce is no longer present which causes the CSP violation. 

The following can be used to demonstrate/replicate the issue: 

```
<!DOCTYPE html>
<html>
<head>
    <meta http-equiv="Content-Security-Policy" content="default-src http://localhost 'nonce-123456' ; child-src 'none'; object-src 'none'; script-src 'nonce-123456';">
    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.js" nonce="123456"></script> <!-- HTML nonce works -->
    <script nonce="123456">
        //This works
        console.log('Inline nonce works');

        //This will also work
        var s = document.createElement('script');
        s.setAttribute('nonce', '123456');
        s.innerHTML = 'console.log("Dynamically generated inline tag works")';
        document.head.appendChild(s);

        // This won't work
        var s2 = document.createElement('script');
        s2.setAttribute('nonce', '123456');
        s2.innerHTML = 'console.log("Dynamically generated inline tag appended via jQuery doesn\'t work")';
        $(document.head).append( s2 ); // This will throw a CSP error    
[...]

PR #4280, opened 2019-01-17
Core: Support passing nonce through jQuery.globalEval

mgol:
### Summary ###

Support passing nonce through jQuery.globalEval:

```js
jQuery.globalEval( code, nonceValue );
```

Fixes gh-4278
Ref gh-3541
Ref gh-4269

### Checklist ###

Docs issue: https://github.com/jquery/api.jquery.com/issues/1123

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -238,8 +238,8 @@ jQuery.extend( {
 	},
 
 	// Evaluates a script in a global context
-	globalEval: function( code ) {
-		DOMEval( code );
+	globalEval: function( code, options ) {
+		DOMEval( code, { nonce: options && options.nonce } );
 	},
 
 	each: function( obj, callback ) {
--- a/src/core/DOMEval.js
+++ b/src/core/DOMEval.js
@@ -10,26 +10,29 @@ define( [
 		noModule: true
 	};
 
-	function DOMEval( code, doc, node ) {
+	function DOMEval( code, node, doc ) {
 		doc = doc || document;
 
-		var i,
+		var i, val,
 			script = doc.createElement( "script" );
 
 		script.text = code;
 		if ( node ) {
 			for ( i in preservedScriptAttributes ) {
-				if ( node[ i ] ) {
-					script[ i ] = node[ i ];
-				} else if ( node.getAttribute( i ) ) {
 
-					// Support: Firefox 64+, Edge 18+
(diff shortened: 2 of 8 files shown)


mgol:
I'll create docs issue if this gets accepted.

mgol:
This is `-1` bytes but that's a clever lie. It turns out you can set `nonce` via `setAttribute`, you just can't trust `getAttribute( "nonce" )`. Thus, I simplified the previous code which allowed to achieve the size decrease. Just applying the refactor without adding support for #4278 would give `-11` bytes. So, effectively, this is `+10` bytes.

mgol:
The biggest question is: is this API OK? Or would you rather see something else?

src/core.js
```diff
@@ -238,8 +238,8 @@ jQuery.extend( {
 	},
 
 	// Evaluates a script in a global context
-	globalEval: function( code ) {
-		DOMEval( code );
+	globalEval: function( code, nonce ) {
```
gibson042: I don't like the signature, one could easily imagine a realm parameter occupying that position. And I don't really feel great about an options object here, but maybe that is the best approach. :\

src/core/DOMEval.js
```diff
@@ -10,26 +10,29 @@ define( [
...
+				// the `nonce` attribute is reset to an empty string whenever it
+				// becomes browsing-context connected.
+				// See https://github.com/whatwg/html/issues/2369
+				// See https://html.spec.whatwg.org/#nonce-attributes
+				// The `node.getAttribute` check was added for the sake of
+				// `jQuery.globalEval` so that it can fake a nonce-containing node
+				// via an object.
+				val = node[ i ] || node.getAttribute && node.getAttribute( i );
+				if ( val ) {
+					script.setAttribute( i, val );
```
gibson042: :+1:

src/core.js
mgol: Oh, the last version doesn't work as we still need the `getAttribute` fallback for non-nonce params. But https://github.com/mgol/jquery/commit/e49e3e865bbab856fe67b6dd75aa844643badbc5 is fine.

timmywil: I prefer the options object, even if there's only one option.

mgol:
Docs issue: https://github.com/jquery/api.jquery.com/issues/1123

PR #4305, opened 2019-02-26
Build: Update jsdom; migrate a test with Symbol polyfill to an iframe test

mgol:
### Summary ###

So far, we've been testing that jQuery element iteration works with polyfilled
Symbol & transpiled for-of via a Node test with jsdom with the Symbol global
removed. Unfortunately, jsdom now requires Symbol to be present for its internal
functionality so such a test is no longer possible. Instead, it's been migrated
to an iframe test with transpiled JavaScript.

This PR also enables us to use ECMAScript 2017 or newer in Node.js code.

~~This PR depends on Node.js 8+ as jsdom dropped support for older versions. [Upstream support for Node.js 6 ends in April 2019](https://github.com/nodejs/Release) so we can land this PR when we start working on jQuery 4.0 on `master`.~~ EDIT: not entirely true, see my comment below.

### Checklist ###


mgol:
Interestingly, this passes Travis so jsdom seems to work even though it doesn't support Node.js 6 officially. Loading the `grunt-karma` task fails but it's only used in a separate Node.js 8 run so it doesn't affect the final status code.

Therefore, we may land it without waiting for jQuery 4.0.

PR #4307, opened 2019-02-28
Build: Update QUnit from 1.23.1 to 2.9.2

mgol:
### Summary ###

Update QUnit from 1.23.1 to 2.9.2, fixing a few remaining issues.

It turns out passing `module=basic` etc. in the query string still works, it's just if you choose a module from the dropdown you'll see a different query string generated. Because of that, integration with TestSwarm still works which I verified locally.

~~I'd still wait with that until we start working on jQuery 4.0 so that we don't need to debug any potential issues in browsers we'll be dropping. Other than that, this should be ready.~~
EDIT: It seems to work so we'll try it now.

### Checklist ###


test/unit/effects.js
```diff
@@ -1971,36 +1971,37 @@ QUnit.test( "Animation callbacks (#11797)", function( assert ) {
...
 		done: function() {
-			assert.step( 4 );
+			assert.step( "done" );
 		},
 		fail: function() {
 			assert.ok( false, "Animation failed" );
 		},
 		always: function() {
-			assert.step( 5 );
+			assert.verifySteps( [ "start", "progress", "last progress", "done" ] );
```
dmethvin: Oh this is really nice, I didn't know about this feature!

PR #4310, opened 2019-03-04
Build: Run the basic test suite in jsdom

mgol:
### Summary ###

The basic test suite is now run in jsdom.

### Checklist ###

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -14,7 +14,15 @@ module.exports = function( grunt ) {
 
 	var fs = require( "fs" ),
 		gzip = require( "gzip-js" ),
-		isTravis = process.env.TRAVIS;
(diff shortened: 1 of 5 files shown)


mgol:
This does require dropping Node.js 6.x support which will happen in about a month (that's the only reason Travis is failing).

mgol:
@dmethvin @timmywil I changed the logic in this PR so that the `karma:jsdom` task is:
* invoked as part of the `test:slow` task
* skipped in Node.js <8

In this way we could merge it even now and have a guarantee this works on 3.x as well. We can remove the old Node logic on `master` once we start working on v4.

Any objections to land it in the current form?

timmywil:
@mgol sounds good to me

PR #4315, opened 2019-03-09
fix selector dependency on init module

midnight-wonderer:
### Summary ###
There is no way people would use the selector module without creating jquery elements.  
I end up importing `core/init` everywhere I import `selector`.  
I believe `selector` module should depend on `core/init` instead of just `core`.

### Checklist ###

### DIFF
--- a/src/selector-native.js
+++ b/src/selector-native.js
@@ -1,5 +1,5 @@
 define( [
-	"./core",
+	"./core/init",
 	"./var/document",
 	"./var/documentElement",
 	"./var/hasOwn",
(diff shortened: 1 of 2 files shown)


midnight-wonderer:
Wow, things are getting messy.

Upon more careful inspection, I realize I should not import `selector` at all.
Only `core/init` should be imported, because it depends on `traversing/findFilter` which depends on `selector`.  
You better add some comment there, or people would gonna import `selector` alone and not `core/init`.
People would never looks for `core/init` because they are looking for what they wanted to use.

PR #4316, opened 2019-03-09
Add comment regarding selector module and custom built

midnight-wonderer:
### Summary ###
The pull request adds a helpful comment, so people save their time by not importing `selector` module in their custom built.  
Related to: https://github.com/jquery/jquery/pull/4315

### Checklist ###

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -1,3 +1,8 @@
+// You should not import this module directly in your custom built.
+// This module has a circular dependency on "./core/init"
+// Consider importing "./core/init" instead.
+// It depends on "./traversing/findFilter" which depends on this module.
+
 define( [ "./selector-sizzle" ], function() {
 	"use strict";
 } );


dmethvin:
People shouldn't be browsing through the source to find modules to include or exclude in their custom build, so a comment here seems unnecessary. Is there something we can improve in the [documentation](https://github.com/jquery/jquery/blob/master/README.md#modules)?

midnight-wonderer:
Oops, I did not build it with grunt.  
I just realize webpack does not officially support.  

If that is the case, I am fine by just closing the PR, and supporting my way of building the module myself.  
However, this at least should serve as a ping to the core team, that the tools keep changing, and other people not use only officially supported build tools.

In my project, each part of jquery has been loaded on demand. It works quite well currently; the only downside is I needs to look into jquery source code from time to time.

dmethvin:
If you have suggestions on how to make the documentation for custom builds and Webpack better, let us know. We could add a note in the readme if that helps. Thanks!

Issue #4322, opened 2019-03-14
Bad performance with Chrome and sibling selector


cguglielmo:
Hi there

We experienced a weird performance issue with jQuery 3.3.1 which does not occur with jQuery 2.2.4. As far as we can tell, only Chrome is affected.

The problem is, that the following code forces the browser to recalculate its style.
`$elem.css('width', 'auto');  
 $elem.css('display', 'inline-block');`

If this is done in a loop let's say for 1000 elements, the browser will recalculate the style a 1000 times! But, it only seems to happen if there is a CSS rule using a sibling selector for the container element of the 1000 nodes. The CSS rule can even be empty, but the rule has to be in a separate style sheet. If the style is embedded in the HTML document, it works fine...

We could track it down to the following line:
`isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",`

Replacing the above line with the following line fixes the problem:
`isBorderBox = extra && jQuery.css( elem, "boxSizing", false, styles ) === "border-box",`

Reading the box-sizing attribute obviously forces the browser to recalculate the style. But honestly, we don't know what this patch could break. It looks like the code is only necessary for IE 9? At least according to issue #3699.

An example to reproduce the issue is attached.
[jQuery Perf.zip](https://github.com/jquery/jquery/files/2967402/jQuery.Perf.zip)

mgol:
Thanks for the report! We fixed a similar issue in #4187 but we missed this one.

We will most likely not be able to avoid layout thrashing in IE but we might be able to do that in other browsers.

@cguglielmo You can check whether your change breaks something by running `npm test` in the repository.

mgol:
PR: #4325

mgol:
@cguglielmo This has been fixed. You can double-check with our dev version, https://code.jquery.com/jquery-git.js (just don't use it in production!). I checked your test case & it no longer does layout thrashing.

PR #4325, opened 2019-03-18
CSS: Avoid forcing a reflow in width/height getters unless necessary

mgol:
### Summary ###

This PR delays reading element `box-sizing` until it's necessary - either because of broken `box-sizing` handling or because we need that information for `extra` processing.

Fixes gh-4322
Ref gh-3991
Ref gh-4010
Ref gh-4185
Ref gh-4187

### Checklist ###

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -118,9 +118,15 @@ function getWidthOrHeight( elem, dimension, extra ) {
 
 	// Start with computed style
 	var styles = getStyles( elem ),
-		val = curCSS( elem, dimension, styles ),
-		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
+
+		// To avoid forcing a reflow, only fetch boxSizing if we need it (gh-4322).
+		// Fake content-box until we know it's needed to know the true value.
+		boxSizingNeeded = !support.boxSizingReliable() || extra,
+		isBorderBox = boxSizingNeeded &&
+			jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
 		valueIsBorderBox = isBorderBox,
+
+		val = curCSS( elem, dimension, styles ),
 		offsetProp = "offset" + dimension[ 0 ].toUpperCase() + dimension.slice( 1 );
 
 	// Support: Firefox <=54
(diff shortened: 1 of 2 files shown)


mgol:
+16 bytes, I'm trying to reduce that.

src/css.js
```diff
@@ -141,10 +147,14 @@ function getWidthOrHeight( elem, dimension, extra ) {
 	// Also use offsetWidth/offsetHeight for when box sizing is unreliable
 	// We use getClientRects() to check for hidden/disconnected.
 	// In those cases, the computed value can be trusted to be border-box
-	if ( ( isBorderBox && !support.boxSizingReliable() || val === "auto" ||
+	if ( ( !support.boxSizingReliable() &&
+		jQuery.css( elem, "boxSizing", false, styles ) === "border-box" ||
```
gibson042: You may get a size reduction by turning this bare condition into an assignment, matching its other two appearances.
```suggestion
		(isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box") ||
```

src/css.js
```diff
@@ -118,9 +118,15 @@ function getWidthOrHeight( elem, dimension, extra ) {
 
 	// Start with computed style
 	var styles = getStyles( elem ),
-		val = curCSS( elem, dimension, styles ),
-		isBorderBox = jQuery.css( elem, "boxSizing", false, styles ) === "border-box",
+
+		// To avoid forcing a reflow, only fetch boxSizing if we need it (gh-4322).
+		// Fake content-box until we know it's needed to know the true value.
+		boxSizingNeeded = !support.boxSizingReliable() || extra,
```
gibson042: `boxSizingNeeded` appears to be a throwaway variable; is the code any smaller if you inline its expression into the `isBorderBox` assignment?

src/css.js
mgol: @gibson042 This `if` is an alternative of conditions and this is just the first of them. Doing the assignment there would mean the value is not set properly for non-IE browsers that fulfill one of the other checks.

src/css.js
mgol: Changing that doesn't influence the gzipped size.

A similar pattern is used in [the setter](https://github.com/jquery/jquery/blob/0ec25abba212efde462a8abcf3376f50116fd6c4/src/css.js#L371), inlining both doesn't change the size either.

src/css.js
gibson042: Right, and that's fine. This is just about compressibility.

mgol: Ah, I see, you meant adding the assignment without actually using it later. Let me check.

mgol: @gibson042 That's still `+16` bytes, exactly as my first version of the PR. With @timmywil's suggestion of just using `isBorderBox` here we get down to `+15` bytes. That reminds the best option then.

PR #4327, opened 2019-03-19
Ajax: Create correct URLs for jQuery.ajax() with cache: false

wenz:
Component: Ajax

Correctly creates the URL for calls to jQuery.ajax() when the "cache" option is set to false.
Based upon discussion on this pull request (it's been a while, so I restarted from scratch): https://github.com/jquery/jquery/pull/3902

Adds 43 characters to .min.js.gz, so I'll "soon-ish" prepare an alternate approach which does not cover all cases, but adds less code.

Fixes gh-3682.

### Summary ###
Correctly creates the URL for calls to jQuery.ajax() when the "cache" option is set to false. Fixes gh-3682.

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -19,7 +19,9 @@ define( [
 var
 	r20 = /%20/g,
 	rhash = /#.*$/,
-	rantiCache = /([?&])_=[^&]*/,
+	rantiCacheParam = /^([^?]*(?=\?)(.(.*(?=&)|)|).)_=\d*&?/,
+	rqueryParameterPrefix = /^([^?]*(?=\?)(.(.*(?=&).|)$|))/,
+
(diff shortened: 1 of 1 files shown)


gibson042:
I'm still of the opinion that a purely aesthetic change does not merit any increase in the size of jquery.min.js.gz, and 43 bytes is right out. But no matter what, a change like this would only be accepted with tests.

dmethvin:
Yeah, 43 bytes is a lot to make a prettier URL that already works so there's not really a bug. I seem to recall experimenting with this in the last go-round and not discovering any cheap way to do it. Something like `URLSearchParams` might be useful but it's not supported in IE11 so we'd be saddled with a fallback and that would make things worse.

Issue #4323, opened 2019-03-16
Preserve CSP nonce on scripts with src attribute in DOM manipulation.


buddh4:
When appending html content containing script tags with `src` attribute, I get a CSP `script-src` violation error. Tested with current master branch. 

Since `domManip` uses `jQuery_evalUrl` for script tags with `src` attribute it violates CSP's with a nonce script rule. I'am aware of https://github.com/jquery/jquery/issues/3969#issuecomment-380921687 but I can't use `crossDomain` since I require sync script loading. The following fix seems only to work for scripts without `src` attribute https://github.com/jquery/jquery/pull/4269/files.

**The problematic line**:

https://github.com/jquery/jquery/blob/5bdc85b82b84e5459462ddad9002f22d1ce74f21/src/manipulation.js#L202

### Hacky fix:

**manipulation/_evalUrl.js:**

Adding a node parameter to `_evalUrl` which will be passed to `jQuery.globalEval`.

```
jQuery._evalUrl = function( url, node ) {
    return jQuery.ajax( {
        url: url,

        // Make this explicit, since user can override this through ajaxSetup (#11264)
        type: "GET",
        dataType: "script",
        cache: true,
        async: false,
        global: false,

        // Only evaluate the response if it is successful (gh-4126)
        // dataFilter is not invoked for failure responses, so using it instead
        // of the default converter is kludgy but it works.
        converters: {
            "text script": function() {}
        },
        dataFilter: function( response ) {
            jQuery.globalEval( response, node );
        }
    } );
};
```

**/manipulation.js:159**

In my tests I noticed that the `nonce` attribute is not accessible anymore after the script nodes were added by `appendChild()` (if CSP is active). A dirty workaround was to backup the nonce attributes before inserting the dom like this:

```
for ( ; i < l; i++ ) {
    node = fragment;

    if ( i !== iNoClone ) {
        node = jQuery.clone( node, true, true );

        // Keep references to cloned scripts for later restoration
        if ( hasScripts ) {

            // Support: Android <=4.0 only, PhantomJS 1 only
            // push.apply(_, arraylike) throws on ancient WebKit
            jQuery.merge( scripts, getAll( node, "script" ) );
        }
    }

    /**
     * Backup nonce attributes, since they won't be accessible after dom insertion
     */
    if ( hasScripts ) {
        scripts.forEach( function( scr ) {
            scr.nonceValue = scr.getAttribute( "nonce" );
        } );
    }


[...]

PR #4328, opened 2019-03-19
Core: Preserve CSP nonce on scripts with src attribute (#4323)

buddh4:
### Summary ###
Fixes https://github.com/jquery/jquery/issues/4323

### Checklist ###

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -199,7 +199,9 @@ function domManip( collection, args, callback, ignored ) {
 
 							// Optional AJAX dependency, but won't run scripts if not present
 							if ( jQuery._evalUrl && !node.noModule ) {
-								jQuery._evalUrl( node.src );
+								jQuery._evalUrl( node.src, {
+									nonce: node.nonce || node.getAttribute( "nonce" )
+								} );
 							}
 						} else {
 							DOMEval( node.textContent.replace( rcleanScript, "" ), node, doc );
--- a/test/data/csp-nonce-external.js
+++ b/test/data/csp-nonce-external.js
@@ -1,0 +1,5 @@
+/* global startIframeTest */
+
+jQuery( function() {
+	$( "body" ).append( "<script nonce='jquery+hardcoded+nonce' src='csp-nonce.js'></script>" );
+} );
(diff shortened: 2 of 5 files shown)


src/manipulation.js
```diff
@@ -199,7 +199,8 @@ function domManip( collection, args, callback, ignored ) {
 
 							// Optional AJAX dependency, but won't run scripts if not present
 							if ( jQuery._evalUrl && !node.noModule ) {
-								jQuery._evalUrl( node.src );
+								var nonce = node.nonce || node.getAttribute( "nonce" );
```
mgol: We always declare variables at the top of its function scope.

test/data/csp-nonce-external.js
```diff
@@ -0,0 +1,5 @@
+/* global startIframeTest */
+
+jQuery( function() {
+	$( "body" ).append( "<script nonce='jquery+hardcoded+nonce' src='csp-nonce.js'><\x2Fscript>" );
```
mgol: The `<\x2Fscript>` hack is not necessary - it's meant to avoid HTML parsing quirks and this is an external script.

test/data/mock.php
```diff
@@ -213,7 +213,6 @@ protected function cspLog( $req ) {
 
 	protected function cspClean( $req ) {
 		file_put_contents( $this->cspFile, '' );
-		unlink( $this->cspFile );
```
mgol: What's the reason for this change? It's meant to clean up after the test so it should stay.

src/manipulation.js
```diff
@@ -199,7 +199,9 @@ function domManip( collection, args, callback, ignored ) {
 
 							// Optional AJAX dependency, but won't run scripts if not present
 							if ( jQuery._evalUrl && !node.noModule ) {
-								jQuery._evalUrl( node.src );
+								jQuery._evalUrl( node.src, {
+									nonce: node.nonce || node.getAttribute( "nonce" )
```
mgol: I hoped this would be compressable with the other similar logic in `DOMEval.js` but there we iterate over more properties so it cannot be generalized so easily.

We'll be able to simplify it when Firefox & Safari implement the `nonce` property.

mgol: cc @gibson042 for compressability efforts

Issue #4300, opened 2019-02-18
Stop shimming `focusin`/`focusout` in jQuery 4.0


mgol:
### Description ###

Firefox has added long-missing support for `focusin`/`focusout` events [in version 63](https://bugzilla.mozilla.org/show_bug.cgi?id=1472887). Other browsers we plan to support in v4 already support the events (apart from Firefox ESR 60 but we'll drop it before jQuery 4.0 is released).

Currently, [we shim `focusin`/`focusout` in all non-IE browsers](https://github.com/jquery/jquery/blob/3.3.1/src/event/focusin.js) both to add support in Firefox and because order of these events is not implemented according to the spec in various browsers. [It used to be wild out there](https://github.com/jquery/jquery/issues/3123) but for now the situation is as follows:

Edge 17+, Chrome, Firefox & Safari dispatch the events in the following order:

1. `blur`
1. `focusout`
1. `focus`
1. `focusin`

IE 11 follows [the spec](https://www.w3.org/TR/uievents/#events-focusevent-event-order) order, i.e.:

1. `focusout`
1. `focusin`
1. `blur`
1. `focus`

jQuery follows the order:

1. `focusout`
1. `blur`
1. `focusin`
1. `focus`

The fact that IE follows the spec order is a result of `blur` & `focus` events being asynchronous there. That's also why we don't shim `focusin`/`focusout` in IE via `focus`/`blur` as in other browsers as that'd make `focusin`/`focusout` asynchronous as well. This was considered in #3123.

As you can notice, the jQuery order doesn't match the spec either, it just fires `blur` after `focusout` & `focus` after `focusin`, the rest is different. Since all of the modern browsers settled on a specific non-standard events order, should we just stop shimming it in jQuery 4.0 and rely on native behavior in all browsers?

### Link to test case ###

https://jsfiddle.net/66znLhfw/2/

dmethvin:
I agree this would be great to eliminate as special cases.  Currently we allow event delegation on "focus" or "blur" events that do not bubble by using the bubbling "focusin" or "focusout". Ideally we'd just make people use "focusin" or "focusout" for those cases. 

Although that would be a breaking change it might be worthwhile to require because people would need to look at their code to ensure it's working properly. This should be possible to detect pretty well with Migrate because non-focusable elements (not natively focusable or no `tabIndex`) shouldn't have focus/blur attached to them. The fix in most cases would simply be to change to "focusin" and "focusout".

mgol:
[...]

PR #4329, opened 2019-03-22
Event: Cleanup leverageNative

gibson042:
One commit to restore the test, one to fix it, and the rest to reduce size.

### Summary ###
Ref 669f720edc4f557dfef986db747c09ebfaa16ef5
* Restore the no-duplication logic in focusin/focusout testing
* Reduce size

```
   raw     gz Sizes
279073  82883 dist/jquery.js
 88068  30594 dist/jquery.min.js

   raw     gz Compared to master @ 669f720edc4f557dfef986db747c09ebfaa16ef5
   +89   +209 dist/jquery.js
  -113    -28 dist/jquery.min.js
```

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -469,7 +469,7 @@ jQuery.event = {
 				// Claim the first handler
 				// dataPriv.set( el, "focus", ... )
 				leverageNative( el, "focus", function( el ) {
-					return el !== safeActiveElement();
+					return el === safeActiveElement();
 				} );
 
 				// Return false to allow normal processing in the caller
@@ -502,7 +502,7 @@ jQuery.event = {
 				// Claim the first handler
 				// dataPriv.set( el, "blur", ... )
 				leverageNative( el, "blur", function( el ) {
-					return el === safeActiveElement();
+					return el !== safeActiveElement();
 				} );
 
 				// Return false to allow normal processing in the caller
@@ -538,7 +538,7 @@ jQuery.event = {
 					dataPriv.get( el, "click" ) === undefined ) {
 
 					// dataPriv.set( el, "click", ... )
-					leverageNative( el, "click", returnFalse );
+					leverageNative( el, "click", returnTrue );
 				}
 
 				// Return false to allow normal processing in the caller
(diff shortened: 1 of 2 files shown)


src/event.js
```diff
@@ -30,7 +30,14 @@ function returnFalse() {
...
 
 // Support: IE <=9 only
-// See #13393 for more info
+// https://bugs.jquery.com/ticket/13393
+function expectSync( elem, type ) {
+
+	// Expect focus to be synchronous when the element is already active,
+	// and blur to be synchronous when the element is not already active
+	// (even in IE)
+	return ( elem === safeActiveElement() ) === ( type === "focus" );
```
mgol: You could perhaps add a support comment for IE <=11 here as other browsers fire those events synchronously so this helper wouldn't be needed?

It seems to me the `Support: IE <=9 only` part should be kept directly above `safeActiveElement` and `expectSync` might get a `Support: IE <=11 only` comment (or `Support: IE <=11+`, whatever we decide the convention to be).

What do you think?

src/event.js
```diff
@@ -574,55 +531,63 @@ jQuery.event = {
...
+						// Cancel the outer synthetic event
+						event.stopImmediatePropagation();
+						event.preventDefault();
+						return result;
+					}
+
+				// If this is an inner synthetic event for an event with a bubbling surrogate
+				// (focus or blur), assume that the surrogate already propagated from triggering the
+				// native event and prevent that from happening again here.
+				// This technically gets the ordering wrong w.r.t. to `.trigger()`, but that seems
```
mgol: Could you explain where the order is wrong considering the `focusin`/`focusout` ordering test passes?

src/event.js
gibson042: Good points; done.

src/event.js
gibson042: Updated the comment. It is subtle (and I may actually be wrong about it), but it will change with #4300 anyway.

mgol:
@timmywil Let's get it landed before `3.4.0`.

Issue #4299, opened 2019-02-18
jQuery: Drop IE<11 support in jQuery 4.0


mgol:
Microsoft has recently announced [dropping IE<11 support](https://techcommunity.microsoft.com/t5/Windows-IT-Pro-Blog/Bringing-Internet-Explorer-11-to-Windows-Server-2012-and-Windows/ba-p/325297) completely in January 2020, including on Windows Server & Embedded versions. IE 11 will be made available to those supported systems that only had access to Internet Explorer 10 until now.

It seems jQuery 4.0 might be where we'd like to drop IE<11 support. When evaluating that, remember we're not releasing 4.0 right now but most likely closer to 2020 so the situation will make it more & more realistic. And with our strategy to reduce browser support only in major releases, if we don't drop those versions we'll be stuck with them for a long time.

Dropping IE<11 support could help a lot with the big planned refactors, like a rewrite of the event system that gets us closer to native and dropping Sizzle in favor of a smaller querySelectorAll wrapper with selector rewriting to work around issues.

Market share of IE<11 seems very low even right now. [StatCounter data](http://gs.statcounter.com/browser-version-partially-combined-market-share/desktop/worldwide/#monthly-201901-201901-bar) shows global IE 10 usage at 0.15% and IE 9 at 0.36%. Even in countries with historically high IE usage like [China](http://gs.statcounter.com/browser-version-partially-combined-market-share/desktop/china/#monthly-201901-201901-bar) IE 10 & IE 9 already have small market share. In [South Korea](http://gs.statcounter.com/browser-version-partially-combined-market-share/desktop/south-korea/#monthly-201901-201901-bar) IE 11 has high usage at 20.1% but IE 10 - only 0.40% and IE 9 - only 0.18%.

[...]

Issue #3950, opened 2018-01-22
Support only last 3 iOS versions and drop Android Browser in jQuery 4.0+


mgol:
### Description ###
We currently support most browsers in their `current` & `current-1` versions. The exceptions are IE, iOS & Android. Wrt to IE it's obvious we can't switch to the "`current` & `current-1`" scheme as this browser is dead and we support it for legacy reasons. Android Browser 4.0-4.3 still has around 6.1% of Android Browser market share (91.8% of which are Android 4.1-4.3) so we can't drop Android Browser just yet.

iOS is different. It doesn't progress as quickly as desktop rolling-release browsers as Safari cannot be updated separately from the OS but people update the OS pretty quickly. Stats show hardly anyone uses an iOS version older than `current-2` at any given moment. [Apple stats](https://developer.apple.com/support/app-store/) provide some insight but they don't separate `current-2` from older versions.

Currently we explicitly support iOS 7+. This means as long as we don't release a new major jQuery version we have to test on more & more iOS versions (currently 4 of them). Could we limit ourselves to a specific number of iOS versions supported in line with most of the desktop browsers? I feel `current` & `current-1` might be too little but perhaps we can limit ourselves to last 3 stable versions?

This would have to happen in jQuery 4.0 & newer as we try to not limit https://jquery.com/browser-support/ in patch/minor releases.

### Link to test case ###
N/A

timmywil:
Decided on last 3 stable for iOS. Android 4.0 will go away.

mgol:
@thecodingdude We'll still have to support IE in the foreseeable future (at least IE 11; currently IE 9+ but it doesn't change *that* much) so we can't do that, unfortunately.

thecodingdude:
I mean, there's already president for removing browsers - your blog post mentions "If you need to support older browsers like Internet Explorer 6-8, Opera 12.1x or Safari 5.1+, use jQuery 1.12." so it would be no different here right? If you want IE 9-11 use jQuery 3.x

IE 11 literally supports no modern web platform features so how you plan on using next generation JS like modules and supporting IE11 is a mystery. Starting from a modern foundation (not IE) would be much easier to attract pull requests and such rather than working with such an old codebase.

Just my thoughts...

mgol:
> how you plan on using next generation JS like modules and supporting IE11 is a mystery

We'd use modules during development but we'd transpile using something like rollup so that the final bundle is one JS script that doesn't use modules.

[...]

Issue #4320, opened 2019-03-13
$.map() fails with large arrays in Сhrome


fantaclaus:
When a large array is passed as an argument to `$.map()` we get an error:

    RangeError: Maximum call stack size exceeded

Tested on Chrome **72** for Windows.
JQuery version: **3.3.1**

Sample code: 

```javascript
$.map(new Array(300000), i => 1)
```

Internally it calls to `concat.apply` and crashes there.

As long as I understand, `apply` tries to push all elements of the array into the stack before invoking the function `concat` and fails on stack shortage.

I believe that any function being invoked thru `apply` with a large array as an argument will fail. 

I've found about 14 usages of `apply` in the code of jQuery. I would suggest you to check out all of these cases.

dmethvin:
Applying `$.map` to 300,000 elements isn't a use case we've considered. Since we're a DOM library, the most common calls relate to arrays of DOM elements and it's not common to have that many DOM elements. Are there situations where the internal uses might be passed hundreds of thousands of elements?

In an earlier ticket you mentioned that this happened with a templating library. Instead of processing one big template perhaps you could process multiple templates in groups.

fantaclaus:
Well, in fact I've just got an error with 130,000 items...

Why do you think that it is uncommon? Someone could develop an intranet business application that work with large tables. A table easily could have about 6,500 rows with 10 columns, So you've got 65,000 cells. Each cell could contain two `<span>` tags, so you've got 130,000 tags...

Anyway, the documentation doesn't mention about any limitations. The function `$.map` is described like a common use function. The native `Array.map` doesn't crash on such arrays, so it is quite unexpected when the jQuery's one does.

> In an earlier ticket you mentioned that this happened with a templating library. Instead of processing one big template perhaps you could process multiple templates in groups.

It was not my ticket.

I'm concerned that the problem is general for jQuery since it uses `apply` method internally. So users could have troubles with some other functions, not only `$.map`. And they really have (according to some closed issues).

dmethvin:
Nobody should be rendering 6,500 rows on a page simultaneously. That is a good case for pagination or other segmentation techniques like a virtual scrolling window.

> So users could have troubles with some other functions, not only $.map. And they really have (according to some closed issues).

There are certainly occasional tickets where people hit these limits in jQuery, but as far as I can recall they have all been in this category of use cases that are unsupported and/or inadvisable.

> Anyway, the documentation doesn't mention about any limitations.

[...]

PR #4347, opened 2019-04-10
Core: Drop support for IE <11, iOS <11, Firefox <65, Android Browser & PhantomJS

mgol:
### Summary ###

Also, update support comments format to match format described in:
https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-69379197
with the change from:
https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-448998379
(open-ended ranges end with `+`).

Further size reductions will be achieved when we drop Firefox 60, iOS 10
and pre-Chromium Edge versions.

Fixes gh-3950
Fixes gh-4299

### Checklist ###


mgol:
Also, update support comments format to match format described in:
https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-69379197
with the change from:
https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-448998379
(open-ended ranges end with `+`).

mgol:
-540 bytes.

I recommend reviewing with whitespace changes ignored (e.g. via [adding `?w=1` to the URL](https://github.com/jquery/jquery/pull/4347/files?w=1)) to see real changes in `xhr.js` that got de-indented.

A few open questions:
1. I removed `support.cors` & `support.ajax` - are we OK with always assuming they're true? I'm not sure if IE 11 allows to disable native XHR as earlier IE did or if we still want to cater to those scenarios (we don't support the ActiveX versions anyway).
1. In a few places (e.g. `src/data/Data.js`) we don't delete stuff from DOM nodes, instead setting them to `undefined`. It's supposedly to help with DOM performance in Blink/WebKit; the ticket referenced there is not available to the public, though. Do we know if this is still true and if it maybe applies to Gecko as well or not? I'd tweak the comment then.
1. `wrapMap.js` references IE 9 directly in two places. Most of the file is covered by the comment the wrapping is needed when used in XHTML; is that not true for `option` as well? We should either move `option` handling to be under the same XHTML-related comment, remove it or update the support comment if it's still needed for IE 11.
1. `curCSS.js` references that `elem.style` is retrieved before the computed style as it fixes some issue in Firefox where it gets wrong values on detached elements. I couldn't find any relevant test or any more details about the supposed issue, could you help, @timmywil, as you committed the change?

src/ajax/xhr.js
```diff
@@ -15,156 +14,109 @@ jQuery.ajaxSettings.xhr = function() {
...
 		// File protocol always yields status code 0, assume 200
-		0: 200,
-
-		// Support: IE <=9 only
-		// #1450: sometimes IE returns 1223 when it should be 204
-		1223: 204
-	},
-	xhrSupported = jQuery.ajaxSettings.xhr();
-
-support.cors = !!xhrSupported && ( "withCredentials" in xhrSupported );
```
dmethvin: The scenarios where I've ever seen this used are `jQuery.support.cors = true`, that is, just setting the value. I wonder if there are a lot of cases where people check to see if it's `true` in their own code? If so it might be safest to set it to `true` here. Here's an example of some code that would break if we didn't: https://github.com/victorquinn/Backbone.CrossDomain/issues/1

mgol: I guess we could add a line with:
```
jQuery.support.cors = true;
```
to `deprecated.js` and add a warning to Migrate for code accessing it?

mgol:
BTW, dropping support for iOS 10 & Firefox 60 (which we'll most likely do before the 4.0 release as we'll drop them around September/October) will cause a further reduction by -272 bytes, getting us to -812 bytes compared to `master` and only `+14` bytes compared to latest `2.2.x`. Here are relevant changes on top of this PR:
https://github.com/mgol/jquery/commit/1f6450fe3cdeca43c3091631b53022cce7ff8cce

src/ajax/xhr.js
jaubourg: It's mainly useful of you want to gracefully degrade and/or fail in browsers without cors which will be none with the new support list. I think it would be better to keep it set to true for a good while in `deprecated.js` before removing it.

mgol: We could also have Migrate fill in those properties as it's easy to do from outside. Technically speaking, external access to `jQuery.support` has never been supported.

mgol:
I ran tests locally on this PR on most of the supported browsers:
Screen Shot 2019-04-10 at 16 48 18

mgol:
Per the discussion at the today's meeting, I changed the format of IE support comments from:
```js
// Support: IE <=9 - 11 only
```
to:
```js
// Support: IE <=9 - 11+
```

mgol:
I added one more commit removing IE 9-related code from `wrapMap.js` - I run some manual tests in XHTML mode and it didn't seem to cause any changes. The original wrapping of `<option>` in `<select>` was added in https://github.com/jquery/jquery/commit/14b09024cbe39bc16200448170ccbd183e3a9ec2 (12.5 years ago!), it was changed to be wrapped in `<select multiple>` in https://github.com/jquery/jquery/commit/264ffbc73be61778484a292392d1b18a8976588f. Removing that wrapping does make more tests fail in IE 9 but IE 11 still passes all the tests.

This removal saved additional `33` bytes. 🍾

mgol:
We're down to `-573` bytes now.

mgol:
@jquery/core I've been thinking that maybe we should just drop Firefox 60 & iOS 10 in this PR as well. We'll do that anyway in October and it's not very likely we'll release 4.0 before that date. Especially dropping iOS 10 would give us a lot - it has a partially broken module support, early versions had partially broken Shadow DOM support, etc. Dropping it would help not only with library size but also it'd be easier to do all the upcoming major changes like the event system refactor, getting rid of Sizzle etc. if we didn't have to care about iOS 10 and breaking any of current workarounds targeted at it.

If, by any chance, we manage to land all those major changes, release an alpha/beta & then stable before we'd normally drop Firefox 60/iOS 10 (not very likely, IMO), we can always update the browser support page for jQuery 4.0 to state explicitly we don't support those browsers. We could later remove that note once they'd be out of our supported browser range anyway.

I already have relevant changes ready at https://github.com/mgol/jquery/commit/724f9558fe41dcd129ea09ff1b562a0479d8436c, I could just merge it to this PR.

What do you think?

dmethvin:
I agree on the timeline, it seems likely that 4.0.0 could wait for those older browsers to drop below the threshold where we'd need to support them. It definitely cleans up some things!

mgol:
I added a commit dropping Firefox 60 & iOS 10. We can drop it if others disagree with me, @dmethvin & @jbedard but I think it makes sense to do it now.

We're down to `-916` bytes!

src/selector-native.js
```diff
@@ -35,11 +35,7 @@ define( [
 
 var hasDuplicate, sortInput,
 	sortStable = jQuery.expando.split( "" ).sort( sortOrder ).join( "" ) === jQuery.expando,
-	matches = documentElement.matches ||
-		documentElement.webkitMatchesSelector ||
-		documentElement.mozMatchesSelector ||
-		documentElement.oMatchesSelector ||
-		documentElement.msMatchesSelector,
+	matches = documentElement.matches || documentElement.msMatchesSelector,
```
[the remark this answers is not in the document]
mgol: Good point about a support comment but it won’t go away as IE needs it.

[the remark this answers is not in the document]
mgol: Comment added. I only mentioned IE as Edge doesn't need it since 4 versions already.

src/manipulation.js
```diff
@@ -172,9 +150,6 @@ function domManip( collection, args, callback, ignored ) {
 
 					// Keep references to cloned scripts for later restoration
 					if ( hasScripts ) {
-
-						// Support: Android <=4.0 only, PhantomJS 1 only
-						// push.apply(_, arraylike) throws on ancient WebKit
 						jQuery.merge( scripts, getAll( node, "script" ) );
```
gibson042: @mgol I was just going back over this PR and wondering about the loss of context here and before other `jQuery.merge` calls... _should_ we switch this to `push.apply( scripts, getAll( node, "script" ) )` or have a comment explaining why it avoids that pattern?

src/manipulation.js
```diff
@@ -475,9 +456,6 @@ jQuery.each( {
 		for ( ; i <= last; i++ ) {
 			elems = i === last ? this : this.clone( true );
 			jQuery( insert[ i ] )[ original ]( elems );
-
-			// Support: Android <=4.0 only, PhantomJS 1 only
-			// .get() because push.apply(_, arraylike) throws on ancient WebKit
 			push.apply( ret, elems.get() );
```
gibson042: Here too; why isn't this `push.apply( ret, elems )`?

src/manipulation.js
mgol: Right... We no longer care about Android 4.0 or PhantomJS 1 but I remember using the `push.apply` approach may lead to issues like #4320. Perhaps a comment to that effect would be useful.

mgol: I created #5298 to track this & other related uses of `push.apply` vs. `jQuery.merge` to consider.

PR #4352, opened 2019-04-15
add .idea to .gitignore

Daniel217D:
### Remove auto generated .idea folder from project###
I think, it can be useful, because many people use development environments such as PhpStorm


timmywil:
Thanks for contributing, but `.idea` is one of many things that we could add to the project `.gitignore`. But the best place for this is your global gitignore.

Issue #4350, opened 2019-04-11
saved.shift is not a function


dsandber:
My app auto-upgraded to 3.4.0 from 3.3.1 and broke with this error:

saved.shift is not a function

I'll append the browser version in a comment shortly.

dsandber:
user_agent: "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/73.0.3683.86 Safari/537.36"

trace: "TypeError: saved.shift is not a function
    at HTMLDivElement.handler (https://inbox-credit.vertical.arbcalls.com/js/vue/app.aaa76e4c45a8eb6f83bf.js:10492:27)
    at HTMLDivElement.dispatch (https://inbox-credit.vertical.arbcalls.com/js/vue/app.aaa76e4c45a8eb6f83bf.js:10244:27)
    at HTMLDivElement.elemData.handle (https://inbox-credit.vertical.arbcalls.com/js/vue/app.aaa76e4c45a8eb6f83bf.js:10051:28)

timmywil:
Sorry, there is not enough information to go on here. You can find help on https://stackoverflow.com. If after sorting out the issue you find there is a change needed in jQuery core, please open a new issue and follow the instructions in the issue template, including providing a test case from https://jsbin.com.

dsandber:
Sorry, because it wasn't triggered directly by my code, and because it doesn't seem to happen consistently, I don't know how to create a test case.

timmywil:
Fair enough, but the issue needs to be fleshed out more regardless.

dsandber:
Luckily "saved.shift" only occurs once in the source-code.  Strange that the comment says IE <= 9 - 11+ but that it is being run on Chrome.


```
dataPriv.set( this, type, jQuery.event.trigger(
  // Support: IE <=9 - 11+
  // Extend with the prototype to reset the above stopImmediatePropagation()
  jQuery.extend( saved.shift(), jQuery.Event.prototype ),
  saved,
  this
) );
```

mgol:
[...]

PR #4354, opened 2019-04-15
Event: Fix handling of multiple async focus events

gibson042:
### Summary ###
Prevent leftover saved data from async focus/blur from overriding outermost trigger handling.

Fixes gh-4350

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -556,7 +556,7 @@ function leverageNative( el, type, expectSync ) {
 				// Interrupt processing of the outer synthetic .trigger()ed event
 				// Saved data should be false in such cases, but might be a leftover capture object
 				// from an async native handler (gh-4350)
-				if ( !( saved.length >= 0 ) ) {
+				if ( !saved.length ) {
 
 					// Store arguments for use when handling the inner native event
 					saved = slice.call( arguments );
@@ -593,15 +593,15 @@ function leverageNative( el, type, expectSync ) {
 
 			// If this is a native event triggered above, everything is now in order
 			// Fire an inner synthetic event with the original arguments
-			} else if ( saved.length >= 0 ) {
+			} else if ( saved.length ) {
 
 				// ...and capture the result
 				dataPriv.set( this, type, { value: jQuery.event.trigger(
 
 					// Support: IE <=9 - 11+
 					// Extend with the prototype to reset the above stopImmediatePropagation()
-					jQuery.extend( saved.shift(), jQuery.Event.prototype ),
-					saved,
+					jQuery.extend( saved[ 0 ], jQuery.Event.prototype ),
+					saved.slice( 1 ),
 					this
 				) } );
 
--- a/src/event.js
+++ b/src/event.js
@@ -559,6 +559,8 @@ function leverageNative( el, type, expectSync ) {
 				if ( !saved.length ) {
 
 					// Store arguments for use when handling the inner native event
+					// There will always be at least one argument (an event object), so this array
+					// will not be confused with a leftover capture object.
 					saved = slice.call( arguments );
 					dataPriv.set( this, type, saved );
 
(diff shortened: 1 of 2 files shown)


src/event.js
```diff
@@ -591,17 +593,17 @@ function leverageNative( el, type, expectSync ) {
 
 			// If this is a native event triggered above, everything is now in order
 			// Fire an inner synthetic event with the original arguments
-			} else if ( saved ) {
+			} else if ( saved.length ) {
 
 				// ...and capture the result
-				dataPriv.set( this, type, jQuery.event.trigger(
+				dataPriv.set( this, type, { value: jQuery.event.trigger(
```
mgol: This would IMO look more readable with `value` moved to the next line. I don't recognize this way of writing objects from our code base.

mgol: Is the change to an object with the `value` field meant to distinguish the saved object from `arguments` saved before? Do I understand it right?

src/event.js
```diff
@@ -554,7 +554,9 @@ function leverageNative( el, type, expectSync ) {
 			if ( ( event.isTrigger & 1 ) && this[ type ] ) {
 
 				// Interrupt processing of the outer synthetic .trigger()ed event
-				if ( !saved ) {
+				// Saved data should be false in such cases, but might be a leftover capture object
+				// from an async native handler (gh-4350)
+				if ( !saved.length ) {
```
mgol: This `if` branch can be executed if `length` is falsy which might suggest it may happen for the `arguments` saved below... but there will always be at least one argument passed as the event argument is always passed.

It's a little fragile but we have a test for that so I guess we're fine.

test/unit/event.js
```diff
@@ -3041,6 +3041,49 @@ QUnit.test( "focus-blur order (#12868)", function( assert ) {
...
+			setTimeout( function() {
+				input.trigger( "focus" );
+			} );
+		} );
+
+	// gain focus
+	input.trigger( "focus" );
+
+	// DOM focus is unreliable in TestSwarm
+	setTimeout( function() {
```
mgol: Shouldn't the `setTimeout` be given some positive timeout value so that this fallback is not executed too fast?

Also, do you think there might be a way to avoid this fallback?

src/event.js
gibson042: Agreed.

gibson042: Yes, exactly.

src/event.js
gibson042: Added source comments to explain.

test/unit/event.js
gibson042: I don't see how to avoid this fallback, but a specific timeout is not necessary because queueing itself ensures that the first focus handler will run before this callback if it is to be invoked at all.

mgol: Makes sense.

BTW, if this is unreliable only in TestSwarm the you could guard the fallback with an additional `QUnit.isSwarm`or whatever it’s named, then it’d be guaranteed it’s not skipped in our Karma tests in Chrome headless.

gibson042: Good idea; done.

mgol:
We'll want to cherry-pick that to master as well.

src/event.js
```diff
@@ -554,9 +554,13 @@ function leverageNative( el, type, expectSync ) {
...
```
stof: wouldn't this break in case `saved` is `false` (set in `dataPriv.set( this, type, false );` on line 570) ?

mgol: @stof No, it works:
```js
var b = false;
!b.length; // true
```
This is because for a boolean `b` `b.length` will aut-box `b` into a `Boolean` instance which doesn't have the `length` field so `b.length` will be `undefined`. A negation of `undefined` is `true`.

I know it's not the most readable code but we care about size and this behavior is covered by tests.

PR #4362, opened 2019-04-17
Event: Stop shimming focusin & focusout events

mgol:
### Summary ###

Latest versions of all browsers now implement focusin & focusout natively
and they all converged on a common event order so it doesn't make much sense
for us to normalize it to a different order anymore.

Note that it means we no longer guarantee that focusin fires before focus
and focusout before blur.

`-123` bytes

Fixes gh-4300

### Checklist ###

* ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~ (I think we never documented event order guarantees)


mgol:
Note: this cannot go in as long as we test on Firefox 60 which doesn't support `focusin` & `focusout`. If we decide to drop this browser as part of #4347 (which I recommend) then this can land.

mgol:
-123 bytes, BTW.

mgol:
I mixed up the Firefox version, this has been implemented since Firefox 52 so we're good to go.

mgol:
@dmethvin Nice catch! PR updated.

mgol:
I also added a new test checking if `focusin` & `focusout` work at all. I think we didn't have such test apart from the one checking the order so we'd actually lose some coverage by removing that test.

mgol:
I also added a new test checking if `focusin` & `focusout` work at all. I think we didn't have such test apart from the one checking the order so we'd actually lose some coverage by removing that test without a replacement.

Issue #4056, opened 2018-04-24
Remove deprecated methods in 4.0.0


mgol:
### Description ###
We've deprecated some methods in 3.x (like `jQuery.camelCase`), we should remember to remove them all in 4.0. Looking at `deprecated.js` should be a good start; other APIs to remove:
- [ ] `jQuery.cssProps`.
- [x] `jQuery.cssNumber` (removed in #4055).

ShashankaNataraj:
@mgol I want to submit a PR for this.. Is this issue still relevant?

mgol:
@ShashankaNataraj It is, though we're not working on jQuery 4.0 yet. You can submit a PR, though. You'd need to look at each of the deprecated APIs; some are in `deprecated.js`, some are only deprecated in docs: [3.0](https://api.jquery.com/category/deprecated/deprecated-3.0/) & [3.3](https://api.jquery.com/category/deprecated/deprecated-3.3/) and some may only have an API issue created but the deprecations are not documented yet: https://github.com/jquery/api.jquery.com/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+deprecate.

For ones that have API docs tickets, we should first document the deprecation before we remove it on `master`; we're a little understaffed when it comes to content sites (like api.jquery.com) so especially there help would be appreciated. Thanks!

mgol:
PR: #4364

PR #4364, opened 2019-04-17
Core: Remove deprecated jQuery APIs

mgol:
### Summary ###

Fixes gh-4056

### Checklist ###


test/unit/css.js
```diff
@@ -1753,9 +1729,6 @@ QUnit.test( "Do not throw on frame elements from css method (#15098)", function(
 
 		assert.expect( !!appearanceName + !!transformName + 1 );
 
-		resetCssPropsFor( "appearance" );
-		resetCssPropsFor( "transform" );
```
mgol: This does make the test a little less isolated but it's already been that way since `finalPropName` has been refactored by @dmethvin to use an internal `vendorProps` map so we're not losing much here.

Another test can influence this one if it sets some transform- or appearance-related props. We can think if there's any way around that but, as I said, it's already an issue.

mgol:
Note to self: separate `jQuery.holdReady` removal to a separate PR, remove `readyWait` there as well.

Issue #4250, opened 2018-12-03
jQuery.ajax: make `dataType: "script"` not execute scripts on unsuccessful HTTP status


mgol:
### Description ###

A followup of PR #4243 and an extension of issue #4126. We should stop evaluating scripts returned with an unsuccessful HTTP status when loaded via `jQuery.ajax` with `dataType: "script"`. The fix in #4243 only targets DOM manipulation.

skrobinson:
I have a possible fix and a test case in commit 6716b75.

mgol:
@skrobinson Your patch skips converters completely for unsuccessful responses; this is not right, it doesn't pass our test suite.

The proper fix will most likely require generalizing what was done in #4243 so that the logic is applied to all requests with `dataType: "script"`.

skrobinson:
@mgol  Thank you for your feedback.  I've re-read #4243 and #4258 and tried again using the pattern in #4258.  I've also improved my testing and have a new attempt at a fix.

mgol:
@skrobinson Let us know if you need any help/direction! (or if you don't have time to submit a PR so that we know we should look at it ourselves)

mgol:
Landed, thanks @skrobinson for keeping up with our review comments!

PR #4379, opened 2019-04-30
Do not execute scripts for unsuccessful http responses #4250

skrobinson:
### Summary ###
Fix for #4250.

### Checklist ###

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,7 @@ matrix:
       env: NPM_SCRIPT=test:browser
 script:
   - npm run $NPM_SCRIPT
+branches:
+  only:
+    - master
+    - /wip-.*/
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,3 @@ matrix:
       env: NPM_SCRIPT=test:browser
 script:
   - npm run $NPM_SCRIPT
-branches:
-  only:
-    - master
-    - /wip-.*/
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -218,7 +218,9 @@ QUnit.assert.ok( true, "mock executed");';
 
 	protected function errorWithScript( $req ) {
 		header( 'HTTP/1.0 404 Not Found' );
-		header( 'Content-Type: application/javascript' );
+        if ( isset( $req->query['scriptReply'] ) ) {
+            header( 'Content-Type: application/javascript' );
+        }
 		if ( isset( $req->query['callback'] ) ) {
 			$callback = $req->query['callback'];
 			echo $callback . '( {"status": 404, "msg": "Not Found"} )';
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -216,6 +216,17 @@ QUnit.assert.ok( true, "mock executed");';
 		unlink( $this->cspFile );
 	}
 
+	protected function errorWithScript( $req ) {
+		header( 'HTTP/1.0 404 Not Found' );
+		header( 'Content-Type: application/javascript' );
+		if ( isset( $req->query['callback'] ) ) {
+			$callback = $req->query['callback'];
+			echo $callback . '( {"status": 404, "msg": "Not Found"} )';
+		} else {
+			echo 'QUnit.assert.ok( false, "Mock return erroneously executed" );';
+		}
+	}
+
 	public function __construct() {
 		$this->cspFile = __DIR__ . '/support/csp.log';
 	}
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -218,7 +218,7 @@ QUnit.assert.ok( true, "mock executed");';
 
 	protected function errorWithScript( $req ) {
 		header( 'HTTP/1.0 404 Not Found' );
-		if ( isset( $req->query['scriptReply'] ) ) {
+		if ( isset( $req->query['withScriptContentType'] ) ) {
 			header( 'Content-Type: application/javascript' );
 		}
 		if ( isset( $req->query['callback'] ) ) {
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -218,9 +218,9 @@ QUnit.assert.ok( true, "mock executed");';
 
 	protected function errorWithScript( $req ) {
 		header( 'HTTP/1.0 404 Not Found' );
-        if ( isset( $req->query['scriptReply'] ) ) {
-            header( 'Content-Type: application/javascript' );
-        }
+		if ( isset( $req->query['scriptReply'] ) ) {
+			header( 'Content-Type: application/javascript' );
+		}
 		if ( isset( $req->query['callback'] ) ) {
 			$callback = $req->query['callback'];
 			echo $callback . '( {"status": 404, "msg": "Not Found"} )';
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -831,7 +831,7 @@ QUnit.module( "ajax", {
 				dataType: "script",
 				url: url( "missing_script_file.js" )
 			} );
-			assert.ok( true, "no error thrown from embedding script with unsuccessful-response src" );
+			assert.ok( true, "no error thrown for missing script" );
 			done();
 		} catch ( e ) {
 			assert.ok( false, "uncaught error" );
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -806,39 +806,90 @@ QUnit.module( "ajax", {
 		};
 	} );
 
-	ajaxTest( "jQuery.ajax() - do not execute scripts from unsuccessful responses (gh-4250)", 2, function( assert ) {
-		return {
-			dataType: "script",
-			url: url( "missing_script_file.js" ),
-			fail: function( _, status, error ) {
-				assert.strictEqual( status, "error", "Check missing script file status" );
-				assert.ok( error.length > 0, "Check for error message" );
-			}
-		};
-	} );
-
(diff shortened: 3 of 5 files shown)


timmywil:
I'd approve without the `wip` commit. Travis will run on PRs.

mgol:
Don't we want converters to be applied for unsuccessful non-script responses?

.travis.yml
```diff
@@ -15,3 +15,7 @@ matrix:
       env: NPM_SCRIPT=test:browser
 script:
   - npm run $NPM_SCRIPT
+branches:
```
mgol: Please revert changes in this file.

timmywil:
Good point. This should be limited to scripts.

dmethvin:
I know of APIs that return non-200 status with JSON.  The only case we want to prevent here is script.

mgol:
I did some attempts in https://github.com/mgol/jquery/commit/0cf386936ad7a6735c283d9e3ea5e690c86127a4 some time ago but I struggled with how to actually detect scripts. This is because a script may have `dataType === "script"` but it may also have none when it was determined to be a script via `content-type` matching.

skrobinson:
@dmethvin Good point.  Should we consider APIs which intentionally return non-200 status scripts?

Is it correct to replace anything other than the default converter?  My current patch would also replace any custom "text script" converter.

mgol:
@skrobinson Returning data on non-200 responses definitely makes sense. Consider a JSON API that sends a 400 with JSON data describing an error in body; that's quite common. Scripts are special in that its response is not really just "converted" but executed as a script, that's why we want to avoid it in this specific example. I think that answers your question:

> Is it correct to replace anything other than the default converter?

with: no, that's not correct. We need to stop applying converters for scripts only.

skrobinson:
> > Is it correct to replace anything other than the default converter?
> with: no, that's not correct. We need to stop applying converters for scripts only.

Yes, I'm convinced about data.  But, should we be careful not to disrupt a non-default "text script" converter?  If the dev-user has defined a "text script" converter with ajaxSetup, should we still replace that converter with a noop?

mgol:
@skrobinson A good question. I think I'm fine with ignoring user-provided script converters for failed HTTP status codes.

skrobinson:
@mgol I'm still learning the GH way and should have brought commit 3e3fc0d to your attention.  It think this incorporates all current feeback.  Please let me know about other edits to be done.

test/unit/ajax.js
```diff
@@ -837,6 +837,41 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - do not execute scripts from unsuccessful responses (gh-4250)", 2, function( assert ) {
+		return {
+			dataType: "script",
+			url: url( "missing_script_file.js" ),
```
mgol: We use `url( "404.txt" )` in a few other places to trigger a 404 so that'd be preferred but here we need more - we need the response to return an executable script & verify it's not executed. Something I did in my attempts that I linked to before: https://github.com/mgol/jquery/commit/0cf386936ad7a6735c283d9e3ea5e690c86127a4.

skrobinson:
We've just started a post-semester remodeling at work and I'll be
unavailable for up to two weeks.  I will return to this PR, but likely
not until end of May.

mgol:
@skrobinson No problem, we're not in a hurry. Let us know when you're ready to continue, I'll be happy to keep reviewing new changes (just ping me directly when you update the PR, please :)).

gibson042:
I think this means that `jQuery._evalUrl` can also be simplified.

mgol:
> I think this means that `jQuery._evalUrl` can also be simplified.

@gibson042 I was thinking about that but it now also supports the options object that's used to pass `nonce` in `domManip`...

But, I guess, we could move the logic back to the converter now since it won't be executed for bad requests. Is that what you had in mind?

test/unit/ajax.js
```diff
@@ -837,6 +837,41 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - do not execute scripts from unsuccessful responses (gh-4250)", 2, function( assert ) {
```
mgol: One more thing that I forgot - we could use a test that sets a custom `text script` converter and makes sure it _does not_ get invoked for 404 scripts.

gibson042:
> But, I guess, we could move the logic back to the converter now since it won't be executed for bad requests. Is that what you had in mind?

Yeah, it should no longer need the `dataFilter` kludge AFAICT.

skrobinson:
@mgol I have pushed updates to my PR branch.

You made a good call asking for more tests.  I found my first `if` was not sufficient.

P.S. I'll look at `jQuery._evalUrl` after this PR lands.

test/unit/ajax.js
```diff
@@ -837,6 +837,92 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - do not execute scripts from unsuccessful responses (gh-4250)", 7, function( assert ) {
+		var globalEval = jQuery.globalEval;
+
+		function request( title, options, overrides ) {
```
mgol: Why do we need both `options` & `overrides`? Can't we drop `options` & rename `overrides` to `options`? Since `beforeSend` & a few others are overriding what's in `options`, there doesn't seem to be a reason to put those things in `options`. But if we don't put them there but just in `overrides` then changing:
```js
jQuery.extend( options, {...}, overrides )
```
to
```js
jQuery.extend( {...}, options, overrides )
```
wouldn't change anything and then we can just merge `options` with `overrides` instead of keeping two objects.

test/unit/ajax.js
```diff
@@ -837,6 +837,92 @@ QUnit.module( "ajax", {
...
+				"script reply",
+				{
+					url: url( "mock.php?action=errorWithScript" )
+				}
+			),
+			request(
+				"script reply with dataType",
+				{
+					dataType: "script",
+					url: url( "mock.php?action=errorWithScript" )
```
mgol: The test with an explicit `dataType: "script"` should request a URL that doesn't return a JavaScript content type to make sure it's still ignored then.

test/unit/ajax.js
```diff
@@ -837,6 +837,92 @@ QUnit.module( "ajax", {
...
+				}
+			),
+			request(
+				"script reply with converter",
+				{
+					converters: { "text script": function() {
```
mgol: Please put `"text script"` in a new line, it reads better. The same in a few places below.

test/unit/ajax.js
```diff
@@ -837,6 +837,92 @@ QUnit.module( "ajax", {
...
+				}
+			),
+			request(
+				"script reply with converter and dataType",
+				{
+					converters: { "text script": function() {
+						assert.ok( false, "No converter for unsuccessful response" );
+					} },
+					dataType: "script",
+					url: url( "mock.php?action=errorWithScript" )
```
mgol: A similar remark here; this should return JavaScript content without the `Content-Type` header.

test/unit/ajax.js
```diff
@@ -837,6 +837,92 @@ QUnit.module( "ajax", {
...
+				}
+			),
+			request(
+				"JSONP reply with dataType",
+				{
+					dataType: "jsonp",
+					url: url( "mock.php?action=errorWithScript" )
```
mgol: Again: this shouldn't send `Content-Type`.

test/unit/ajax.js
```diff
@@ -837,6 +837,92 @@ QUnit.module( "ajax", {
...
+					assert.ok( false, "Unanticipated success" );
+				}
+			}, overrides );
+		}
+
+		return [
+			request(
+				"HTML reply",
+				{
+					url: url( "404.js" )
```
mgol: Please use `404.txt` as other files do as we handle it separately in the Node.js server (used for Karma tests); here & below.

test/unit/ajax.js
skrobinson: @mgol Just for clarification... I should edit these tests to return a non-JS content type ?  Or I should add tests so that both JS and non-JS content types are returned by the server?

mgol: @skrobinson As far as I remember, I was thinking about replacing this test but I guess having both variants would be good.

mgol:
@skrobinson Look at Travis status, the build is red.

mgol:
Cool! Please post a comment when the PR is updated. Meanwhile, to clean up the comment thread, I'll mark our comments as off-topic.

skrobinson:
@mgol  This PR is ready for another review.

RE: the many test cases.  I copied the pattern used for the "jQuery.ajax() - cache" tests, once I understood how those tests worked.

test/data/mock.php
```diff
@@ -216,6 +216,19 @@ protected function cspClean( $req ) {
 		unlink( $this->cspFile );
 	}
 
+	protected function errorWithScript( $req ) {
+		header( 'HTTP/1.0 404 Not Found' );
+        if ( isset( $req->query['scriptReply'] ) ) {
```
mgol: I'd change the name from `scriptReply` to `withScriptContentType`; otherwise it looks a bit too similar to the `errorWithScript` name.

mgol:
@skrobinson before applying further updates, could you rebase this PR over current `master`? The current base is from April this year & a lot has changed in the setup since then.

skrobinson:
@mgol  These latest changes are complete and I've rebased onto today's master.

mgol:
Landed, thanks @skrobinson for keeping up with our review comments!

PR #4381, opened 2019-04-30
Build: ESLint: forbid unused function parameters

mgol:
### Summary ###

This commit requires all function parameters to be used, not just the last one.
In cases where that's not possible as we need to match an external API, there's
an escape hatch of prefixing an unused argument with `_`.

This change makes it easier to catch unused AMD dependencies and unused
parameters in internal functions the API of which we may change at will, among
other things.

Unused AMD dependencies have been removed as part of this commit.

What do you think? If you agree this change is worthwhile, should we add that rule to `eslint-config-jquery`?

### Checklist ###

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -372,7 +372,7 @@ module.exports = function( grunt ) {
 
 		// Ask for permission the first time
 		if ( insight.optOut === undefined ) {
-			insight.askPermission( null, function( error, result ) {
+			insight.askPermission( null, function( _error, result ) {
 				exec( result );
 			} );
 		} else {
--- a/dist/.eslintrc.json
+++ b/dist/.eslintrc.json
@@ -7,9 +7,10 @@
 		// That is okay for the built version
 		"no-multiple-empty-lines": "off",
 
-		// Because sizzle is not compatible to jquery code style
+		// Sizzle is not compatible with jQuery code style
 		"no-nested-ternary": "off",
 		"no-unused-expressions": "off",
+		"no-unused-vars": "off",
 		"lines-around-comment": "off",
 		"space-in-parens": "off",
 		"camelcase": "off",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "@babel/plugin-transform-for-of": "7.4.3",
     "commitplease": "3.2.0",
     "core-js-bundle": "3.0.0",
-    "eslint-config-jquery": "1.0.1",
+    "eslint-config-jquery": "2.0.0",
     "grunt": "1.0.4",
     "grunt-babel": "8.0.0",
     "grunt-cli": "1.3.2",
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -838,7 +838,7 @@ jQuery.extend( {
 	}
 } );
 
-jQuery.each( [ "get", "post" ], function( i, method ) {
+jQuery.each( [ "get", "post" ], function( _i, method ) {
 	jQuery[ method ] = function( url, data, callback, type ) {
 
 		// Shift arguments if data argument was omitted
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -119,7 +119,7 @@ boolHook = {
 	}
 };
 
-jQuery.each( jQuery.expr.match.bool.source.match( /\w+/g ), function( i, name ) {
+jQuery.each( jQuery.expr.match.bool.source.match( /\w+/g ), function( _i, name ) {
 	var getter = attrHandle[ name ] || jQuery.find.attr;
 
 	attrHandle[ name ] = function( elem, name, isXML ) {
--- a/src/core.js
+++ b/src/core.js
@@ -4,7 +4,6 @@
 
 define( [
 	"./var/arr",
-	"./var/document",
 	"./var/getProto",
 	"./var/slice",
 	"./var/concat",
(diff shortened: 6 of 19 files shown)


mgol:
@Krinkle AMD is the biggest use case but we're likely to get rid of it soon anyway in favor of standard ES modules. It's also useful for lots of internal functions, though. As @timmywil said, it doesn't help with functions with signatures bound by external contracts but we have lots of internal functions and it's easy to miss a specific parameter is not required anymore and could be removed with size gains. It forces you to think each time you need to prefix a parameter whether you can remove it or not. Therefore, I see gains beyond AMD.

In this PR I only had to use the `_` workaround in 15 places in the entire jQuery source, 8 of which are `jQuery.each` & 1 `jQuery.map`, leaving only 6 "regular" cases.

If you think this rule is not a good fit for QUnit, I could leave it as an override here in jQuery repo, although we're not using any overrides at the moment - but I see [QUnit is](https://github.com/qunitjs/qunit/blob/master/.eslintrc.json) and quite a lot at that.

mgol:
I wrote a comment and it disappeared, argh...

jQuery's AMD is the biggest beneficent of this new rule but we'll most likely migrate to ES modules so it will stop being so in the future. I see other gains, though - while, as @timmywil noted, functions bound by external contracts need the `_` prefixes, there are lots of internal ones and it's sometimes easy to miss a particular parameter may no longer be used & can be removed.

In my PR I only had to add the `_` prefix in 15 places, 8 of which are `jQuery.each` & 1 `jQuery.map`, leaving 6 "regular" cases.

If you think this rule is not a good fit for QUnit, I may just leave an override here but we haven't been using overrides so far. I see [QUnit is](https://github.com/qunitjs/qunit/blob/master/.eslintrc.json), though, and quite a lot at that; QUnit also uses the `eslint:recommended` config which we don't. I wonder how big a gain for QUnit is that it uses the `jquery` config; in the past we thought of it as a preset common for all jQuery Foundation projects but that hasn't materialized. At the moment, I think, it's mostly a few jQuery projects to which similar people contribute: Core, Migrate, Color, and QUnit.

mgol:
I wrote a comment and it disappeared, argh...

jQuery's AMD is the biggest beneficent of this new rule but we'll most likely migrate to ES modules so it will stop being so in the future. I see other gains, though - while, as @timmywil noted, functions bound by external contracts need the `_` prefixes, there are lots of internal ones and it's sometimes easy to miss a particular parameter may no longer be used & can be removed.

In my PR I only had to add the `_` prefix in 15 places, 8 of which are `jQuery.each` & 1 `jQuery.map`, leaving 6 "regular" cases.

If you think this rule is not a good fit for QUnit, I may just leave an override here but we haven't been using overrides so far. I see [QUnit is](https://github.com/qunitjs/qunit/blob/master/.eslintrc.json), though, and quite a lot at that; QUnit also uses the `eslint:recommended` config which we don't. I wonder how big a gain for QUnit is that it uses the `jquery` config; in the past we thought of it as a preset common for all jQuery Foundation projects but that hasn't materialized. At the moment, I think, it's mostly a few jQuery projects to which similar people contribute: Core, Migrate, Color, and QUnit.

mgol:
I wrote a comment and it disappeared, argh...

jQuery's AMD is the biggest beneficent of this new rule but we'll most likely migrate to ES modules so it will stop being so in the future. I see other gains, though - while, as @timmywil noted, functions bound by external contracts need the `_` prefixes, there are lots of internal ones and it's sometimes easy to miss a particular parameter may no longer be used & can be removed.

In my PR I only had to add the `_` prefix in 15 places, 8 of which are `jQuery.each` & 1 `jQuery.map`, leaving 6 "regular" cases.

If you think this rule is not a good fit for QUnit, I may just leave an override here but we haven't been using overrides so far. I see [QUnit is](https://github.com/qunitjs/qunit/blob/master/.eslintrc.json), though, and quite a lot at that; QUnit also uses the `eslint:recommended` config which we don't. I wonder how big a gain for QUnit is that it uses the `jquery` config; in the past we thought of it as a preset common for all jQuery Foundation projects but that hasn't materialized. At the moment, I think, it's mostly a few jQuery projects to which similar people contribute: Core, Migrate, Color, and QUnit.

mgol:
I wrote a comment and it disappeared, argh...

jQuery's AMD is the biggest beneficent of this new rule but we'll most likely migrate to ES modules so it will stop being so in the future. I see other gains, though - while, as @timmywil noted, functions bound by external contracts need the `_` prefixes, there are lots of internal ones and it's sometimes easy to miss a particular parameter may no longer be used & can be removed.

In my PR I only had to add the `_` prefix in 15 places, 8 of which are `jQuery.each` & 1 `jQuery.map`, leaving 6 "regular" cases.

If you think this rule is not a good fit for QUnit, I may just leave an override here but we haven't been using overrides so far. I see [QUnit is](https://github.com/qunitjs/qunit/blob/master/.eslintrc.json), though, and quite a lot at that; QUnit also uses the `eslint:recommended` config which we don't. I wonder how big a gain for QUnit is that it uses the `jquery` config; in the past we thought of it as a preset common for all jQuery Foundation projects but that hasn't materialized. At the moment, I think, it's mostly a few jQuery projects to which similar people contribute: Core, Migrate, Color, and QUnit.

Krinkle:
@mgol Yeah, makes sense to share with e.g. jquery migrate. Using the preset for that makes sense.

I'd say don't worry about QUnit in that case, we can disable things as needed, no problem.

mgol:
For anyone reading this, GitHub has messed the order of the comments and some older comments display below the newer ones.

trentmwillis:
My two cents: I don't think it would impact QUnit much, if it becomes a pain we can opt-out as @Krinkle mentions above.

mgol:
I submitted https://github.com/jquery/eslint-config-jquery/pull/14 for the config changes. Please review, after publishing an update I'll change this PR.

mgol:
Cherry-picked to `3.x-stable` in d7e13f128a4965f5d2bf510dc2314faa266f9dd7.

Issue #4384, opened 2019-05-03
Getting children of <object> tag with contents() not working correctly


patocallaghan:
### Description ###
**Version:** jQuery 3.4.1

Since https://github.com/jquery/jquery/commit/0ba8e38d0c4ab4a4fb9054e7a713630be9743aff it appears that using `contents()` to get the children of an `<object>` is not working as expected.

Taking this example HTML

```html
<object type="application/x-shockwave-flash" width="200" height="300" id="penguin">
    <param name="movie" value="flash/penguin.swf">
    <param name="quality" value="high">
    <img src="images/penguin.jpg" width="200" height="300" alt="Penguin">
</object>
```

I'd expect the following to happen

```js
jQuery(objectHtml).contents().length // => 3
```

But what actually happens is:

```js
jQuery(objectHtml).contents().length // => 0
```

This happens because on this line https://github.com/jquery/jquery/blob/110802c7f22b677ef658963aa95ebdf5cb9c5573/src/traversing.js#L148-L149 it only checks if `elem.contentDocument` is not  `undefined`. But in this scenario `elem.contentDocument` is `null` so the conditional is truthy and enters the `if` block and returns `null` for the contents.

If we change the conditional to be `elem.contentDocument != null` it works correctly.

### Link to test case ###
https://codepen.io/patocallaghan/pen/bybaWv?editors=0011

I have a PR to fix it but I wanted to open an issue first.

mgol:
Thanks for the report! This is a regression from jQuery 3.3.1 so I added the 3.4.2 milestone.

mgol:
PR: #4385

PR #4385, opened 2019-05-03
Fix retrieving children of `<object>` using `contents()`

patocallaghan:
Fix for https://github.com/jquery/jquery/issues/4384

### Summary ###
Since https://github.com/jquery/jquery/commit/0ba8e38d0c4ab4a4fb9054e7a713630be9743aff, using `contents()` to get the children of an `<object>` is not working.

Taking this example HTML

```html
<object type="application/x-shockwave-flash" width="200" height="300" id="penguin">
    <param name="movie" value="flash/penguin.swf">
    <param name="quality" value="high">
    <img src="images/penguin.jpg" width="200" height="300" alt="Penguin">
</object>
```

I'd expect the following to happen

```js
jQuery(objectHtml).contents().length // => 3
```

But what actually happens is:

```js
jQuery(objectHtml).contents().length // => 0
```

Changing the conditional to `elem.contentDocument != null` handles this case and the function returns the correct children.

### Checklist ###

### DIFF
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -145,7 +145,7 @@ jQuery.each( {
 		return siblings( elem.firstChild );
 	},
 	contents: function( elem ) {
-		if ( typeof elem.contentDocument !== "undefined" ) {
+		if ( elem.contentDocument != null ) {
 			return elem.contentDocument;
 		}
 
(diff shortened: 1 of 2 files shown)


mgol:
Note to self: remember to cherry-pick this to `3.4-stable`.

mgol:
I merged it and the newly added test fails in IE: http://swarm.jquery.org/result/3328701

mgol:
PR with a fix for IE 11: #4390

Issue #4386, opened 2019-05-04
Rely on documentMode for most IE workarounds?


mgol:
### Description ###

With jQuery 4.0 dropping support for lots of legacy browsers, currently (on `master`) all of our support tests only fail in IE. This is not a surprise as IE 11 is the only browser we support that's not getting updates. I'd like to consider relying on `document.documentMode` to detect IE in such cases.

First versions of jQuery relied on user agent parsing for applying workarounds. Later, that was changed to rely on support tests. What problems did the original approach have?
1. If a browser fixed the bug, the workaround was still applied in it, needlessly penalizing it.
2. Another browser using a similar engine may also have a bug but the workaround wasn't applied there if its UA didn't match our check.
3. Another browser using a similar engine may not have the same bug as a browser with a similar user agent that we test on but it'd be penalized anyway.
4. An update to the browser may change its user agent, breaking our check.
5. The workaround may rely on the bug itself, meaning that if the browser fixed that bug, jQuery would break in it. This may prevent browsers from fixing issues.

Let's address those points one by one:
1. There are no future IE versions planned. Edge with EdgeHTML was its continuation but its development stopped as well (and I wouldn't plan to cover Edge with this approach). While we can't have guarantees what will happen in the future, I'd rather not optimize for improbable edge cases - and this point is just about needless workarounds, in such an unlikely event an updated IE would still not break jQuery.
2. There's enough code on the internet (e.g. in various web frameworks like AngularJS) in which IE workarounds are guarded by `documentMode` checks that any Trident-based browser with removed `documentMode` would face lots of troubles anyway. Also, we don't have to support such browsers apart from IE. Plus, with Trident being closed source forks can't freely edit the engine for their needs as they can do with e.g. Chromium.
3. Modern browsers are being based on modern engines, Trident isn't one of them. See point (1) - I'd rather not optimize for improbable edge cases.
4. A `documentMode` check is not a user-agent check. Parsing user agent strings is inherently fragile, here we'd just check for thuthiness of `document.documentMode`. Any IE update removing `documentMode` would break lots of code in IE already.
5. To make sure that in the unlikely event IE fixes a bug we workaround it won't break us, we can run tests in Chrome Headless with `document.documentMode` set to `11`.

I'd like to note that historically we haven't relied on support tests fully. In some cases we lumped a few issues under a single support test that didn't carefully check them all. I'd have to look a bit more for specific links to discussions, though.

What would be the gains of relying on `documentMode`?
[...]

Issue #3041, opened 2016-04-05
.width() in fullscreen mode in iframe is too large in IE 11


heporap:
.width() is incorrect value in fullscreen mode of iframe.

I got 12800 on IE11. IE Edge is OK.

```html
<iframe src="iframe.html" allowFullscreen webkitAllowFullscreen></iframe>
```

iframe.html, html
```html
<div id="fs" style="width:100%;height:100px;background-color:green;color:black;">
	<p><input type="button" value="fullscreen" id="run"></p>
	<p><input type="button" value="test" id="test"></p>
	<p id="output"></p>
</div>
```

iframe.html, javascript
```javascript
jQuery(function($){
	$('#test').click(function(){
		$('#output').text( $('#fs').width() );
	});
	$('#run').click(function(){
		pageEnterFullScreen( $('#fs')[0] );
	});
	
});

function pageEnterFullScreen(element){
	
	if( element.requestFullscreen ){
		element.requestFullscreen();
		
	}else if( element.msRequestFullscreen ) {
		element.msRequestFullscreen();
		
	}else if(element.mozRequestFullScreen ){
		element.mozRequestFullScreen();
	
	}else if(element.webkitRequestFullscreen ){
		element.webkitRequestFullscreen();
		
	}
	
}
```

mgol:
Could you create a test case on jsfiddle.net or jsbin.com? Also, I assume you mean it wirks fine in Edge and not the edge mode of IE? (There's no such thing as IE Edge)

heporap:
I could not create iframe page with allowfullscreen attribute on jsfiddle.net. I made a sample in github pages instead. 
http://heporap.github.io/testcase/iframe_fullscreen.html

IE 11 of Windows 8 outputs 128000, and IE Edge of Windows 10 outputs 1920.

dmethvin:
Ref gh-1764 and its PR gh-2401. There is no feature check (it would be pretty much impossible I think) so we're just [blindly multiplying  by 100](https://github.com/jquery/jquery/blob/master/src/css.js#L129-134) if we see fullscreen mode. Maybe IE11 fixed this?

mgol:
[...]

PR #4387, opened 2019-05-04
Core: Remove IE-specific support tests, rely on document.documentMode

mgol:
### Summary ###

~~To make sure none of the IE-targeted workarounds depends on IE bugs we also
have a new special test run on Chrome Headless with document.documentMode
set to 11 to force all those workarounds to be applied in Chrome & make sure
tests still pass.~~

**Edit:**
1. After recent changes we're back at -412 bytes
1. Also, update some tests to IE-sniff when deciding whether to skip a test.

Fixes gh-4386

### Checklist ###


Gruntfile.js
```diff
@@ -163,62 +210,30 @@ module.exports = function( grunt ) {
...
 				singleRun: true
 			},
+
 			main: {
 
 				// The Chrome sandbox doesn't work on Travis.
 				browsers: [ isTravis ? "ChromeHeadlessNoSandbox" : "ChromeHeadless" ]
 			},
 
+			mainWithDocumentMode: {
```
mgol: Do you think this mode is worth it or that we should just assume IE is not going to change too much?

test/unit/support.js
```diff
@@ -58,53 +58,17 @@ testIframe(
...
+			chrome: {},
+			safari: {},
+			firefox: {},
+			ios: {}
 		};
 
 	if ( /edge\//i.test( userAgent ) ) {
 		expected = expectedMap.edge;
-	} else if ( document.documentMode ) {
+	} else if ( /trident\//i.test( userAgent ) ) {
```
mgol: I had to change this check back to be UA-based as Chrome with faked `documentMode` would still pass the support tests that IE fails.

There's no difference at the moment but if we had to add some the problem would surface.

test/unit/fake-document-mode.js
```diff
@@ -0,0 +1,7 @@
+QUnit.module( "fake-document-mode", { afterEach: moduleTeardown } );
```
mgol: I haven't added this test to the list where all others are specified as it only makes sense in the mode with faked `documentMode`.

test/unit/support.js
```diff
@@ -58,53 +58,17 @@ testIframe(
...
+			chrome: {},
+			safari: {},
+			firefox: {},
+			ios: {}
 		};
 
 	if ( /edge\//i.test( userAgent ) ) {
 		expected = expectedMap.edge;
-	} else if ( document.documentMode ) {
+	} else if ( /trident\/7\.0/i.test( userAgent ) ) {
```
mgol: I had to change this check back to be UA-based as Chrome with faked documentMode would still pass the support tests that IE fails.

There's no difference at the moment but if we had to add some the problem would surface.

mgol:
It turns out my workaround for Chrome with `documentMode` emulation failing the "SVG dimensions (border-box)" test was incorrect; IE doesn't support `offset*` properties on SVGs either. The workaround really depends on IE incorrectly reporting content-box dimensions for `border-box` elements. I reverted this change (we're down to `-415` bytes), you can see the failure on Travis now.

There are two ways we can work around this:
1. We accept that in this case we depend on the bug existing in IE and on it not getting fixed there. That means the special "Chrome with `documentMode` emulation" test run doesn't make sense anymore.
2. We restore the `boxSizingReliable` test but guard it in the additional `documentMode` check to avoid triggering layout outside of IE. This will reduce size gains from `-415` bytes to `-201` bytes, although I suspect this can be reduced by removing some now redundant CSS from the support test.

Note that in the past we've been bitten by an assumption that IE won't fix a bug when IE had a bug with dimensions in fullscreen mode inside iframes (see #1764) and we applied a workaround that multiplied the dimensions by 100 based on the `documentMode` check which we had to revert later (see #3041) when IE fixed it contrary to earlier claims. The probability of IE fixing something now is perhaps lower than 3 years ago, especially when it comes to something more complex than this simple iframe sizing bug but there's still some risk.

dmethvin:
I prefer the first option. It seems like the "fullscreen in iframes" bug was trivial to fix but something tells me that IE11 isn't going to fix its unreliable box sizing fix ever. If that seems like too much risk we could just guard it with `documentMode` to avoid forcing layout anywhere but IE. The sad thing about that solution is that IE11 will probably suffer the most performance hit from forcing layout.

gibson042:
I also prefer the first option. As I mentioned in https://github.com/jquery/jquery/issues/4386#issuecomment-489331660 , if anyone ever _does_ release an update or fork of IE that fixes the issue, we can re-add support tests then.

Gruntfile.js
mgol: This has been removed now.

test/unit/support.js
mgol: Reverted since we're not doing this special Chrome run.

mgol:
PR updated and is ready for final review.

I removed the special Chrome with fake docMode run, I also modified some tests to IE-sniff when deciding whether to skip instead of performing sometimes complex (& easier to do wrong) support tests where possible.

test/data/testinit.js
```diff
@@ -256,14 +256,11 @@ if ( !window.__karma__ ) {
...
-	script.type = "module";
-	script.text = "QUnit.moduleTypeSupported = true";
-	document.head.appendChild( script ).parentNode.removeChild( script );
-}
-moduleTypeSupported();
+// Support: IE 11+
+// A variable to make it easier to skip specific tests in IE, mostly
+// testing integrations with newer Web features not supported by it.
+QUnit.isIE = /trident\/7\.0/i.test( window.navigator.userAgent );
+QUnit.testUnlessIE = QUnit.isIE ? "skip" : "test";
```
gibson042: Should we structure this as a function rather than a method name?
```suggestion
QUnit.testUnlessIE = QUnit.isIE ?
	QUnit.skip :
	QUnit.test;
```
```patch
-QUnit[ QUnit.testUnlessIE ]( "css(--customProperty)", function( assert ) {
+QUnit.testUnlessIE( "css(--customProperty)", function( assert ) {
```

mgol: Good idea, that makes much more sense.

mgol: PR updated.

mgol:
@gibson042 PR updated. I also added a support comment for e0d941156900a6bff7c098c8ea7290528e468cf8 that was missing.

mgol:
@gibson042 PR updated (changes are in the new fixup commit). I also added a support comment for e0d941156900a6bff7c098c8ea7290528e468cf8 that was missing.

PR #4389, opened 2019-05-06
Build: Fix the regex parsing AMD var-modules

mgol:
### Summary ###

Fix the regex parsing AMD var-modules.

The previous regex caused the final jQuery binary to have syntax errors for
var-modules with names starting with "return". For example, the following module
wouldn't work when the file is named `returnTrue.js`:

```js
define( function() {
	"use strict";
	return function returnTrue() {
		return true;
	};
} );
```

### Checklist ###

* ~~New tests have been added to show the fix or feature works~~ N/A
* ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~ N/A

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -65,7 +65,7 @@ module.exports = function( grunt ) {
 		if ( /.\/var\//.test( path.replace( process.cwd(), "" ) ) ) {
 			contents = contents
 				.replace(
-					/define\([\w\W]*?return/,
+					/define\(\s*(["'])[\w\W]*?\1[\w\W]*?return/,
 					"var " +
 					( /var\/([\w-]+)/.exec( name )[ 1 ] ) +
 					" ="


mgol:
@timmywil The value of the `contents` variable is not what's in the file, it gets the name & the dependencies array inserted during compilation. For the `returnTrue` example, the `contents` value before the replacement is:
```js
define( 'var/returnTrue',[],function() {

"use strict";

return function returnTrue() {
	return true;
};

} );
```
After the replacement it is:
```js
var returnTrue =True',[],function() {

"use strict";

return function returnTrue() {
	return true;
};
```

See the updated bin: https://jsbin.com/mufamok/1/edit?html,js,console,output

mgol:
This PR is a result of my experiments around #4387. In commit https://github.com/mgol/jquery/commit/9a9c3f0b2bbb0920dbdf61cec6fdab1f34a3cab6 I had to change the regexp because it was mangling my newly-separated `returnTrue` module. This code will not land in the end but I thought I'll at least submit the regex fix separately to avoid future issues like that.

timmywil:
Ah, I forgot it inserts the name. Ok.

build/tasks/build.js
```diff
@@ -65,7 +65,7 @@ module.exports = function( grunt ) {
 		if ( /.\/var\//.test( path.replace( process.cwd(), "" ) ) ) {
 			contents = contents
 				.replace(
-					/define\([\w\W]*?return/,
+					/define\(\s*(["'])[\w\W]*?\1[\w\W]*?return/,
```
gibson042: Simpler suggestion: https://jsbin.com/wozinogigo/1/edit?js,console,output
```suggestion
					/define\([\w\W]*?\breturn\b/,
```

mgol: Technically, there could be a directory named `return`. Not the best idea, perhaps, but my regex should work for all of those weird cases.

mgol:
Cherry-picked to `3.x-stable` in 36b59c9661b7c24bb0271f813d2a02cac44de81a.

PR #4395, opened 2019-05-13
Selector: Inline Sizzle into the selector module

mgol:
### Summary ###

This PR removes Sizzle from jQuery. The first commit inlines Sizzle mostly as-is, the subsequent commit fixes ESLint violations; later commits mostly remove workarounds for old browsers.

Currently the PR reduces the gzipped jQuery size by `1201` bytes compared to `master`.

I'm opening it as a draft as I still need to backport Sizzle tests. It's ready for initial review, though (reviewing by commit might be easier).

I've checked that all the tests pass in Chrome 74 & IE 11. All Sizzle tests are also passing for `jQuery.find` except for the `Sizzle.noConflict` one - but `jQuery.find` has always stripped `noConflict` out of its internal Sizzle version.

### Checklist ###

* [x] New tests have been added to show the fix or feature works (Sizzle tests ported in #4406)


mgol:
I've checked that all the tests pass in Chrome 74 & IE 11. All Sizzle tests are also passing for `jQuery.find` except for the `Sizzle.noConflict` one - but `jQuery.find` has always stripped `noConflict` out of its internal Sizzle version.

timmywil:
This looks like it's on the right track. I'd like to see it with the tests migrated.

mgol:
I submitted a draft PR porting Sizzle tests in #4406. I'll be adding more commits to it but it might be easier for you to review commit by commit even now instead of all at once when I'm done.

mgol:
@jquery/core During my work on #4406, I've reminded myself we support many more custom selectors than just the ones listed in the 3.4.0 blog post as deprecated. We listed: `:first`, `:last`, `:eq`, `:even`, `:odd`, `:lt`, `:gt`, and `:nth` but we also support non-standard (linked to their API pages):
- [`:contains`](https://api.jquery.com/contains-selector/) - no easy replacement?
- [`[attr!="value"]`](https://api.jquery.com/attribute-not-equal-selector/) - can be replaced by users to `:not([attr="value"])`
- [`:selected`](https://api.jquery.com/selected-selector/) - can be replaced with `:checked` which natively supports selected options
- [`:input`](https://api.jquery.com/input-selector/) - no easy replacement other than listing all the matching elements?
and probably others, these are just the ones I had to deal with so far.

What do we plan to do with those selectors? Do we want to implement internal replacement (where possible, `:contains` is out, I think)?

dmethvin:
Some of this can be handled by Migrate, or at least the easy cases of them. I wonder if we want *any* of custom selectors in 4.0 or should just have new custom methods by similar names where needed?

It would be useful to have *very clear* instructions on how to convert from old to new, even if it means making new core methods to simplify things. So `$("div:contains(h1) > p")` is hard to rewrite unless we create a new method so it would be possible to say `$("div").contains("h1").children("p")`.

mgol:
> I wonder if we want _any_ of custom selectors in 4.0 or should just have new custom methods by similar names where needed?

(porting some info from the Gitter discussion) We still need to maintain our own custom parsing just to reliably apply the ID hack to have a correct scoping mechanism. When we have parsing, adding some re-mapping may not be that hard in many of those cases (like `:selected`), so we can try. `:contains` is a bit different here.

timmywil:
Porting a little more from the Gitter conversation, `:contains` will be much easier to support than POS selectors. I think we can revisit which selectors we really want to keep, but I expect we'll be able to support all but POS in the upcoming rewrite.

timmywil:
As far as this PR is concerned, I think it's good to go as a first step.

test/data/testinit.js
```diff
@@ -45,7 +45,7 @@ this.q = function() {
 /**
  * Asserts that a select matches the given IDs
  * @param {String} message - Assertion name
- * @param {String} selector - Sizzle selector
+ * @param {String} selector - DOM selector
```
Krinkle: DOM Selector or CSS Selector?

Krinkle:
Also, if not already, ensure all entries in AUTHORS.txt are present in the jQuery one.

test/data/testinit.js
mgol: It's not a standard CSS selector, that's it was named a Sizzle one. I wanted to get rid of such Sizzle mentions. Would "jQuery selector" be OK?

mgol:
> Also, if not already, ensure all entries in AUTHORS.txt are present in the jQuery one.

@Krinkle Good point. I'll need to sort them into the jQuery list based on comparing their first contribution to Sizzle with other first contributions to jQuery, I guess.

mgol:
OK, I think this is ready to review. I removed the draft status.

mgol:
@Krinkle `AUTHORS.txt` updated.

test/data/testinit.js
Krinkle: Yes, that's a good idea.

PR #4399, opened 2019-05-15
You cannot send a request when AJAX passes in a Json object

lyf2nb:
### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -564,6 +564,11 @@ jQuery.extend( {
 				s.crossDomain = true;
 			}
 		}
+	
+		//You cannot send a request when AJAX passes in a Json object
+		if(typeof(s.data) == "object" && Object.prototype.toString.call(s.data).toLowerCase() == "[object object]" && !s.data.length){
+			s.data = JSON.stringify(s.data);
+		}
 
 		// Convert data if not already a string
 		if ( s.data && s.processData && typeof s.data !== "string" ) {


mgol:
Thanks for the PR but we need to understand the issue you're trying to solve first. Please submit an issue & fill in all relevant details before we can look into a PR.

lyf2nb:
> Thanks for the PR but we need to understand the issue you're trying to solve first. Please submit an issue & fill in all relevant details before we can look into a PR.

https://github.com/jquery/jquery/issues/4403#issue-444718004

dmethvin:
It sounds like you might want to send a JSON string and set the content type. I'm not suggesting we start a help thread here but StackOverflow may have an answer. If not, open a new thread there.

https://stackoverflow.com/questions/12693947/how-to-send-json-instead-of-a-query-string-with-ajax

lyf2nb:
I know what it means, but why not have Jquery support JSON objects instead of serialized JSON strings

lyf2nb:
I know what it means, but why not have Jquery support JSON objects instead of serialized JSON strings
On 5/16/2019 10:07，Dave Methvin<notifications@github.com> wrote：

It sounds like you might want to send a JSON string and set the content type. I'm not suggesting we start a help thread here but StackOverflow may have an answer. If not, open a new thread there.

https://stackoverflow.com/questions/12693947/how-to-send-json-instead-of-a-query-string-with-ajax

PR #1615, opened 2014-07-06
Fix #15173: Support `responseURL` attribute for jqXHR

xKerman:
As I reported in [#15173](http://bugs.jquery.com/ticket/15173), `responseURL` attribute is defined in WHATWG XMLHttpRequest spec and is already implemented in Firefox Aurora and Chrome dev channel.  However we can not get that attribute information from `jqXHR` currently.
This pull request adds `responseURL` attribute for jqXHR.

Tested browsers:
- Chrome:
  - Stable channel (Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/35.0.1916.153 Safari/537.36)
  - Dev channel (Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/37.0.2062.3 Safari/537.36)
- Firefox
  - Stable (Mozilla/5.0 (Macintosh; Intel Mac OS X 10.9; rv:30.0) Gecko/20100101 Firefox/30.0)
  - Aurora (Mozilla/5.0 (Macintosh; Intel Mac OS X 10.9; rv:32.0) Gecko/20100101 Firefox/32.0)
- Safari
  - Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_3) AppleWebKit/537.76.4 (KHTML, like Gecko) Version/7.0.4 Safari/537.76.4

PR #4405, opened 2019-05-19
Adding responseURL to jqXHR

### LINKED ISSUE #4339: Adding responseURL field to jqXHR
### Description ###

The XMLHttpRequest object has the [responseURL property](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseURL), which is very useful to detect redirect and especially where the request got redirected.

jQXHR objects however does not expose this property and since there are no possibility to get the underlying XMLHttpRequest object (unless you override jQuery.ajaxSettings.xhr function), it is not possible for user functions to get the url to which the request was redirected at all.

This feature were already discussed [here](https://bugs.jquery.com/ticket/15173) and [here](https://github.com/jquery/jquery/pull/1615), but the feature were never implemented.

csmadhav:
### Summary ###

Fixes: #4339
I've taken some references from #1615 (which needed to be resurrected)
original PR was from @xKerman
Since recently support for some browsers was dropped in jquery respective changed were also added.
Please ignore the previous PR (closed already)

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -447,8 +447,6 @@ jQuery.extend( {
 			jqXHR = {
 				readyState: 0,
 
-				responseURL: "",
-
 				// Builds headers hashtable if needed
 				getResponseHeader: function( key ) {
 					var match;
@@ -713,7 +711,7 @@ jQuery.extend( {
 		}
 
 		// Callback for when everything is done
-		function done( status, nativeStatusText, responseURL, responses, headers ) {
+		function done( status, nativeStatusText, responses, headers, responseURL ) {
 			var isSuccess, success, error, response, modified,
 				statusText = nativeStatusText;
 
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -447,6 +447,8 @@ jQuery.extend( {
 			jqXHR = {
 				readyState: 0,
 
+				responseURL: "",
+
 				// Builds headers hashtable if needed
 				getResponseHeader: function( key ) {
 					var match;
@@ -711,7 +713,7 @@ jQuery.extend( {
 		}
 
 		// Callback for when everything is done
-		function done( status, nativeStatusText, responses, headers ) {
+		function done( status, nativeStatusText, responseURL, responses, headers ) {
 			var isSuccess, success, error, response, modified,
 				statusText = nativeStatusText;
 
@@ -798,6 +800,7 @@ jQuery.extend( {
 			// Set data for the fake xhr object
 			jqXHR.status = status;
 			jqXHR.statusText = ( nativeStatusText || statusText ) + "";
+			jqXHR.responseURL = responseURL;
 
 			// Success/Error
 			if ( isSuccess ) {
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -1,5 +1,4 @@
 import jQuery from "../core.js";
-import support from "../var/support.js";
 
 import "../ajax.js";
 
@@ -11,10 +10,7 @@ var xhrSuccessStatus = {
 
 		// File protocol always yields status code 0, assume 200
 		0: 200
-	},
-	xhrSupported = jQuery.ajaxSettings.xhr();
-
-support.responseURL = !!xhrSupported && ( "responseURL" in xhrSupported );
+	};
 
 jQuery.ajaxTransport( function( options ) {
 	var callback;
@@ -79,7 +75,7 @@ jQuery.ajaxTransport( function( options ) {
 							complete(
 								xhrSuccessStatus[ xhr.status ] || xhr.status,
 								xhr.statusText,
-								support.responseURL ? xhr.responseURL : "",
+								xhr.responseURL,
 
 								// For XHR2 non-text, let the caller handle it (gh-2498)
 								( xhr.responseType || "text" ) === "text" ?
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -1,4 +1,5 @@
 import jQuery from "../core.js";
+import support from "../var/support.js";
 
 import "../ajax.js";
 
@@ -10,7 +11,10 @@ var xhrSuccessStatus = {
 
 		// File protocol always yields status code 0, assume 200
 		0: 200
-	};
+	},
+	xhrSupported = jQuery.ajaxSettings.xhr();
+
+support.responseURL = !!xhrSupported && ( "responseURL" in xhrSupported );
 
 jQuery.ajaxTransport( function( options ) {
 	var callback;
@@ -68,12 +72,14 @@ jQuery.ajaxTransport( function( options ) {
 
 								// File: protocol always yields status 0; see #8605, #14207
 								xhr.status,
-								xhr.statusText
+								xhr.statusText,
+								""
 							);
 						} else {
 							complete(
 								xhrSuccessStatus[ xhr.status ] || xhr.status,
 								xhr.statusText,
+								support.responseURL ? xhr.responseURL : "",
 
 								// For XHR2 non-text, let the caller handle it (gh-2498)
 								( xhr.responseType || "text" ) === "text" ?
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -68,20 +68,17 @@ jQuery.ajaxTransport( function( options ) {
 
 								// File: protocol always yields status 0; see #8605, #14207
 								xhr.status,
-								xhr.statusText,
-								""
+								xhr.statusText
 							);
 						} else {
 							complete(
 								xhrSuccessStatus[ xhr.status ] || xhr.status,
-								xhr.statusText,
-								xhr.responseURL,
-
-								// For XHR2 non-text, let the caller handle it (gh-2498)
+								xhr.statusText, // For XHR2 non-text, let the caller handle it (gh-2498)
 								( xhr.responseType || "text" ) === "text" ?
 									{ text: xhr.responseText } :
 									{ binary: xhr.response },
-								xhr.getAllResponseHeaders()
+								xhr.getAllResponseHeaders(),
+								xhr.responseURL
 							);
 						}
 					}
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -215,6 +215,14 @@ QUnit.assert.ok( true, "mock executed");';
 		file_put_contents( $this->cspFile, '' );
 		unlink( $this->cspFile );
 	}
+	
+	public function responseURL($req) {
+		if ( isset($req->query["url"]) ) {
+			header( "Location: $req->query[url]" );
+			exit;
+		}
+		echo "OK";
+	}
 
 	protected function errorWithScript( $req ) {
 		header( 'HTTP/1.0 404 Not Found' );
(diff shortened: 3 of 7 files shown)


src/ajax/xhr.js
```diff
@@ -15,7 +16,10 @@ var xhrSuccessStatus = {
 
 		// File protocol always yields status code 0, assume 200
 		0: 200
-	};
+	},
+	xhrSupported = jQuery.ajaxSettings.xhr();
+
+support.responseURL = !!xhrSupported && ( "responseURL" in xhrSupported );
```
mgol: From all of our supported (on `master`) browsers, only IE doesn't support `responseUrl`. Since #4387, we identify IE directly & don't define support tests if only IE is affected.

src/ajax/xhr.js
```diff
@@ -73,12 +77,14 @@ jQuery.ajaxTransport( function( options ) {
...
 								xhr.status,
-								xhr.statusText
+								xhr.statusText,
+								""
 							);
 						} else {
 							complete(
 								xhrSuccessStatus[ xhr.status ] || xhr.status,
 								xhr.statusText,
+								support.responseURL ? xhr.responseURL : "",
```
mgol: What would happen if we just passed it through without checking for support? If IE 11 doesn't crash but just passes `undefined` through, that'd be fine for me.

test/data/mock.php
```diff
@@ -215,6 +215,14 @@ protected function cspClean( $req ) {
 		file_put_contents( $this->cspFile, '' );
 		unlink( $this->cspFile );
 	}
+	
+	public function responseURL($req) {
+		if ( isset($req->query["url"]) ) {
+			header( "Location: $req->query[url]" );
```
mgol: Does that also set the `302` status code, as code in `middleware-mockserver` does?

test/unit/ajax.js
```diff
@@ -1408,6 +1408,103 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - responseURL", 13, function( assert ) {
+		function expectedUrl( url ) {
+			return jQuery.support.responseURL ? url : "";
+		}
+
+		/**
+		 * this function generates complete
```
mgol: Please start sentences with an uppercase letter. Also, we usually use single-line comments (`//`) even for comments spanning multiple lines.

test/unit/ajax.js
```diff
@@ -1408,6 +1408,103 @@ QUnit.module( "ajax", {
...
+		 * absolute URL without any relative path components
+		 * example:
+		 * input: http://localhost:9876/base/test/data/testinit.js/../../../test/data/mock.php?action=responseURL&155821644568982930
+		 * output: http://localhost:9876/base/test/data/mock.php?action=responseURL&155821644568982930
+		 */
+		function removeRelativePathComponents( url ) {
```
mgol: Why is this function necessary? We usually just use our existing `url` function.

test/unit/support.js
```diff
@@ -56,18 +56,33 @@ testIframe(
...
 		userAgent = window.navigator.userAgent,
 		expectedMap = {
-			edge: {},
-			ie_11: {},
-			chrome: {},
-			safari: {},
-			firefox: {},
-			ios: {}
+			edge: {
+				responseURL:true
```
mgol: There should be a space after a colon. But this will be removed anyway as we don't need this support property.

test/unit/support.js
```diff
@@ -56,18 +56,33 @@ testIframe(
...
+			},
+			ios: {
+				responseURL:true
+			}
 		};
 
 	if ( /edge\//i.test( userAgent ) ) {
+		version = parseInt( ( /edge\/([0-9.]+)/i ).exec( navigator.userAgent )[ 1 ] );
 		expected = expectedMap.edge;
+		expected.responseURL = version >= 14;
```
mgol: We only support last 2 Edge versions, currently 17 & 18. This is not needed.

Please revert all changes done to this file.

mgol (commented):
Thanks for the PR! I added some comments.

src/ajax/xhr.js
csmadhav: Okay! I was doing it for IE 11 & Edge < 14, removing

src/ajax/xhr.js
csmadhav: removed this condition as well, IE doesn't crash and passes undefined

test/data/mock.php
csmadhav: Yes

test/unit/ajax.js
csmadhav: This is needed for the case when there is no redirect, in that case, `responseURL` should be equal to the initial URL only. The URL returned by `url` function contains some relative path components like `../../` (checkout the example input to this function in comments) so what happens is that after getting the response, `responseURL` value doesn't contain `../` and its cleaned because of which I am not able to assert it. help me with alternatives here

test/unit/support.js
csmadhav: Yes, I didn't know that, fixed now

test/unit/ajax.js
csmadhav: hey @mgol can you suggest anything here?

mgol: @csmadhav I’m at a conference; I’ll have a look next week unless I get to it before.

mgol: OK, it seems the function can stay but:
1. I'd rename it to `normalizeUrl`; shorter & still fully conveys the meaning.
1. The implementation seems too complex for what it does, how about:

```js
function normalizeUrl( url ) {
	var finalUrl = url;
	while ( finalUrl.indexOf( '/../' ) > -1 ) {
		finalUrl = normalizedUrl.replace(/\/[^/]+\/\.\.\//g, '/');
	}
	return finalUrl;
}
```

test/unit/ajax.js
```diff
@@ -1408,6 +1408,98 @@ QUnit.module( "ajax", {
...
+			url = cleaned + url.substring( match.index + match[ 0 ].length - 1 );
+			if ( ( url.match( /\.\.\//g ) || [] ).length !== 0 ) {
+				url = removeRelativePathComponents( url );
+			}
+			return url;
+		}
+
+		var successURL = removeRelativePathComponents( url( "mock.php?action=responseURL" ) ),
+			errorURL = "http://example.invalid",
+			redirectAndSuccessURL = url( "mock.php?action=responseURL" ) + "&url=" + encodeURIComponent( successURL ),
```
mgol: The whole string should be passed to `url`, not just a part of it.

test/unit/ajax.js
```diff
@@ -1408,6 +1408,98 @@ QUnit.module( "ajax", {
...
+			var splits = url.substring( 0, match.index ).split( "/" );
+			cleaned = splits.slice( 0, splits.length - count - 1 ).join( "/" );
+			url = cleaned + url.substring( match.index + match[ 0 ].length - 1 );
+			if ( ( url.match( /\.\.\//g ) || [] ).length !== 0 ) {
+				url = removeRelativePathComponents( url );
+			}
+			return url;
+		}
+
+		var successURL = removeRelativePathComponents( url( "mock.php?action=responseURL" ) ),
```
mgol: You may use this normalization for _comparison_ but not when triggering a request. The `url` function logic is needed for the two ways we run unit tests (PHP & Node.js-based). You have to use it unaltered when you pass the URL to jQuery but you may normalize the URL _in assertions_, when comparing it to the desired one if it's necessary.

mgol (changes requested):
Sorry for the delay. See my latest comments.

test/unit/ajax.js
csmadhav: Yes, that would solve!

src/ajax.js
```diff
@@ -451,6 +451,8 @@ jQuery.extend( {
 			jqXHR = {
 				readyState: 0,
 
+				responseURL: "",
```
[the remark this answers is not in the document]
csmadhav: it makes sense to change it to `undefined`

src/ajax.js
```diff
@@ -714,7 +716,7 @@ jQuery.extend( {
 		}
 
 		// Callback for when everything is done
-		function done( status, nativeStatusText, responses, headers ) {
+		function done( status, nativeStatusText, responseURL, responses, headers ) {
```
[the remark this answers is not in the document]
csmadhav: I've added it in this API for now, I agree that 5 parameters are quite a lot we should really think of some other API, help me here.

csmadhav:
sorry for updating this late.

src/ajax/xhr.js
```diff
@@ -73,13 +73,12 @@ jQuery.ajaxTransport( function( options ) {
 						} else {
 							complete(
 								xhrSuccessStatus[ xhr.status ] || xhr.status,
-								xhr.statusText,
-
-								// For XHR2 non-text, let the caller handle it (gh-2498)
+								xhr.statusText, // For XHR2 non-text, let the caller handle it (gh-2498)
```
mgol: This comment shouldn't have been moved, please move it back below.

test/unit/ajax.js
```diff
@@ -1523,6 +1523,93 @@ QUnit.module( "ajax", {
...
+				finalUrl = finalUrl.replace( /\/[^/]+\/\.\.\//g, "/" );
+			}
+			return finalUrl;
+		}
+
+		var successURL = url( "mock.php?action=responseURL" ),
+			errorURL = "http://example.invalid",
+			redirectAndSuccessURL = url( "mock.php?action=responseURL&url=" + encodeURIComponent( successURL ) ),
+			redirectAndErrorURL = url( "mock.php?action=responseURL&url=" + encodeURIComponent( errorURL ) ),
+			jsonpURL = baseURL + "mock.php?action=jsonp&callback=?",
```
mgol: Is there any reason why this uses `baseURL` directly instead of the `url` function?

src/ajax.js
```diff
@@ -711,7 +711,7 @@ jQuery.extend( {
 		}
 
 		// Callback for when everything is done
-		function done( status, nativeStatusText, responses, headers ) {
+		function done( status, nativeStatusText, responses, headers, responseURL ) {
```
mgol: Unfortunately, this function signature is public API as that's the shape of `completeCallback` as described here: https://api.jquery.com/jQuery.ajaxTransport/

This is a new parameter at the end so it's not a breaking change. However, we need to think if this is a good API as 5 unnamed parameters is quite a lot.

(from my comment from https://github.com/jquery/jquery/pull/4405#discussion_r305066924)

mgol (changes requested):
See newest comments.

mgol:
Adding the "Needs review" label back to discuss https://github.com/jquery/jquery/pull/4405#discussion_r364510504

src/ajax.js
mgol: FYI: I opened #4634 to make adding `responseURL` possible without adding more parameters here.

mgol:

mgol:
@csmadhav I know it's been a while, but we finally landed `jQuery.ajaxTransport` tests in #5793 and the object-based API in #4634, so this PR is now unblocked. Would you be interested rebasing over the latest `main` & adding `jQuery.ajaxTransport` tests?

mgol:
Closing in favor of #5826.

PR #4406, opened 2019-05-20
Selector: Port Sizzle tests to jQuery

mgol:
### Summary ###

This PR ports Sizzle tests to jQuery. Reviewing by commit might be the best. Sizzle tests are located in three files named [`extending.js`](https://github.com/jquery/sizzle/blob/2.3.4/test/unit/extending.js), [`utilities.js`](https://github.com/jquery/sizzle/blob/2.3.4/test/unit/utilities.js) & [`selector.js`](https://github.com/jquery/sizzle/blob/2.3.4/test/unit/selector.js). TODO list:
- [x] Review tests from `extending.js`: these are only tests for custom pseudos which we'll be removing so I haven't backported them.
- [x] Review tests from `utilities`.js: I backported everything that wasn't already present with the exception of `Sizzle.attr` a.k.a. `jQuery.find.attr`. This is an internal API & I think jQuery attribute tests are thorough enough that we don't need to add these tests. Let me know if you disagree.
- [x] Review tests from `selector.js`.

### Checklist ###


mgol:
I ported many more tests, pushed in new commits. Still a lot to go but we're getting closer.

mgol:
I added a few more commits. It's not 100% ready yet but it's close so I'd appreciate a review.

I'll finish remaining pieces & address feedback when I'm back from vacation.

timmywil:
This is a lot to review. Honestly, I'm okay with merging when you feel it is ready and then we can increment in smaller PRs from there.

mgol:
As discussed at the meeting today, I'll merge it soon (tomorrow or on Wednesday) unless someone wants to have a look.

Some final remarks/questions from me (you can review them instead if you want):
1. I updated `selector-native` slightly so that it passes some backported tests it used to fail - some Sizzle logic wasn't backported there.
1. Re `test/unit/extending.js`: it contains tests for custom pseudos which we'll soon drop so I haven't backported it
1. Re `test/unit/utilities.js`: it tests a few APIs that we already have tets for on the jQuery side; it also contains tests for `Sizzle.attr`, though. I assume they're not needed as jQuery attr tests should be enough.
1. Tests using `Sizzle.matchesSelector` have been migrated by replacing Sizzle.matchesSelector(selector, elem)` by `jQuery(elem).is(selector)`. `Sizzle.matches` tests weren't backported as it's a private API that no code in jQuery or Sizzle uses.
1. Lots of tests using Sizzle have been migrated to use `jQuery.find` - should we document that method? If not, should we migrate those tests to use `jQuery`? Some queries need to be modified then. jQuery.find accepts as many as four parameters; do we want to support them all? Some tests are using all of them right now.
1. Safari has surprisingly good support for CSS selectors that other browsers don't support. E.g. `:not(complex selector)` works in Safari. Therefore, I UA-sniff in tests & run some tests just in Safari OR if Sizzle is present (by checking for `jQuery.find.compile` as existing tests do).

Issue #4453, opened 2019-08-09
Leverage the :scope pseudo-class instead of the ID hack where possible


mgol:
### Description ###

The [`:scope` pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:scope) has surprisingly good browser support: Chrome, Firefox & Safari have supported if for a long time; only IE & Edge lack support. We could leverage this pseudo-class to get rid of the ID hack in most cases. Adding a temporary ID may cause layout thrashing which was reported [a few times](https://github.com/jquery/sizzle/issues/405) in [the past](https://github.com/jquery/jquery/issues/4332).

We can't completely eliminate the ID hack in modern browses as sibling selectors require us to change context to the parent and then `:scope` stops applying to what we'd like. But it'd still improve performance in the vast majority of cases.

I've attached the `3.4.2` milestone as the patch seems simple enough that I'd like to backport it to Sizzle. It'd be a nice perf win in a patch release.

### Link to test case ###

N/A

mgol:
PR: #4454

mgol:
The PR landed on `master` but it also needs to land in Sizzle if we want it in jQuery 3.4.2. Therefore, I'm reopening this issue for now to investigate if that's feasible.

mgol:
Sizzle PR: https://github.com/jquery/sizzle/pull/453

mgol:
I added the "Needs review" label to indicate the Sizzle PR needs to be reviewed (as we don't track such a label there).

mgol:
The Sizzle PR landed. Reopening since we still need to update Sizzle in jQuery 3.x.

mgol:
Sizzle update merged in #4641. The issue is now fixed both on `master` & on `3.x-stable`.

Issue #4332, opened 2019-03-24
Temporary Sizzle nodes trigger mutationObserver and no easy way to identify their nature. Related performance problems on running against different window.


exteon:
### Description ###

We are adding an element via jQuery to an inner frame (different window object) which is watched by a `MutationObserver` set on `window.document`.

When adding the element, a blank node with an id of `#sizzle<numeric_id>`  is created on the document.

What is the documented purpose of this node? We would like to be able to either avoid its creation or a reliable way to determine its nature and ignore it (matching id against `sizzle\d+` seems a bit contrived, undocumented and prone to break in the future...!?).

Searching for this revealed it's some internal functionality in jQuery related to supporting IE. However we  are actually seing this in Chrome/FF so we are at a loss for any documentation of this behavior.

**Update:** the manipulations on these nodes seem to happen even in the most innocent-looking of operations, such as calling `:is(':hidden')` on any node. It's a storm of DOM changes for any operation.

**Update2:** We took a look at the jQuery source files and did some more behavior testing; it's apparent to us that `Sizzle.setDocument()` (which does this) is called on every invocation on any selector. There is no way this can not be a performance hog. Also we found it also creates `<div></div>` nodes without id directly in the document, so there is no way to tell the nature of the created nodes.

gibson042:
Your observations are correct. Creation of test elements such as those from https://github.com/jquery/sizzle/blob/67e9691e8689b824387affb9da06dc663da63598/src/sizzle.js#L634 cannot be avoided, and giving them unpredictable ids is intentional because it keeps multiple instances of jQuery and/or Sizzle from confusing each other's data.

`setDocument` itself is necessary because almost every DOM implementation has bugs that Sizzle must detect in order to fix, and those bugs are different across e.g. HTML and XML DOMs within the same browser. `setDocument` aborts when it detects that the context document has not changed since the last invocation, but we do not cache beyond that so switching between frames incurs the penalty.

This could potentially be addressed, but it is an unusual case that we haven't optimized for (and would not optimize for unless it could be done without significantly increasing gzipped file size). And since a future release of jQuery will replace Sizzle anyway, we are unlikely to devote much effort to this. But if you wish to open a pull request against Sizzle, I promise to help you work through it. Absent that, I would probably instrument the mutation observer to be aware of and ignore the automatically-created elements (all of which should relate to selection, and all of which are also removed immediately).

exteon:
Ok, thanks, notes:

[...]

PR #4454, opened 2019-08-09
Selector: Leverage the :scope pseudo-class where possible

mgol:
### Summary ###

The [`:scope` pseudo-class][1] has surprisingly good browser support: Chrome,
Firefox & Safari have supported if for a long time; only IE & Edge lack support.
This commit leverages this pseudo-class to get rid of the ID hack in most cases.
Adding a temporary ID may cause layout thrashing which was reported a few times
in [the past.

We can't completely eliminate the ID hack in modern browses as sibling selectors
require us to change context to the parent and then `:scope` stops applying to
what we'd like. But it'd still improve performance in the vast majority of
cases.

[1]: https://developer.mozilla.org/en-US/docs/Web/CSS/:scope

Fixes gh-4453
Ref gh-4332
Ref jquery/sizzle#405

### Checklist ###

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -154,7 +154,7 @@ function selectorError( msg ) {
 }
 
 function find( selector, context, results, seed ) {
-	var m, i, elem, nid, match, groups, newSelector, canUseScope,
+	var m, i, elem, nid, match, groups, newSelector,
 		newContext = context && context.ownerDocument,
 
 		// nodeType defaults to 9, since context defaults to document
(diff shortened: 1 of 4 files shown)


mgol:
Currently, the PR adds 49 bytes gzipped. I'd welcome suggestions how to compress it but I'd argue even at this size increase it's worth it due to the fact it removes layout thrashing from the place people don't expect it may be, i.e. regular DOM querying.

mgol:
We'll be able to reduce the space impact from +49 to +22 bytes once we drop support for EdgeHTML and we can remove the support test in favor of an `isIE` check.

mgol:
Reduced to +43 bytes.

mgol:
Reduced to +37 bytes (see the newest commit).

mgol:
I added a test to ensure Mutation Observers are not fired for non-sibling selections.

PR #4459, opened 2019-08-19
Core: Use Array#flat where supported

aelafifi:
### Summary ###

Solve `$.map()` issue when working with large arrays on Chrome browser (#4320)

### Checklist ###


aelafifi:
I've used the native `.flat()` function wherever it's supported.
Otherwise, `.concat.apply()` will be called, because after trying implementing  a custom logic to flat the array, it takes a very long time to handle it in the browser, and as mentioned @ [Won't Fix](https://github.com/jquery/jquery/wiki/Won't-Fix):
> * Serious performance implications of fixing a relatively rare problem

Other solution will be good (from the functional side) is to put the flatting logic into the main loop in `$.map()` itself, but I think it's a bad design, especially in this case since it's very rare!

src/var/flat.js
```diff
@@ -0,0 +1,14 @@
+define( [
+	"./arr",
+	"./concat"
+], function( arr, concat ) {
+	"use strict";
+
+	return arr.flat ? function() {
```
mgol: Please add a support comment, i.e. a comment starting with `Support:`, mentioning affected browsers & explaining where the fallback is used. Here's an example of a support comment:
https://github.com/jquery/jquery/blob/b334ce7735ae453bd5643b251f36c3cedce4b3de/src/css.js#L323-L328
Here, judging by MDN data:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat
you need to mention IE, Edge & Safari (all with version ranges).

mgol:
Thanks for the PR! See my comment & please sign our CLA.

src/var/flat.js
```diff
@@ -0,0 +1,14 @@
+define( [
+	"./arr",
+	"./concat"
+], function( arr, concat ) {
+	"use strict";
```
mgol: Please de-indent the contents of this function. We plan to migrate to ES modules and we'd like to avoid massive indentation changes then.

src/var/flat.js
```diff
@@ -0,0 +1,18 @@
+define( [
+	"./arr",
+	"./concat"
+], function( arr, concat ) {
+
+"use strict";
+
+// Support: IE, chrome < 69, firefox < 62, opera < 56, safari < 12
```
mgol: The format of support comments should follow the syntax described in:
https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-69379197
(examples in the table down the comment) with my modification:
https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-448998379

In particular: browser names start with an uppercase letter, please mind the spacing, each browsers needs a version range specified & each of them needs to end with `only` or with `+`. Also, there's no need to include Chrome, Firefox & Opera here as the issue there is only in versions we don't support. You should add Edge, though.

src/var/flat.js
aelafifi: Done, please re-check it and tell me if it should be documented some way different.

src/var/flat.js
```diff
@@ -0,0 +1,18 @@
...
+// IE and other browsers earlier versions doesn't have `flat` method
```
mgol: I'd phrase it differently:
```suggestion
// Provide fallback for browsers without Array#flat.
```

mgol:
Also, it'd be great if the PR title & commit messages followed the format from https://contribute.jquery.org/commits-and-pull-requests/#commit-guidelines

We have a pre-commit hook that enforces that & it should catch those transgressions, do you have Node.js installed locally?

src/var/flat.js
aelafifi: IE & Edge doesn't support Array#flat at all.
Could you suggest its format, because I don't know how to define version range while the range is undefined (opened)!

mgol: Did you have a look at the second comment I linked above? It provides a few examples. Basically, for Edge I'd expect `Edge 12 - 18+` or `Edge 18+`, for IE: `IE 11+`.

aelafifi:
> Also, it'd be great if the PR title & commit messages followed the format from https://contribute.jquery.org/commits-and-pull-requests/#commit-guidelines

Thanks, I'll check it.

> We have a pre-commit hook that enforces that & it should catch those transgressions, do you have Node.js installed locally?

Yes, I've Node.js v12.6.0

mgol:
@aelafifi Can you make sure all usages of the `concat.apply( [], something )` pattern are converted? I see at least one other usage in `manipulation`.

src/core.js
```diff
@@ -395,7 +395,7 @@ jQuery.extend( {
 		}
 
 		// Flatten any nested arrays
-		return concat.apply( [], ret );
+		return flat.call( ret );
```
mgol: I think it might be smaller if the function just accepted a parameter: `flat( ret )`, without using `call`. You may try such an implementation & see if running `grunt` shows a smaller size increase.

aelafifi: Done

test/unit/core.js
```diff
@@ -791,6 +791,17 @@ QUnit.test( "jQuery.map", function( assert ) {
...
+	result = jQuery.map( [ [ [ 1, 2 ], 3 ], 4 ], function( v, k ) {
+		return v;
+	} );
+	assert.equal( result.length, 3, "Array flatten only one level down" );
+	assert.ok( Array.isArray( result[ 0 ] ), "Array flatten only one level down" );
+
+	result = jQuery.map( Array( 300000 ), function( v, k ) {
+		return k;
+	} );
+	assert.equal( result.length, 300000, "Able to map 300000 records without any problems (#4320)" );
```
mgol: As I understand, this will fail in browsers without `Array#flat` support. So we need to guard those tests with a check for `Array.prototype.flat`. Other browsers need dummy assertions so that the count passed to `assert.expect` matches. See an example:
https://github.com/jquery/jquery/blob/b334ce7735ae453bd5643b251f36c3cedce4b3de/test/unit/selector.js#L24-L30

src/var/flat.js
```diff
@@ -0,0 +1,18 @@
...
+], function( arr, concat ) {
+
+"use strict";
+
+// Support: IE 11+, Edge 18+
+// Provide fallback for browsers without Array#flat.
+return arr.flat ? function( array ) {
+
+	// Flat with depth = 1 to give the same result of concat.apply.
+	return arr.flat.call( array, 1 );
```
mgol: 1 is the default, you don't need to specify it. Also, the comment above this line is not really necessary.

mgol:
I run some benchmarks: `flat` is the fastest in Chrome, Firefox & Safari: https://jsperf.com/flat-vs-concat-vs-iteration

aelafifi:
I'll work on them as soon as possible,
I have a question about the PR name, why it's related to component `Selector`?

timmywil:
@aelafifi You're right that it's more universal than selector, but the main use for this historically has been to array-ify the result of selections.

mgol:
@aelafifi Sorry, my mistake, it should have been `Core` as it's used in other places as well, e.g. manipulation.

mgol:
I'll remove the "Needs review" label for now as the PR needs some further work.

aelafifi:
Sorry for being late.
I've updated the points above, please check them and tell me if there is any thing wrong.

timmywil:
That’s not what I meant. We’re not using concat anywhere else so it doesn’t need to be its own var file.

aelafifi:
Great, it's done.

test/unit/core.js
```diff
@@ -791,6 +791,22 @@ QUnit.test( "jQuery.map", function( assert ) {
...
 	assert.equal( result.join( "" ), "00012223", "Array results flattened (#2616)" );
+
+	result = jQuery.map( [ [ [ 1, 2 ], 3 ], 4 ], function( v, k ) {
+		return v;
+	} );
+	assert.equal( result.length, 3, "Array flatten only one level down" );
+	assert.ok( Array.isArray( result[ 0 ] ), "Array flatten only one level down" );
+
+
+	if ( Array.prototype.flat ) {
```
mgol: Please add a support comment similar to the one in `src/var/flat.js`:
```js
// Support: IE 11+, Edge 18+
// Skip the test in browsers without Array#flat.
```

aelafifi: Added :smiley:

mgol:
Landed on `master` & `3.x-stable`, changes will be in jQuery 3.5.0. Thanks!

PR #4466, opened 2019-08-21
Build: Remove the external directory, read from node_modules directly

mgol:
### Summary ###

Now that Sizzle is gone & we use npm, we can read from node_modules directly
and skip the setup that copies some files to the external directory.

### Checklist ###


mgol:
Both the Karma test run (as evidenced by this PR) and manual QUnit running via the PHP setup works with this PR.

mgol:
@dmethvin `package-lock.json` is not committed to the repo due to various npm instability issues. That said, all 3 packages that we source from `node_modules` directly now are installed in specific versions in `package.json` and we only load their source so even without a lockfile we know exactly what we use.

PR #4467, opened 2019-08-21
Ajax: Simplify jQuery.ajaxSettings.xhr

mgol:
### Summary ###

Previously, jQuery.ajaxSettings.xhr, contents were wrapped in a try-catch
as we defined jQuery.support.ajax & jQuery.support.cors executed during the
jQuery load and we didn't want to crash if IE had native XHR disabled (which
is possible). While jQuery hasn't supported the ActiveX-based XHR since 2.0,
jQuery with XHR disabled could still be used for its other features in such
a crippled browser.

Since gh-4347, jQuery.support.ajax & jQuery.support.cors no longer exist, so
we don't need the try-catch anymore.

Fixes gh-1967
Ref gh-4347

-8 bytes

### Checklist ###

### DIFF
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -6,9 +6,7 @@ define( [
 "use strict";
 
 jQuery.ajaxSettings.xhr = function() {
-	try {
-		return new window.XMLHttpRequest();
-	} catch ( e ) {}
+	return new window.XMLHttpRequest();
 };
 
 var xhrSuccessStatus = {


src/ajax/xhr.js
```diff
@@ -6,9 +6,7 @@ define( [
 "use strict";
 
 jQuery.ajaxSettings.xhr = function() {
-	try {
-		return new window.XMLHttpRequest();
-	} catch ( e ) {}
+	return new window.XMLHttpRequest();
```
dmethvin: In environments like jsdom the user would need to ensure there's a fake XHR stub in `window` or jQuery will throw an error here. That's a bit of a different behavior than currently, where it returns `undefined`. Any attempt to use `$.ajax` will fail either way, but I wonder if anyone calls `$.ajaxSettings.xhr()` to detect whether XHR is enabled. Probably not?

jaubourg: A call to `$.ajax` would fail anyway. Only difference is the exception might actually be a little more informative with this change. I'm pretty certain nobody actually calls `$.ajaxSettings.xhr()` as a means to control if there is an XHR implementation and the change does not preclude from overriding `$.ajaxSettings.xhr` if needs be.

It's a +1 for me.

PR #4469, opened 2019-08-26
Tests: fix karma debug tasks, no jQuery global is available

timmywil:
### Summary ###

In the process of testing another PR, I noticed that the debug tasks (not `karma:main`) were all broken because testrunner expected a jQuery global, which is removed with `noConflict` in testinit. I'm not sure how `karma:main` was even working unless another jQuery somehow leaked into the globals there.

### Checklist ###

### DIFF
--- a/test/data/testrunner.js
+++ b/test/data/testrunner.js
@@ -4,12 +4,16 @@
 
 // Store the old counts so that we only assert on tests that have actually leaked,
 // instead of asserting every time a test has leaked sometime in the past
-var oldCacheLength = 0,
+var ajaxSettings,
+	oldCacheLength = 0,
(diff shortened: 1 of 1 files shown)


test/data/testrunner.js
```diff
@@ -4,12 +4,16 @@
...
-var oldCacheLength = 0,
+var ajaxSettings,
+	oldCacheLength = 0,
 	oldActive = 0,
 
 	expectedDataKeys = {},
-	splice = [].splice,
+	splice = [].splice;
+
+if ( window.jQuery ) {
```
mgol: Won't it cause silent ajax failures since a variable we need won't be declared?

PR #4470, opened 2019-08-26
Tests: Fix offset fractions tests in Chrome for Android

mgol:
### Summary ###

This PR backports a looser assertion from `master` where the browsers passes
offset tests. This should fix the TestSwarm failures that we have in Chrome for Android, see: http://swarm.jquery.org/result/3341025

### Checklist ###

### DIFF
--- a/test/unit/offset.js
+++ b/test/unit/offset.js
@@ -806,7 +806,7 @@ QUnit.test( "fractions (see #7730 and #7885)", function( assert ) {
 	// Support: Chrome <=45 - 46
 	// In recent Chrome these values differ a little.
 	assert.ok( Math.abs( result.top - expected.top ) < 0.25, "Check top within 0.25 of expected" );
-	assert.equal( result.left, expected.left, "Check left" );
+	assert.ok( Math.abs( result.left - expected.left ) < 0.25, "Check left within 0.25 of expected" );
 
 	div.remove();
 } );


test/unit/offset.js
```diff
@@ -806,7 +806,7 @@ QUnit.test( "fractions (see #7730 and #7885)", function( assert ) {
 	// Support: Chrome <=45 - 46
 	// In recent Chrome these values differ a little.
 	assert.ok( Math.abs( result.top - expected.top ) < 0.25, "Check top within 0.25 of expected" );
-	assert.equal( result.left, expected.left, "Check left" );
+	assert.ok( Math.abs( result.left - expected.left ) < 0.25, "Check left within 0.25 of expected" );
```
dmethvin: On Windows I've seen 0.33 pixels as the difference:
```
Expected:   11
Result:     11.33331298828125
Difference: +0.33331298828125
at http://localhost/jquery/test/unit/offset.js:484:9     at http://localhost/jquery/test/data/testinit.js:267:10
```
This was on Windows. Could we go with 0.5 as the limit?

mgol: @dmethvin Are you sure you had no zoom or interfering extensions? I'm fine with increasing the limit if there's a need but:
1. We run tests on Chrome on Windows a lot and it's not exceeding the limit there.
2. We have the same (`0.25`) limits set on `master` as I'm trying to do here (& as is already done above).

If those limits are not enough, we should increase them on `master` as well.

timmywil: `.33` sounds like a lot.

mgol: @timmywil That's true. Even our 0.25 limit is very high, on TestSwarm I see the diff of `0.0078125` so even `0.01` would suffice: http://swarm.jquery.org/result/3341025

dmethvin: I agree it's a lot, but I'm seeing it consistently on my new IBM Thinkpad. Screen is not zoomed. I suspect it has to do with the hi-res screen and the scaling it does, but I'm not sure. 
Annotation 2019-09-03 221104

mgol: @dmethvin Can we look into this issue separately? This PR mainly resolves the current TestSwarm breakage by aligning tests with `master`. Any further changes should be done on `master` first. Apart from that, I'd love to try to debug this issue further first. As Timmy mentioned, `0.33` is a lot so if it's not some buggy environment, we may have to look into fixing it as such a delta may break lots of apps, I'd rather not allow us to be so inaccurate.

dmethvin: @mgol Sure, I don't think it needs to hold up this PR.

Issue #4441, opened 2019-07-29
IE & Edge crash with "Permission denied" on the `contents() for <frame />` test


mgol:
### Description ###

IE 11 as well as Edge (even 17-18) often crash on the [`traversing: contents() for <frame />`](https://github.com/jquery/jquery/blob/47835965bd100a3661d8299d8b769ceeb8b6ce48/test/unit/traversing.js#L883-L898) test with the "Permission denied" error. IE crashes way often than Edge. This is probably related to my Sizzle removal PR (#4395) but I wasn't able to pinpoint the exact reason so far. The error points to [line `174` in `src/selector.js`](https://github.com/jquery/jquery/blob/47835965bd100a3661d8299d8b769ceeb8b6ce48/src/selector.js#L174). An example failing run: http://swarm.jquery.org/result/3337866.

It's hard to debug as it's clearly a race condition. If I run this test in isolation, it always passes.

I think I've seen a similar failure on the `3.4-stable` branch as well but it was happening way less often.

We should figure this out before the `4.0.0` release.

### Link to test case ###

N/A, it's a core test already in the repo.

mgol:
Unfortunately, while in the vast majority of cases the test case fails in IE, when you run that particular test in isolation it almost never fails. This makes debugging tricky.

mgol:
Some conclusions from my investigation:

1. The error is triggered by [this line](https://github.com/jquery/jquery/blob/47835965bd100a3661d8299d8b769ceeb8b6ce48/test/unit/traversing.js#L891):
    ```js
    var contents = container.find( "#test-frame" ).contents();
    ```
1. This is not the only test failing with "Permission denied". I initially thought this one failure triggers more of them sometimes but commenting out that one test results in frequent failures in some other tests, e.g. the [`traversing: sort direction`](https://github.com/jquery/jquery/blob/47835965bd100a3661d8299d8b769ceeb8b6ce48/test/unit/traversing.js#L900-L924) test. The stack trace doesn't point to anything that could help.

mgol:
I found the cuplrit. IE/Edge sometimes crash when comparing documents between frames... but only if one of the strict equality operators (`===` or `!==`) is used. Shallow equivalents (`==` or `!=`) work fine... PR incoming.

This means the issue might as well exist on 3.x as well, maybe it just triggers the condition less often. I'm changing the milestone to 3.4.2 & I'll submit a PR to Sizzle as well.

mgol:
[...]

PR #4471, opened 2019-08-26
Selector: Use shallow document comparisons to avoid IE/Edge crashes

mgol:
### Summary ###

IE/Edge sometimes crash when comparing documents between frames using the strict
equality operator (`===` & `!==`). Funnily enough, shallow comparisons
(`==` & `!=`) work without crashing.

Fixes gh-4441

While I cannot be 100% certain this PR will solve the issue, I modified the code locally, wrapping the place where the browser crashes in a try-catch where I assigned both compared documents to global variables. I then compared them using all 4 operators manually in the console and the strict one were always crashing. I reproduced the same behavior in IE 11 & Edge 18.

I can't really write a test for it as the crashes are not consistent - the only consistent thing was that if it crashed once on a particular comparison, it keeps crashing on strict-comparing those two documents.

This fixes a failure in traversing tests but the code is in Selector so I'm marking the component as such.

### Checklist ###

* ~~New tests have been added to show the fix or feature works~~ Not possible to test it reliably IMO

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -172,14 +172,7 @@ function find( selector, context, results, seed ) {
 
 	// Try to shortcut find operations (as opposed to filters) in HTML documents
 	if ( !seed ) {
-
-		// Support: IE 11+, Edge 17 - 18+
-		// IE/Edge sometimes throw a "Permission denied" error when strict-comparing
-		// two documents; shallow comparisons work.
-		// eslint-disable-next-line eqeqeq
-		if ( ( context ? context.ownerDocument || context : preferredDoc ) != document ) {
-			setDocument( context );
-		}
+		setDocument( context );
 		context = context || document;
 
 		if ( documentIsHTML ) {
@@ -461,15 +454,7 @@ find.matches = function( expr, elements ) {
 };
 
 find.matchesSelector = function( elem, expr ) {
-
-	// Set document vars if needed
-	// Support: IE 11+, Edge 17 - 18+
-	// IE/Edge sometimes throw a "Permission denied" error when strict-comparing
-	// two documents; shallow comparisons work.
-	// eslint-disable-next-line eqeqeq
-	if ( ( elem.ownerDocument || elem ) != document ) {
-		setDocument( elem );
-	}
+	setDocument( elem );
 
 	if ( documentIsHTML &&
 		!nonnativeSelectorCache[ expr + " " ] &&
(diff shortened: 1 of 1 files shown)


src/selector.js
```diff
@@ -173,7 +173,11 @@ function find( selector, context, results, seed ) {
 	// Try to shortcut find operations (as opposed to filters) in HTML documents
 	if ( !seed ) {
 
-		if ( ( context ? context.ownerDocument || context : preferredDoc ) !== document ) {
+		// Support: IE 11+, Edge 17 - 18+
+		// IE/Edge sometimes throw a "Permission denied" error when strict-comparing
+		// two documents; shallow comparisons work.
+		// eslint-disable-next-line eqeqeq
+		if ( ( context ? context.ownerDocument || context : preferredDoc ) != document ) {
```
mgol: This line is the only place where I've seen IE/Edge test failures. I though, though, that it may be worth changing all document comparisons to make sure we're not missing some potential errors.

I could also create an internal utility function to compare documents & put the comment there. That would most likely increase the size, though (currently this PR is at -5 bytes).

mgol:
I'll submit a Sizzle issue after we have consensus here.

src/selector.js
mgol: A version with an utility function would be +15 bytes instead of -5: https://github.com/mgol/jquery/commit/53d084000bdb1e68bf78312781a72e2e0db4379c.

src/selector.js
```diff
@@ -425,7 +429,11 @@ function setDocument( node ) {
 		doc = node ? node.ownerDocument || node : preferredDoc;
 
 	// Return early if doc is invalid or already selected
-	if ( doc === document || doc.nodeType !== 9 ) {
+	// Support: IE 11+, Edge 17 - 18+
+	// IE/Edge sometimes throw a "Permission denied" error when strict-comparing
+	// two documents; shallow comparisons work.
```
mgol: If we want to go for this solution, do you think we should avoid this duplicate comment in some way?

mgol:
What's funny is that I can reproduce the issue easily (not 100% of times but pretty often) in IE 11 & Edge 18 & I assume by TestSwarm failures that it's reproducible in Edge 17 as well - but I cannot reproduce it in IE 9 or 10. Seems they regressed?

timmywil:
I still approve.

PR #4478, opened 2019-09-08

dmethvin:
An upcoming release of Migrate will generate warnings for calls to `.css()` that pass numbers rather than strings, see https://github.com/jquery/jquery-migrate/issues/296 . At the moment, core's `.offset()` setter passes numbers rather than `px` strings so it would throw warnings.

This is the dumbest possible implementation I could think of (and `+18`)  but there are several concerns I have with trying to optimize this code. First, there is an undocumented `using` option that is used by jQuery UI. Second, it's possible for options other than `top` and `left` to be passed through to `.css()` even though that's also undocumented; I don't know if anyone does that.

For the purposes of removing this warning for 3.x I want to just fix it so that Migrate can avoid spurious warnings, and not cause any other problems. For 4.x we could try something fancier.

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -64,6 +64,12 @@ jQuery.offset = {
 			options.using.call( elem, props );
 
 		} else {
+			if ( typeof props.top === "number" ) {
+				props.top += "px";
+			}
+			if ( typeof props.left === "number" ) {
+				props.left += "px";
+			}
 			curElem.css( props );
 		}
 	}


dmethvin:
@mgol thoughts on how to solve this? We will need some change in core if we want to give a warning in Migrate.

mgol:
Hmm, I don’t recall me closing this PR & GitHub says I did it 4 days ago...

Let me think about it.

dmethvin:
I think this may have been automatically closed by GitHub when the branch
was renamed? I can submit the PR against master.

>

mgol:
Ah, that's right, all PRs against `3.4-stable` must have been auto-closed. I don't see any other recent closed PRs that were submitted against that branch so you must have been the only victim; sorry! I should have checked.

Please resubmit against `3.x-stable` then.

PR #4485, opened 2019-09-16
Core: Implement .even() & .odd() to replace POS :even & :odd

mgol:
### Summary ###

Implement .even() & .odd() to replace POS :even & :odd. The current version adds 28 bytes; it'd be slightly smaller (+25 bytes) to rely on `.filter()` but that'd make `core` depend on `traversing/findFilter`.

### Checklist ###

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -100,6 +100,18 @@ jQuery.fn = jQuery.prototype = {
 		return this.eq( -1 );
 	},
 
+	even: function() {
+		return this.filter( function( i ) {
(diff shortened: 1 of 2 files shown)


mgol:
@jquery/core If we want to make the migration away from POS easier, it might be worth releasing the next jQuery as `3.5.0` instead of `3.4.2` with `.odd()`/`.even()` implemented. Thoughts?

dmethvin:
This definitely will make migration easier.

PR #4508, opened 2019-10-10

dmethvin:
An upcoming release of Migrate will generate warnings for calls to .css() that pass numbers rather than strings, see jquery/jquery-migrate#296 . At the moment, core's .offset() setter passes numbers rather than px strings so it would throw warnings.

This is the dumbest possible implementation I could think of (and `+18`) but there are several concerns I have with trying to optimize this code. First, there is an undocumented `using` option that is used by jQuery UI. Second, it's possible for options other than top and left to be passed through to .css() even though that's also undocumented; I don't know if anyone does that.

For the purposes of removing this warning for 3.x I want to just fix it so that Migrate can avoid spurious warnings, and not cause any other compat problems. For 4.x we could try something fancier.

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -63,6 +63,12 @@ jQuery.offset = {
 			options.using.call( elem, props );
 
 		} else {
+			if ( typeof props.top === "number" ) {
+				props.top += "px";
+			}
+			if ( typeof props.left === "number" ) {
+				props.left += "px";
+			}
 			curElem.css( props );
 		}
 	}


src/offset.js
```diff
@@ -63,6 +63,12 @@ jQuery.offset = {
 			options.using.call( elem, props );
 
 		} else {
+			if ( typeof props.top === "number" ) {
+				props.top += "px";
```
mgol: Why not append `"px"` at the point of definition instead? That should be smaller.

mgol: Does UI expect to have bare numbers here?

dmethvin: Yes, unfortunately. We could redefine `.offset()` to take non-px values but that seems like a bigger change with even more consequences. On the `.css()` API it seems more reasonable to insist on strings with units but here the numbers are always supposed to represent pixels. If we require `.offset()` to give strings with `px` appended then I can see people wanting to use arbitrary units.

mgol: Right, makes sense. I'd experiment whether something like `[ "top", "left" ].forEach(...)` isn't smaller but if it's not, LGTM, especially for jQuery 3.x.

We could try to clean that up for 4.0, just please, before merging this one, create an issue with the "Blocker" label & the `4.0.0` milestone so that we don't forget.

timmywil:
Fixes https://github.com/jquery/jquery/issues/4525

src/offset.js
mgol: Timmy created an issue so I have no further reservations.

mgol:
A slightly smaller version for `master`: #4651.

Issue #4435, opened 2019-07-16
Bug of find element with empty attribute in v3.4


hirofumii:
### Description ###
Hi.
After updating jQuery to 3.4, I encountered strange behavior on Edge and IE11.

Finding elements with empty attributes was successful in v3.3.
But, v3.4 and later will fail on Edge and IE11.

```html
<div id="test">
    <input name="">
</div>
```

```js
var $input = $("#test").find("[name='']");

console.log($input.length); // This will return 0, only on Edge and IE11 after updating to v3.4
```

### Link to test case ###
https://jsbin.com/jitatiduqi/1/edit?html,css,js,output

If this is a bug, I'm glad if a fix is ​​considered.
Thanks.

timmywil:
We should look into the behavior change, but it's not surprising as Edge and IE often fill in undefined attributes with empty string, making them difficult to differentiate.

dmethvin:
Probably related to #3550

mgol:
I added an update at #3550. @hirofumii can you check if your issue still exists with https://code.jquery.com/jquery-git.js instead of your current copy of jQuery?

timmywil:
I checked this against jquery-git in IE11 on browserstack and it's still failing.

timmywil:
Ok, I figured out the difference. It seems it's due to [a recent optimization in Sizzle](https://github.com/jquery/sizzle/commit/71fe25c4c9a10d8827b845d398ca8fb7da163749) where we tried to avoid manipulating the DOM when it wasn't necessary. Turns out you get the right result if you add the ID of the context to the selector: https://jsbin.com/vasufuv/1/edit?html,js,console,output

hirofumii:
Thanks, @mgol.

I tested it in my environment just in case, but unfortunately the problem was not solved.
Apparently there is another cause (surely Sizzle?).

timmywil:
@hirofumii See my comment above. It is indeed Sizzle.

hirofumii:
@timmywil Oh, sorry! I saw the comment and intended to add 😄, but not...

mgol:
Sizzle PR: https://github.com/jquery/sizzle/pull/450

mgol:
The PR linked above is not one for this issue.

mgol:
[...]

PR #4510, opened 2019-10-13
Selector: Make empty attribute selectors work in IE again

mgol:
### Summary ###

qSA in IE often (but not always) doesn't find elements with an empty name
attribute selector (`[name=""]`). ~~Assigning a temporary name attribute to
the document root (& removing it afterwards) seems to resolve the issue.~~
EDIT: Detect that & fall back to Sizzle traversal.

Fixes gh-4435

### Checklist ###


mgol:
+21 bytes

This only fixes the issue in IE, not Edge. See the comment at https://github.com/jquery/jquery/issues/4435#issuecomment-541442015 for the rationale.

mgol:
We could also accept the perf hit & assign such a temporary name attribute in all browsers, at least once. Hopefully that wouldn't cause style recalculation as the final value would be the same as the initial one, worth checking.

src/selector.js
```diff
@@ -436,6 +437,15 @@ function setDocument( node ) {
 	documentElement = document.documentElement;
 	documentIsHTML = !jQuery.isXMLDoc( document );
 
+	// Support: IE 11+
+	// IE 11 doesn't find elements on a `[name='']` query in some cases.
+	// Adding a temporary attribute to the document before the selection works
+	// around the issue.
+	if ( isIE && documentElement ) {
+		documentElement.setAttribute( "name", "" );
```
mgol: I guess I should check if there's such an attribute already set and - if it is - restore it afterwards. `documentElement` with a `name` attribute would be quite a weird thing to do but it's not impossible.

timmywil: I wonder if the problem is already resolved if the `documentElement` already has a `name` attribute.

mgol: @timmywil It looks like the problem doesn't exist if an _empty_ name attribute is already present on the document. However, a non-empty attribute doesn't help so this is all needed.

mgol:
PR updated.

I build a test case similar to https://jsbin.com/cocaqis/1/edit?html,js but with one iframe being available immediately in the DOM (with similar DOM contents) and another similar one being added after a few seconds. If there was no workaround, all of them failed. If I added a workaround in the main doc, all succeeded. If I added one in the first iframe, the selection in the main DOM failed & then the selection in both iframes succeeded. If I added it to the second one, it was failing until the selection happened in that second iframe, then all started suceeding.

Also, if I see success, after refreshing the page (with "Always refresh from server" checked in the Network DevTools tab) I see every selection succeeding as well so the workaround even survives page refreshes.

Of course, IE being IE, it's not impossible that the workaround stops working after a while but for now I think we can work on the assumption it survives long enough. If that proves not enough, we'll depend on someone submitting an issue.

mgol:
+30 bytes, BTW.

mgol:
Sizzle version of the PR: https://github.com/jquery/sizzle/pull/461

mgol:
PR updated. `[name='']` now falls back to Sizzle traversal & both IE 11 & Edge 15-18 pass all tests. +49 bytes.

src/selector/rbuggyQSA.js
```diff
@@ -23,6 +25,18 @@ if ( isIE ) {
...
 
+// Support: IE 11+, Edge 15 - 18+
+// IE 11/Edge don't find elements on a `[name='']` query in some cases.
+// Adding a temporary attribute to the document before the selection works
+// around the issue.
+// Interestingly, IE 10 & older don't seem to have the issue.
+input.setAttribute( "name", "" );
+testEl.appendChild( input );
+if ( !testEl.querySelectorAll( "[name='']" ).length ) {
+	rbuggyQSA.push( "\\[" + whitespace + "*name" + whitespace + "*=" +
```
gibson042: Does the `=` here also need to cover other operators (e.g., `[*^$]?=`)? Cf. checks for [empty strings after ^= or $= or *=](https://github.com/jquery/sizzle/blob/a528ae520245b13a6a879041e4a2b135c7d6e522/src/sizzle.js#L816) and [case-sensitivity of name attribute](https://github.com/jquery/sizzle/blob/a528ae520245b13a6a879041e4a2b135c7d6e522/src/sizzle.js#L858).

src/selector/var/whitespace.js
```diff
@@ -0,0 +1,6 @@
+define( function() {
+	"use strict";
+
+	// http://www.w3.org/TR/css3-selectors/#whitespace
```
gibson042: Might as well make this HTTPS now.

mgol: Nice catch. I fixed the URL in the other PR but since this declaration has moved I could have missed that URL.

src/selector/rbuggyQSA.js
mgol: Thanks, I'll have a look. Unfortunately, I merged the Sizzle PR before seeing this comment so it might need a followup PR later.

mgol: @gibson042 What's the meaning of `!` inside of a regex there? Native CSS selectors don't support negated attribute selector `[name!=value]` so they'd go the Sizzle route anyway.

gibson042: > @gibson042 What's the meaning of `!` inside of a regex there? Native CSS selectors don't support negated attribute selector `[name!=value]` so they'd go the Sizzle route anyway.

Looks like it's there to make a longer repeat of https://github.com/jquery/sizzle/blob/a528ae520245b13a6a879041e4a2b135c7d6e522/src/sizzle.js#L87 for gzip.

mgol: @gibson042 I checked the selectors you mentioned, the issue is all of them (`*=`, `^=`, `$=`) require the matched string to be non-empty to return any result. Therefore, all browsers return no results for them so there's nothing to workaround for IE here.

PR #4513, opened 2019-10-14
CSS: Reduce css/support.js

gibson042:
Ref gh-4506

### DIFF
--- a/src/css/support.js
+++ b/src/css/support.js
@@ -1,8 +1,9 @@
 define( [
+	"../core",
 	"../var/document",
 	"../var/documentElement",
 	"../var/support"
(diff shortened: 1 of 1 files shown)


mgol:
This implementation fails the CSP check as you can see on Travis. I had similar issues when trying to optimize it.

Note that this support test will most likely go away before we release jQuery 4.0 as I expect we'll drop EdgeHTML support by then (although, I cannot be 100% sure). But if we're able to optimize it, we could backport the optimization to the `3.x-stable` branch where it won't go away.

mgol:
@gibson042 Is this good to close or do you have any other idea for how to make this size optimization work?

PR #4095, opened 2018-06-10
Remove unnecessary eslint exception

edg2s:
The linked-to issue was fixed 2 years ago.

### Summary ###
Removes an old eslint exception and comment explaining it

### Checklist ###

* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [ ] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #4520, opened 2019-10-17
Build: Enable reportUnusedDisableDirectives in ESLint

edg2s:
### Summary ###
The reportUnusedDisableDirectives option in ESLint will report an error when in inline disable directive is no longer required, such as what happened gh-4095.

### Checklist ###

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -79,7 +79,8 @@ module.exports = function( grunt ) {
 			options: {
 
 				// See https://github.com/sindresorhus/grunt-eslint/issues/119
-				quiet: true
+				quiet: true,
+				reportUnusedDisableDirectives: true
 			},
 
 			// We have to explicitly declare "src" property otherwise "newer"
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -79,8 +79,7 @@ module.exports = function( grunt ) {
 			options: {
 
 				// See https://github.com/sindresorhus/grunt-eslint/issues/119
-				quiet: true,
-				reportUnusedDisableDirectives: true
+				quiet: true
 			},
 
 			// We have to explicitly declare "src" property otherwise "newer"
--- a/.eslintrc-browser.json
+++ b/.eslintrc-browser.json
@@ -3,6 +3,8 @@
 
 	"extends": "jquery",
 
+	"reportUnusedDisableDirectives": true,
+
 	"parserOptions": {
 		"ecmaVersion": 5
 	},
--- a/.eslintrc-node.json
+++ b/.eslintrc-node.json
@@ -3,6 +3,8 @@
 
 	"extends": "jquery",
 
+	"reportUnusedDisableDirectives": true,
+
 	"parserOptions": {
 		"ecmaVersion": 2018
 	},


Gruntfile.js
```diff
@@ -77,7 +77,8 @@ module.exports = function( grunt ) {
 			options: {
 
 				// See https://github.com/sindresorhus/grunt-eslint/issues/119
-				quiet: true
+				quiet: true,
+				reportUnusedDisableDirectives: true
```
mgol: Is it possible to configure in the ESLint configuration file instead of here?

mgol:
I like it. One question.

mgol:
@edg2s Do you want to update it or should we take over?

edg2s:
Adding reportUnusedDisableDirectives in the config was only added in ESLint 6.3.0. This repo currently only requires version 6.0.1 (via grunt-eslint 22.0.0).

mgol:
@edg2s We can update ESLint if that helps, I don’t mind including it in a PR.

mgol:
@edg2s You said before this doesn't work with the currently used ESLint version. Did anything change?

edg2s:
Sorry, I assumed you would update ESLint separately. I can do it here though.

mgol:
@edg2s I've noticed `grunt-eslint` has been already updated on `master` in November, it's just that your branch is based on an extremely old version of `master`. Can you please rebase over latest `master`?

edg2s:
Done.

mgol:
I merged the PR to both `master` & `3.x-stable` and, unfortunately, only afterwards I noticed:
1. This doesn't error but only print warnings; those warnings are silenced by `grunt eslint` and they get only printed if you run `node_modules/.bin/eslint .` directly.
1. there's one `eslint-disable` comment that is legitimate but we still want to keep it - it's above the [`version` definition](https://github.com/jquery/jquery/blob/46f9810b73a7ad446d7c3711faf92f56b67df3c1/src/core.js#L18-L19) because in some custom compilations that string may get large and we validate the build jQuery file with ESLint as well.

I'm not sure how to proceed here due to the above. We can catch some things through these warnings but only when calling the ESLint binary directly and we have to ignore at least this one `max-len` warning.

dmethvin:
Seems to be a common-but-not-enough request: https://github.com/eslint/eslint/issues/12703

The workaround they suggest is `--max-warnings 0` assuming we don't get warnings we want to ignore.

edg2s:
A couple of options:
1. Remove the directive, and disable reportUnusedDisableDirectives on `grunt eslint:dist`
```js
dist: {
	options: {
		reportUnusedDisableDirectives: false
	},
	src: [ "dist/jquery.js", "dist/jquery.min.js" ]
},
dev: {
	src: [ "src/**/*.js", "Gruntfile.js", "test/**/*.js", "build/**/*.js" ]
},

```
2. Have the build script that swaps out @VERSION insert the directive (and only if the resulting line is longer than 100 chars).

mgol:
@dmethvin Their reasoning sounds reasonable. We could work with it only if we pinned our ESLint version so that patch updates don't make our build fail. Currently we have it pinned via `package-lock.json` but I'd rather have it explicit in the `package.json` if possible.

@edg2s I'd go for the first option; way smaller complexity. This doesn't resolve the issue that `grunt eslint` doesn't print the warnings but we can leave that for a separate PR.

edg2s:
Yeah, you would have to use it in combination with max-warnings and version pinning.

edg2s:
Another issue is that grunt-eslint is giving use warnings for files that listed in `.eslintignore`. In the upstream issue there are some solutions that involve parsing the ignore file, but that can be done more cleanly by just fetching the parsed results from CLIEngine:

```js
src: [ "src/**/*.js", "Gruntfile.js", "test/**/*.js", "build/**/*.js" ]

	// Ignore files from .eslintignore
	// See https://github.com/sindresorhus/grunt-eslint/issues/119
	.concat(
		new CLIEngine()
			.getConfigForFile( "Gruntfile.js" )
			.ignorePatterns.map( p => "!" + p )
	)
```

Issue #4529, opened 2019-10-22
Firefox 70 fails the "dimensions: width/height on a table row with phantom borders (gh-3698)" test


mgol:
### Description ###

The just released Firefox 70 (& newer) fails the "dimensions: width/height on a table row with phantom borders (gh-3698)" test. As Travis uses the newest version of Firefox, it started failing on `master`.

### Link to test case ###

See the relevant test case in the jQuery repo, a screenshot:
<img width="633" alt="Screen Shot 2019-10-22 at 20 00 16" src="https://user-images.githubusercontent.com/1758366/67315319-9f232400-f506-11e9-9296-26fc695cd072.png">

timmywil:
Just a note, we keep track of swarm failures in the meeting notes. Those tests can be in flux a lot between releases and I don't think we need to create a GH issue for each failure. All failures are always addressed before a release can go out.

mgol:
This doesn't fail on Swarm as we don't have Firefox 70 there yet; it fails on Travis as I've recently added Firefox testing there & Firefox 70 has been released today.

I think these issues make sense; they e.g. allow to assign an issue to a particular person, they also make it easier to distinguish flakes from consistent issues (this is the latter case). It looks like a real issue to me, it fails on stable, beta & nightly and I meant to look at it tomorrow.

mgol:
To clarify, I'm not saying we have to create issues for all those TestSwarm failures. But we can if we feel the need IMO.

mgol:
One additional reason to allow issue creation in such cases is that this creates a space to dump the investigation results which I plan to do soon.

mgol:
Results from my investigation:
The failing test case was added in https://github.com/jquery/jquery/pull/3738/files; it works on the following HTML:

```html
<table id="gh3698" style="border-collapse: separate; border-spacing: 0; background: #baffc9"><tbody>
	<tr style="margin: 0; border: 10px solid black; padding: 0">
		<td style="margin: 0; border: 0; padding: 0; height: 42px; width: 42px;"></td>
	</tr>
</tbody></table>
```

[...]

PR #4536, opened 2019-10-27
Tests: Don't test synchronous XHR on unload in Chrome

mgol:
### Summary ###

Chrome 78 dropped support for synchronous XHR requests inside of beforeunload, unload, pagehide, and visibilitychange event handlers. This made an AJAX test checking exactly that fail in Chrome.
See https://bugs.chromium.org/p/chromium/issues/detail?id=952452

BTW, I can't reliably reproduce the failure in Chrome on macOS when tested via the PHP setup. It does fail in Chrome 78 via Karma & in Edge Chromium in all ways, though. It also reliably fails in TestSwarm. Therefore, we need to drop that test in Chrome.

### Checklist ###

### DIFF
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -2001,14 +2001,19 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
 		};
 	} );
 
-	testIframe(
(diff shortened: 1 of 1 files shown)


mgol:
Firefox tests fail due to #4529 but this PR makes them succeed in Chrome again.

PR #4541, opened 2019-11-14
Core: Migrate from AMD to ES modules 🎉

mgol:
### Summary ###

Migrate source to ES modules. RequireJS is still used to load tests, this may perhaps be changed in a separate PR.

Reviewing by commit might be the easiest.

@timmywil I'd appreciate especially a review of the build process after my changes as custom compilation works slightly differently now (for example, I had to add an explicit import of `./core/parseHTML.js` to `jquery.js` as otherwise excluding AJAX cut out that module as well and a lot depends on it). If you could locally test some scenarios you had in mind when you initially wrote that & make sure they still work, that'd help.

TODO:
- [ ] Add AMD compilation to `amd/`
- [ ] (optional) Add back AMD mode to QUnit

### Checklist ###


test/jquery.js
```diff
@@ -21,12 +22,13 @@
...
 
-		// AMD loading is asynchronous and incompatible with synchronous test loading in Karma
+		// ES modules loading is asynchronous and incompatible with synchronous
+		// test loading in Karma.
 		if ( !window.__karma__ ) {
 			QUnit.config.urlConfig.push( {
-				id: "amd",
-				label: "Load with AMD",
-				tooltip: "Load the AMD jQuery file (and its dependencies)"
+				id: "jsmodules",
```
mgol: What do you think about this parameter name?

timmywil: Perhaps "esmodules" to make it more clear that the modules are standard es format.

mgol: Renamed.

mgol:
This PR now adds 89 bytes to the minified (non-gzipped) build & 5 bytes to the minified gzipped build. The gzipped difference mostly comes from the addition of the `./core/parseHTML.js` import to `jquery.js`, without it it added 36 bytes to minified (non-gzipped) and actually reduced (!) the minified gzipped build by 10 bytes. This is only a result of re-shuffling the final file, apparently Uglify was compiling the previous one better.

I'm OK with such a delta given the gains from using the modern module format & removing our brittle compile process hacks as it's now mostly relying on Rollup but maybe it's possible to optimize it a little.

mgol:
OK, apparently we can go back to -10 bytes minified gzipped (& minimal unminified diff from the current Git compilation) if we move the `./core/parseHTML.js` from directly after the `./core.js` import to just before the `./ajax/load.js` one.

src/ajax.js
```diff
@@ -615,7 +611,8 @@ jQuery.extend( {
 			// Add or update anti-cache param if needed
 			if ( s.cache === false ) {
 				cacheURL = cacheURL.replace( rantiCache, "$1" );
-				uncached = ( rquery.test( cacheURL ) ? "&" : "?" ) + "_=" + ( nonce++ ) + uncached;
+				uncached = ( rquery.test( cacheURL ) ? "&" : "?" ) + "_=" +
+					( nonce.nonce++ ) + uncached;
```
gibson042: Assuming the `nonce` property name doesn't matter, you might get a little more compression by renaming it (e.g., `nonce.value++` or `nonce.data++` or `nonce.length++` or `nonce.expando++` or `nonce.guid++`).

mgol: Thanks, I'll try.

mgol: `guid` reduces size differences from `-10` bytes to `-12` bytes, I'll use that.

mgol:
@rart This is planned for jQuery 4.0.0 which will take at least several more months. Migrating this to the `3.x-stable` line while preserving backwards compatibility of the package structure might require non-trivial amount of work so at the moment we don't have plans to do it.

rart:
@mgol thanks for the quick reply.

Have you guys considered an `4.0.0.rc` sort of release/publish? (alpha, beta, rc)
In our case, we're in the need of the ES modules formats since, we use rollup and in the latest `rollup`/`@rollup/plugin-*`, jQuery bundling is throwing so it'd be great to be able to pull this. I assume this will also help with tree shaking.

Slightly off topic, but is there a recommended or sample rollup config for bundling jQuery (3.4.1)? Using the `namedExports` property with the `commonjs` plugin seems not to be helping.

mgol:
@rart There is still significant work planned for jQuery 4.0.0 so we don't plan any pre-release soon; it'd be too far from the final release. If you want to use ES modules from `master`, you can always copy our `src` directory. Just remember it's not stable so you'll need to be careful with upgrades.

Issue #4397, opened 2019-05-14
Bug with .focus and text nodes in 3.4


ambischof:
### Description ###
There's a new breaking change in 3.4 when you attempt to call `.focus(fn)` on a selector containing a text node.

Consider the following: 
```
$('<span>foo</span>hi<span>bar</span>').focus(function() {})
``` 
in v3.3.1 this is all fine, but in v3.4.0 and v3.4.1 you get 
```Uncaught RangeError: Maximum call stack size exceeded```

Of note, no one would do this on purpose, as it doesn't make sense, and even this example is a bit contrived (actual case I encountered at least had form elements), but given various libraries out of my control, it managed to happen for me. 

This is the case in question: https://github.com/gitana/alpaca/issues/705

This is clearly outside the boundaries of normal usage, however, is it possible for the `.focus(fn)` could have some checks to not fail quite so critically?

According to the [jQuery function docs](https://api.jquery.com/jQuery/#jQuery2) it _sounds_ like text nodes should generally be tolerated?
> When passing HTML to jQuery(), note that text nodes are not treated as DOM elements. With the exception of a few methods (such as .content()), they are generally ignored or removed. E.g:

### Link to test case ###
In this jsfiddle with v3.4.1, it produces an error
https://jsfiddle.net/anneb574/r2pambcz/1/ 

but this jsfiddle with v3.3.1 it doesn't
https://jsfiddle.net/anneb574/r2pambcz/3/

timmywil:
What a well-structured issue. Test cases and everything. Thanks!

We can look into whether the call stack error is avoidable. `focus` did see some significant changes in 3.4.

dmethvin:
This is like a holy trinity of nightmare scenarios. There aren't a lot of ways to get a text node into a set, but `$()` and `.contents()` are two. If you call any filtering method after that the text nodes would be thrown out. 

The event system does not want or expect the target of an event to be a text node. It's not supposed to happen, and for the few cases where browsers erroneously did it in the past we [fixed the issue when the browser created it](https://github.com/jquery/jquery/blob/2.2-stable/src/event.js#L478-L482). That fix wouldn't work in this case anyway 1) because those text nodes are orphans and `event.target` would be `null` which also is not a respectable value and 2) manually triggered events go through a different and less realistic path.

I'll also point out that triggering focus on three nodes in succession as done in this test case is pretty strange. When the whole thing is done only one will have focus, I assume it's the last one but the async focus behavior of IE may do something different.

[...]

PR #4558, opened 2019-12-03
Event: Only attach events to objects that accept data - for real

mgol:
### Summary ###

There was a check in jQuery.event.add that was supposed to make it a noop
for objects that don't accept data like text or comment nodes. The problem was
the check was incorrect: it assumed `dataPriv.get( elem )` returns a falsy
value for an `elem` that doesn't accept data but that's not the case - we get
an empty object then. The check was changed to use `acceptData` directly.

Fixes gh-4397

It's funny that this has been the behavior for at least the past few years so this code has been wrong for a long time. 😱 

We'll also need to CP that to `3.x-stable`

+2 bytes

### Checklist ###

### DIFF
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -2811,6 +2811,15 @@ QUnit.test( "preventDefault() on focusin does not throw exception", function( as
 	}, QUnit.config.testTimeout / 4 || 1000 );
 } );
 
+QUnit.test( ".on('focus', fn) on a text node doesn't throw", function( assert ) {
+	assert.expect( 1 );
+
+	jQuery( document.createTextNode( "text" ) )
+		.on( "focus", function() {} );
+
+	assert.ok( true, "No crash" );
+} );
+
 QUnit.test( "Donor event interference", function( assert ) {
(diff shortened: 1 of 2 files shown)


dmethvin:
Wow, this does go back a ways!

test/unit/event.js
```diff
@@ -2811,6 +2811,15 @@ QUnit.test( "preventDefault() on focusin does not throw exception", function( as
 	}, QUnit.config.testTimeout / 4 || 1000 );
 } );
 
+QUnit.test( ".on('focus', fn) on a text node doesn't throw", function( assert ) {
+	assert.expect( 1 );
+
+	jQuery( "<div></div>text<span></span>" )
```
gibson042: I'd prefer something more obvious here such as `jQuery( document.createTextNode("text") )`, but this works as long as the behavior doesn't change (do we test that, though?).

mgol: @gibson042 Your suggestion looks good to me. What do you mean by "as long as the behavior doesn't change", though? What behavior?

Issue #4542, opened 2019-11-15
jQuery does not respect the crossOrigin attribute on the script tag


gaohuia:
Don't know if it is a new issue for you, I just met it in our site.

### Description ###

Just look at the following HTML code.  There are two script tags in the page, one is statically written and the other is added by jQuery.  And they are almost the same except a small difference on the src.

Steps:

1. Put this page on a html server, and get a link to that page.
2. Open your browser, I used chrome v78.
3. New tab and press F12 to open the dev tool.
4. Change to the Network tab.
5. Open the link you got in 1st step.
6. Observe the network.

Expect: Both `jquery-3.4.1.min.js` && `jquery-3.4.1.min.js?cors` should be issued with the http header "Origin"
Result: Only `jquery-3.4.1.min.js` is issued with the "Origin" header, the other one is not.

When this issue happened, the browser caches  the response of `jquery-3.4.1.min.js?cors` which has no CORS headers. And if other pages use this script with SRI, it will be blocked. Because it matches the cache but the cache does not contain a CORS header like "Access-Control-Allow-Origin".

```html
<!doctype html>
<html lang="en">
<head>
    <script src="https://code.jquery.com/jquery-3.4.1.min.js" crossorigin="anonymous"></script>
    <script>
        $(function(){
            $("#container").html('<script src="//code.jquery.com/jquery-3.4.1.min.js?cors" crossorigin="anonymous" ><\/script>');
            alert('done');
        });
    </script>

</head>
<body>
    <div id="container"></div>
</body>
</html>
```

mgol:
It seems we might just need to add support for the [`crossOrigin` property on script elements](https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement) to [src/core/DOMEval.js#L3-L8](https://github.com/jquery/jquery/blob/5a3e0664d261422f11a78faaf101d70c73b3a5a8/src/core/DOMEval.js#L3-L8). If possible, we should also write an iframe test that ensures this fixes the issue you mention.

Would you like to try to submit a PR for that?

gaohuia:
It seems to be related with the transport of "script".
I'll try to make a PR later.

timmywil:
Moving to 4.0 as this is a breaking change.

PR #4563, opened 2019-12-10
Manipulation: Respect script crossorigin attribute in DOM manipulation

gaohuia:
### Summary ###
jQuery does not respect the crossOrigin attribute on the script tag #4542

### Checklist ###

### DIFF
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -63,6 +63,7 @@ jQuery.ajaxTransport( "script", function( s ) {
 		return {
 			send: function( _, complete ) {
 				script = jQuery( "<script>" )
+					.attr( "crossOrigin", s.crossOrigin || null )
 					.attr( s.scriptAttrs || {} )
 					.prop( { charset: s.scriptCharset, src: s.url } )
 					.on( "load error", callback = function( evt ) {
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -63,7 +63,6 @@ jQuery.ajaxTransport( "script", function( s ) {
 		return {
 			send: function( _, complete ) {
 				script = jQuery( "<script>" )
-					.attr( "crossOrigin", s.crossOrigin || null )
 					.attr( s.scriptAttrs || {} )
 					.prop( { charset: s.scriptCharset, src: s.url } )
 					.on( "load error", callback = function( evt ) {
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -158,8 +158,7 @@ function domManip( collection, args, callback, ignored ) {
 							if ( jQuery._evalUrl && !node.noModule ) {
 								jQuery._evalUrl( node.src, {
 									nonce: node.nonce || node.getAttribute( "nonce" ),
-									crossOrigin: node.crossOrigin ||
-										node.getAttribute( "crossOrigin" )
+									crossOrigin: node.crossOrigin
 								}, doc );
 							}
 						} else {
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -158,9 +158,8 @@ function domManip( collection, args, callback, ignored ) {
 							if ( jQuery._evalUrl && !node.noModule ) {
 								jQuery._evalUrl( node.src, {
 									nonce: node.nonce || node.getAttribute( "nonce" ),
-									crossOrigin: node.crossOrigin || node.getAttribute(
-										"crossOrigin"
-									)
+									crossOrigin: node.crossOrigin ||
+										node.getAttribute( "crossOrigin" )
 								}, doc );
 							}
 						} else {
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -157,7 +157,10 @@ function domManip( collection, args, callback, ignored ) {
 							// Optional AJAX dependency, but won't run scripts if not present
 							if ( jQuery._evalUrl && !node.noModule ) {
 								jQuery._evalUrl( node.src, {
-									nonce: node.nonce || node.getAttribute( "nonce" )
+									nonce: node.nonce || node.getAttribute( "nonce" ),
+									crossOrigin: node.crossOrigin || node.getAttribute(
+										"crossOrigin"
+									)
 								}, doc );
 							}
 						} else {
--- a/src/manipulation/_evalUrl.js
+++ b/src/manipulation/_evalUrl.js
@@ -10,7 +10,7 @@ jQuery._evalUrl = function( url, options, doc ) {
 		cache: true,
 		async: false,
 		global: false,
-		crossOrigin: options.crossOrigin || null,
+		scriptAttrs: options.crossOrigin ? { "crossOrigin": options.crossOrigin } : null,
 
 		// Only evaluate the response if it is successful (gh-4126)
 		// dataFilter is not invoked for failure responses, so using it instead
--- a/src/manipulation/_evalUrl.js
+++ b/src/manipulation/_evalUrl.js
@@ -10,6 +10,7 @@ jQuery._evalUrl = function( url, options, doc ) {
 		cache: true,
 		async: false,
 		global: false,
+		crossOrigin: options.crossOrigin || null,
 
 		// Only evaluate the response if it is successful (gh-4126)
 		// dataFilter is not invoked for failure responses, so using it instead
--- a/src/manipulation/_evalUrl.js
+++ b/src/manipulation/_evalUrl.js
@@ -10,7 +10,7 @@ jQuery._evalUrl = function( url, options, doc ) {
 		cache: true,
 		async: false,
 		global: false,
-		scriptAttrs: options.crossOrigin ? { "crossOrigin": options.crossOrigin } : null,
+		scriptAttrs: options.crossOrigin ? { "crossOrigin": options.crossOrigin } : undefined,
 
 		// Only evaluate the response if it is successful (gh-4126)
 		// dataFilter is not invoked for failure responses, so using it instead
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -61,11 +61,11 @@ class MockServer {
 
 		if ( !empty( $req->query['callback'] ) ) {
 			$headers = array_combine(
-				array_map( 'strtolower', array_keys( $req->headers ) ), 
+				array_map( 'strtolower', array_keys( $req->headers ) ),
 				array_values( $req->headers )
 			);
 
-			echo "{$req->query['callback']}(" . json_encode( $req->headers ) . ")";	
+			echo "{$req->query['callback']}(" . json_encode( $req->headers ) . ")";
 		} else {
 			echo 'QUnit.assert.ok( true, "mock executed" );';
 		}
@@ -246,17 +246,6 @@ QUnit.assert.ok( true, "mock executed");';
 			echo $callback . '( {"status": 404, "msg": "Not Found"} )';
 		} else {
 			echo 'QUnit.assert.ok( false, "Mock return erroneously executed" );';
-		}
-	}
-
-	protected function cors( $req ) {
-		if ( isset( $req->headers['ORIGIN'] ) ) {
-			$origin = $req->headers['ORIGIN'];
-			header( "Access-Control-Allow-Methods: GET" );
-			header( "Access-Control-Allow-Origin: " . $origin );
-			echo 'corsCallback( true )';
-		} else {
-			echo 'corsCallback( false )';
 		}
 	}
 
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -235,14 +235,14 @@ QUnit.assert.ok( true, "mock executed");';
 		}
 	}
 
-	protected function cors($req) {
-		if (isset($req->headers['ORIGIN'])) {
+	protected function cors( $req ) {
+		if ( isset( $req->headers['ORIGIN'] ) ) {
 			$origin = $req->headers['ORIGIN'];
-			header("Access-Control-Allow-Methods: GET");
-			header("Access-Control-Allow-Origin: " . $origin);
-			corsCallback(true);
+			header( "Access-Control-Allow-Methods: GET" );
+			header( "Access-Control-Allow-Origin: " . $origin );
+			echo 'corsCallback( true )';
 		} else {
-			corsCallback(false);
+			echo 'corsCallback( false )';
 		}
 	}
 
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -235,6 +235,17 @@ QUnit.assert.ok( true, "mock executed");';
 		}
 	}
 
+	protected function cors($req) {
+		if (isset($req->headers['ORIGIN'])) {
+			$origin = $req->headers['ORIGIN'];
+			header("Access-Control-Allow-Methods: GET");
+			header("Access-Control-Allow-Origin: " . $origin);
+			corsCallback(true);
+		} else {
+			corsCallback(false);
+		}
+	}
+
 	public function __construct() {
 		$this->cspFile = __DIR__ . '/support/csp.log';
 	}
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -65,7 +65,7 @@ class MockServer {
 				array_values( $req->headers )
 			);
 
-			echo "{$req->query['callback']}(" . json_encode( $req->headers ) . ")";
+			echo "{$req->query['callback']}(" . json_encode( [ 'headers' => $req->headers ] ) . ")";
 		} else {
 			echo 'QUnit.assert.ok( true, "mock executed" );';
 		}
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -65,7 +65,7 @@ class MockServer {
 				array_values( $req->headers )
 			);
 
-			echo "{$req->query['callback']}(" . json_encode( [ 'headers' => $req->headers ] ) . ")";
+			echo "{$req->query['callback']}(" . json_encode( [ 'headers' => $headers ] ) . ")";
 		} else {
 			echo 'QUnit.assert.ok( true, "mock executed" );';
 		}
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -65,7 +65,7 @@ class MockServer {
 				array_values( $req->headers )
 			);
 
-			echo "{$req->query['callback']}(" . json_encode( [ 'headers' => $headers ] ) . ")";
+			echo $req->query['callback'] . "(" . json_encode( [ 'headers' => $headers ] ) . ")";
 		} else {
 			echo 'QUnit.assert.ok( true, "mock executed" );';
 		}
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -54,7 +54,21 @@ class MockServer {
 		} else {
 			header( 'Content-type: text/html' );
 		}
-		echo 'QUnit.assert.ok( true, "mock executed" );';
+
+		if ( !empty( $req->query['cors'] ) ) {
+			header( "Access-Control-Allow-Origin: *" );
+		}
+
+		if ( !empty( $req->query['callback'] ) ) {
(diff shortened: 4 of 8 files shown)


mgol:
Thanks for the PR, I'll have a closer look later. In the meantime, please sign our CLA, this is a requirement for us moving forwards with the PR.

src/ajax/script.js
```diff
@@ -46,6 +46,7 @@ jQuery.ajaxTransport( "script", function( s ) {
 		return {
 			send: function( _, complete ) {
 				script = jQuery( "<script>" )
+					.attr( "crossOrigin", s.crossOrigin || null )
 					.attr( s.scriptAttrs || {} )
```
jaubourg: That's what the `scriptAttrs` option if for. You can add any attribute in there with no need to introduce a new option. It was designed specifically to avoid option-creep.

The correct route would be to add `crossOrigin` into the `scriptAttrs` object.

gaohuia: Updated.

src/manipulation.js
```diff
@@ -162,7 +162,9 @@ function domManip( collection, args, callback, ignored ) {
 							// Optional AJAX dependency, but won't run scripts if not present
 							if ( jQuery._evalUrl && !node.noModule ) {
 								jQuery._evalUrl( node.src, {
-									nonce: node.nonce || node.getAttribute( "nonce" )
+									nonce: node.nonce || node.getAttribute( "nonce" ),
+									crossOrigin: node.crossOrigin ||
```
mgol: I don't think you need to read both the attribute & the property. Support for both seems similar:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script
https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement
(look at the "Browser compatibility" tables below).

The reason why for `nonce` we needed both the property & the attribute is that in Chrome (& latest standards) `getAttribute( "nonce" )` doesn't return an up to date `nonce` value:
https://github.com/whatwg/html/issues/2369
https://html.spec.whatwg.org/#nonce-attributes
but, at the same time, the `nonce` property is not supported in Firefox & Edge so we need the `getAttribute` fallback for them.

I'd just leave `crossOrigin: node.crossOrigin` here. But please make sure the tests still pass.

gaohuia:
Thanks for your explaination.

test/unit/core.js
```diff
@@ -197,6 +197,34 @@ QUnit.test( "globalEval execution after script injection (#7862)", function( ass
...
+	var done = assert.async();
+	assert.expect( 1 );
+
+	var timeout = null;
+
+	Globals.register( "corsCallback" );
+	window.corsCallback = function( response ) {
+		assert.ok( response === true, "Cors" );
+
+		if ( timeout != null ) {
```
mgol: The `if` is not necessary, `clearTimeout` on `null` is a no-op.

test/unit/core.js
```diff
@@ -197,6 +197,34 @@ QUnit.test( "globalEval execution after script injection (#7862)", function( ass
...
+			window.clearTimeout( timeout );
+		}
+
+		done();
+	};
+
+	var src = baseURL + "mock.php?action=cors";
+	src = src.replace( "localhost", "127.0.0.1" );
```
mgol: Is this simulating a cross-origin request by fetching a script from `127.0.0.1` at a `localhost` host? If that's the case, please add a comment above this line (with an empty line preceeding the comment).

test/unit/core.js
```diff
@@ -197,6 +197,34 @@ QUnit.test( "globalEval execution after script injection (#7862)", function( ass
...
+		assert.ok( response === true, "Cors" );
+
+		if ( timeout != null ) {
+			window.clearTimeout( timeout );
+		}
+
+		done();
+	};
+
+	var src = baseURL + "mock.php?action=cors";
```
mgol: We typically put all var declarations near the top of the function under one `var` keyword, one declaration per line if immediately assigned, unassigned vars all in one line (the first one). Please see other tests for comparison.

Here I'd expect `assert.expect( 1 );` to be the first line, then an empty line & then all vars declared with one var statement.

test/unit/core.js
```diff
@@ -197,6 +197,34 @@ QUnit.test( "globalEval execution after script injection (#7862)", function( ass
...
+	src = src.replace( "localhost", "127.0.0.1" );
+	var html = "<script type=\"text/javascript\" src=\"" + src + "\" crossOrigin=\"anonymous\"><\/script>";
+
+	jQuery( document.body ).append( html );
+	timeout = window.setTimeout( function() {
+		assert.ok( false, "Cors" );
```
mgol: This assertion fires for me in Safari 13, Edge 18 & IE 11. It correctly doesn't fire only in Firefox 72 & Chrome 79.

We need to figure out why it fails in other browsers. For IE we might have to skip the test as it doesn't support the crossorigin attribute (by switching from `QUnit.test` to `QUnit.testUnlessIE`). Other browsers should pass it, though.

test/unit/core.js
```diff
@@ -197,6 +197,34 @@ QUnit.test( "globalEval execution after script injection (#7862)", function( ass
...
+		if ( timeout != null ) {
+			window.clearTimeout( timeout );
+		}
+
+		done();
+	};
+
+	var src = baseURL + "mock.php?action=cors";
+	src = src.replace( "localhost", "127.0.0.1" );
+	var html = "<script type=\"text/javascript\" src=\"" + src + "\" crossOrigin=\"anonymous\"><\/script>";
```
mgol: HTML is case-insensitive, so:
```suggestion
	var html = "<script type=\"text/javascript\" src=\"" + src + "\" crossorigin=\"anonymous\"><\/script>";
```

mgol:
@gaohuia Your PR now contains my cherry-picked PRs... Can you instead rebase your PR over current master so that it only contains your code? My commits should disappear then:
```
git fetch --all --tags -p
git rebase origin/master
```
If your upstream is named differently than `origin`, substitute the name.

In the future, it's a good idea to not modify code on the `master` branch but keep `master` synced with upstream instead and use a different branch for PRs. See this link for more info:
https://contribute.jquery.org/commits-and-pull-requests/#never-commit-on-master

This is for the future sake only, let's keep this PR as it is as you can't change the source branch when the PR is open.

mgol:
@gaohuia You can see that jQuery Slim tests failed on your PR (see the Travis link). Your unit test should be moved to `unit/ajax.js` where tests related to the ajax module should lie. Then that test will be excluded for the Slim build & it should pass again.

Also, remember about https://github.com/jquery/jquery/pull/4563#discussion_r364368782 - we still need to get the tests to succeed in IE, Edge & Safari.

gaohuia:
Hi @mgol I didn't see any failed tests in safari 13, can you try again?

mgol:
@gaohuia Yes, Safari 13 is still failing. Note that Travis only runs tests on Chrome & Firefox, you need to test on the other browsers manually.

See e.g. Safari 13:
Screen Shot 2020-01-17 at 23 03 56
and Edge 18:
Screen Shot 2020-01-17 at 23 09 56

mgol:
Hmm, the above test results were achieved in our PHP setup, the one that runs on our TestSwarm as well: http://swarm.jquery.org/project/jquery

When I run it on the Karma setup, by running `grunt karma:chrome-debug`, opening http://localhost:9876 in Safari & clicking "DEBUG", all tests pass.

mgol:
However, in Chrome & Firefox tests pass in both modes.

test/unit/ajax.js
```diff
@@ -2840,6 +2840,32 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
 		} );
 	} );
 
+	QUnit.test( "jQuery.append with crossOrigin attribute", function( assert ) {
```
mgol: The attribute is lowercase so:
```suggestion
	QUnit.test( "jQuery.append with crossorigin attribute", function( assert ) {
```

test/unit/ajax.js
```diff
@@ -2840,6 +2840,32 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
...
+		Globals.register( "corsCallback" );
+		window.corsCallback = function( response ) {
+			assert.ok( response === true, "Cors" );
+			window.clearTimeout( timeout );
+			done();
+		};
+
+		var src = baseURL + "mock.php?action=cors";
+
+		// To simulate a cross-origin request
```
mgol:
```suggestion
		// Simulate a cross-origin request
```

test/unit/ajax.js
```diff
@@ -2840,6 +2840,32 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
 		} );
 	} );
 
+	QUnit.test( "jQuery.append with crossOrigin attribute", function( assert ) {
+		var done = assert.async();
```
mgol: We typically put all var declarations near the top of the function under one `var` keyword, one declaration per line if immediately assigned, unassigned vars all in one line (the first one). Please see other tests for comparison.

Here I'd expect `assert.expect( 1 );` to be the first line, then an empty line & then all vars declared with one var statement.

test/unit/ajax.js
```diff
@@ -2840,6 +2840,32 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
...
+
+		var src = baseURL + "mock.php?action=cors";
+
+		// To simulate a cross-origin request
+		src = src.replace( "localhost", "127.0.0.1" );
+		var html = "<script type=\"text/javascript\" src=\"" + src + "\" crossorigin=\"anonymous\"><\/script>";
+
+		jQuery( document.body ).append( html );
+		timeout = window.setTimeout( function() {
+			assert.ok( false, "Cors" );
```
mgol: This assertion fires for me in Safari 13, Edge 18 & IE 11. It correctly doesn't fire only in Firefox 72 & Chrome 79.

We need to figure out why it fails in other browsers. For IE we might have to skip the test as it doesn't support the crossorigin attribute (by switching from `QUnit.test` to `QUnit.testUnlessIE`). Other browsers should pass it, though.

From my recent investigation: it seems the tests pass in Safari in Karma but not via the PHP setup. On the other hand, they succeed in both modes in Chrome & Firefox.

test/middleware-mockserver.js
```diff
@@ -238,6 +238,15 @@ var mocks = {
...
 			resp.end( "QUnit.assert.ok( false, \"Mock return erroneously executed\" );" );
 		}
+	},
+	cors: function( req, resp ) {
+		if ( typeof req.headers[ "origin" ] != "undefined" ) {
+			resp.setHeader( "Access-Control-Allow-Methods", "GET" );
+			resp.setHeader( "Access-Control-Allow-Origin", req.headers[ "origin" ] );
+			resp.end( "corsCallback(true)" );
+		} else {
+			resp.end( "corsCallback(false)" );
```
mgol:
```suggestion
			resp.end( "corsCallback( false )" );
```

test/middleware-mockserver.js
```diff
@@ -238,6 +238,15 @@ var mocks = {
 		} else {
 			resp.end( "QUnit.assert.ok( false, \"Mock return erroneously executed\" );" );
 		}
+	},
+	cors: function( req, resp ) {
+		if ( typeof req.headers[ "origin" ] != "undefined" ) {
+			resp.setHeader( "Access-Control-Allow-Methods", "GET" );
+			resp.setHeader( "Access-Control-Allow-Origin", req.headers[ "origin" ] );
+			resp.end( "corsCallback(true)" );
```
mgol:
```suggestion
			resp.end( "corsCallback( true )" );
```

test/data/mock.php
```diff
@@ -229,6 +229,17 @@ protected function errorWithScript( $req ) {
 		}
 	}
 
+	protected function cors($req) {
+		if (isset($req->headers['ORIGIN'])) {
```
mgol:
```suggestion
		if ( isset( $req->headers['ORIGIN'] ) ) {
```

test/data/mock.php
```diff
@@ -229,6 +229,17 @@ protected function errorWithScript( $req ) {
...
+			$origin = $req->headers['ORIGIN'];
+			header("Access-Control-Allow-Methods: GET");
```
mgol:
```suggestion
			header( "Access-Control-Allow-Methods: GET" );
```

test/data/mock.php
```diff
@@ -229,6 +229,17 @@ protected function errorWithScript( $req ) {
...
+			header("Access-Control-Allow-Origin: " . $origin);
```
mgol:
```suggestion
			header( "Access-Control-Allow-Origin: " . $origin );
```

test/data/mock.php
```diff
@@ -229,6 +229,17 @@ protected function errorWithScript( $req ) {
...
+			corsCallback(true);
```
mgol: There's a bug here, this will not actually do anything. It should be:
```suggestion
			echo 'corsCallback( true )';
```

test/data/mock.php
```diff
@@ -229,6 +229,17 @@ protected function errorWithScript( $req ) {
...
+		} else {
+			corsCallback(false);
```
mgol: It should be:
```suggestion
			echo 'corsCallback( false )';
```

test/data/mock.php
```diff
@@ -229,6 +229,17 @@ protected function errorWithScript( $req ) {
 		}
 	}
 
+	protected function cors($req) {
```
mgol:
```suggestion
	protected function cors( $req ) {
```

mgol:
I figured out the issue. Karma runs on `http://localhost:9876` so a replacement of `localhost` to `127.0.0.1` worked fine there. However, on my local PHP setup I'm relying on virtual hosts so the page runs on `http://jquery.l` so the replacement failed and the loaded script was, in effect, using the same domain. Apparently, in Chrome & Firefox that works but Safari 13 & Edge 18 just don't execute the script.

Ultimately, we cannot rely on the URL where the tests are loaded to be `localhost`. On our TestSwarm, for example, tests are loaded via URLs like:
http://builds.jenkins.jquery.com/jquery/ff2819911da6cbbed5ee42c35d695240f06e65e3/test/index.html
Replacing it via an IP won't work.

We need to remove the replacement of `localhost` to `127.0.0.1` from the test. The mere inclusion of the `crossorigin` attribute makes the script loaded with the `Origin` header so the current test checking whether `corsCallback` was invoked with `true` would fail without current source changes; that may serve as some kind of a check. Unfortunately, if we cut out the `Access-Control-Allow-Methods` & `Access-Control-Allow-Origin` the test still passes; I'm not sure if there's a way around that, though, unfortunately.

test/data/mock.php
mgol: Hmm, actually, this mistake was why the PHP tests completely failed for me... With these changes they work even in Safari.

Taking that & what I wrote in:
https://github.com/jquery/jquery/pull/4563#issuecomment-575833262
into account, perhaps the current check is fine, we should just add a comment above the `localhost` to `127.0.0.1` replacement saying that full cross-origin verification is only done when tests run from `localhost`, e.g. when run via Karma (as we can depend on `localhost` resolving to `127.0.0.1`) and that in other cases we only do a more limited check that the Origin header is sent.

test/unit/ajax.js
```diff
@@ -2840,6 +2840,32 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
...
 
+	QUnit.test( "jQuery.append with crossOrigin attribute", function( assert ) {
+		var done = assert.async();
+		assert.expect( 1 );
+
+		var timeout = null;
+
+		Globals.register( "corsCallback" );
+		window.corsCallback = function( response ) {
+			assert.ok( response === true, "Cors" );
```
mgol: This could be more descriptive:
```suggestion
			assert.ok( response === true, "Callback fired" );
```

test/unit/ajax.js
mgol: A bit more descriptive:
```suggestion
			assert.ok( false, "Callback should have fired" );
```

gaohuia:
Thanks, I learnt a lot in this PR.

RiZKiT:
Regarding Firefox, if you request a script with `crossorigin="anonymous"` but do not provide a `Access-Control-Allow-Origin`header, the script will not be executed, on purpose.

gaohuia:
@mgol Can you help to review it?

mgol (changes requested):
See latest comments.

src/manipulation/_evalUrl.js
```diff
@@ -10,6 +10,7 @@ jQuery._evalUrl = function( url, options, doc ) {
 		cache: true,
 		async: false,
 		global: false,
+		scriptAttrs: options.crossOrigin ? { "crossOrigin": options.crossOrigin } : null,
```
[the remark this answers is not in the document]
mgol: The decision from the meeting is to treat it as a breaking change. This means we'll release it in jQuery 4.0 but not before, unless we figure out a backwards-compatible way to land this there.

test/middleware-mockserver.js
```diff
@@ -243,6 +243,15 @@ var mocks = {
 		} else {
 			resp.end( "QUnit.assert.ok( false, \"Mock return erroneously executed\" );" );
 		}
+	},
+	cors: function( req, resp ) {
+		if ( typeof req.headers[ "origin" ] != "undefined" ) {
+			resp.setHeader( "Access-Control-Allow-Methods", "GET" );
```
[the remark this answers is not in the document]
gaohuia: Should be good. Will remove it.

test/middleware-mockserver.js
```diff
@@ -243,6 +243,15 @@ var mocks = {
...
+			resp.setHeader( "Access-Control-Allow-Origin", req.headers[ "origin" ] );
```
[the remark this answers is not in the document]
gaohuia: I think so.

test/middleware-mockserver.js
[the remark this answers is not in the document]
mgol: @gaohuia Let's resolve threads only when a change is pushed to make it easier for us to track what's left to do.

gaohuia:
@mgol can you tell me how to run the karma tests with the real mock.php (I mean run it with php)? I added the same logic to "mock.php", but just not sure if it works fine.

mgol:
@gaohuia PHP tests are not run via Karma. To have all tests passing, you need to have Apache or nginx configured with PHP and open `/test/?hidepassed` in a browser, tests will run.

If this is too much, you can run a quick static PHP server in the jQuery directory:
```
php -S 0.0.0.0:7001
```
and then open http://localhost:7001/test/?hidepassed. Note that with this basic setup some tests will fail as PHP can only handle one request at a time and some tests depend on being able to run requests in parallel. But it might be enough for your purposes.

gaohuia:
Thanks @mgol , i have nginx+fpm locally.  the mock.php seems working good.

mgol (approved):
Looks good to me! Just one suggestion.

In my testing, the new test succeeds on BrowserStack in IE 11 in Windows 10 but not in Windows 7. I'm not sure why, maybe the Windows 7 copy is not up to date. We'll worry about that when/if we start testing on Karma in IE.

Thanks for following up with our code review remarks for this past months!

One note, @gaohuia: we're treating this as a potentially breaking change as code using the `crossorigin` attribute on same-domain scripts would now switch to using the script tag injection method which is asynchronous. Because of that, for now we'll only land it on the `master` branch which is the future jQuery 4.0. We can consider backporting this change to the `3.x-stable` branch if this incompatibility is resolved, maybe by only passing `scriptAttrs` in `jQuery._evalUrl` for cross-domain scripts? We need to be careful here, though, so at least for now it's 4.0-only.

mgol:
Landed, thanks!

PR #4572, opened 2019-12-16
Ajax: Deprecate AJAX event aliases, inline event/alias.js into deprecated.js

mgol:
### Summary ###

Deprecate AJAX event aliases, inline event/alias.js into deprecated.js.

Regular event aliases are already deprecated, this brings the AJAX ones in line with all the others.

As for inlining - it doesn't make sense to me to have almost all deprecated stuff in `deprecated.js` but only event aliases in `event/alias.js` - but still only imported from `deprecated.js`.

If accepted, this should be cherry-picked to `3.x-stable` as well.

### Checklist ###


src/deprecated.js
```diff
@@ -22,9 +23,39 @@ jQuery.fn.extend( {
...
+
+		// Handle event binding
+		jQuery.fn[ name ] = function( data, fn ) {
+			return arguments.length > 0 ?
+				this.on( name, null, data, fn ) :
+				this.trigger( name );
+		};
+	} );
+
+jQuery.each( [
```
mgol: I think I need to wrap it with an `if ( jQuery.ajax )` condition to better support custom builds.

mgol: Correction: I'll instead create a `src/deprecated` directory and will move it there. This, together with a proper modification of the [`removeWith` section in Gruntfile.js](https://github.com/jquery/jquery/blob/c1ee33aded44051b8f1288b59d2efdc68d0413cc/Gruntfile.js#L60-L69) will allow to cut this code out when excluding AJAX.

The same has to be done for the event alias part.

mgol: Done.

mgol:
Ready for another review.

mgol:
Landed on `master` at 23d53928f383b0e7440bf4b08b7524e6af232fad and on `3.x-stable` at 7a3cf9c03cc34d5493383852f94d96fe4a3486ef.

Issue #4568, opened 2019-12-15
Drop support for Edge Legacy (EdgeHTML-based) in 4.0?


mgol:
### Description ###

We've touch the topic a few time but I thought it'd be good to have any decision we make publicly available on GitHub.

Our [browser support policy](https://jquery.com/browser-support/) states we support "(Current - 1) and Current" versions of Edge. Right now it's Edge with EdgeHTML 17 & 18<sup>[1]</sup> but the new Chromium-based Edge version (v79<sup>[2]</sup>) is [slated for a public release on January 15, 2020](https://blogs.windows.com/msedgedev/2019/11/04/edge-chromium-release-candidate-get-ready/). Version 80 should follow a few weeks afterwards as they'd be following the [Chromium release calendar](https://www.chromium.org/developers/calendar). Technically speaking, that means jQuery will officially no longer support the legacy EdgeHTML-based Edge versions (as the two latest versions will be 79 & 80, both Chromium-based). However, I feel an engine switch is a big enough event that we should make any decisions about dropping support - even if technically we don't have to - more consciously. Therefore, I wanted to explain how I see the Edge situation after January 15, 2020.

The public release of the new Edge will probably not mean auto-updating everyone on Windows 10 to it but I expect the next Windows 10 "feature update" (which should be released around April/May 2020) will include it.

Microsoft still supports EdgeHTML 16, 17 & 18 on Enterprise & Education versions of Windows 10 so those OS lines may take a bit more to update (installing the Chromium-based Edge is supported in all of them, though). Microsoft will also still support EdgeHTML-based WebView in addition to the new Chromium-based one; there's no timeline to remove the old WebView. Some platforms, like Windows 10 on ARM or Xbox will not get the new Edge at the same time as most desktop Windows versions will.

Dropping EdgeHTML supports would bring us some advantages: it'd decrease the minified gzipped jQuery size by 230 bytes as of now and it'd allow us to restrict some workarounds needed by IE & Edge to just IE by a documentMode check. That fits nicely with other legacy workarounds removals in 4.0, making jQuery faster especially for modern browsers.

I'd keep testing jQuery 3.x with at least EdgeHTML 18 even though technically it will fall out of our supported browsers list soon.

[1]: For support purposes we use EdgeHTML versions rather than Edge app versions which are now at v44
[2]: Chromium-based Edge aligns its major version with Chromium, getting rid of current two separate versioning strategies for the app & its web engine

### Link to test case ###

N/A

[...]

PR #4573, opened 2019-12-16
Selector: Drop support for EdgeHTML (i.e. non-Chromium Microsoft Edge)

mgol:
### Summary ###

**NOTE:** This is a draft PR. No decision has been made yet.

Also, restrict some workarounds that were applied unconditionally in all
browsers to run only in IE now. This slightly increases the size but reduces
the performance burden on modern browsers that don't need the workarounds.

Also, clean up some comments & remove some obsolete workarounds.

Fixes gh-4568

-240 bytes

### Checklist ###


src/ajax.js
```diff
@@ -542,7 +542,7 @@ jQuery.extend( {
 		if ( s.crossDomain == null ) {
 			urlAnchor = document.createElement( "a" );
 
-			// Support: IE <=8 - 11+, Edge 12 - 17 only
+			// Support: IE <=8 - 11+
```
mgol: I removed mentions of Edge from support comments as well. Among other things, this is because many IE workarounds are now hidden behind an `isIE` check so a support comment mentioning Edge would be confusing as Edge wouldn't execute that code.

IE is quite special here as that's the only browser which we detect as a browser to work around its bugs instead of running support tests.

mgol:
Closing for now until we really plan to do it.

mgol:
New PR: #4792

PR #4576, opened 2019-12-18
fix $(document).offset() error

shishiv30:
the document didn't have getClientRects, 
we should check whether the getClientRects is undefined before use it

### Checklist ###

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -84,7 +84,7 @@ jQuery.fn.extend( {
 		// Support: IE <=11+
 		// Running getBoundingClientRect on a
 		// disconnected node in IE throws an error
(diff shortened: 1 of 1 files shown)


mgol:
Offset doesn't make sense on `document` so it's not supported there; see https://jquery.com/upgrade-guide/3.0/#breaking-change-invalid-input-to-the-offset-method

aleen42:
But this is a break change, resulting in scripting error, rather than returning a safe `undefined`. I hope it can be fixed.

mgol:
Yes, it’s a breaking change but that’s exactly what major version updates are for. In this case, masking invalid input may lead to hard to debug issues; we decided it’s better to just throw early.

PR #4577, opened 2019-12-31
Build:Tests: Fix custom build tests, verify on Travis 

mgol:
### Summary ###

This PR fixes unit tests for the following builds:
1. The no-deprecated build: `custom:-deprecated`
1. The current slim build: `custom:-ajax,-effects`
1. The future (#4553) slim build: `custom:-ajax,-callbacks,-deferred,-effects`

It also adds separate Karma runs for the no-deprecated & current slim builds. Once the new slim build configuration lands, we'll have to modify what we test on Travis as well.

This PR can also be cherry-picked to `3.x-stable`.

### Checklist ###

* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [x] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~

PR #4578, opened 2019-12-31
Build: Create a `grunt custom:slim` alias for the Slim build

mgol:
### Summary ###

This PR adds a custom `slim` alias for the official Slim build. It will allow to achieve the same build locally in an easier way and will also de-duplicate the slim build configuration in changes like #4577, making us sure we really test the slim build even after its definition changes.

This PR interferes with #4577 so whichever one lands first, the other one will need to get updated (e.g. if #4577 lands first, this PR will need to modify the new Travis config to use the new alias).

### Checklist ###

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -291,9 +291,27 @@ module.exports = function( grunt ) {
 	// Becomes:
 	//
 	//   grunt build:*:*:+ajax:-dimensions:-effects:-offset
+	//
+	// There's also a special "slim" alias that resolves to the jQuery Slim build
+	// configuration:
+	//
+	//   grunt custom:slim
 	grunt.registerTask( "custom", function() {
 		const args = this.args;
-		const modules = args.length ? args[ 0 ].replace( /,/g, ":" ) : "";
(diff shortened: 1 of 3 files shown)


build/tasks/build.js
```diff
@@ -291,9 +291,27 @@ module.exports = function( grunt ) {
...
 	grunt.registerTask( "custom", function() {
 		const args = this.args;
-		const modules = args.length ? args[ 0 ].replace( /,/g, ":" ) : "";
+		const modules = args.length ?
+			args[ 0 ]
+				.split( "," )
+
+				// Replace "slim" with respective exclusions meant for
+				// the official slim build
+				.reduce( ( acc, elem ) => acc.concat(
```
mgol: `flatMap` would be so useful here... Alas, it's Node 12+ and we'll have to support at least v10 for the time being.

mgol:
Landed on `master` at 9b9ed469b43e9fa6e2c752444470ae4c87d03d57 & on `3.x-stable` at 4cbdc745cc32f0fb8865139df7caad075b05d2b8.

PR #4585, opened 2020-01-08
Attributes: Refactor val(): don't strip carriage return, isolate IE workarounds

mgol:
### Summary ###

Before this change, `val()` was stripping out carriage return characters from
the returned value. No test has relied on that. The logic was different for
option elements as its custom defined hook was omitting this stripping logic.

This commit gets rid of the carriage return removal and isolates the IE-only
select val getter to be skipped in other browsers.

-16 bytes

### Checklist ###


mgol:
-16 bytes

PR #4593, opened 2020-01-16
CSS: Remove the opacity CSS hook

mgol:
### Summary ###

The consequence is `.css( "opacity" )` will now return an empty string for
detached elements in standard-compliant browsers and "1" in IE & the legacy
Edge. That behavior is shared by most other CSS properties which we're not
normalizing either.

### Checklist ###


mgol:
-25 bytes

Issue #3256, opened 2016-08-01
Event types colliding with Object prototype properties don't work


gibson042:
We use `event.type` [as a key](https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/event.js#L294) to get the array of event handlers, which throws an exception when we instead receive an `Object.prototype` property and treat it like an array: https://jsfiddle.net/9t4kqaqr/

Our standard fix is appending a space (cf. [delegated selector match tracking](https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/event.js#L370-L371) and [Sizzle `createCache`](https://github.com/jquery/sizzle/blob/582b63c607b344ce5838d89eaf57e693461948b2/src/sizzle.js#L349-L367)), but we should also consider `Map` for such purposes (ponyfilling as needed), especially if [Data.js](https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/data/Data.js) can be updated accordingly—one of @rwaldron's original goals when creating it.

mgol:
> but we should also consider `Map` for such purposes

Why not `Object.create(null)`? This one we wouldn't have to ponyfill.

gibson042:
> Why not `Object.create(null)`?

I thought we still supported environments where that fails, but things look good! :+1:

mgol:
Yup, and the changes required in that case will be smaller. :)

I'm only not sure if PhantomJS 1.x supports Object.create.

PetarMax:
I've just rediscovered this bug. Is there a reason this hasn't been fixed yet?

gibson042:
No one has yet gotten around to it. But a pull request would be welcome!

timmywil:
I think we can implement the easy fix with spaces immediately, then consider using Map in 4.0.

mgol:
PR: #4603. I also fixed similar issues in manipulation & data modules using `Object.create(null)`.

rjaros:
In my opinion this is a breaking change that should not go into 3.5.0 minor upgrade version. It can break a lot of dependent projects, like this one: https://github.com/snapappointments/bootstrap-select/issues/2430

mgol:
@rjaros Please submit a new issue with details.

mgol:
[...]

PR #4603, opened 2020-02-05
Data:Event:Manipulation: Prevent collisions with Object.prototype properties

mgol:
## Summary ###

Make sure events & data keys matching Object.prototype properties work.
A separate fix for such events on cloned elements was added as well.

The commit also removes two unnecessary checks:
1. A check for `events` in private data in `.trigger()` - only the `handle`
   key is needed.
2. A fallback of the `events` private data to an empty object in
   `jQuery.event.dispatch` because that method is called from the `handle`
   function in element's private data and both `events` & `handle` are set and
   removed at the same time.

Fixes gh-3256

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -133,7 +133,7 @@ jQuery.event = {
 
 		// Init the element's event structure and main handler, if this is the first
 		if ( !( events = elemData.events ) ) {
-			events = elemData.events = {};
+			events = elemData.events = Object.create( null );
 		}
 		if ( !( eventHandle = elemData.handle ) ) {
 			eventHandle = elemData.handle = function( e ) {
@@ -294,7 +294,9 @@ jQuery.event = {
 			// Make a writable jQuery.Event from the native event object
 			event = jQuery.event.fix( nativeEvent ),
 
-			handlers = ( dataPriv.get( this, "events" ) || {} )[ event.type ] || [],
+			handlers = (
+					dataPriv.get( this, "events" ) || Object.create( null )
+				)[ event.type ] || [],
 			special = jQuery.event.special[ event.type ] || {};
 
 		// Use the fix-ed jQuery.Event rather than the (read-only) native event
--- a/src/event/trigger.js
+++ b/src/event/trigger.js
@@ -99,7 +99,9 @@ jQuery.extend( jQuery.event, {
 				special.bindType || type;
 
 			// jQuery handler
-			handle = ( dataPriv.get( cur, "events" ) || {} )[ event.type ] &&
+			handle = (
+					dataPriv.get( cur, "events" ) || Object.create( null )
+				)[ event.type ] &&
 				dataPriv.get( cur, "handle" );
 			if ( handle ) {
 				handle.apply( cur, data );
--- a/test/unit/data.js
+++ b/test/unit/data.js
@@ -998,3 +998,18 @@ QUnit.test( ".data(prop) does not create expando", function( assert ) {
 		}
 	}
 } );
+
+QUnit.test( "keys matching Object.prototype properties  (gh-3256)", function( assert ) {
+	assert.expect( 2 );
+
+	var div = jQuery( "<div/>" );
+
+	assert.strictEqual( div.data( "hasOwnProperty" ), undefined,
+		"hasOwnProperty not matched (before forced data creation)" );
+
+	// Force the creation of a data object for this element.
+	div.data( { foo: "bar" } );
+
+	assert.strictEqual( div.data( "hasOwnProperty" ), undefined,
+		"hasOwnProperty not matched (after forced data creation)" );
+} );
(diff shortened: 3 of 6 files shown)


src/event.js
```diff
@@ -298,7 +298,7 @@ jQuery.event = {
 
 		var i, j, ret, matched, handleObj, handlerQueue,
 			args = new Array( arguments.length ),
-			handlers = ( dataPriv.get( this, "events" ) || {} )[ event.type ] || [],
+			handlers = dataPriv.get( this, "events" )[ event.type ] || [],
```
mgol: I'm depending here on the fact `jQuery.event.dispatch` (where this code resides) is only called from within `eventData.handle` of the element which means the element already has private `events` data set. Let me know if you think it's still needed for some reason; it'd be good to write a test then.

src/event/trigger.js
```diff
@@ -99,8 +99,7 @@ jQuery.extend( jQuery.event, {
 				special.bindType || type;
 
 			// jQuery handler
-			handle = ( dataPriv.get( cur, "events" ) || {} )[ event.type ] &&
-				dataPriv.get( cur, "handle" );
+			handle = dataPriv.get( cur, "handle" );
```
mgol: What was the reason for the `events` check here when we only use `handle`?

mgol:
-19 bytes because of removals of fallbacks where they didn't seem needed IMO. I'd appreciate a review of those cases to be sure.

src/event.js
dmethvin: At some point `jQuery.event.dispatch` would have been called from internal functions like [jQuery.event.simulate](https://github.com/jquery/jquery/blob/17eeebb3c04c72e5d00520c191cce858844aedec/src/event.js#L605-L626) in the 1.7.x era to support oldIE. The tests for that were pulled out though, so I suspect this guard isn't needed anymore.

src/event/trigger.js
dmethvin: Just an optimization I suspect. If there are no events of that type for this element, we can avoid the call to `handle`. However, the call to get the data object isn't free so it probably doesn't save anything. In the 1.7.x era this was a direct lookup on an object so it might have been worth the difference. Today I don't see it being important.

mgol: Ah, so this is an optimization that prevents calling `handle` for elements that have some events data but not for an event we're triggering right now? The `handle` function is basically some simple checks & then calling `jQuery.event.dispatch`. That function is not particularly light so I'll need to check if it's worth removing this optimization.

BTW, `Object.prototype` collisions are not a big issue here as the worst that could happen is code triggering a colliding event name does a bit more work than necessary. That's why it'd be hard to write a test that fails on the `master` version of this code.

src/event.js
mgol: @dmethvin Do you think we could remove `jQuery.event.simulate` in 4.0? The whole `jQuery.event` is not a documented API but I'm not sure which parts of it are used outside; certainly `jQuery.event.special` is. `jQuery.event.simulate` was only used internally for simulating `focusin`/`focusout` but we removed the patch in #4362.

mgol: Note to self: check jQuery Mousewheel, in particular:
https://github.com/jquery/jquery-mousewheel/blob/aba08a57e4816c93f639ddc6c95088316b3ceb24/jquery.mousewheel.js#L218

gibson042: I'm good with keeping this change and revisiting if it causes problems.

src/event/trigger.js
gibson042: I think it is worth keeping a guard in place to ensure that `handle` is empty if the event type is not registered.

test/unit/data.js
```diff
@@ -998,3 +998,18 @@ QUnit.test( ".data(prop) does not create expando", function( assert ) {
...
+QUnit.test( "keys matching Object.prototype properties  (gh-3256)", function( assert ) {
+	assert.expect( 2 );
+
+	var div = jQuery( "<div/>" );
+
+	assert.strictEqual( div.data( "hasOwnProperty" ), undefined,
+		"hasOwnProperty not matched (before forced data creation)" );
+
+	// Query the collection to force the creation of a data object for this element.
+	div.data();
```
gibson042: I think this should be even stronger.
```suggestion
	div.data( { foo: "bar" } );
```

src/event/trigger.js
mgol: I restored the previous check, just with `Object.create( null )` as a fallback.

test/unit/data.js
mgol: Good point, updated.

src/event.js
mgol: I restored the safeguard, just changing the fallback to `Object.create( null )`. With other changes I had to make, restoring this safeguard actually _reduced_ the gzipped size by 1 byte so why even risk. 🤷‍♂

mgol:
Adding "Needs review" back as a few things have changed since I reported the PR.

It's now +0 bytes.

PR #4612, opened 2020-02-24
Core:Ajax: Align nonce & global with master, fix an AMD issue

mgol:
### Summary ###

This PR aligns the `3.x-stable` branch with `master` in two aspects:
1. It migrates the nonce module to return an object instead of a primitive
variable. This had to be changed on `master` as in ES modules you export
live read-only bindings to variables, meaning you can't increment the nonce
directly. Also, the way it was done so far was working differently in AMD & the
single built file - in the built file one nonce variable was declared, accessed
and incremented. In AMD mode separate instances were create for each module
that depend on the nonce module, creating unintended nonce clashes.
2. Whether the `noGlobal` parameter was set to `true` is now checked using the
typeof operator to align with `master`.

+8 bytes

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -610,7 +610,8 @@ jQuery.extend( {
 			// Add or update anti-cache param if needed
 			if ( s.cache === false ) {
 				cacheURL = cacheURL.replace( rantiCache, "$1" );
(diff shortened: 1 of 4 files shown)


mgol:
+8 bytes

PR #4614, opened 2020-02-24
Release: Use an in-repository dist README fixture

mgol:
### Summary ###

Use a dist README fixture kept in the jQuery repository instead of modifying
an existing one. This makes the jQuery repository the single source of truth
when it comes to jQuery releases and it makes it easier to make changes to
README without worrying how it will affect older jQuery lines.

https://github.com/jquery/jquery-release/pull/103 is a pre-requisite to make this PR work.

### Checklist ###

### DIFF
--- a/build/fixtures/README.md
+++ b/build/fixtures/README.md
@@ -1,0 +1,62 @@
+# jQuery
+
+> jQuery is a fast, small, and feature-rich JavaScript library.
+
+For information on how to get started and how to use jQuery, please see [jQuery's documentation](http://api.jquery.com/).
+For source files and issues, please visit the [jQuery repo](https://github.com/jquery/jquery).
+
+If upgrading, please see the [blog post for @VERSION](@BLOG_POST_LINK). This includes notable differences from the previous version and a more readable changelog.
+
+## Including jQuery
+
+Below are some of the most common ways to include jQuery.
+
+### Browser
+
+#### Script tag
+
+```html
+<script src="https://code.jquery.com/jquery-@VERSION.min.js"></script>
+```
+
+#### Babel
+
+[Babel](http://babeljs.io/) is a next generation JavaScript compiler. One of the features is the ability to use ES6/ES2015 modules now, even though browsers do not yet support this feature natively.
+
+```js
+import $ from "jquery";
+```
+
+#### Browserify/Webpack
+
+There are several ways to use [Browserify](http://browserify.org/) and [Webpack](https://webpack.github.io/). For more information on using these tools, please refer to the corresponding project's documention. In the script, including jQuery will usually look like this...
+
+```js
+var $ = require( "jquery" );
+```
+
+#### AMD (Asynchronous Module Definition)
+
+AMD is a module format built for the browser. For more information, we recommend [require.js' documentation](http://requirejs.org/docs/whyamd.html).
+
+```js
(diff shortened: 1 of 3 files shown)


build/release.js
```diff
@@ -4,30 +4,35 @@ var fs = require( "fs" );
...
+	];
+	const filesToCommit = [
+		...distFiles,
+		"src/core.js"
+	];
+	const cdn = require( "./release/cdn" );
+	const dist = require( "./release/dist" );
 
-		npmTags = Release.npmTags;
+	const npmTags = Release.npmTags;
```
mgol: @timmywil We had a talk a while ago about the `onevar` rule in ESLint. You said you don't want to remove/change the rule so that code using the preset doesn't need to adapt. However:

1. The rule is *not* actually enabled in the jQuery preset. AFAIK it was removed because Jörn didn't want it in jQuery UI or QUnit due to Git diff issues etc.
2. While we adhere to this way of declaring variables in jQuery code, I'd argue it's actually harmful when used for `const`/`let`. This is because for `var`s we were emulating how the JS engine interprets those declarations; in case of `const`/`let` the declarations are not hoisted and it's common to put them in the middle of the function where they are needed.
3. Other ES6 code in our repo already uses separate declarations per variable so we have some history here. That also means we don't have history of requiring this rule for `const`/`let` so not applying it shouldn't cause issues with code refactorings.

build/release.js
```diff
@@ -4,30 +4,35 @@ var fs = require( "fs" );
...
 
 	function setSrcVersion( filepath ) {
 		var contents = fs.readFileSync( filepath, "utf8" );
 		contents = contents.replace( /@VERSION/g, Release.newVersion );
 		fs.writeFileSync( filepath, contents, "utf8" );
 	}
 
+	// Stop the release process if any of the asynchronous operations fails.
+	process.on( "unhandledRejection", reason => {
```
mgol: I wonder if we should put this error handling in jquery-release somehow...

build/fixtures/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+#### Browserify/Webpack
+
+There are several ways to use [Browserify](http://browserify.org/) and [Webpack](https://webpack.github.io/). For more information on using these tools, please refer to the corresponding project's documention. In the script, including jQuery will usually look like this...
+
+```js
+var $ = require( "jquery" );
+```
+
+#### AMD (Asynchronous Module Definition)
```
mgol: I'll submit a separate PR adding info about our new ES modules. Since we only want it on the 4.x branch, this PR will be useful here. :)

build/release.js
timmywil: The `onevar` rule can technically differentiate between `var`, `const`, and `let`. I'd be immediately in favor of specifying that we only require `onevar` for `var`. As for `var`, I'm not sure it's worth the large diffs and blame overrides.

mgol: @timmywil Makes sense, I didn't know it can be configured just for `var`. We can add it to our own ESLint config since it was rejected as a general rule for the jQuery preset.

mgol: @timmywil Done in #4615.

build/release.js
mgol: Done in https://github.com/jquery/jquery-release/pull/103.

PR #4648, opened 2020-03-28
Build:Event: Make sure all source modules' exports are used

mgol:
### Summary ###

Make sure all source modules' exports are used.

To achieve that, use `eslint-plugin-import`'s `no-unused-modules` rule.

Also, explicitly import `event/trigger.js` from `jquery.js`; so far it was only
imported from ajax.js, making it mistakenly skipped in the
`custom:slim,-deprecated` build.

-5 bytes! :)

### Checklist ###

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -13,6 +13,10 @@
 	"rules": {
 		"import/extensions": [ "error", "always" ],
 		"import/no-cycle": "error",
+		"import/no-unused-modules": [ "error", {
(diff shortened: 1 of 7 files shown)


mgol:
-5 bytes! :)

timmywil:
So `rhtml` wasn't used anywhere? I guess that's an artifact from removing stuff for 4.0.

I wonder, tho, if we should leave `parseXML`'s default export for those using the modules directly.

PR #4650, opened 2020-03-29
Ajax: Overwrite s.contentType with content-type header value, if any

### LINKED ISSUE #4119: Invalid encoding of %20 char sequence in $.ajax post, put body

wenz:
### Summary ###
Overwrite s.contentType with content-type header value, if any
Fixes #4119 

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -862,7 +862,8 @@ jQuery.each( [ "get", "post" ], function( _i, method ) {
 } );
 
 jQuery.ajaxPrefilter( function( s ) {
-	for ( var i in s.headers ) {
+	var i;
+	for ( i in s.headers ) {
 		if ( i.toLowerCase() === "content-type" ) {
 			s.contentType = s.headers[ i ] || "";
 		}
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -862,11 +862,9 @@ jQuery.each( [ "get", "post" ], function( _i, method ) {
 } );
 
 jQuery.ajaxPrefilter( function( s ) {
-	if ( s.data && s.processData ) {
-		for ( var i in s.headers ) {
-			if ( i.toLowerCase() === "content-type" ) {
-				s.contentType = s.headers[ i ] || "";
-			}
+	for ( var i in s.headers ) {
+		if ( i.toLowerCase() === "content-type" ) {
+			s.contentType = s.headers[ i ] || "";
 		}
 	}
 } );
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -861,4 +861,14 @@ jQuery.each( [ "get", "post" ], function( _i, method ) {
 	};
 } );
 
+jQuery.ajaxPrefilter( function( s ) {
+	if ( s.data && s.processData ) {
+		for ( var i in s.headers ) {
+			if ( i.toLowerCase() === "content-type" ) {
+				s.contentType = s.headers[ i ] || "";
+			}
+		}
+	}
+} );
+
 export default jQuery;
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -1442,6 +1442,19 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - override contentType with header (gh-4119)", 1, function( assert ) {
+		return {
+			url: "bogus.html",
+			contentType: "application/json",
+			headers: { "content-type": "application/x-www-form-urlencoded" },
+			beforeSend: function( _, s ) {
+				assert.strictEqual( s.contentType, "application/x-www-form-urlencoded", "contentType is overwritten" );
+				return false;
+			},
+			error: true
+		};
+	} );
+
 	ajaxTest( "jQuery.ajax() - data - no processing POST", 1, function( assert ) {
 		return {
 			url: "bogus.html",
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -1448,7 +1448,8 @@ QUnit.module( "ajax", {
 			contentType: "application/json",
 			headers: { "content-type": "application/x-www-form-urlencoded" },
 			beforeSend: function( _, s ) {
-				assert.strictEqual( s.contentType, "application/x-www-form-urlencoded", "contentType is overwritten" );
+				assert.strictEqual( s.contentType, "application/x-www-form-urlencoded",
+					"contentType is overwritten" );
 				return false;
 			},
 			error: true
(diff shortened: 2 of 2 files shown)


wenz:
Is there a way to re-run the jsf-clabot? My Git email settings were incorrect; now, the commit is correctly linked to my GitHub account.

mgol:
@wenz You need to amend the existing commit with new author information & then `git push --force-with-lease`. You can also modify the commit message to be more precise.

wenz:
@mgol done - thanks for the pointer!

src/ajax.js
```diff
@@ -861,4 +861,14 @@ jQuery.each( [ "get", "post" ], function( _i, method ) {
 	};
 } );
 
+jQuery.ajaxPrefilter( function( s ) {
+	if ( s.data && s.processData ) {
```
mgol: I don't think this check is needed. The value of `contentType` shouldn't depend on whether data is passed & being processed.

test/unit/ajax.js
```diff
@@ -1410,6 +1410,22 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - don't escape %20 in contentType (gh-4119)", 1, function( assert ) {
```
mgol: We weren't escaping `%20` in contentType but in data. How about:
```suggestion
	ajaxTest( "jQuery.ajax() - '%20' in data with a header Content-Type override (gh-4119)", 1, function( assert ) {
```
? A bit long but I don't have shorter ideas at the moment.

test/unit/ajax.js
```diff
@@ -1410,6 +1410,22 @@ QUnit.module( "ajax", {
...
+		return {
+			url: "bogus.html",
+			contentType: "application/x-www-form-urlencoded",
+			headers: { "content-type": "application/json" },
```
mgol: We should also have a test with these values reversed to make sure the override still happens.

test/unit/ajax.js
mgol: This is testing one intended side-effect of the added prefilter but we should also test it more explicitly. Please also add tests checking that the `contentType` field gets correctly overwritten.

src/ajax.js
wenz: You are right - done! (Also saves a few bytes.)

test/unit/ajax.js
wenz: I've added my suggestion, based on yours.

test/unit/ajax.js
wenz: Done; added a new test.

test/unit/ajax.js
wenz: I struggled with implementing that. I cannot use beforeSend since that seems to be called prior to the prefilter. The XHR object (and also jqXHR) does not give me read access to the request headers sent. Do you have any ideas or pointers for me?

src/ajax.js
```diff
@@ -861,4 +861,12 @@ jQuery.each( [ "get", "post" ], function( _i, method ) {
 	};
 } );
 
+jQuery.ajaxPrefilter( function( s ) {
+	for ( var i in s.headers ) {
```
mgol: Our code style requires separate variable declarations at the beginning of the function. So:
```suggestion
	var i;
	for ( i in s.headers ) {
```

test/unit/ajax.js
mgol: @wenz Are you sure? It should be called after the prefilter.

wenz: Here's the test I tried within an `ajaxTest()` call, which always fails (s.contentType is "application/json"):

```
return {
	url: "bogus.html",
	contentType: "application/json",
	headers: { "content-type": "application/x-www-form-urlencoded" },
	beforeSend: function( _, s ) {
		assert.strictEqual( s.contentType, "application/x-www-form-urlencoded", "contentType is overwritten" );
		return false;
	},
	error: true
};
```
Is there something obvious I'm doing wrong?

The %20 tests indirectly check whether the prefilter works, so at least we have coverage of that code (?).

mgol: @wenz That test passes for me on your branch. How are you running the tests? You need to either run `grunt` first or check the `esmodules` checkbox which will make tests use jQuery source files directly instead of the built file.

wenz: I did run `grunt`, but most probably I messed up somewhere else. I just tried from scratch, and indeed now the test passes. Thank you for your patience! :)

test/unit/ajax.js
```diff
@@ -1442,6 +1442,19 @@ QUnit.module( "ajax", {
 		};
 	} );
 
+	ajaxTest( "jQuery.ajax() - override contentType with header (gh-4119)", 1, function( assert ) {
+		return {
+			url: "bogus.html",
+			contentType: "application/json",
+			headers: { "content-type": "application/x-www-form-urlencoded" },
+			beforeSend: function( _, s ) {
+				assert.strictEqual( s.contentType, "application/x-www-form-urlencoded", "contentType is overwritten" );
```
mgol: This line is quite long, could you break it?
```suggestion
				assert.strictEqual( s.contentType, "application/x-www-form-urlencoded",
					"contentType is overwritten" );
```

timmywil:
Yes, this might change behavior, but I consider this a bug fix. You'd have to set the header and then expect that header to be ignored.

mgol:
> You'd have to set the header and then expect that header to be ignored.

You could have a header bag that you add to the request via a helper local to the project & then rely on being able to override that via `dataType` (although this override was never complete in the end). But yes, it does seem to be an edge case.

mgol:
Landed on `master` in 7fb90a6beaeffe16699800f73746748f6a5cc2de & on `3.x-stable` in 065143c2e93512eb0c82d1b344b71d06eb7cf01c.

Thanks for the PR & keeping up with our feedback!

Issue #4525, opened 2019-10-21
Ensure .offset sends px-ed strings to .css()


timmywil:
### Description ###

We deprecated passing numbers to CSS, which migrate warns about, but offset passes the numbers internally. Append the 'px' before passing to `.css()`.

mgol:
Fixed for `3.5.0` in #4508 but we still need a fix for 4.0.

mgol:
PR: #4651

mgol:
We've decided to fix this on the Migrate side by not warning against properties from the whitelist that will still have `px` auto-appended. More details in https://github.com/jquery/jquery/pull/4651#issuecomment-608384853.

PR #4651, opened 2020-04-01

mgol:
### Summary ###

This is an alternative to #4508 that's smaller by 4 bytes on master & by 2 bytes on `3.x-stable`.

An upcoming release of Migrate will generate warnings for calls to `.css()`
that pass numbers rather than strings, see jquery/jquery-migrate#296. At
the moment, core's `.offset()` setter passes numbers rather than px strings
so it would throw warnings. This commit fixes that.

Fixes gh-4525
Ref gh-4508

Co-authored-by: Dave Methvin <dave.methvin@gmail.com>

### Checklist ###

### DIFF
--- a/src/offset.js
+++ b/src/offset.js
@@ -12,6 +12,7 @@ jQuery.offset = {
 		var curPosition, curLeft, curCSSTop, curTop, curOffset, curCSSLeft, calculatePosition,
 			position = jQuery.css( elem, "position" ),
 			curElem = jQuery( elem ),
+			numberProps = {},
 			props = {};
 
 		// Set position first, in-case top/left are set even on static elem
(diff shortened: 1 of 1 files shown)


mgol:
This adds 16 bytes to the bundle.

I keep thinking we might want to instead change Migrate to only warn for those properties that are not on the px-appending whitelist for 4.0, i.e. ones not matched by the regex in [`src/css/isAutoPx.js`](https://github.com/jquery/jquery/blob/90fed4b453a5becdb7f173d9e3c1492390a1441f/src/css/isAutoPx.js).

Then Core size would not suffer and Migrate users would focus on the things that will really break in jQuery 4.0 if unfixed.

Let's discuss this on Monday.

mgol:
We've decided to follow the steps from my last comment so no Core changes are necessary. Once Migrate is updated, we may also remove the previous Core `3.x` patch from #4508.

PR #4656, opened 2020-04-06
Event: Use only one focusin/out handler per matching window & document

mgol:
### Summary ###

The `doc` variable in:
https://github.com/jquery/jquery/blob/3.4.1/src/event/focusin.js#L30
matched `document` for `document` & `window` for `window`, creating two separate
wrapper event handlers & calling handlers twice if at least one `focusout` or
`focusin` handler was attached on *both* `window` & `document`, or on `window`
& another regular node.

Also, fix the "focusin from an iframe" test to actually verify the behavior
from commit 1cecf64e5aa415367a7dae0b55c2dd17b591442d - the commit that
introduced the regression - to make sure we don't regress on either front.

+5 bytes.

Test updates should be cherry-picked to `master` without any source changes.

### Checklist ###

### DIFF
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -2555,7 +2555,9 @@ testIframe(
 	function( assert, framejQuery, frameWin, frameDoc ) {
 		assert.expect( 1 );
 
-		var input = jQuery( frameDoc ).find( "#frame-input" );
+		var done = assert.async(),
+			focus = false,
+			input = jQuery( frameDoc ).find( "#frame-input" );
 
 		// Create a focusin handler on the parent; shouldn't affect the iframe's fate
 		jQuery( "body" ).on( "focusin.iframeTest", function() {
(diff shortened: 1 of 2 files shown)


test/unit/event.js
```diff
@@ -2552,34 +2552,69 @@ QUnit.test( "focusin using non-element targets", function( assert ) {
...
+		input[ 0 ].focus();
 
 		// Remove body handler manually since it's outside the fixture
 		jQuery( "body" ).off( "focusin.iframeTest" );
+
+		setTimeout( function() {
+
+			// DOM focus is unreliable in TestSwarm
+			if ( QUnit.isSwarm && !focus ) {
+				assert.ok( true, "GAP: Could not observe focus change" );
```
mgol, line 2582: @gibson042 BTW, I copied this from other similar assertions but... what is GAP?

mgol:
Tests backported to `master` in #4657.

test/unit/event.js
gibson042: My attempt to make clear that this is not a true passing assertion, but rather a failure to verify the desired assertion at all... a **gap** between intent and capability.

mgol: Ah. Since it’s all uppercase, I thought it’s an acronym. 😅

Issue #4652, opened 2020-04-04
Focusout event on $(window) & $(document) : events call twice


dutrieux:
### Description ###
If I attach `focusout` event on the `$(window)` and `$(document)`, each events are called twice.

I use jQuery v3.4.1 and I have the same problem with Firefox (v74.0.1) and Chrome (v80.0.3987.163).
If I use native Js, each events are called once.

### Link to test case ###
https://jsbin.com/rilamarodi/1/edit?html,css,js,output
open console, focus input field and go out (remove focus) to see on console that each events are called twice.

![focusout](https://user-images.githubusercontent.com/1622751/78463055-df531900-76d8-11ea-868b-085496cd109d.gif)

mgol:
Thanks for the report. To add to that, on Firefox if I focus the input & then either focus another application or Firefox DevTools, I see 4 occurrences of the document focusout & 5 of window focusout.

I initially thought it's related to the big refactor of focus event handling in #4279 but the issue is even in 3.0.0.

dutrieux:
>  but the issue is even in 3.0.0.

in 2.2.4 too !

mgol:
OK, so this issue first occurred in jQuery 1.11.0/2.1.0. It was working fine in 2.0.3. It was introduced in PR #1369, commit 1cecf64e5aa415367a7dae0b55c2dd17b591442d. That commit broke `focusout` handling on non-element targets, though, a fix was added in c2aca17d457d302cb1683f925b9e5ee93f0984ea so that's the first commit where you can see this issue.

The issue doesn't exist on `master` which is future jQuery 4.0.0 because we removed the `focusin`/`focusout` shim there in PR #4362. We didn't backport the removal to the `3.x` line as it changes the relative order between `focusin`/`focusout` & `focus`/`blur` events.

Taking into account that:
* the issue should be relatively easy to work around
* this is the first report I see about it
* this change landed 9 years ago
it might be best to leave it as-is for the 3.x branch.

If you find a simple patch fixing the current 3.x implementation, we'd review the PR, though.

Marking for discussion at today's team meeting.

mgol:
Oh, actually this is pretty simple: the `doc` variable in https://github.com/jquery/jquery/blob/3.4.1/src/event/focusin.js#L30 will match `document` for `document` & `window` for `window` and we'd like them to be grouped by document. A fix incoming, it will be quite simple.

[...]

PR #4657, opened 2020-04-06
Tests: Use only one focusin/out handler per matching window & document

mgol:
### Summary ###

Backport tests from a jQuery 3.x fix that's not needed on `master`.

Also, fix the "focusin from an iframe" test to actually verify the behavior
from commit 1cecf64e5aa415367a7dae0b55c2dd17b591442d - the commit that
introduced the regression - to make sure we don't regress on either front.

The main part of the modified test was checking that focusin handling in an
iframe works and that's still checked. The test was also checking that it
doesn't propagate to the parent document, though, and, apparently, in IE it
does. This one test is now blacklisted in IE.

(cherry picked from 9e15d6b469556eccfa607c5ecf53b20c84529125)
(cherry picked from 1a4f10ddc37c34c6dc3a451ee451b5c6cf367399)

Ref gh-4652
Ref gh-4656

### Checklist ###

### DIFF
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -2560,30 +2560,72 @@ testIframe(
 	function( assert, framejQuery, frameWin, frameDoc ) {
 		assert.expect( 1 );
 
-		var input = jQuery( frameDoc ).find( "#frame-input" );
+		var done = assert.async(),
+			focus = false,
+			input = jQuery( frameDoc ).find( "#frame-input" );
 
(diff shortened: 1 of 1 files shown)


mgol:
I've noticed I was missing a cleanup for `window` & `document` handlers in the newly added test. Fixed here, I'll need to backport that to `3.x-stable` as well.

PR #4670, opened 2020-04-16
Improved trim performance in Selector.js

wonseop:
The native JavaScript trim() method works much faster than regular expressions in the case of removing whitespace from the beginning and the end of a string.

e.g. String.prototype.trim vs Trim String With Sizzle's RegEx
       (https://jsperf.com/string-prototype-trim-vs-trim-string-with-sizzle-regex)
- Testing in Chrome 81.0.4044 / Windows 10 0.0.0
   : The native trim() method works 97% faster in my PC.
- Testing in IE 11.0.0 / Windows 10 0.0.0
   : The native trim() method works 94% faster in my PC.

The trim using the regular expression of selector.js is the same as Sizzle, but it seems desirable to remove it for the following reasons.
- The current version and the later of IE9 supports String.prototype.trim. 
  jQuery v2.X has stopped supporting the prior versions in IE8.
- Since Sizzle is removed from 4.0, there is no need to keep the code.

However, since regular expressions for trim are also used for other purposes, the only portion of the codes with the improved performance with the trim () method are changed.

### Checklist ###

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -8,6 +8,7 @@ import push from "./var/push.js";
 import whitespace from "./selector/var/whitespace.js";
 import rbuggyQSA from "./selector/rbuggyQSA.js";
 import support from "./selector/support.js";
+import trim from "./var/trim.js";
 
 // The following utils are attached directly to the jQuery object.
 import "./selector/contains.js";
@@ -268,7 +269,7 @@ function find( selector, context, results, seed ) {
 	}
 
 	// All others
-	return select( selector.replace( rtrim, "$1" ), context, results, seed );
+	return select( trim.call( selector ), context, results, seed );
 }
 
 /**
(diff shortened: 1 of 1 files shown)


mgol:
Thanks for the pull request!

Unfortunately, this improvement is not correct. The main reason why we no longer use `String.prototype.trim` internally is that it recognizes [more whitespace](https://tc39.es/ecma262/#prod-WhiteSpace) than the [selectors spec](https://www.w3.org/TR/selectors-3/#whitespace). For example, the following will find 0 elements (`00A0` is a code point for "non-breaking space"):
```js
document.querySelectorAll('body\u{00A0}')
```
while the following will match one - the document body:
```js
document.querySelectorAll('body\u{00A0}'.trim())
```

When we remove `jQuery.trim` from Core, we'll also remove the `trim.js` file as it's only used there.

mgol:
Also, measuring performance in micro-benchmarks like the one you linked to is tricky; code like the one you shared is quite likely to be optimized by the browser to no code at all as soon as it notices it produces no side effects.

wonseop:
Thank you so much for your detailed explanation.

PR #4672, opened 2020-04-17
Build: Correct code indentations based on jQuery Style Guide

wonseop:
### Summary ###

- Correct code indentations based on Core Style Guide
   (https://contribute.jquery.org/style-guide/js/#spacing)
    
- Add rules to `src/.eslintrc.json` to enable "enforcing consistent indentation",
   with minimal changes to the current code.

### Checklist ###

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -12,7 +12,10 @@
 
 	"rules": {
 		"import/extensions": [ "error", "always" ],
-		"import/no-cycle": "error"
+		"import/no-cycle": "error",
+		"indent": [ "error", "tab", {
+			"outerIIFEBody": 0
+		} ]
 	},
 
 	"overrides": [
@@ -23,7 +26,17 @@
 				"sourceType": "script"
 			},
 			"rules": {
-				"no-unused-vars": "off"
+				"no-unused-vars": "off",
+				"indent": [ "error", "tab", {
+
+					// Unlike other codes, "wrapper.js" is implemented in UMD.
+					// So it required a specific exception for jQuery's UMD
+					// Code Style. This makes that indentation check is not
+					// performed for 1 depth of outer FunctionExpressions
+					"ignoredNodes": [
+						"Program > ExpressionStatement > CallExpression > FunctionExpression > *"
+					]
+				} ]
 			},
 			"globals": {
 				"jQuery": false,
--- a/src/deprecated/event.js
+++ b/src/deprecated/event.js
@@ -28,7 +28,8 @@ jQuery.fn.extend( {
 	}
 } );
 
-jQuery.each( ( "blur focus focusin focusout resize scroll click dblclick " +
+jQuery.each(
+	( "blur focus focusin focusout resize scroll click dblclick " +
 	"mousedown mouseup mousemove mouseover mouseout mouseenter mouseleave " +
 	"change select submit keydown keypress keyup contextmenu" ).split( " " ),
 	function( _i, name ) {
@@ -39,4 +40,5 @@ jQuery.each( ( "blur focus focusin focusout resize scroll click dblclick " +
 				this.on( name, null, data, fn ) :
 				this.trigger( name );
 		};
-	} );
+	}
+);
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -101,16 +101,14 @@ jQuery.fn.extend( {
 			// Can add propHook for "elements" to filter or add form elements
 			var elements = jQuery.prop( this, "elements" );
 			return elements ? jQuery.makeArray( elements ) : this;
-		} )
-		.filter( function() {
+		} ).filter( function() {
 			var type = this.type;
 
 			// Use .is( ":disabled" ) so that fieldset[disabled] works
 			return this.name && !jQuery( this ).is( ":disabled" ) &&
 				rsubmittable.test( this.nodeName ) && !rsubmitterTypes.test( type ) &&
 				( this.checked || !rcheckableType.test( type ) );
-		} )
-		.map( function( _i, elem ) {
+		} ).map( function( _i, elem ) {
 			var val = jQuery( this ).val();
 
 			if ( val == null ) {
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -45,7 +45,8 @@ var
 
 	// Anchor tag for parsing the document origin
 	originAnchor = document.createElement( "a" );
-	originAnchor.href = location.href;
+
+originAnchor.href = location.href;
 
 // Base "constructor" for jQuery.ajaxPrefilter and jQuery.ajaxTransport
 function addToPrefiltersOrTransports( structure ) {
@@ -426,8 +427,8 @@ jQuery.extend( {
 			// Context for global events is callbackContext if it is a DOM node or jQuery collection
 			globalEventContext = s.context &&
 				( callbackContext.nodeType || callbackContext.jquery ) ?
-					jQuery( callbackContext ) :
-					jQuery.event,
+				jQuery( callbackContext ) :
+				jQuery.event,
 
 			// Deferreds
 			deferred = jQuery.Deferred(),
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -8,9 +8,9 @@ jQuery.ajaxSettings.xhr = function() {
 
 var xhrSuccessStatus = {
 
-		// File protocol always yields status code 0, assume 200
-		0: 200
-	};
+	// File protocol always yields status code 0, assume 200
+	0: 200
+};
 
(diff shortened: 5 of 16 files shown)


wonseop:
I don't know why the Travis CI's test fails. 
This is an indentation correction, and has no effect on the behavior.

mgol:
Thanks for the PR!

I'm wondering if we shouldn't just enable the `indent` option, at least in `src/.eslintrc.json`. I'm sure there are some violations but I've heard that rule has improved a lot recently so maybe it'd be feasible for us to use it. That way we won't need to apply such small indentation fixes but we'll be sure it's fine all the time.

Docs for the indent rule: https://eslint.org/docs/rules/indent.

Would you like to have a look at that?

wonseop:
If you add the following to ".eslintrc-node.json", enforcing consistent indentation is possible in that case.

```
    ...
    "rules": {
        "indent": ["error", "tab", {
            "flatTernaryExpressions": true,
            "outerIIFEBody": 0
        }],
    ...
````
If you check "src/selector.js" in master branch using CLI, you can get the following result.

```
$ eslint -c .eslintrc.json src/selector.js
 
/home/wskim/Projects/jquery/src/selector.js
  606:1  error  Expected indentation of 5 tabs but found 1 space  indent
  607:1  error  Expected indentation of 4 tabs but found 0        indent
 
✖ 2 problems (2 errors, 0 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.
```

However, there are the following issues to apply the above settings.
Currently, when calling "grunt eslint:dev", eslint does not work properly in "src/**/*.js", "Gruntfile.js", and "test/**/*.js".
If you want to use ".eslintrc.json" when calling "grunt eslint:dev", the following code should be added to "Gruntfile.js".

```
        ...
        eslint: {
            ...
            dev: {            // 90 line
                options: {    // new options for eslint:dev
 
                    // Add config
                    // https://www.npmjs.com/package/grunt-eslint#custom-config-and-rules
                    configFile: '.eslintrc.json',
                },
                src: [ "src/**/*.js", "Gruntfile.js", "test/**/*.js", "build/**/*.js" ]
            }
        ...
```

Currently, the code styles of "src/**/*.js" and "test/**/*.js" are very different from ".eslintrc.json",
so 2477 errors occur with only the default settings,
and when the "indent" option is applied, 7486 errors occur.

In my opinion, applying eslint right away and fixing all errors seems difficult, and it seems to require step-by-step correction.
I think it would be nice to proceed like this:

  1. Partially add rules of ".eslintrc-node.json"
  2. Fix source code based on ".eslintrc-node.json"
  3. Repeat 1 and 2
  4. Fix "grunt eslint:dev" to work with ".eslintrc-node.json"

Please review the above.

File: src/.eslintrc.json [hunk 1]
```diff
@@ -23,7 +27,10 @@
 				"sourceType": "script"
 			},
 			"rules": {
-				"no-unused-vars": "off"
+				"no-unused-vars": "off",
+				"indent": [ "error", "tab", {
+					"ignoredNodes": [ "Program > ExpressionStatement > CallExpression > FunctionExpression > *"	]
```
mgol, line 32: Can you explain this line?

src/.eslintrc.json
```diff
@@ -12,7 +12,11 @@
 
 	"rules": {
 		"import/extensions": [ "error", "always" ],
-		"import/no-cycle": "error"
+		"import/no-cycle": "error",
+		"indent": [ "error", "tab", {
+			"MemberExpression": 1,
```
mgol, line 17: `1` seems to be the default so this shouldn't be necessary.

src/.eslintrc.json
```diff
@@ -12,7 +12,11 @@
...
+			"outerIIFEBody": 0
```
mgol, line 18: Please use `"off"`:
```suggestion
			"outerIIFEBody": "off"
```

src/deprecated/event.js
```diff
@@ -30,13 +30,12 @@ jQuery.fn.extend( {
...
-	function( _i, name ) {
-
-		// Handle event binding
-		jQuery.fn[ name ] = function( data, fn ) {
-			return arguments.length > 0 ?
-				this.on( name, null, data, fn ) :
-				this.trigger( name );
-		};
-	} );
+	"change select submit keydown keypress keyup contextmenu" ).split( " " ), function( _i, name ) {
```
mgol, line 33: This line is a bit long; could we leave it as it was before?

src/serialize.js
```diff
@@ -102,29 +102,29 @@ jQuery.fn.extend( {
...
-
-			if ( Array.isArray( val ) ) {
-				return jQuery.map( val, function( val ) {
-					return { name: elem.name, value: val.replace( rCRLF, "\r\n" ) };
-				} );
-			}
-
-			return { name: elem.name, value: val.replace( rCRLF, "\r\n" ) };
-		} ).get();
+			.filter( function() {
```
mgol, line 105: Could you also move the `.map` a few lines above to the next line so that all callbacks are indented the same way?

src/.eslintrc.json
wonseop: As it was implemented according to the JSON grammar, I cannot add comments.

Unlike other codes, `wrapper.js` is implemented in UMD, and is implemented as follows.

```
// jQuery's UMD Code Style
// https://contribute.jquery.org/style-guide/js/#full-file-closures.
( function( factory ) {

    // Indentation: 1

} ( function() {

// Indentation: 0

} ) );
```

This style violates new eslint indentation rule.

```
// New eslint rule
( function( factory ) {

// Immediately Invoked Function Expression's indentation: 0

} ( function() {

    // Normal function's indentation: 1

} ) );
```

Therefore, `wrapper.js` required an specific exception.
I added "ignoredNodes" rule that indentation check is not performed for 1 depth of outer FunctionExpressions by analyzing the syntax tree of `wrapper.js` at https://esprima.org/demo/parse.html.

```
// Specific exception for UMD
( function( factory ) {

    var a;          // <-- ignore (1 depth)

    function b() {  // <-- ignore (1 depth)
        var c;      // <-- check! (2 depth ~)
    }

} ( function() {

var a;          // <-- ignore (1 depth)

function b() {  // <-- ignore (1 depth)
    var c;      // <-- check! (2 depth ~)
}

} ) );
```

I think it's for the best...

src/.eslintrc.json
wonseop: I will delete it

src/.eslintrc.json
wonseop: When I change the value to "off", grunt does not work.

```
$ grunt
Running "eslint:dev" (eslint) task
Warning: src/.eslintrc.json:
	Configuration for rule "indent" is invalid:
	Value "off" should be integer.
```

src/deprecated/event.js
wonseop: I will modify it again to minimize code changes.

src/serialize.js
wonseop: I will fix it

src/.eslintrc.json
mgol: You can add comments to ESLint config files! It’s not a true JSON, it’s “JSON with comments”. And we already use comments in some of them.

wonseop: I add comments.

src/deprecated/event.js
```diff
@@ -28,7 +28,8 @@ jQuery.fn.extend( {
 	}
 } );
 
-jQuery.each( ( "blur focus focusin focusout resize scroll click dblclick " +
+jQuery.each(
+	( "blur focus focusin focusout resize scroll click dblclick " +
```
mgol, line 32: Was this change necessary?

wonseop: The current code needs to be modified as follows according to the new rules, and changes of 9 lines occur.

```
jQuery.each( ( "blur focus focusin focusout resize scroll click dblclick " +
	"mousedown mouseup mousemove mouseover mouseout mouseenter mouseleave " +
	"change select submit keydown keypress keyup contextmenu" ).split( " " ),
function( _i, name ) {   // <-- start

	// Handle event binding
	jQuery.fn[ name ] = function( data, fn ) {
		return arguments.length > 0 ?
			this.on( name, null, data, fn ) :
			this.trigger( name );
	};
} );                   // <-- end
```

wonseop: If there is a better way, please let me know.

mgol: Thanks, the current way makes sense to me then.

mgol:
@timmywil We didn't discuss it yet, I wanted a final team approval so let me add the label back.

mgol:
Landed, thanks, @wonseop!

PR #4647, opened 2020-03-25
Manipulation: Skip the select wrapper for <option> outside of IE 9

mgol:
### Summary ###

Skip the select wrapper for <option> outside of IE 9.

~~+29~~ +20 bytes.

### Checklist ###

* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [ ] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #4642, opened 2020-03-14
Manipulation: Make jQuery.htmlPrefilter an identity function

mgol:
### Summary ###

Make tests pass with jQuery.htmlPrefilter being an identity function.

### Checklist ###

* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* ~~New tests have been added to show the fix or feature works~~
* [x] Grunt build and unit tests pass locally with these changes
* ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~

PR #4685, opened 2020-04-23
Tests: Add tests for recently fixed manipulation XSS issues

mgol:
### Summary ###

Add tests for recently fixed manipulation XSS issues.
Ref gh-4642
Ref gh-4647

### Checklist ###

### DIFF
--- a/test/unit/manipulation.js
+++ b/test/unit/manipulation.js
@@ -2912,3 +2912,52 @@ testIframe(
 	// See https://web.archive.org/web/20171203124125/https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13246371/
 	QUnit[ /\bedge\//i.test( navigator.userAgent ) ? "skip" : "test" ]
 );
+
+QUnit.test( "Sanitized HTML doesn't get unsanitized", function( assert ) {
+
+	var container,
+		counter = 0,
+		assertCount = 13,
+		done = assert.async( assertCount );
+
+	assert.expect( assertCount );
+
+	Globals.register( "xss" );
+	window.xss = sinon.spy();
+
+	container = jQuery( "<div></div>" );
+	container.appendTo( "#qunit-fixture" );
+
+	function test( htmlString ) {
+		var currCounter = counter,
+			div = jQuery( "<div></div>" );
+
+		counter++;
+
+		div.appendTo( container );
+		div.html( htmlString );
+
+		setTimeout( function() {
+			assert.ok( window.xss.withArgs( currCounter ).notCalled,
+				"Insecure code wasn't executed, input: " + htmlString );
+			done();
+		}, 1000 );
+	}
+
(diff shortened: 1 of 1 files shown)


test/unit/manipulation.js
```diff
@@ -2912,3 +2912,52 @@ testIframe(
...
+		counter++;
+
+		div.appendTo( container );
+		div.html( htmlString );
+
+		setTimeout( function() {
+			assert.ok( window.xss.withArgs( currCounter ).notCalled,
+				"Insecure code wasn't executed, input: " + htmlString );
+			done();
+		}, 1000 );
```
gibson042, line 2944: It would be nice to reduce this forced one-second frequency... what would you think about something like so?
```js
div.appendTo( container );
div.html( htmlString );
container.append("<img src=x onerror=\"xssDone(counter)\">")
```

mgol: I'll try that, thanks!

mgol: @gibson042 Now that I think about it... How do we know this `onerror` callback will be fired as the last one? 404 responses may not be cached AFAIK so there's a risk that one will fire before the other ones are done.

gibson042: Well, `src="#"` would guarantee that the invalid source is cached.

mgol: That will resolve to the current page, though, so it won't be a 404.

Also, currently all the tests succeed even if cache is disabled in the browser DevTools. I'd like to keep that working.

mgol: @gibson042 I'm thinking about ways to reduce this delay but if I don't find anything reliable that'd work with cache disabled, I think I'll just merge it, we can always apply improvements later.

gibson042: We don't need a 404, just need a response that isn't a successful image. And I don't see any cache concerns, but if you want to merge without the acceleration then I won't object.

mgol: Good point about not needing 404s. I tested your suggestion, though, and - if I understood it correctly - it looks like I was right about race conditions. See the code at:
https://github.com/mgol/jquery/commit/7cd593f2ef49e2d4d14cce55f7435691ca33ad73
I reverted the two security fixes locally & all assertions should fail but it's pretty random for me; on each refresh different ones are failing for me. I tried the `esmodules` tests.

mgol: An example test run below, I only run this single test. Note that the order of reported assertions is also pretty random.
Screen Shot 2020-04-28 at 23 46 05

gibson042: Yeah, raciness confirmed. I don't think that's fatal to the testing, but it would require restructuring. No worries.

mgol:
Looks like one of the added tests fails in iOS 8 - 12. A followup fix PR, PTAL: #4694.

PR #4461, opened 2019-08-19
Core: Deprecate jQuery.trim

ShashankaNataraj:
### Summary ###

Component: Core
 
- Moved `jQuery.trim()` from the `core.js` file into the `deprecated.js` file
- Moved corresponding unit test from `test/unit/core.js` to `test/unit/deprecated.js`
 
Fixes #4363

### Checklist ###

* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [x] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #4695, opened 2020-04-29
Deprecated: Remove jQuery.trim

mgol:
### Summary ###

The API has been deprecated in 3.5.0 so it can be removed in 4.0.0.

Ref gh-4461

### Checklist ###

### DIFF
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -1,6 +1,5 @@
 import jQuery from "./core.js";
 import slice from "./var/slice.js";
-import trim from "./var/trim.js";
 
 import "./deprecated/ajax-event-alias.js";
 import "./deprecated/event.js";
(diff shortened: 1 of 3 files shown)


mgol:
@dmethvin `String.prototype.trim.apply(s)` works if `s` is a number or an object, etc. However, it throws if `s` is `null` or `undefined`. So it's not a complete replacement.

Maybe we need to update the note in Migrate?

dmethvin:
We can just add a note in the `warnings.md` there explaining that something like `String.prototype.trim.apply(s==null? "" : s)` works perhaps? In any case I think we can handle it in Migrate.

mgol:
@dmethvin that sounds good!

mgol:
@dmethvin I submitted a PR with the text change: https://github.com/jquery/jquery-migrate/pull/358.

Issue #3804, opened 2017-10-01
Non-passive even listener violation


neutralboy:
[Here's the link to the picture on how it looks](https://i.imgur.com/7Djcxy7.png)

Im using Jquery with UIkit and it threw me this error when I used a side nav.
It also throws a google chrome error to [this link.](https://www.chromestatus.com/feature/5745543795965952)

mgol:
jQuery doesn't support passive event listeners yet; this is planned for 4.0 and tracked in #2871. I'm closing it as a duplicate.

Issue #2871, opened 2016-01-26
Consider adding support for passive event listeners


RByers:
Blink has shipped support for [EventListenerOptions](https://dom.spec.whatwg.org/#dictdef-eventlisteneroptions) and we expect to soon ship support for the [passive option](https://github.com/RByers/EventListenerOptions/blob/gh-pages/explainer.md).

Ideally jquery users would be able to mark event listeners as passive, so they can get the same performance benefits.  Also @scottgonzalez mentions it would be good for jquery to polyfill the capturing API as well.

Thoughts?  Feel free to file any issues / questions with the spec [here](https://github.com/RByers/EventListenerOptions/issues).

dmethvin:
What kind of support are you looking for?

I think it would be really hard to expose either `capture` or `passive` via the classic `.on()` API which has been built with the assumption (by both us and users) of a single multiplexed native event handler. Special event hooks can take advantage of being able to capture events before bubbling jQuery handlers see them. All of jQuery's DOM manipulation knows about jQuery events so it can clone them and/or do teardown when the elements go away.

We could create a *new* API which is what #1735 is about, but there are [several hard issues to solve](https://github.com/jquery/jquery/issues/1735#issuecomment-70377782) if we want that to be the foundation for `.on()`. If these use cases are limited to scroll scenarios it seems like a plugin might be the best approach for now. That way we can avoid solving problems that don't (yet) exist for the general case.

Also, what is a reliable way to check for `passive` support? Do all non-`passive` implementations reliably throw an exception if they see an object there instead of a boolean? Are there (or might there be in the future) implementations that support `{ capture: true }`  but just ignore other options they don't understand?

RByers:
> What kind of support are you looking for?

I don't presume to know enough about event handling in jQuery and the typical usage patterns to have a valuable opinion on what the right design is.  But at a high-level, the thing you want for maximizing scroll performance is some way for the code attaching the listener to disable the use of `preventDefault`.

This could even be just more syntax for the event name list.  Eg. some knob that lets developers opt-in to passive-by-default for `touchstart` and `touchmove` events (`wheel` is currently harder due to the lack of equivalent to `touch-action`), then some escape hatch if necessary like `.on("touchstart:withPreventDefault", handler)`.

[...]

PR #4700, opened 2020-04-30
Update event.js manage passive event listener

SoraWebsoft:
### Summary ###
Correct version of manage passive event Listener to correct Best Practice for Google / Lighthouse
See video for view impact on web page : https://www.youtube.com/watch?v=NPM6172J22g

Issue  #3804 - Non-passive even listener violation

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -14,7 +14,28 @@ import "./selector.js";
 var
 	rkeyEvent = /^key/,
 	rmouseEvent = /^(?:mouse|pointer|contextmenu|drag|drop)|click/,
-	rtypenamespace = /^([^.]*)(?:\.(.+)|)/;
+	rtypenamespace = /^([^.]*)(?:\.(.+)|)/,
+	passiveSupported = false,
+	passiveListeners = [],
+	options = null;
+
+// Passive Listener Check
+// We check if the web browser manage passive listeners
+try {
+	options = Object.defineProperty( {}, "passive", { get: function() {
+			passiveSupported = true;
+
+			// If web browser manage passive listeners, then we pass in array with all events to put in passive listener
+			passiveListeners = [ "wheel", "mousewheel", "touchmove", "touchstart", "touchend" ];
+		}
+	} );
+	window.addEventListener( "test", options, options );
+	window.removeEventListener( "test", options, options );
+} catch ( err ) {
+	passiveSupported = false;
+}
+
+// END Passive Listener Check
 
 function returnTrue() {
 	return true;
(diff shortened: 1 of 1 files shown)


mgol:
Thanks for the PR but this is not so simple. We need to support non-passive event listeners as well. Fixing the issue will require a significant refactor of the whole event subsystem in jQuery which is planned for 4.0.0 & tracked in #2871.

SoraWebsoft:
> Thanks for the PR but this is not so simple. We need to support non-passive event listeners as well. Fixing the issue will require a significant refactor of the whole event subsystem in jQuery which is planned for 4.0.0 & tracked in #2871.

I have tested this in my jQuery 3.5.0 version and it's work. 
That's why I wanted to share this patch.

Regards.

src/event.js
```diff
@@ -14,7 +14,28 @@ import "./selector.js";
...
+			passiveSupported = true;
+
+			// If web browser manage passive listeners, then we pass in array with all events to put in passive listener
+			passiveListeners = [ "wheel", "mousewheel", "touchmove", "touchstart", "touchend" ];
+		}
+	} );
+	window.addEventListener( "test", options, options );
+	window.removeEventListener( "test", options, options );
+} catch ( err ) {
+	passiveSupported = false;
```
dmethvin, line 35: Do we need this since it's already initialized to `false`?

src/event.js
```diff
@@ -14,7 +14,28 @@ import "./selector.js";
 var
 	rkeyEvent = /^key/,
 	rmouseEvent = /^(?:mouse|pointer|contextmenu|drag|drop)|click/,
-	rtypenamespace = /^([^.]*)(?:\.(.+)|)/;
+	rtypenamespace = /^([^.]*)(?:\.(.+)|)/,
+	passiveSupported = false,
+	passiveListeners = [],
+	options = null;
+
+// Passive Listener Check
```
dmethvin, line 22: Don't really need these comments.

src/event.js
```diff
@@ -189,7 +210,21 @@ jQuery.event = {
...
-						elem.addEventListener( type, eventHandle );
+
+						// Passive Listener, If web browser manage passive listeners
+						if ( passiveSupported === true ) {
+
+							// If current listener (type) is in passive listener list
+							if ( passiveListeners.indexOf( type ) >= 0 ) {
+
+								// We declare it to passive listener {passive: true}
+								elem.addEventListener( type, eventHandle, { passive: true } );
```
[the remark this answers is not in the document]
mgol: Exactly. It's a frequent use case to call `preventDefault()` on a wheel event.

Also, touch event listeners are passive by default in latest Chrome, Firefox & Safari so for touch events this is not even needed.

PR #4705, opened 2020-05-04
Build: Followups after introducing ES modules compiled via Rollup

mgol:
### Summary ###

This PR cleans up a few comments & configurations that are out of date
after the migration to ES modules backed by a Rollup-based compilation.

Also, de-indent AMD modules. This will preserve a more similar
structure to the one on 3.x-stable where the body of the main `define`
wrapper is not indented.

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -576,7 +576,7 @@ jQuery.extend( {
 		}
 
 		// We can fire global events as of now if asked to
-		// Don't fire events if jQuery.event is undefined in an ESM-usage scenario (#15118)
+		// Don't fire events if jQuery.event is undefined in an ESM-usage scenario (trac-15118)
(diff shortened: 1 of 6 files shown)


src/ajax.js
```diff
@@ -575,7 +575,7 @@ jQuery.extend( {
 		}
 
 		// We can fire global events as of now if asked to
-		// Don't fire events if jQuery.event is undefined in an AMD-usage scenario (#15118)
+		// Don't fire events if jQuery.event is undefined in an ESM-usage scenario (#15118)
```
dmethvin, line 578:
```suggestion
		// Don't fire events if jQuery.event is undefined in an ESM-usage scenario (trac-15118)
```

mgol: Good point. I might do a separate PR in the future to update all the references.

PR #4708, opened 2020-05-05
Serialize: jQuery.param: Allow serialization of standalone keys

### LINKED ISSUE #4707: $.param - cannot unambiguously serialize null relative to empty string
### Description ###
In many application environments there is a distinction between the `null` value and the empty string value; it could be useful to provide an option to allow developers to serialize data via jQuery.param in a way that disambiguates between those two conditions.

No application-specific approach is suggested here, but as an example, an environment might wish to choose that `?x=1&y=&z` infers `x=1, y='', z=null`

Allowing serialization with abbreviated keys (i.e. `$.param({'x': null, 'y': 0}, ...) === '?x&y=0'`) would also allow some clients to save a marginal character in application environments where the server identifies query-string keys by splitting on ampersand (`&`) characters and already provides a default null/empty value.

### Link to test case ###
An illustrative test case is provided here:

https://github.com/jayaddison/jquery/blob/79825abf26c5430dbe27003dbc9bdc59be8dc78b/test/unit/serialize.js#L79-L80

### References ###
- https://bugs.jquery.com/ticket/8653#comment:3
- https://tools.ietf.org/html/rfc3986#section-3.4
- https://github.com/psf/requests/issues/2651
- https://bugs.python.org/issue18857
- https://stackoverflow.com/questions/4557387/is-a-url-query-parameter-valid-if-it-has-no-value
- https://github.com/sindresorhus/query-string#falsy-values

Edit: add reference to `sindresorhus/query-string`

jayaddison:
### Summary ###
This change introduces an optional `standaloneKeys` parameter to the [jQuery.param](https://api.jquery.com/jQuery.param/) function, allowing a developer to specify that they would like to allow null-valued keys to be abbreviated (rendered without a trailing `=`) in the serialized output. 

Resolves #4707

### Checklist ###

### DIFF
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -52,7 +52,7 @@ function buildParams( prefix, obj, traditional, add ) {
 
 // Serialize an array of form elements or a set of
 // key/values into a query string
-jQuery.param = function( a, traditional ) {
+jQuery.param = function( a, traditional, standaloneKeys ) {
 	var prefix,
 		s = [],
 		add = function( key, valueOrFunction ) {
@@ -62,7 +62,8 @@ jQuery.param = function( a, traditional ) {
 				valueOrFunction() :
 				valueOrFunction;
 
-			s[ s.length ] = encodeURIComponent( key ) + "=" +
+			s[ s.length ] = encodeURIComponent( key ) +
+				( standaloneKeys && value == null ? "" : "=" ) +
 				encodeURIComponent( value == null ? "" : value );
 		};
 
--- a/test/unit/serialize.js
+++ b/test/unit/serialize.js
@@ -1,7 +1,7 @@
 QUnit.module( "serialize", { afterEach: moduleTeardown } );
 
 QUnit.test( "jQuery.param()", function( assert ) {
-	assert.expect( 24 );
+	assert.expect( 25 );
 
 	var params;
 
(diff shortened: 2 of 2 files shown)


dmethvin:
Given that it's been about a decade since we've needed to change serialization, I'll ask whether there is some IETF or other standard that governs this. I'm not even sure how to get a `null` or `undefined` value out of an HTML form element, they generally initialize to an empty string.

jayaddison:
Thanks @dmethvin - certainly worth reviewing this one carefully.

Along with one or two other reading references, there's a link to [RFC 3986](https://tools.ietf.org/html/rfc3986) in the linked issue #4707, and as far as I can tell from that, this should be valid.  Valid is necessary but not sufficient for being acceptable, however :)

Although many use cases will indeed be using serialization of data from input fields, I'm considering scenarios where an application may want to serialize a JavaScript datastructure (potentially containing nulls) in an unambiguous and/or abbreviated form.

This brings to mind that I should add test coverage on recursive & nested object cases, so I'll look into that soon.

mgol:
@jayaddison I don't see anything in the linked RFC section that would justify this change, could you provide more details?

jayaddison:
@mgol Thanks for taking a look.  Yes indeed, nothing in the RFC itself suggests this functionality; it's a proposal that I'm making based on a desire to reliably serialize and subsequently deserialize objects containing null values from a JavaScript application.

Currently this is challenging because there is no robust way to distinguish between an empty string and a null value during serialization operations -- the objects `{a: ''}` and `{a: null}` end up serializing to the same data on-disk/on-the-wire.

These are generally treated as distinct conditions in JavaScript and in many database applications - and my suggestion is that we could safely and backwards-compatibly introduce this measure in order to allow unambiguous data representation for nulls.

I've added some test coverage in b20a747c3ea4152b0a46c945146a4e370fab75d3 to illustrate the collision problem (and to verify the fix).

timmywil:
Thank you for your contribution. We discussed this in the meeting and decided not to implement this in jQuery. If the behavior makes it into the RFC or a widely used specification, we will reconsider.

jayaddison:
@timmywil Understood; thanks for taking the time to discuss and review the change.  I'll reach out again if this becomes standardized more widely.

Issue #4409, opened 2019-05-27
Have Trusted Types API built directly into the jQuery Core Files


ayumihamsaki:
The latest version of jQuery V3.4.1 has 5 `innerHTML()` which are not sanitized by default. Does the jQuery team plan on addressing these DOM XSS things in the near future with regards to the new Trusted Types API see here: https://github.com/WICG/trusted-types

mgol:
Thanks for the report. What would you expect jQuery to do here?

mgol:
Note, though, that jQuery doesn't support experimental APIs as they may change and then we're left with obsolete code. It's fine to experiment & think of the approaches but we'd wait with landing any actual code in the library for the spec to stabilize & to be implemented in at least 2 major browser engines.

dmethvin:
Yeah, this is a tough one. The team has discussed it for years and had issues determining how to make things safer without seriously redesigning the API. Most of our HTML goes through `jQuery.parseHTML()` and `jQuery.htmlPrefilter()` so it is currently possible to try and sanitize through that. There's the one fast path for `.html(string)` that assigns directly to `.innerHTML` but that already goes through the prefilter and could just be taken out if needed. The others are Sizzle feature checks for browsers we won't support in 4.0 anyway.

Perhaps we could recognize a Trusted Type and pass it through as if were HTML, but not actually do any manipulation with it, e.g. `$("div").append(someTrustedType)`. (As an aside, I would hope that a Trusted Type doesn't return an HTML representation when you call `.toString()` on it.)

Seems like they've taken old browsers into account with a [simple polyfill](https://github.com/WICG/trusted-types#tinyfill) that just returns strings if the browser doesn't support Trusted Types.

But given where the implementations are at, it is probably too early to add any code.

ayumihamsaki:
I will add an extra comment as you added a '**needs info**' label. I totally understand that the Trusted Types API is right now in a trial. I thought maybe this issue could be dealt with in your future updates like in V4.0.0 and by then the API would have changed from a trial / draft spec to a more stable spec.

When I did a quick play around with Trusted Types API and jQuery V3.4.1 I got the following results:

![58434496-dc989800-80b3-11e9-9f2f-2eabe3d329a4](https://user-images.githubusercontent.com/17784082/58437472-0789e880-80c2-11e9-9675-e1297f5b80ec.png)

DomPurifer have added this API see here for further details: https://github.com/cure53/DOMPurify#what-about-dompurify-and-trusted-types

[...]

PR #4723, opened 2020-05-31
Manipulation: Avoid concatenating strings in buildFragment

mgol:
### Summary ###

Concatenating HTML strings in buildFragment is a possible security risk as it
creates an opportunity of escaping the concatenated wrapper. It also makes it
impossible to support secure HTML wrappers like
[trusted types](https://web.dev/trusted-types/). It's safer to create wrapper
elements using `document.createElement` & `appendChild`.

The previous way was needed in jQuery <4 because IE <10 doesn't accept table
parts set via `innerHTML`, even if the element which contents are set is
a proper table element, e.g.:
```js
tr.innerHTML = "<td></td>";
```
The whole structure needs to be passed in one HTML string. jQuery 4 drops
support for IE <11 so this is no longer an issue; in older version we'd have to
duplicate the code paths.

IE <10 needed to have `<option>` elements wrapped in
`<select multiple="multiple">` but we no longer need that on master which makes
the `document.createElement` way shorter.

jQuery 1.x sometimes needed to have more than one element in the wrapper that
would precede parts wrapping HTML input so descending needed to use `lastChild`.
Since all wrappers are single-element now, we can use `firstChild` which
compresses better as it's used in other places in the code as well.

All these improvements, apart from making logic more secure, decrease the
gzipped size by 55 bytes.

Ref gh-4409

### Checklist ###

### DIFF
--- a/src/manipulation/buildFragment.js
+++ b/src/manipulation/buildFragment.js
@@ -1,6 +1,8 @@
 import jQuery from "../core.js";
 import toType from "../core/toType.js";
 import isAttached from "../core/isAttached.js";
+import arr from "../var/arr.js";
+import document from "../var/document.js";
 import rtagName from "./var/rtagName.js";
 import rscriptType from "./var/rscriptType.js";
 import wrapMap from "./wrapMap.js";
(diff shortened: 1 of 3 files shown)


mgol:
My mistake, see #4724 instead.

Issue #4417, opened 2019-06-07
Exception when element is removed on blur


paul-f:
### Description ###

Running Chrome 73.0.3683.86 on Windows 7 with jQuery 3.4.1.

jQuery tries to read a property of an undefined, causing an exception. This occurs when an element is removed when it loses focus, during a focusout handler, responding to a jQuery triggered blur() event. The three key operations are all performed via jQuery: $(element).blur(), $(element).on("focusout"), $(element).remove(). 

The exception occurs on line 5467; here is the code up to that line:
```js
					dataPriv.set( this, type, saved );

					// Trigger the native event and capture its result
					// Support: IE <=9 - 11+
					// focus() and blur() are asynchronous
					notAsync = expectSync( this, type );
					this[ type ]();
					result = dataPriv.get( this, type );
					if ( saved !== result || notAsync ) {
						dataPriv.set( this, type, false );
					} else {
						result = {};
					}
					if ( saved !== result ) {

						// Cancel the outer synthetic event
						event.stopImmediatePropagation();
						event.preventDefault();
						return result.value;
					}
```

In the last line, result is undefined. That is because during `this[ type ]()` (which issues the event), the element is removed, causing jQuery to remove its saved information from dataPriv in cleanData() line 6046:
					elem[ dataPriv.expando ] = undefined;

Thus, after `this[ type ]()`, result is undefined, it does not equal saved, so the `dataPriv.set()` of the conditional is executed and result is left undefined. Presumably, `result = ()` needs to be set outside the conditional. This problem is also timing-sensitive -- with the certain break points it may not occur.

This didn't happen in 3.3.1. Note also that jQuery's `blur()` must be used, rather than normal JS `blur()`, in order to go through this code path.

### Link to test case ###
jsfiddle uses 3.3.1, so I didn't set up a test case.

*EDIT by @mgol: I wrapped code pieces in backticks to fix Markdown formatting

mgol:
Thanks for the report!

We advise JS Bin for creating test cases as JSFiddle no longer supports IE; you control the whole HTML there. But even in JSFiddle you can attach a proper jQuery version by uploading a script or adding a script tag in HTML. So please still add a test case. Thanks!

paul-f:
Here's a jsbin link;
https://jsbin.com/wuqumigiya/edit?html,js,console,output

Note that the problem occurs on Chrome, not on Firefox. It is timing/sequence-related.

gibson042:
[...]

PR #4725, opened 2020-06-01
Handle potential null result for leverageNative

radar:
### Summary ###

Possible fix for #4417.

### Checklist ###

----

I am not sure where or how to test that this fix works. Help / direction would be appreciated.

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -561,7 +561,7 @@ function leverageNative( el, type, expectSync ) {
 						// Cancel the outer synthetic event
 						event.stopImmediatePropagation();
 						event.preventDefault();
-						return result.value;
+						return result && result.value;
 					}
 
 				// If this is an inner synthetic event for an event with a bubbling surrogate


mgol:
Thanks for the PR!

The most important part here is how to test this change, though, the source change itself is simple and it has already been suggested by @gibson042 in #4417.

Do you want to write a test?

radar:
I am not sure where to begin writing a test.

Perhaps I could copy over the HTML + JS from https://jsbin.com/wuqumigiya/edit?html,js,console and assert within `test/unit/event.js` that the error does not happen there.

What do you think?

mgol:
jQuery only runs unit tests in CI, we don't have any integration test setup. This poses problems in situations like this one as some focus-related bugs are specific to native events and emulating the behavior in unit tests is non-trivial.

I don't know how to write the test, I tried for a while in the past but I didn't manage to do it before I run out of time. So I can't provide a very concrete advise, unfortunately.

You would need to add a test in [`test/unit/event.js`](https://github.com/jquery/jquery/blob/master/test/unit/event.js), you can have a look at existing focus tests there, maybe you'll find some inspiration.

ryuran:
@radar @mgol After several months It style broken even if the fix is known. I hope it will be merged soon.
We currently are not able to update jQuery because lot of our tests fail on this `value`of undefined.

mgol:
@ryuran We still need to figure out a test for this. If you can help with that then the PR will surely be merged sooner.

mgol:
I submitted a version with a test so I'll close this PR. Sorry! See #4799.

PR #4752, opened 2020-07-18
Build: Use the US spelling of "favor"

Necmttn:
For reference: https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -308,7 +308,7 @@ Martijn W. van der Lee <martijn@vanderlee.com>
 Devin Wilson <dwilson6.github@gmail.com>
 Damian Senn <jquery@topaxi.codes>
 Zack Hall <zackhall@outlook.com>
-Vitality Terziev <vitaliyterziev@gmail.com>
+Vitaliy Terziev <vitaliyterziev@gmail.com>
 Todor Prikumov <tono_pr@abv.bg>
 Bernhard M. Wiedemann <jquerybmw@lsmod.de>
 Jha Naman <createnaman@gmail.com>
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -308,7 +308,7 @@ Martijn W. van der Lee <martijn@vanderlee.com>
 Devin Wilson <dwilson6.github@gmail.com>
 Damian Senn <jquery@topaxi.codes>
 Zack Hall <zackhall@outlook.com>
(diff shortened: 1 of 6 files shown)


AUTHORS.txt
```diff
@@ -308,7 +308,7 @@ Martijn W. van der Lee <martijn@vanderlee.com>
 Devin Wilson <dwilson6.github@gmail.com>
 Damian Senn <jquery@topaxi.codes>
 Zack Hall <zackhall@outlook.com>
-Vitaliy Terziev <vitaliyterziev@gmail.com>
+Vitality Terziev <vitaliyterziev@gmail.com>
```
mgol, line 311: That doesn't look intended.

test/data/jquery-1.9.1.js
```diff
@@ -4328,7 +4328,7 @@ setDocument = Sizzle.setDocument = function( node ) {
 		assert( function( div ) {
 
 			// Select is set to empty string on purpose
-			// This is to test IE's treatment of not explictly
+			// This is to test IE's treatment of not explicitly
```
mgol, line 4331: This is jQuery 1.9.1; it's fixed in time and should never be changed.

test/index.html
```diff
@@ -35,7 +35,7 @@
 			amd = QUnit.config.amd || QUnit.urlParams.amd;
 
 		// Workaround: Remove call to `window.__karma__.loaded()`
-		// in favour of calling `window.__karma__.start()` from `loadTests()`
```
mgol, line 38: It's not incorrect, this is just British spelling. I think in principle we follow American English, though, so it looks like a good change.

Necmttn:
Hey @mgol, just reverted the changes you pointed. 
thank you for checking.

test/unit/selector.js
```diff
@@ -2125,7 +2125,7 @@ QUnit.test( "jQuery.escapeSelector", function( assert ) {
 	assert.equal( jQuery.escapeSelector( "\x20\x21\x78\x79" ), "\\ \\!xy",
 		"Escapes non-word ASCII characters" );
 
-	// Astral symbol (U+1D306 TETRAGRAM FOR CENTRE)
```
mgol, line 2128: This is the official name for the character so it shouldn't be changed: https://www.fileformat.info/info/unicode/char/1d306/index.htm

mgol:
Landed, thanks!

Issue #4756, opened 2020-07-21
QtWebKit: jQuery thinks that DOM elements collection is a function


natipo:
### Description ###
After upgrade from jQuery 3.2.1 to 3.3.0 [wkhtmltopdf ](https://wkhtmltopdf.org/index.html) tool stopped working due to the following error in jQuery:
```javascript
TypeError: 'undefined' is not a function (evaluating 'e.getAttribute("type")')
```

Browser to reproduce: QtWeb 3.8.5 (http://www.qtweb.net/)

The actual problem is inside `isFunction( obj )`. The current fix for web browsers where DOM nodes are being classified as a function is not enough for QtWebKit because types such as `NodeList` or `HTMLCollection` don't have attribute `nodeType` and the `typeof` command returns `function`.

Output using QtWeb JavaScript console:
```javascript
> typeof document.getElementsByTagName("p")
"function"
> document.getElementsByTagName("p").nodeType
undefined
> document.getElementsByTagName("p").toString()
"[object NodeList]"
```
Many systems depend on this wkhtmltopdf tool in terms of converting HTML page to PDF. Since jQuery 4 is about to drop support for old browsers, I would like to bring back support for QtWebKit 4 in jQuery 3.x so I could upgrade the system to the latest and secure version of jQuery 3.x.

### Link to test case ###

1. Create a simple index.html page.
```html
<!DOCTYPE html>
<html>
	<head>
		<script src="https://code.jquery.com/jquery-3.5.1.min.js"></script>
		<script type="text/javascript">
			$( document ).ready(function() {				
				$( "p" ).append( "<p>My second paragraph.</p>" );
			});
		</script>
	</head> 
	<body>
		<h1>My First Heading</h1>
		<p>My first paragraph.</p>
	</body>
</html>
```
2. Download portable version of QtWeb 3.8.5 (http://www.qtweb.net/download.html)
3. Open above page in QtWeb and notice that second paragraph tag is not added, but it should be.
4. Additionally you may enable console in QtWeb `Tools -> Enable Web Inspector` in order to see error details.

mgol:
Thanks for the report. We don't support QtWebKit but I'm not opposed to a simple patch just for the `3.x` line if it's all that's needed.

For reference, the commit that caused this issue is a16339b8933f115da3661f3d3f64854c3fa60bdc. See the linked issues for an explanation about the change.

mgol:
Fixed by #4757.

PR #4757, opened 2020-07-21
Core: Bring back QtWebKit support for jQuery 3.x

natipo:
Fix allows jQuery to recognize that DOM types such as HTMLCollection or NodeList are not functions.

Fixes gh-4756

### Checklist ###

### DIFF
--- a/src/var/isFunction.js
+++ b/src/var/isFunction.js
@@ -1,13 +1,17 @@
-define( function() {
+define( [
+	"./toString"
+], function() {
 	"use strict";
 
 	return function isFunction( obj ) {
 
-		// Support: Chrome <=57, Firefox <=52, QtWeb <=3.8.5, WebKit <=534.34,
-		// wkhtmltopdf tool <=0.12.5
+		// Support: Chrome <=57, Firefox <=52
 		// In some browsers, typeof returns "function" for HTML <object> elements
 		// (i.e., `typeof document.createElement( "object" ) === "function"`).
-		// We don't want to classify *any* DOM node or DOM collection as a function.
+		// We don't want to classify *any* DOM node as a function.
+		// Support: QtWeb <=3.8.5, WebKit <=534.34, wkhtmltopdf tool <=0.12.5
+		// Plus for old WebKit, typeof returns "function" for HTML collections
+		// (e.g., `typeof document.getElementsByTagName("div") === "function"`). (gh-4756)
 		return typeof obj === "function" && typeof obj.nodeType !== "number" &&
 			[ "[object HTMLCollection]", "[object NodeList]" ]
 				.indexOf( toString.call( obj ) ) === -1;
--- a/src/var/isFunction.js
+++ b/src/var/isFunction.js
@@ -3,12 +3,14 @@ define( function() {
 
 	return function isFunction( obj ) {
 
-		// Support: Chrome <=57, Firefox <=52
+		// Support: Chrome <=57, Firefox <=52, QtWeb <=3.8.5, WebKit <=534.34,
+		// wkhtmltopdf tool <=0.12.5
 		// In some browsers, typeof returns "function" for HTML <object> elements
 		// (i.e., `typeof document.createElement( "object" ) === "function"`).
-		// We don't want to classify *any* DOM node as a function.
+		// We don't want to classify *any* DOM node or DOM collection as a function.
 		return typeof obj === "function" && typeof obj.nodeType !== "number" &&
-			[ "[object HTMLCollection]", "[object NodeList]" ].indexOf( obj.toString() ) === -1;
+			[ "[object HTMLCollection]", "[object NodeList]" ]
+				.indexOf( toString.call( obj ) ) === -1;
 	};
 
 } );
--- a/src/var/isFunction.js
+++ b/src/var/isFunction.js
@@ -1,6 +1,4 @@
-define( [
-	"./toString"
-], function( toString ) {
+define( function() {
 	"use strict";
 
 	return function isFunction( obj ) {
--- a/src/var/isFunction.js
+++ b/src/var/isFunction.js
@@ -1,6 +1,6 @@
 define( [
 	"./toString"
-], function() {
+], function( toString ) {
 	"use strict";
 
 	return function isFunction( obj ) {
--- a/src/var/isFunction.js
+++ b/src/var/isFunction.js
@@ -7,7 +7,8 @@ define( function() {
 		// In some browsers, typeof returns "function" for HTML <object> elements
 		// (i.e., `typeof document.createElement( "object" ) === "function"`).
 		// We don't want to classify *any* DOM node as a function.
-		return typeof obj === "function" && typeof obj.nodeType !== "number";
+		return typeof obj === "function" && typeof obj.nodeType !== "number" &&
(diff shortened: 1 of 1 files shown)


mgol:
Thanks for the PR. If you run the full test suite on QtWebKit, does it pass all tests with this PR? We're not officially supporting QtWebKit and I'd rather avoid multiple subsequent PRs fixing other issues if they exist.

natipo:
I just run the full test suite on QtWebKit. Only 39 out of 1061 failed. Actual number of failed tests should be lower if we include polyfills e.g. for `bind`. These failing tests don't seem to be critical, and most of them are styling related. This is an awesome result for officially not supported browser. I am not expecting to fix all issues for QtWebKit, but this one because it is a blocker for me, and everyone who depends on wkhtmltopdf tool and would like to upgrade jQuery.

src/var/isFunction.js
```diff
@@ -7,7 +7,8 @@ define( function() {
 		// In some browsers, typeof returns "function" for HTML <object> elements
 		// (i.e., `typeof document.createElement( "object" ) === "function"`).
 		// We don't want to classify *any* DOM node as a function.
-		return typeof obj === "function" && typeof obj.nodeType !== "number";
+		return typeof obj === "function" && typeof obj.nodeType !== "number" &&
+			[ "[object HTMLCollection]", "[object NodeList]" ].indexOf( obj.toString() ) === -1;
```
mgol, line 11: Are you certain these are the only two needed additions?

mgol, line 11: Please add a support comment mentioning affected environments & their versions. Examples are all the comments starting with `// Support:`, you can find many examples in the code base.

mgol:
Also, we'll need unit tests for this change, even if we won't actively test on that browser in CI. Ideally just invoke a few APIs that are crashing currently because of this issue.

src/var/isFunction.js
natipo: I believe these are the only two really needed.
- `NodeList` is returned when we query elements on `document`.
- `HTMLCollection` is returned when we access node `children` property.


**My investigation**
On Mozilla website (https://developer.mozilla.org/en-US/docs/Web/API) under _**Interfaces**_ section I can see many different `*List` or `*Collection` types. However such collections have to be accessed in a special non generic way.
For example `<select>` tag could return different result depending on how we access its children nodes:
- Generic way to access children
```javascript
> $("select")[0].children.toString()
"[object HTMLCollection]"
```
- Specific way, and it only works because element type is `HTMLSelectElement`
```javascript
> $("select")[0].options.toString()
"[object HTMLOptionsCollection]"
```
@mgol please let me know if you are aware of jQuery accessing DOM elements other way than just querying `document` or accessing `children` property? If so then we should also cover such cases. Otherwise I would keep just these two.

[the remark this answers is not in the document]
natipo: Thank you for the comment, although I am not sure if I understood. Do you mean something like this: `toString.call( obj )`?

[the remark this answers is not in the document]
mgol: @natipo Yes, exactly.

natipo:
There are many existing tests which are failing on QtWeb due to this issue.
image
Within the fix they are passing :)
image
I believe there is no need to add more tests for this particular PR.

src/var/isFunction.js
```diff
@@ -3,12 +3,14 @@ define( function() {
 
 	return function isFunction( obj ) {
 
-		// Support: Chrome <=57, Firefox <=52
+		// Support: Chrome <=57, Firefox <=52, QtWeb <=3.8.5, WebKit <=534.34,
+		// wkhtmltopdf tool <=0.12.5
```
mgol, line 7: I would add a separate support comment below the existing one as the issue it covers is different. See an example here:
https://github.com/jquery/jquery/blob/fa0058af426c4e482059214c29c29f004254d9a1/src/event.js#L366-L370

src/var/isFunction.js
```diff
@@ -3,11 +3,14 @@ define( function() {
...
 		// In some browsers, typeof returns "function" for HTML <object> elements
 		// (i.e., `typeof document.createElement( "object" ) === "function"`).
-		// We don't want to classify *any* DOM node as a function.
-		return typeof obj === "function" && typeof obj.nodeType !== "number";
+		// We don't want to classify *any* DOM node or DOM collection as a function.
+		return typeof obj === "function" && typeof obj.nodeType !== "number" &&
+			[ "[object HTMLCollection]", "[object NodeList]" ]
+				.indexOf( toString.call( obj ) ) === -1;
```
mgol, line 13: You need to import `toString` via AMD's `define` dependency array, see what the top portions of other files look like.

I was surprised this didn't blow up but that's because it exists on `window` as well.

src/var/isFunction.js
natipo: Agree. Checked in already.

src/var/isFunction.js
natipo: Done, thanks.

src/var/isFunction.js
```diff
@@ -7,7 +9,12 @@ define( function() {
 		// In some browsers, typeof returns "function" for HTML <object> elements
 		// (i.e., `typeof document.createElement( "object" ) === "function"`).
 		// We don't want to classify *any* DOM node as a function.
-		return typeof obj === "function" && typeof obj.nodeType !== "number";
+		// Support: QtWeb <=3.8.5, WebKit <=534.34, wkhtmltopdf tool <=0.12.5
+		// Plus for old WebKit, typeof returns "function" for HTML collections
+		// (e.g., `typeof document.getElementsByTagName("div") === "function"`). (gh-4756)
+		return typeof obj === "function" && typeof obj.nodeType !== "number" &&
+			[ "[object HTMLCollection]", "[object NodeList]" ]
+				.indexOf( toString.call( obj ) ) === -1;
```
gibson042, line 17: I'd like to see something a bit less fine-tuned.
```suggestion
			typeof obj.item !== "function";
```

src/var/isFunction.js
natipo: Not applicable anymore due to new, more generic solution i.e. `obj.item !== "function`.

mgol:
@gibson042 what explicit test would you like to see here? This is an internal util, bugs in it only manifest through usage of other official APIs. In a way, this is already as explicit as possible.

natipo:
Actually I could add following unit test for this PR. It simulates old WebKit DOM collection so we can reproduce this issue even in new Chrome. It is a bit tricky though so please let me know what do you think.
```javascript
QUnit.test( "jQuery should not classify DOM collection as a function (#gh-4756)", function( assert ) {
	assert.expect( 4 );

	// Simulate DOMCollection object from WebKit 534.34 with 3 elements
	// 1. Should be mistaken with function, i.e. typeof obj === "function"
	function obj( a, b, c ) { }
	// 2. Should have length
	assert.equal( obj.length, 3, "Collection length should be 3." );
	// 3. Should be like collection, i.e. typeof obj.item === "function"
	obj.item = function( i ) { return obj[ i ]; };
	// 4. Should have 3 valid items
	obj[ 0 ] = obj[ 1 ] = obj[ 2 ] = 0;

	// Call isFunction( obj ) through isArrayLike
	jQuery.each( obj, function( i, x ) { assert.equal( x, obj[ i ], "Could read element at index: " + i ); } );
} );
```
Without fix:
image

With fix:
image

Issue #3969, opened 2018-02-07
getScript requires 'unsafe-inline' CSP rule


koczkatamas:
jQuery should be able to load scripts via `$.getScript` without causing `unsafe-inline` CSP violations.

Look at the following test case:

**index.html**

```html
<meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self' https://code.jquery.com 'sha256-DOU0/sVUxitk2MNb5zXL6o3XIe+ZQ5oKG5jmGDO9PmQ='">
<script src="https://code.jquery.com/jquery-3.3.1.js"></script>
<input id="btnJQuery" type="button" value="Load script via jQuery" />
<input id="btnScriptTag" type="button" value="Load script via script tag" />

<script>
  console.log("start");
  
  $("#btnJQuery").on("click", function() {
    $.getScript("payload.js");
  });
  
  $("#btnScriptTag").on("click", function() {
    let scriptEl = document.createElement("script");
    scriptEl.onload = function() { console.log('onload', arguments); };
    scriptEl.onerror = function() { console.log('onerror', arguments); };
    scriptEl.src = "payload.js";
    document.head.appendChild(scriptEl);  
  });
</script>
```

**payload.js**
```
console.log("OK");
```

If you click on the `Load script via jQuery` button the following error is shown in the browser's console (Chrome in this case):

> jquery-3.3.1.js:111 Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' https://code.jquery.com 'sha256-DOU0/sVUxitk2MNb5zXL6o3XIe+ZQ5oKG5jmGDO9PmQ='". Either the 'unsafe-inline' keyword, a hash ('sha256-oKrRnv4NSECe1z+2Q8HSC47J9uP6ALT0+UguOrbK7UU='), or a nonce ('nonce-...') is required to enable inline execution.

On the other hand if you click on the `Load script via script tag` button then the script is loaded without issues and the text `OK` is shown in the console.

**Note:** I am reporting this issue, because third-party libraries use `$.getScript()` and they consider this a jQuery issue (rightly) and won't fix this by their own because they expect the fix from jQuery.

For example see this issue: https://github.com/jackocnr/intl-tel-input/issues/541

## Security considerations

Using constructs which require `unsafe-inline` or `unsafe-eval` are generally not recommended and the security community considers them as bad practice.

[...]

PR #4763, opened 2020-07-29
Ajax: Avoid CSP errors in the script transport for async requests

mgol:
### Summary ###

Until now, the AJAX script transport only used a script tag to load scripts
for cross-domain requests or ones with `scriptAttrs` set. This commit makes
it also used for all async requests to avoid CSP errors arising from usage
of inline scripts. This also makes `jQuery.getScript` not trigger CSP errors
as it uses the AJAX script transport under the hood.

For sync requests such a change is impossible and that's what `jQuery._evalUrl`
uses. Fixing that is tracked in gh-1895.

The commit also makes other type of requests using the script tag version of the
script transport set its type to "GET", namely async scripts & ones with
`scriptAttrs` set in addition to the existing cross-domain ones.

Fixes gh-3969

### Checklist ###

### DIFF
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -32,7 +32,10 @@ jQuery.ajaxPrefilter( "script", function( s ) {
 	if ( s.cache === undefined ) {
 		s.cache = false;
 	}
-	if ( s.crossDomain ) {
+
+	// These types of requests are handled via a script tag
+	// so force their methods to GET.
+	if ( s.crossDomain || s.async || s.scriptAttrs ) {
 		s.type = "GET";
 	}
 } );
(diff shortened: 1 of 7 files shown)


mgol:
I believe this is not a breaking change. Can we include it in 3.6.0?

src/ajax/script.js
```diff
@@ -32,16 +32,20 @@ jQuery.ajaxPrefilter( "script", function( s ) {
 	if ( s.cache === undefined ) {
 		s.cache = false;
 	}
-	if ( s.crossDomain ) {
+
+	// These types of requests are handled via a script tag
+	// so force their methods to GET.
+	if ( s.async || s.crossDomain || s.scriptAttrs ) {
```
Krinkle, line 38: Did you optimize this order for gzip compression? If not, I suspect placing `s.crossDomain` first might give us a byte or two more.

mgol:
Adding the `Behavior Change` label per @dmethvin's comment: https://github.com/jquery/jquery/pull/4763#pullrequestreview-458847152

src/ajax/script.js
mgol: I checked various reorderings & what you propose (+ a few others) were the smallest. Smaller by... one byte compared to my original version. 😄

+10 bytes now

mgol: PR updated.

PR #4754, opened 2020-07-18
Ajax: Drop the json to jsonp auto-promotion logic

mgol:
### Summary ###

Previously, `jQuery.ajax` with `dataType: 'json'` with a provided callback was
automatically converted to a jsonp request unless one also specified
`jsonp: false`. Today the preferred way of interacting with a cross-domain
backend is CORS which works in all browsers jQuery 4 will support.

Auto-promoting JSON requests to JSONP ones introduces a security issue as the
developer may be unaware they're not just downloading data but executing code
from a remote domain.

This commit disables the auto-promoting logic.

BREAKING CHANGE: to trigger a JSONP request, it's now required to specify
`dataType: "jsonp"`; previously some requests with `dataType: "json"` were
auto-promoted to JSONP.

Fixes gh-1799
Fixes gh-3376

### Checklist ###

* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [x] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #4764, opened 2020-07-29
Tests: Recognize callbacks with dots in the Node.js mock server

mgol:
### Summary ###

This aligns the Node.js server with the previous PHP one in accepting `mock.php`
as a callback which is triggered by a recently added test. This prevents the
request crashing on that Node.js server and printing a JS error:
```
TypeError: Cannot read property '1' of null
```

Ref gh-4754

### Checklist ###

### DIFF
--- a/test/middleware-mockserver.js
+++ b/test/middleware-mockserver.js
@@ -101,7 +101,7 @@ var mocks = {
 		} else if ( req.query.callback ) {
 			callback = Promise.resolve( req.query.callback );
 		} else if ( req.method === "GET" ) {
-			callback = Promise.resolve( req.url.match( /^.+\/([^\/?.]+)\?.+$/ )[ 1 ] );
+			callback = Promise.resolve( req.url.match( /^.+\/([^\/?]+)\?.+$/ )[ 1 ] );
 		} else {
 			callback = getBody( req ).then( function( body ) {
 				return body.trim().replace( "callback=", "" );


Krinkle:
This is usually recommended against because it allows things like `document.<something>`, e.g. `document.body.lastChild.firstChild.click` or some such.

However, for the local mock proxy that seems fine 👍

timmywil:
We're going to check why the filename is used as the callback name.

test/middleware-mockserver.js
```diff
@@ -101,7 +101,7 @@ var mocks = {
 		} else if ( req.query.callback ) {
 			callback = Promise.resolve( req.query.callback );
 		} else if ( req.method === "GET" ) {
-			callback = Promise.resolve( req.url.match( /^.+\/([^\/?.]+)\?.+$/ )[ 1 ] );
+			callback = Promise.resolve( req.url.match( /^.+\/([^\/?]+)\?.+$/ )[ 1 ] );
```
mgol, line 104: Note to self: check why one test is using the file name on the server as the callback name as these should be unrelated. Maybe there's a better way to fix this.

mgol:
@timmywil @Krinkle I've just had another look. Code I'm changing doesn't check the query parameter but a REST-like path. Since we only serve it as `mock.php`, that's the only path you can get via this method so a dot seems safe.

On the other hand, we do not validate the callback parameter value in any way either in the PHP server or in the Node.js one, e.g. the following URL:
http://builds.jenkins.jquery.com/jquery/c18dc49699bc27481a4af36ed1a0ee1b19c6eb03/test/data/mock.php?action=jsonp&callback=document.body.innerHTML=
returns:
```js
document.body.innerHTML=({ "data": {"lang": "en", "length": 25} })
```

Therefore, this PR should be safe to land; we can address callback validation separately if needed. But, since this was the behavior for ages, this shouldn't be a huge issue in practice...

mgol:
Landed on `master` in df6858df2ed3fc5c424591a5e09b900eb4ce0417 and on `3.x-stable` in 4c572a7fee5cb45fd6f328933eee7c95e7de51cd.

Issue #3235, opened 2016-07-15
Remove `event.which` property fill


dmethvin:
IE8 seems to be the last browser that didn't provide this natively. It's a bit tricky to test because it needs a real event and not a simulated one. 

https://github.com/jquery/jquery/pull/3233#discussion_r70875393

See also gh-2337, if we're not filling the property and just passing it on I don't know that we need to formally deprecate it in jQuery or generate a Migrate warning. If browsers ever tried to deprecate it via a console warning, the pass-through we do when copying the property would be the responsibility of the caller. If anything I'd say we add a note in our docs to use `event.button` or `event.buttons` where possible.

mgol:
PR: #4765

mgol:
Why is this marked as a behavior change if all the browsers already support it?

mgol:
@dmethvin Sorry, I didn't notice this issue & your assignment before I made the PR. I reassigned the issue to myself.

dmethvin:
@mgol No problem, thanks for doing it! Should just be a case of deleting the `which` entry in event.js, I wouldn't think it changes any tests.

Issue #4755, opened 2020-07-20
Keyboard event.which incorrectly deprecated or documentation out of date?


ChrisHSandN:
### Description ###
The TypeScript definitions [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jquery/misc.d.ts#L4163) state `event.which` is:

```js
@deprecated ​ Deprecated since 3.3. See \`{@link https://github.com/jquery/api.jquery.com/issues/821 }\`.
```

But on the deprecated 3.3 list there is no mention of deprecation:
https://api.jquery.com/category/deprecated/deprecated-3.3/

On the `event.which` page there is no mention of deprecation:
https://api.jquery.com/event.which/

On the `keydown` page there is no mention of deprecation and it even specifically advises to use `event.which`:
https://api.jquery.com/keydown/
> To determine which key was pressed, examine the event object that is passed to the handler function. While browsers use differing properties to store this information, jQuery normalizes the .which property so you can reliably use it to retrieve the key code.

***
As an aside, surely while the native `event.which` may have been [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which) this is the JQuery `event.which` and this kind of cross browser normalisation is exactly why we use JQuery in the first place?

5 years after those issues were resolved and there is still no completely [reliable native property to use](https://caniuse.com/#search=KeyboardEvent).

mgol:
Thanks for the report. The linked API issue is the one that tracks documenting this.

As for normalizing, jQuery is indeed trying to normalize browser differences but at the same time we try to align with where the standards are going. Instead of undeprecating `which`, we'd rather make sure the replacements can be used.

@dmethvin What do you think?

timmywil:
We're considering un-deprecating this property. Event if `KeyboardEvent.code` and `KeyboardEvent.key` were supported in all our supported browsers, they are not direct replacements.

dmethvin:
For key-related events, this is all `event.which` does:
```js
  if ( event.which == null && rkeyEvent.test( event.type ) ) {
    return event.charCode != null ? event.charCode : event.keyCode;
  }
```
[...]

PR #4765, opened 2020-07-29
Event: Remove the event.which shim

mgol:
### Summary ###

All supported browsers implement this property by themselves. The shim was only
needed for IE <9.

Fixes gh-3235
Ref gh-4755

-101 bytes!

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -11,10 +11,7 @@ import nodeName from "./core/nodeName.js";
 import "./core/init.js";
 import "./selector.js";
 
-var
-	rkeyEvent = /^key/,
-	rmouseEvent = /^(?:mouse|pointer|contextmenu|drag|drop)|click/,
-	rtypenamespace = /^([^.]*)(?:\.(.+)|)/;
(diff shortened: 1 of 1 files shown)


mgol:
Since this was only needed for IE <9, perhaps we could land it in `3.x-stable` as well?

Krinkle:
Code LGTM, but:

@dmethvin wrote at https://github.com/jquery/jquery/issues/3235:
> It's a bit tricky to test because it needs a real event and not a simulated one.

Have we tested this on a standalone page in IE9-11 to confirm once more that it sets `e.which` correctly for key events and mouse buttons?

dmethvin:
If it sets `e.which` at all, that value is used. Since IE was the source of the property it's unlikely they have taken it out in the meantime so I think we're safe.

mgol:
Landed on `master` in 1a5fff4c169dbaa2df72c656868bcf60ed4413d0 & on `3.x-stable` in 2bf9793efc9ff6ab5370119673744e9b43e3b872.

Issue #4771, opened 2020-08-18
JSONP and parsing responseText upon 409 Conflict


fras2560:
Bug Reports:
  Unable to parse responseText of JSONP in a error handler.
  *  What do you expect to happen? Able to parse the JSON
  *  What actually happens? Get a parsing error
  *  Which browsers are affected - Chrome/Firefox?

Currently using Backbone and JQuery. Recently have upgraded from 3.3.1 to 3.5.1. Essentially, backbone is using a jsonp request to delete a model but expecting a 409 due to the model being unable to be deleted. It works as expected and end up in the error handler but the responseJson is not present. I have walked through the stack trace difference between 3.3.1 & 3.5.1. The reason seems to be a parsing error:

`error: Error: jQuery35109401972421271656_1597775421415 was not called at Function.error (3.5.1/jquery.js:334:9) at s.converters.script json (3.5.1/jquery.js:10232:12) at ajaxConvert (jquery/3.5.1/jquery.js:9260:19) at done (3.5.1/jquery.js:9736:15) at XMLHttpRequest.<anonymous> (3.5.1/jquery.js:10047:9)
message: "jQuery35109401972421271656_1597775421415 was not called"
stack: "Error: jQuery35109401972421271656_1597775421415 was not called
state: "parsererror"`

From walking through the stacktrace it seems related to - https://github.com/jquery/jquery/commit/50871a5a85cc802421b40cc67e2830601968affe

mgol:
Thanks for the report.

We considered #4379 to be a bug fix but JSONP is a bit special, indeed - under the hood it's a script but it simulates JSON responses in an environment without a CORS setup and sending JSON payloads on error responses is quite typical there. Maybe we need to revisit the JSONP case.

It'd help if you could provide a minimal test case or - even better - a test case for the jQuery test suite, something like https://github.com/jquery/jquery/blob/82b87f6f0e45ca4e717b4e3a4a20a592709a099f/test/unit/ajax.js#L809-L922 but for JSONP & a different expected result. Would you want to try that?

fras2560:
@mgol  Thanks for the quick response. I will give it a try to make a test case or at the very least a minimal test case. I hope I correctly described the situation but will confirm for sure (bit  complicated with Backbone, Server, and JQuery).

mgol:
It’d help if the test case didn’t contain external libraries. If this is
just about JSONP in error scenarios, any 4xx or 5xx status code should work
for the test case.

Thanks!
-- 
[...]

PR #4773, opened 2020-08-20
Ajax: Execute JSONP error script responses

fras2560:
Fixes: #4771 
However, it should not parse the response if just a script

### DIFF
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -233,6 +233,19 @@ QUnit.assert.ok( true, "mock executed");';
 		}
 	}
 
+	protected function errorWithJSONPAndResponseBodyScript( $req ) {
+		header( 'HTTP/1.0 404 Not Found' );
+		if ( isset( $req->query['withScriptContentType'] ) ) {
+			header( 'Content-Type: application/javascript' );
+		}
+		if ( isset( $req->query['callback'] ) ) {
+			$callback = $req->query['callback'];
+			echo $callback . '( {"status": 404, "msg": "Not Found"} )';
+		} else {
+			echo 'QUnit.assert.ok( false, "Mock return erroneously executed" );';
+		}
+	}
+
 	public function __construct() {
 		$this->cspFile = __DIR__ . '/support/csp.log';
 	}
--- a/test/data/mock.php
+++ b/test/data/mock.php
@@ -233,19 +233,6 @@ QUnit.assert.ok( true, "mock executed");';
 		}
 	}
 
-	protected function errorWithJSONPAndResponseBodyScript( $req ) {
-		header( 'HTTP/1.0 404 Not Found' );
-		if ( isset( $req->query['withScriptContentType'] ) ) {
-			header( 'Content-Type: application/javascript' );
-		}
-		if ( isset( $req->query['callback'] ) ) {
-			$callback = $req->query['callback'];
-			echo $callback . '( {"status": 404, "msg": "Not Found"} )';
-		} else {
-			echo 'QUnit.assert.ok( false, "Mock return erroneously executed" );';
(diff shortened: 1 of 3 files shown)


timmywil:
Thanks for opening a PR. Is there an issue already open to go with this?

fras2560:
@timmywil yes PR #4771

timmywil:
Thanks

test/data/mock.php
```diff
@@ -233,6 +233,19 @@ protected function errorWithScript( $req ) {
 		}
 	}
 
+	protected function errorWithJSONPAndResponseBodyScript( $req ) {
```
mgol, line 236: This function shouldn't be needed since the above `errorWithScript` already does the same, doesn't it?

If you had to add something here, `mock.php` is a PHP file but we also have Karma-based tests with a Node.js server which is what gets run on Travis, for example (or if you run `npm test` locally). That server is defined in `test/middleware-mockserver.js` and it needs to mirror all the definitions here. As the tests already used the `mock.php` name in the URLs, the Node.js server also responds on them.

test/unit/ajax.js
```diff
@@ -806,6 +806,46 @@ QUnit.module( "ajax", {
...
+				assert.ok( true, "Converter for unsuccessful response" );
+			}
+		};
+
+		function request( title, options ) {
+			var testMsg = title + ": expected JSON response body";
+			return jQuery.extend( {
+				beforeSend: function() {
+					jQuery.globalEval = function() {
+						assert.ok( true, "Should eval" );
```
mgol, line 823: Maybe you could check here that it evals what you expect in the test?

test/unit/ajax.js
```diff
@@ -806,6 +806,46 @@ QUnit.module( "ajax", {
...
+					};
+				},
+				complete: function() {
+					jQuery.globalEval = globalEval;
+				},
+				// error is the significant assertion
+				error: function( xhr ) {
+					assert.strictEqual( xhr.responseBody, { id: 1 }, testMsg );
```
mgol, line 831: Where do you get this `{ id: 1 }` from?

test/unit/ajax.js
```diff
@@ -806,6 +806,46 @@ QUnit.module( "ajax", {
...
+				},
+				success: function() {
+					assert.ok( false, "Unanticipated success" );
+				}
+			}, options );
+		}
+		return [
+			request(
```
mgol, line 839: In this case, where you only have one request, wrapping it in a `request` function doesn't really help much. Also, you can just return an object, `ajaxTest` wraps it in a one-element array by itself. You can see many other AJAX tests are structured like that.

mgol (changes requested):
Thanks for the PR! This is a very good start. I added some comments.

mgol:
Landed on `master` in a1e619b03a557b47c3e26a5e74af12b63a0d5e73 and on `3.x-stable` in 3bae54aa50b7485b1f18501dff038a61507012a5.

mgol:
The new test failed on master, probably due to #4754. Investigating...

mgol:
Actually, it's because of #4763. The current solution passes the tests on 3.x but it wouldn't pass a cross domain one and #4763 made more requests like cross domain.

I'm working on a fix, it will need to be applied to both branches.

mgol:
From what I see, JSONP error responses have never been passed to the error callback for cross domain requests. So we can probably leave it as-is on `3.x-stable`, this would be a feature request and with JSONP being a legacy communication method perhaps it's fine this way.

mgol:
Fix for `master` available at #4778. The `3.x-stable` branch is fine as-is.

PR #4778, opened 2020-08-26
Ajax: Make responseJSON work for erroneous same-domain JSONP requests

mgol:
### Summary ###

Don't use a script tag for JSONP requests unless for cross-domain requests
or if scriptAttrs are provided. This makes the `responseJSON` property available
in JSONP error callbacks.

This fixes a regression from jQuery 3.5.0 introduced in gh-4379 which made
erroneous script responses to not be executed to follow native behavior.

The 3.x-stable branch doesn't need this fix as it doesn't use script tags for
regular async requests.

Ref gh-4771
Ref gh-4773
Ref gh-4379

+6 bytes

### Checklist ###

* ~~New tests have been added to show the fix or feature works~~ This fixes an existing test from #4773

### DIFF
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -3,6 +3,22 @@ import document from "../var/document.js";
 
 import "../ajax.js";
 
+function canUseScriptTag( s ) {
+
+	// A script tag can only be used for async, cross domain or forced-by-attrs requests.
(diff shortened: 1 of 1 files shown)


mgol:
+6 bytes

PR #4792, opened 2020-09-21
Core: Drop support for Edge Legacy (i.e. non-Chromium Microsoft Edge)

mgol:
### Summary ###

Drop support for EdgeHTML (i.e. non-Chromium Microsoft Edge).

Also, restrict some workarounds that were applied unconditionally in all
browsers to run only in IE now. This slightly increases the size but reduces
the performance burden on modern browsers that don't need the workarounds.

Also, clean up some comments & remove some obsolete workarounds.

Fixes gh-4568

-255 bytes 🎉

### Checklist ###


src/ajax.js
```diff
@@ -543,7 +543,7 @@ jQuery.extend( {
 		if ( s.crossDomain == null ) {
 			urlAnchor = document.createElement( "a" );
 
-			// Support: IE <=8 - 11+, Edge 12 - 17 only
+			// Support: IE <=8 - 11+
```
mgol, line 546: I removed mentions of Edge from support comments as well. Among other things, this is because many IE workarounds are now hidden behind an `isIE` check so a support comment mentioning Edge would be confusing as Edge wouldn't execute that code.

IE is quite special here as that's the only browser which we detect as a browser to work around its bugs instead of running support tests.

mgol:
Travis fails because tests on Firefox ESR are run on version 68 at the moment. That version doesn't support the `nonce` property yet as that was only fixed in v75. Support for Firefox 68 ESR [ends tomorrow](https://wiki.mozilla.org/Release_Management/Calendar) so I can re-run Travis tomorrow or the day after that.

src/core/isAttached.js
```diff
@@ -4,16 +4,17 @@ import documentElement from "../var/documentElement.js";
 import "../selector/contains.js"; // jQuery.contains
 
 var isAttached = function( elem ) {
-		return jQuery.contains( elem.ownerDocument, elem );
+		return jQuery.contains( elem.ownerDocument, elem ) ||
+			elem.getRootNode( composed ) === elem.ownerDocument;
```
mgol, line 8: Reorganizing the definitions decreased the gzipped size a bit.

src/selector.js
```diff
@@ -229,12 +229,13 @@ function find( selector, context, results, seed ) {
...
-					if ( newContext !== context || !support.scope ) {
+					// Outside of IE, if we're not changing the context we can
+					// use :scope instead of an ID.
+					if ( newContext !== context || isIE ) {
 
-						// Capture the context ID, setting it first if necessary
-						if ( ( nid = context.getAttribute( "id" ) ) ) {
+						// Capture the context ID, setting it first if necessary.
+						nid = context.getAttribute( "id" );
+						if ( nid ) {
```
timmywil, line 238: I actually like this in the `if` better as it's consitent with the expando setting a few lines below.

timmywil (approved):

src/selector.js
mgol: @timmywil Makes sense, I've just reverted this change.

PR #4793, opened 2020-09-21
Build: Explicitly exclude the queue module from the slim build

mgol:
### Summary ###

The queue module is not present in the slim build as it depends on deferred
and our Gruntfile specifies excluding deferred should also exclude queue:
https://github.com/jquery/jquery/blob/3.5.1/Gruntfile.js#L66
This commit makes this exclusion explicit so that the queue module never
accidentally gets re-included in the slim build if it stopped importing from
the deferred module directly.

### Checklist ###

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -60,7 +60,7 @@ module.exports = function( grunt ) {
 		const done = this.async();
 
 		try {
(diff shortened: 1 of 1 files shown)


timmywil (approved):
LGTM

PR #4808, opened 2020-11-10
Dimensions: Add offset prop fallback to FF for unreliable TR dimensions

timmywil:
Fixes gh-4529

### Summary ###

Firefox incorrectly (or perhaps correctly) includes table borders in computed
dimensions, but they are the only one. Workaround this by testing for it and
falling back to offset properties.

This applies to latest FF and so needs to go on master as well.

### Checklist ###

+203 bytes

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -12,6 +12,7 @@ import swap from "./css/var/swap.js";
 import curCSS from "./css/curCSS.js";
 import adjustCSS from "./css/adjustCSS.js";
 import finalPropName from "./css/finalPropName.js";
+import support from "./css/support.js";
 
 import "./core/init.js";
 import "./core/ready.js";
@@ -134,23 +135,24 @@ function getWidthOrHeight( elem, dimension, extra ) {
 	}
 
 
-	if ( ( isIE &&
-		(
-
-			// Support: IE 9 - 11+
-			// Use offsetWidth/offsetHeight for when box sizing is unreliable.
-			// In those cases, the computed value can be trusted to be border-box.
-			isBorderBox ||
-
-			// Support: IE 10 - 11+
-			// IE misreports `getComputedStyle` of table rows with width/height
-			// set in CSS while `offset*` properties report correct values.
-			nodeName( elem, "tr" )
-		) ||
+	if ( (
 
 		// Fall back to offsetWidth/offsetHeight when value is "auto"
 		// This happens for inline elements with no explicit setting (gh-3571)
-		val === "auto" ) &&
+		val === "auto" ||
+
+		// Support: IE 9 - 11+
+		// Use offsetWidth/offsetHeight for when box sizing is unreliable.
+		// In those cases, the computed value can be trusted to be border-box.
+		( isIE && isBorderBox ) ||
+
+		// Support: IE 10 - 11+
+		// IE misreports `getComputedStyle` of table rows with width/height
+		// set in CSS while `offset*` properties report correct values.
+		// Support: Firefox 70+
+		// Firefox includes border widths
+		// in computed dimensions for table rows. (gh-4529)
+		( !support.reliableTrDimensions() && nodeName( elem, "tr" ) ) ) &&
 
 		// Make sure the element is visible & connected
 		elem.getClientRects().length ) {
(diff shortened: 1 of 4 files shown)


src/css.js
```diff
@@ -134,23 +135,24 @@ function getWidthOrHeight( elem, dimension, extra ) {
...
-			// IE misreports `getComputedStyle` of table rows with width/height
-			// set in CSS while `offset*` properties report correct values.
-			nodeName( elem, "tr" )
-		) ||
+	if ( (
 
 		// Fall back to offsetWidth/offsetHeight when value is "auto"
 		// This happens for inline elements with no explicit setting (gh-3571)
-		val === "auto" ) &&
+		val === "auto" ||
```
mgol, line 142: Nice optimization to skip the support test when not needed.

src/css/support.js
```diff
@@ -0,0 +1,53 @@
...
+import support from "../var/support.js";
+
+( function() {
+
+var reliableTrDimensionsVal,
+	div = document.createElement( "div" );
+
+// Finish early in limited (non-browser) environments
+if ( !div.style ) {
+	return;
```
mgol, line 13: Reaching here will make `support.reliableTrDimensions()` crash. I guess this is similar to what we do on `3.x-stable`...

src/css/support.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+jQuery.extend( support, {
```
mgol, line 16: Is it worth using `jQuery.extend` here? Wouldn't assigning directly to `support` be smaller?

src/css/support.js
```diff
@@ -0,0 +1,53 @@
...
+
+	// Support: Firefox 70+
+	// Only Firefox includes border widths
+	// in computed dimensions. (gh-4529)
+	reliableTrDimensions: function() {
```
mgol, line 21: Do we need the entirety of this test when we only test against Firefox behavior?

mgol (commented):
Support tests are coming back! Logic looks good to me, I just hope we can limit the size increase a bit. Some of it is expected, modifying an existing test is cheaper than creating a new one when there's none at the moment.

I added some comments.

src/css/support.js
timmywil: I thought about that. An error is probably a good thing in that case, but perhaps there's a better place for the error?

src/css/support.js
timmywil: Probably worth switching to assignment. Was an artifact from copying over.

src/css/support.js
timmywil: Looking at the test I'm not sure what to remove. We need at least 3 elements, creating them with `createElement` prevents other issues as we've seen, the absolute styling is good practice to hide the table, and so on. Also, I removed the `isIE` check for this case so it's technically also covering IE11. I suppose I should note that in the Support comment in this module as well.

mgol: Yeah, an IE comment would be good. If this test was possible to get shortened in a way that it would still cover non-IE browsers then we could check for IE using our `isIE` variable as we've done so far. But if the test cannot be made shorter, covering IE here is fine.

src/css/support.js
timmywil: Switching to assignment saved us 8 bytes.

src/css/support.js
mgol: It might be OK to leave it here.

src/css.js
```diff
@@ -134,23 +135,24 @@ function getWidthOrHeight( elem, dimension, extra ) {
...
+		// Use offsetWidth/offsetHeight for when box sizing is unreliable.
+		// In those cases, the computed value can be trusted to be border-box.
+		( isIE && isBorderBox ) ||
+
+		// Support: IE 10 - 11+
+		// IE misreports `getComputedStyle` of table rows with width/height
+		// set in CSS while `offset*` properties report correct values.
+		// Support: Firefox 70+
+		// Firefox includes border widths
+		// in computed dimensions for table rows. (gh-4529)
```
mgol, line 154: Back when we had lots of support comment, when code was guarded by one, we usually didn't comment what gets worked around as the support comment above the support test was explaining that anyway. Maybe we can remove this comment in that case here.

mgol: Although, I see we have a similar comment on `3.x-stable` here so I'm fine either way.

mgol (approved):
Looks good, just one doubt about a comment. Thanks! 🙂

Issue #4382, opened 2019-05-01
bug with focus() method in a specific case in 3.4.0


zdravkov:
Hi all,

I just noticed and odd change in the  behavior of the focus method in jQuery 3.4.0 compared with 3.3.1, 2.**, and 1.*.* versions in a specific case when we try to change the focus to the child element in _focusin_ event of its parent. Here is the code and here you can see a [code dojo](https://dojo.telerik.com/@zdravkov/eCuJEhuH/2) with the issue:
```
        <span>Check the console</span>
        <div id="toolbar" tabindex="0">
                 <a tabIndex="0" id="foo" >foo</a>
        </div>
        <script>
            $(document).ready(function () {
                var toolbar = $('#toolbar');

                toolbar.on("focusin", function (ev) {
                        var element = $(this).find("a");

                        if (element.length) {
                            element[0].focus();
                            console.log(document.activeElement)
                        }
                    })

                toolbar[0].focus();
                //toolbar[0].focus(); this works

                console.log('With 3.3.1 here we expect to see "foo":', document.activeElement.id);

            });
        </script>
```
If we use older version or the native focus() method the issue is not present.

Greetings.
Plamen

mgol:
Thanks for the report. The issue doesn't exist on [`master`](https://code.jquery.com/jquery-git.js) (future jQuery 4.0) where we removed our custom `focusin` patch but it still exists on [`3.4-stable`](https://code.jquery.com/jquery-3.x-git.js).

cc @gibson042

timmywil:
@jquery/core Is this something we want to get into 3.4.1?

mgol:
@timmywil While I think this issue should be fixed in a patch update to 3.4, IMO #4350 & #4356 are way more serious and it's been 3 weeks since the 3.4.0 release so I'd rather not postpone 3.4.1 further. We can always do 3.4.2 with a fix for this issue & any other that may pop up in the meantime.

caseyjhol:
More examples here:

- jQuery v3.3.1 (working): https://fresh-capable-sense.glitch.me/jquery-3.3.1.html
- jQuery v3.4.0 (not working): https://fresh-capable-sense.glitch.me/jquery-3.4.0.html
- jQuery v3.5.1 (not working): https://fresh-capable-sense.glitch.me/jquery-3.5.1.html
- Editor: https://glitch.com/edit/#!/fresh-capable-sense

The "Focus this" button element should be focused.

ygj6:
[...]

PR #4813, opened 2020-12-01
Event: Make focus re-triggering not focus the original element back

mgol:
### Summary ###

If during a focus handler another focus event is triggered:

```js
elem1.on( "focus", function() {
	elem2.trigger( "focus" );
} );
```

due to their synchronous nature everywhere outside of IE the hack added in
gh-4279 to leverage native events causes the native `.focus()` method to be
called last for the initial element, making it steal the focus back. To resolve
this, we now skip calling the native method if focus was changed.

A side effect of this change is that now `focusin` will only propagate to the
document for the last focused element. This is a change in behavior but it also
aligns us better with how this works with native methods.

Fixes gh-4382
Ref gh-4279

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -11,7 +11,9 @@ import nodeName from "./core/nodeName.js";
 import "./core/init.js";
 import "./selector.js";
 
-var rtypenamespace = /^([^.]*)(?:\.(.+)|)/;
+var currentFocusTarget,
+	focusTriggersInProgress = 0,
+	rtypenamespace = /^([^.]*)(?:\.(.+)|)/;
 
 function returnTrue() {
 	return true;
@@ -581,6 +583,13 @@ function leverageNative( el, type, expectSync ) {
 			// Fire an inner synthetic event with the original arguments
 			} else if ( saved.length ) {
 
+				// Track the number of focus triggers in progress to know when
+				// `currentFocusTarget` can be cleared to save memory.
+				if ( type === "focus" ) {
+					focusTriggersInProgress++;
+					currentFocusTarget = this;
+				}
+
 				// ...and capture the result
 				dataPriv.set( this, type, {
 					value: jQuery.event.trigger(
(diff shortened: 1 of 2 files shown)


mgol:
+58 bytes. @gibson042 any ideas for size improvements?

dmethvin (approved):
Wow, I forgot about `_default`. The special event plumbing really helps here. I think programmatic forcing of focus is an anti-pattern in general, but this doesn't seem to be too bad as far as code goes.

mgol:
> Wow, I forgot about `_default`.

Yeah, [it's only used for `click` at the moment](https://github.com/jquery/jquery/blob/3.5.1/src/event.js#L512-L520), I'm adding a second entry here. 🙂

mgol:
OK, I pushed another version of this change that just returns `true` in the `focus` & `blur` `_default` unconditionally, it's only +6 bytes. This works because we already call the native method in `leverageNative` and we don't need to call it again. For `click`, `_default` already returns `true` in situations where `leverageNative` is used.

All the tests pass, focus is changed and the `focusin` event propagates to `document` properly. Am I missing anything? If @gibson042 was able to look at it, that'd be great.

@dmethvin can you also re-review this?

src/event.js
```diff
@@ -746,6 +746,12 @@ jQuery.each( { focus: "focusin", blur: "focusout" }, function( type, delegateTyp
 			return true;
 		},
 
+		// Suppress native focus or blur as it's already being fired
+		// in leverageNative.
+		_default: function() {
```
dmethvin, line 751: Yeah I hadn't thought about that. The `_default` action happens at the very very very end of this process so as long as `leverageNative` always does the work for us earlier we don't need to do it here.

dmethvin (approved):

timmywil (approved):
LGTM, I trust @dmethvin here

gibson042 (approved):
I love this solution.

mgol:
Landed on `master` in dbcffb396c2db61ff96edc4162602e850797d61f & on `3.x-stable` in 2fadbc0a9839893f434cfca42587593c8794375d.

Issue #4784, opened 2020-08-31
jQuery.parseXML should add detailed information about parse errors


MatthiasSchmalz:
### Description ###
If an invalid XML is passed to jQuery.parseXML, currently the thrown parse error only contain the input data but not the error reason.
However the internally created XML document contains the parser errors in the element parsererror as innerText and could be added. This is helpful for debugging errors in XML files

### Link to test case ###
https://jsfiddle.net/o678xuz3/

mgol:
Thanks for the report. However, the test case attached doesn't work, it even doesn't include jQuery. Please share a test case where the issue and what you'd expect instead can be reproduced.

MatthiasSchmalz:
I'm sorry I have not described the testcase properly.
It is using jQuery which is available implicitely and it uses the jQuery.parseXML. It logs the error message that is raised by the parser to the console.
Additionally it logs the error which the internally used XML parser produces and which should be returned.

mgol:
What do you propose us to change? `jQuery.parseXML` returns the parsed XML and doesn't crash in case of errors. Do you propose that an error we log to the console contained contents of `xml.getElementsByTagName( "parsererror" )` as well?

MatthiasSchmalz:
I propose to add the xml.getElementsByTagName( "parsererror" )[0].innerText to the text of the thrown Error.

timmywil:
I like this idea. I'd much prefer to either show the parserror content (with provided snippet) or a generic message without trying to add the data to the string.

mgol:
PR: #4816

PR #4816, opened 2020-12-01
Core: Report browser errors in parseXML

mgol:
### Summary ###

Fixes gh-4784

In IE 11 we'd just throw an empty error now, in others - the reported error. We wouldn't include input data in the thrown error anymore.

+10 bytes, although I cheated a bit - I also removed an assignment of `xml` to `undefined` in `catch` as any error would happen before the assignment so the variable would already hold the `undefined` value. Without this optimization, it'd be +13 bytes.

That +10 can be changed to -3 if we just call:
```js
jQuery.error( parserErrorElem && parserErrorElem.textContent )
```
but that has two issues. Here's the returned element `parsererror` (after pretty-printing) in Blink & WebKit:
```xml
<parsererror
	xmlns="http://www.w3.org/1999/xhtml"
	style="display: block; white-space: pre; border: 2px solid #c77; padding: 0 1em 0 1em; margin: 1em; background-color: #fdd; color: black">
	<h3>This page contains the following errors:</h3>
	<div style="font-family:monospace;font-size:12px">error on line 14 at column 9: Opening and ending tag mismatch: th line 0 and tr
</div>
	<h3>Below is a rendering of the page up to the first error.</h3>
</parsererror>
```
and here's the one in Gecko:
```xml
<parsererror xmlns="http://www.mozilla.org/newlayout/xml/parsererror.xml">
	XML Parsing Error: mismatched tag. Expected: &lt;/th&gt;.
Location: http://sandbox.l/test.xhtml
Line Number 14, Column 6:
	<sourcetext>			&lt;/tr&gt;\n--------------------------^</sourcetext>
</parsererror>
```

If we just print `textContent`, in Chrome the text will miss spaces in some places:
```
This page contains the following errors:error on line 14 at column 9: Opening and ending tag mismatch: th line 0 and tr
Below is a rendering of the page up to the first error.
```
That may not be enough a reason to add bytes but here's what ends up in Firefox:
```
XML Parsing Error: mismatched tag. Expected: </th>.
Location: http://sandbox.l/test.xhtml
Line Number 14, Column 6:			</tr>
--------------------------^
```
Firefox prints ASCII art which is misaligned now. With the code from this PR, here's what gets printed:
```
XML Parsing Error: mismatched tag. Expected: </th>.
Location: http://sandbox.l/test.html
Line Number 14, Column 6:
			</tr>
--------------------------^
```

On the other hand, it actually uses tabs to pad `</tr>` and the alignment depends on tab being rendered as 8 characters... So maybe we can give up on that.

**EDIT:** Actually, this will mostly be printed in the DevTools of the respective browser, here: Firefox. So it should show up correctly in the console.

Let's make a decision at the next meeting.

### Checklist ###

### DIFF
--- a/src/core/parseXML.js
+++ b/src/core/parseXML.js
@@ -13,13 +13,15 @@ jQuery.parseXML = function( data ) {
 		xml = ( new window.DOMParser() ).parseFromString( data, "text/xml" );
 	} catch ( e ) {}
 
-	parserErrorElem = xml.getElementsByTagName( "parsererror" )[ 0 ];
+	parserErrorElem = xml && xml.getElementsByTagName( "parsererror" )[ 0 ];
 	if ( !xml || parserErrorElem ) {
-		jQuery.error( parserErrorElem &&
-			jQuery.map( parserErrorElem.childNodes, function( el ) {
-				return el.textContent;
-			} ).join( "\n" )
-		);
+		jQuery.error( "Invalid XML: " + (
+			parserErrorElem ?
+				jQuery.map( parserErrorElem.childNodes, function( el ) {
+					return el.textContent;
+				} ).join( "\n" ) :
+				data
+		) );
 	}
 	return xml;
 };
--- a/src/core/parseXML.js
+++ b/src/core/parseXML.js
@@ -2,7 +2,7 @@ import jQuery from "../core.js";
 
 // Cross-browser xml parsing
 jQuery.parseXML = function( data ) {
-	var xml;
+	var xml, parserErrorElem;
 	if ( !data || typeof data !== "string" ) {
 		return null;
 	}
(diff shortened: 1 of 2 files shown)


dmethvin (commented):
I wonder if we should preserve the old `Invalid XML:` prefix for compat reasons and just check for that in the unit test? We don't control the browser's reporting of XML errors so we can't ensure it will show any specific error message. What do you think? In either case it is an improvement!

src/core/parseXML.js
```diff
@@ -11,12 +11,15 @@ jQuery.parseXML = function( data ) {
...
 		xml = ( new window.DOMParser() ).parseFromString( data, "text/xml" );
-	} catch ( e ) {
-		xml = undefined;
-	}
+	} catch ( e ) {}
 
-	if ( !xml || xml.getElementsByTagName( "parsererror" ).length ) {
-		jQuery.error( "Invalid XML: " + data );
+	parserErrorElem = xml.getElementsByTagName( "parsererror" )[ 0 ];
+	if ( !xml || parserErrorElem ) {
```
campersau, line 17: I guess the `!xml` check should be around `xml.getElementsByTagName` instead?

mgol:
@dmethvin 
> I wonder if we should preserve the old `Invalid XML:` prefix for compat reasons and just check for that in the unit test? We don't control the browser's reporting of XML errors so we can't ensure it will show any specific error message. What do you think? In either case it is an improvement!

I'm not sure either way. Judging by the messages I cited in the PR description, the Chrome error kind of misses the context so the `Invalid XML: ` prefix would help. If we don't include user data in the error, though, in IE the message just ends like that with nothing appearing after the colon. Should we preserve that for IE? @timmywil was against but I don't see the risk - we're sure `data` is a string at this point and throwing an error is safe with any provided string.

src/core/parseXML.js
mgol: Good catch! Fixed.

mgol:
I updated the PR with proposed changes; unfortunately, now it's at +25 bytes.

timmywil (approved):

mgol:
@dmethvin does the current version look good to you?

dmethvin (approved):
LGTM!

mgol:
Landed on:
* `master` in 8969732518470a7f8e654d5bc5be0b0076cb0b87
* `3.x-stable` in 54d98835b6e35e3dcb92a9f30bcac82ad2039487 with a followup fixup in 1ec36332fca66d97a5c3bc7ac1b70692429f5be6.

PR #4824, opened 2021-01-04
Tests: Fix tests for not auto-executing scripts without dataType

mgol:
### Summary ###

Two issues are fixed in testing for responses with a script Content-Type not
getting auto-executed unless an explicit `dataType: "script"` is provided:
* the test is now using a correct "text/javascript" Content-Type; it was using
  "text/html" until now which doesn't really check if the fix works
* the Node.js based version of the tests didn't account for an empty `header`
  query string parameter

Ref gh-2432
Ref gh-2588
Ref 39cdb8c9aa0fde68f733553ba050a2ba9d86474c

### Checklist ###

### DIFF
--- a/test/middleware-mockserver.js
+++ b/test/middleware-mockserver.js
@@ -62,7 +62,7 @@ var mocks = {
 	script: function( req, resp ) {
 		if ( req.query.header === "ecma" ) {
 			resp.writeHead( 200, { "content-type": "application/ecmascript" } );
-		} else if ( req.query.header ) {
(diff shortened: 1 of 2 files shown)


timmywil (approved):
LGTM

mgol:
Landed on `master` in d38528b17a846b7ca4513b41150a05436546292d & on `3.x-stable` in 7298e04faee8b01b335148cf5b91f84af9cf5826.

PR #4864, opened 2021-03-24
Tests: Make more tests run natively in Chrome & Firefox

mgol:
### Summary ###

Chrome & Firefox now support complex `:not()` selectors so those test can run
in them even without custom jQuery selector code. In the past, it was only
possible in Safari, now we only need to exclude IE.

### Checklist ###

### DIFF
--- a/test/unit/selector.js
+++ b/test/unit/selector.js
@@ -1137,14 +1137,14 @@ QUnit.test( "pseudo - :not", function( assert ) {
 	assert.t( ":not() failing interior", "#qunit-fixture p:not(.foo)", [ "firstp", "ap", "sndp", "en", "sap", "first" ] );
 	assert.t( ":not() failing interior", "#qunit-fixture p:not(#blargh)", [ "firstp", "ap", "sndp", "en", "sap", "first" ] );
 
-	if ( QUnit.jQuerySelectors || this.safari ) {
+	if ( QUnit.jQuerySelectors || !QUnit.isIE ) {
 		assert.t( ":not() failing interior", "#qunit-fixture p:not(div.foo)", [ "firstp", "ap", "sndp", "en", "sap", "first" ] );
 		assert.t( ":not() failing interior", "#qunit-fixture p:not(p.foo)", [ "firstp", "ap", "sndp", "en", "sap", "first" ] );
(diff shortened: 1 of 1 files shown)


timmywil (approved):

test/unit/selector.js
```diff
@@ -1137,14 +1137,14 @@ QUnit.test( "pseudo - :not", function( assert ) {
...
-	if ( QUnit.jQuerySelectors || this.safari ) {
+	if ( QUnit.jQuerySelectors || !QUnit.isIE ) {
 		assert.t( ":not() failing interior", "#qunit-fixture p:not(div.foo)", [ "firstp", "ap", "sndp", "en", "sap", "first" ] );
 		assert.t( ":not() failing interior", "#qunit-fixture p:not(p.foo)", [ "firstp", "ap", "sndp", "en", "sap", "first" ] );
 		assert.t( ":not() failing interior", "#qunit-fixture p:not(div#blargh)", [ "firstp", "ap", "sndp", "en", "sap", "first" ] );
 		assert.t( ":not() failing interior", "#qunit-fixture p:not(p#blargh)", [ "firstp", "ap", "sndp", "en", "sap", "first" ] );
 	} else {
-		// Support: Chrome 75+, Firefox 67+
-		// Chrome/Firefox don't support `:not(complex selector)`.
+		// Support: IE 11+
```
Krinkle, line 1146: IE 11 was the last IE, I think!

mgol: Yes, but our syntax allows either `11+` or `11 only` and the team decided the former is better, just in case there is a newer version one day. The `only` versions are mostly when we verified a newer version fixes the issue which cannot be done here.

Krinkle: Ah, I didn't realize there was a technical rule that parses these. np, not worth making more complicated then :)

mgol: We don't have any tooling but we have a grammar defined by @gibson042 in https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-69379197 with my amendment in https://github.com/jquery/contribute.jquery.org/issues/95#issuecomment-448998379.

mary-doc (changes requested):
@matankb

matankb:
Not sure why I was tagged...

PR #4870, opened 2021-04-08
Build: Take core-js-bundle from the external directory as well

mgol:
### Summary ###

Build: Take core-js-bundle from the external directory as well

That package was missed in gh-4865 as it only broke browsers needing the
polyfill which is just IE at the moment. Thus, it broke Core tests in IE only.

Ref gh-4865

### Checklist ###


timmywil (approved):
LGTM

PR #4873, opened 2021-04-12
This is what I think about jQuery supporting antisemitic political movements where donations are fraudulently spent on mansions.

Matthew-Bonner:
This is what I think about jQuery supporting antisemitic political movements where donations are fraudulently spent on mansions.


mgol:
Stop spamming our repo or we’ll have to ban you in the jQuery GitHub org.

timmywil:
This is inappropriate to say the least. Blocked and reported.

Issue #4867, opened 2021-03-29
You cannot programmatically focus an element after removing its focus handler


JoolsCaesar:
### Description ###
I'm seeing this in the latest chromium browsers Windows Chrome & Edge 89, using jquery 3.6.

If you add and remove an element's focus handler, subsequent calls to $element.focus(), will do nothing.
HTML:
```html
<div tabindex="-1" class="fdiv"></div>
```
CSS:
```css
.fdiv{
  height:100px;
  width:100px;
  border:1px solid black;
}
.fdiv:focus{
  background-color:green;
}
```
See the linked to test case below. I have a simple div that turns green when you focus it and some js that will focus said div after 2 seconds.
Before setting this timeout I add and then immediately remove an empty focus handler:
```js
//Using https://cdnjs.cloudflare.com/ajax/libs/jquery/3.6.0/jquery.min.js

const $origDiv = $('.fdiv');

// Comment out the following line to fix
$origDiv.on('focus',function(){}).off('focus');

setTimeout(function(){
	$origDiv.focus();
},2000)
```

If you run the fiddle, the div should turn green (gain focus) after 2 seconds, but it doesn't!
If you comment out the focus handler line, all works as expected.

### Link to test case ###
https://jsfiddle.net/JoolsCaesar/ut371acd/

---- 

<sup>EDIT by @mgol: I fixed syntax highlighting in code blocks.</sup>

JoolsCaesar:
FWIW I tracked this down to the bit in jquery:
```
function leverageNative( el, type, expectSync ) {

	// Missing expectSync indicates a trigger call, which must force setup through jQuery.event.add
	if ( !expectSync ) {
		if ( dataPriv.get( el, type ) === undefined ) {
			jQuery.event.add( el, type, returnTrue );
		}
		return;
	}
```
`dataPriv.get( el, type )` where `el` is the element we're attempting to focus and type is `'focus'` returns `false`, so we just drop out without doing anything.

mgol:
Thanks for the report! The code you mentioned exists since jQuery 3.4.0, I suspect it only became a problem after #4813 landed.

JoolsCaesar:
Makes sense. I spotted this after upgrading from 3.0.0 to 3.6.0, so couldn't narrow it down much.

mgol:
[...]

PR #4885, opened 2021-05-07
Event: Don't break focus triggering after `.on(focus).off(focus)`

mgol:
### Summary ###

The `_default` function in the special event settings for focus/blur has
always returned `true` since gh-4813 as the event was already being fired
from `leverageNative`. However, that only works if there's an active handler
on that element; this made a quick consecutive call:

```js
elem.on( "focus", function() {} ).off( "focus" );
```

make subsequent `.trigger( "focus" )` calls to not do any triggering.

The solution, already used in a similar `_default` method for the `click` event,
is to check for the `dataPriv` entry on the element for the focus event
(similarly for blur).

Fixes gh-4867

+10 bytes

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -746,10 +746,10 @@ jQuery.each( { focus: "focusin", blur: "focusout" }, function( type, delegateTyp
 			return true;
 		},
 
-		// Suppress native focus or blur as it's already being fired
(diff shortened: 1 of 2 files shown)


mgol:
+10 bytes

Issue #4904, opened 2021-07-14
HTML style comment breaks script


ovarn:
### Description ###
$.append fails when I try to append script element with HTML style comment.

There is no errors when I try to do the same thing using vanila JS:
```
var script = document.createElement('script');
script.text = '<!-- html style comment in javascript -->';
document.querySelector('html').append(script); // no errors
```
When I do it using jQuery:
```
var script = document.createElement('script');
script.text = '<!--html style comment in javascript -->';
$('html').append(script); // Uncaught SyntaxError: Unexpected identifier
```
The root cause:
The _domManip_ function does RegExp replacement before calling _DOMEval_ and removes _&lt;!--_ and _-->_:

https://github.com/jquery/jquery/blob/a684e6ba836f7c553968d7d026ed7941e1a612d8/src/manipulation.js#L30
https://github.com/jquery/jquery/blob/a684e6ba836f7c553968d7d026ed7941e1a612d8/src/manipulation.js#L164

### Link to test case ###
https://codepen.io/ovarn/pen/vYmgjzP (open browser console to see the error)

mgol:
Thanks for the report. The purpose of this regex is to both clean HTML comments and CDATA sections in XML documents, e.g.:
```xml
<script>
<![CDATA[
console.log("<");
]]>
</script>
```
HTML comments themselves should not be an issue as ECMAScript now requires HTML-style comment delimiters to also serve as JS comment ones; a relevant test case from [Compat Tables](https://kangax.github.io/compat-table/es6/):
```js
--> A comment
<!-- Another comment
var a = 3; <!-- Another comment
return a === 3;
```
Unfortunately, IE 11 doesn't support this feature. 😞 

Using regexes to parse JS is always going to have edge cases, we've learned that the hard way when we had to release [jQuery 3.5](https://jquery.com/upgrade-guide/3.5/). However, we need to cut out CDATA & HTML comments to be able to execute inline scripts that may contain them and using a full parser would be too slow & huge.

If there's any other way to remove CDATA & HTML comments without hitting issues like this one, we'd gladly replace the current implementation.

mgol:
Hmm, since the spec requires `<!--` and `-->` to serve as single-line comment delimiters, maybe we should just tweak the regex to match not just those tokens but also what's after them until the end of the line?

mgol:
Looking at it again, it seems we could skip using `rcleanScript` completely outside of IE and in IE fix the regex to match full lines.

[...]

PR #4905, opened 2021-07-15
Manipulation: Don't remove HTML comments from scripts

mgol:
### Summary ###

When evaluating scripts, jQuery strips out the possible wrapping HTML comment
and a CDATA section. However, all supported browsers are already doing that
when loading JS via appending a script tag to the DOM which is how we've been
doing `jQuery.globalEval` since jQuery 3.0.0. jQuery logic was imperfect, e.g.
it just stripped the `<!--` and `-->` markers, respectively at the beginning or
the end of the script contents. However, browsers are also stripping everything
following those markers in the same line, treating them as single-line comments
delimiters; this is now also mandated by ECMAScript 2015 in Annex B. Instead
of fixing the jQuery logic, just let the browser do its thing.

We still need to strip CDATA sections for backwards compatibility. This
shouldn't be needed as in XML documents they're already not visible when
inspecting element contents and in HTML documents they have no meaning but
we're preserving that logic for backwards compatibility. This will be removed
completely in 4.0.

Fixes gh-4904

-8 bytes

4.0 version of this PR: #4906

### Checklist ###

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -40,7 +40,8 @@ var
 
 	// checked="checked" or checked
 	rchecked = /checked\s*(?:[^=]|=\s*.checked.)/i,
-	rcleanScript = /^\s*<!(?:\[CDATA\[|--)|(?:\]\]|--)>\s*$/g;
+
+	rcleanScript = /^\s*<!\[CDATA\[|\]\]>\s*$/g;
 
 // Prefer a tbody over its parent table for containing new rows
 function manipulationTarget( elem, content ) {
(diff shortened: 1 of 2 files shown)


mgol:
4.0 version of this PR: #4906

mgol:
-8 bytes

timmywil (approved):

gibson042 (approved):

src/manipulation.js
```diff
@@ -195,6 +196,12 @@ function domManip( collection, args, callback, ignored ) {
 								}, doc );
 							}
 						} else {
+
+							// Clean the CDATA sections from script contents. This shouldn't be
```
gibson042, line 200:
```suggestion
							// Unwrap a CDATA section containing script contents. This shouldn't be
```

timmywil (approved):

PR #4906, opened 2021-07-15
Manipulation: Don't remove HTML comments from scripts

mgol:
### Summary ###

When evaluating scripts, jQuery strips out the possible wrapping HTML comment
and a CDATA section. However, all supported browsers are already doing that
when loading JS via appending a script tag to the DOM which is how we've been
doing `jQuery.globalEval` since jQuery 3.0.0. jQuery logic was imperfect, e.g.
it just stripped the `<!--` and `-->` markers, respectively at the beginning or
the end of the script contents. However, browsers are also stripping everything
following those markers in the same line, treating them as single-line comments
delimiters; this is now also mandated by ECMAScript 2015 in Annex B. Instead
of fixing the jQuery logic, just let the browser do its thing.

We also used to strip CDATA sections. However, this shouldn't be needed as in
XML documents they're already not visible when inspecting element contents and
in HTML documents they have no meaning. We've preserved that behavior for
backwards compatibility in 3.x but we're removing it for 4.0.

Fixes gh-4904

-44 bytes

3.x version of this PR: #4905

### Checklist ###

* [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com (@mgol: we should document this)

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -25,9 +25,7 @@ var
 
 	// Support: IE <=10 - 11+
 	// In IE using regex groups here causes severe slowdowns.
-	rnoInnerhtml = /<script|<style|<link/i,
-
-	rcleanScript = /^\s*<!(?:\[CDATA\[|--)|(?:\]\]|--)>\s*$/g;
+	rnoInnerhtml = /<script|<style|<link/i;
 
 // Prefer a tbody over its parent table for containing new rows
 function manipulationTarget( elem, content ) {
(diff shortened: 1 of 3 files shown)


mgol:
-8 bytes

mgol:
-44 bytes

mgol:
-8 bytes applied to the 3.x version, not here

timmywil (approved):
Makes sense. LGTM

gibson042 (approved):

timmywil (approved):

goarblt:
hello
https://www.evil.com

Issue #4926, opened 2021-09-07
The compatibility of different browsers should be considered about test case css(--customProperty)


fecore1:
### Description ###
The compatibility of different browsers should be considered about test case [css(--customProperty)](https://github.com/jquery/jquery/blob/main/test/unit/css.js#L1780)

`$elem.css( "--prop2" )` has different val on different browsers:

- firefox 91: `$elem.css( "--prop2" ) => 'val2'`
- chrome 93: `$elem.css( "--prop2" ) => ' val2'`
- edge 93: `$elem.css( "--prop2" ) => ' val2'`

As i know, `jQuery.css` is based on `getComputedStyle` and `getPropertyValue`, and that have different behaves on different browsers.

### Link to test case ###
Please open test case bellow use different browsers and attention the value of `--prop2`

https://codepen.io/fecore1/pen/KKqNxQx

mgol:
Thanks for the report. There was a recent spec change that require CSS Custom Properties values to be trimmed and it's just been implemented in Firefox 91 so far. I expect it to take some time to roll out to all supported browsers, especially Safari so maybe we should, indeed, trim it until all supported browsers adjust. It sounds easy enough to do, just trim the value.

BTW, Firefox behavior is still buggy as it only trims leading spaces and leaves trailing spaces as they are, they have a bug for this: https://bugzilla.mozilla.org/show_bug.cgi?id=1728165.

Would you like to try to submit a fix by yourself?

mgol:
BTW, [we have tests for whitespace handling of CSS Custom Properties](https://github.com/jquery/jquery/blob/175db73ec7938e774d9e93d3afdfb35a24466b47/test/unit/css.js#L1780) (a.k.a. CSS Variables) and the Firefox 91 change is why our Travis tests are failing: https://app.travis-ci.com/github/jquery/jquery/jobs/535647620

timmywil:
I'm in favor of trimming if that's what all browsers will do eventually.

fecore1:
@mgol Thanks for the quick response. 

> There was a recent spec change that require CSS Custom Properties values to be trimmed and it's just been implemented in Firefox 91 so far. 

Yes, so in my option, test case [css(--customProperty)](https://github.com/jquery/jquery/blob/main/test/unit/css.js#L1780) is  more appropriate to be modified  to `assert.equal( $elem.css( "--prop2" ), "val2", "Trim whitespace" );` because Travis tests has failed now and all browsers will implement the spec change eventually.

[...]

PR #4927, opened 2021-09-09
Core:Manipulation: Add basic TrustedHTML support

mgol:
### Summary ###

This is a draft to show the changes necessary to support TrustedHTML (see https://web.dev/trusted-types/ for a nice introduction to trusted types). This builds on other work needed for such support (although useful even without supporting trusted types), including #4642 and #4724.

I went out of my way to add support in a way that does *not* rely on any APIs from this new spec as it's Blink-only, at least for now.

I'm not sure if this is something we'd like to do given limited browser support but I feel seeing what kind of code changes are necessary may help us with reaching the decision.

Fixes gh-4409

### Checklist ###


mgol:
+38 bytes but I haven't spent a lot of time optimizing for size.

mgol:
Travis fails due to #4926 which is unrelated.

src/core/init.js
```diff
@@ -50,19 +70,22 @@ var rootjQuery,
...
-						context && context.nodeType ? context.ownerDocument || context : document,
+						context && context.nodeType ?
+							context.ownerDocument || context :
+							document,
 						true
 					) );
 
 					// HANDLE: $(html, props)
-					if ( rsingleTag.test( match[ 1 ] ) && jQuery.isPlainObject( context ) ) {
+					if ( rsingleTag.test( htmlDeserialized || match[ 1 ] ) &&
```
koto, line 80: Just a small size optimization: 

TT stringify (we rely on that when creating `match` from `selector` on line 47). 

We could leverage that and call `rSingleTag.test(match[1])` directly, that way `htmlDeserialized` variable is not needed at all - unless there is some linter check that prevents calling RegExp.test with a non-string.

That said, this change would enable an interesting trick, in which you pass to jQuery an object that stringifies differently on every attempt - but I'm not sure jQuery aims to be resilient to this kind of attacks from a code that calls jQuery (there's probably 100 more vectors like that).

test/data/trusted-html.html
```diff
@@ -0,0 +1,44 @@
...
+<script>
+	var i, input, elem, tags,
+		results = [],
+		policy = trustedTypes.createPolicy( "jquery-test-policy", {
+			createHTML: function( html ) {
+				return html;
+			}
+		} ),
+		inputs = [
+			[ "<div></div>", "<div class='test'></div>", [ "div" ] ],
```
koto, line 20: Does it make sense to also add a test for `TrustedHTML`s that wrap over a non-obvious HTML, e.g. just some text? From what I can tell, different branches are used then.

koto (approved):

koto:
Looks great!

test/data/trusted-html.html
mgol: Yes, good point. That will only apply to the second array items as the first ones are passed directly to `jQuery` which also supports selectors - and since for TrustedHTML we cannot carve out a part of an input string to later parse, I opted to restrict support for TrustedHTML-wrapped strings to ones that are start with `<` and end with `>` - i.e. what was already going through the fast path skipping the regex match.

src/core/init.js
mgol: Good point, I'll try that. I didn't consider that `RegExp#test` stringifies its input.

test/data/trusted-html.html
koto: So, I assume that `TrustedHTML` wrapping `something` will just get stringified somewhere and be treated as a selector? That's OK, I don't think there would be expectations for jQuery to do anything else. I think it just makes sense to have an explicit test for that behavior.

mgol: To be clear, this limitation only applies to (pseudo-code) `jQuery(TrustedHTML('foo'))`; `jQuery(document.body).append(TrustedHTML('foo'))` should append the text `foo` just as without wrapping in TrustedHTML; I'll add a test for that.

koto: Oh, I forgot about https://api.jquery.com/jQuery/#jQuery-object behavior :) Yeah, that's fine, thanks for clarifying.

mgol:
PR updated, we're down to +30 bytes.

koto (approved):

test/data/trusted-html.html
koto: Sounds good!

timmywil (approved):

mgol:
@timmywil @koto I also added one small change in the tests (in the fixup commit) that makes the tests run even in browsers with no TrustedHTML support - there, it uses a simple object wrapper with a `toString` function returning the HTML string. It also shows that the changes in this PR are generic and would work with a number of similar APIs as well.

goarblt:
hello
https://www.evil.com

PR #4925, opened 2021-09-04
update irc to libera and fix LAMP dead link

fecore1:
### Summary ###

update readme:
- irc.freenode.net works broken, and update irc address to libera based on the infomation from  https://github.com/jquery/irc.jquery.org/issues/24
- fix LAMP dead link

### Checklist ###

* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [x] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #4930, opened 2021-09-13
CSS: Trim whitespace surrounding CSS Custom Properties values

fecore1:
### Summary ###

There was a recent spec change that require CSS Custom Properties values to be trimmed, and workaround trim it until all supported browsers adjust.

Fixes [#4926](https://github.com/jquery/jquery/issues/4926)

### Checklist ###


timmywil:
Don't we want to limit trimming to custom vars? I'm not sure it matters, but this trims all values.

src/css/curCSS.js
```diff
@@ -9,7 +9,10 @@ function curCSS( elem, name, computed ) {
 
 	// getPropertyValue is needed for `.css('--customProperty')` (gh-3144)
 	if ( computed ) {
-		ret = computed.getPropertyValue( name ) || computed[ name ];
+		ret = computed.getPropertyValue( name );
+
+		// trim whitespace (issue #4926)
+		ret = ( ret && ret.trim() ) || computed[ name ];
```
mgol, line 15: This trims all values. While this generally happens for regular properties, it seems safer to just trim where needed, i.e. for custom properties.

We have the `rcustomProp` regex defined in `css.js`; let's move it to a separate file `css/var/rcustomProp.js` (you can see other files in `var` for a pattern how it's used) and use both here and in `css.js` to detect custom properties and only trim in that case.

mgol (changes requested):
Thanks for the PR! This is very close to what we need, see comments for the blockers.

We're also considering whether to trim in the setter as well. We can consider this separately to this PR, though, especially that we haven't reached a decision there yet.

src/css/curCSS.js
fecore1: > This trims all values. While this generally happens for regular properties, it seems safer to just trim where needed, i.e. for custom properties.

Yes, It's a good idea because the changes will be minimal. I will add restrictions.

fecore1:
> [CLA Not Signed](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/19401567/167174/4930/#/?version=2)
> 
> 
> * [❌](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/19401567/167174/4930/#/?version=2) - fecore1 The commit ([04d7e47](https://github.com/jquery/jquery/commit/04d7e47aeddf48e9ee84ace3a5798c147a0ac70f) ,b6a958e91519eda18ebde889111ca8f6a4e14ba6) is not authorized under a signed CLA. [Please click here to be authorized](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/19401567/167174/4930/#/?version=2). For further assistance with EasyCLA, [please submit a support request ticket](https://jira.linuxfoundation.org/servicedesk/customer/portal/4).

hi, @mgol  I'm a little confused as to whether I need to sign an EasyCLA? 

BTW I've signed the CLA, and this is not my first PR. Another [PR](https://github.com/jquery/jquery/pull/4925) which has be merged.

mgol:
@fecore1 yes; we had to change the CLA system we use; see https://github.com/jquery/jquery/issues/4931 for more details.

test/unit/css.js
```diff
@@ -1739,8 +1739,9 @@ QUnit.testUnlessIE( "css(--customProperty)", function( assert ) {
 		"        --prop1:val1;\n" +
 		"        --prop2: val2;\n" +
 		"        --prop3:val3 ;\n" +
-		"        --prop4:\"val4\";\n" +
-		"        --prop5:'val5';\n" +
+		"        --prop4: val4 ;\n" +
```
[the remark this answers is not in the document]
mgol: @fecore1 Why did you resolve this? I don't see such a test added here.

mgol:
@fecore1 Can you sign the new CLA? Unfortunately, as outlined in #4931, we're not allowed to merge any new contribution (which includes updates to existing PRs) without such a signature.

fecore1:
> @fecore1 Can you sign the new CLA? Unfortunately, as outlined in #4931, we're not allowed to merge any new contribution (which includes updates to existing PRs) without such a signature.

done

test/unit/css.js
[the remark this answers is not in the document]
fecore1: @mgol Sorry for missing submissions, and you can resolve this if added test are okay

mgol (approved):
Looks awesome, thanks!

It's at +14 bytes now which is quite reasonable.

test/unit/css.js
```diff
@@ -1738,9 +1738,13 @@ QUnit.testUnlessIE( "css(--customProperty)", function( assert ) {
...
-		"        --prop3:val3 ;\n" +
-		"        --prop4:\"val4\";\n" +
-		"        --prop5:'val5';\n" +
+		"        --prop3:   val3;\n" +
+		"        --prop4:val4 ;\n" +
+		"        --prop5:val5   ;\n" +
+		"        --prop6: val6 ;\n" +
+		"        --prop7:   val7   ;\n" +
+		"        --prop8:\"val8\";\n" +
+		"        --prop9:'val9';\n" +
```
gibson042, line 1747: Should there be test cases with [CSS whitespace](https://www.w3.org/TR/css-syntax-3/#whitespace) other than U+0020 SPACE?

mgol: Good point. That also makes me think, we should perhaps replace CSS whitespace instead of calling `.trim()`, shouldn't we?

gibson042: Yes, definitely.

src/css/curCSS.js
```diff
@@ -1,16 +1,23 @@
...
 
 	computed = computed || getStyles( elem );
 
 	// getPropertyValue is needed for `.css('--customProperty')` (gh-3144)
 	if ( computed ) {
 		ret = computed.getPropertyValue( name ) || computed[ name ];
 
+		// trim whitespace for custom property (issue #4926)
+		if ( isCustomProp ) {
+			ret = ret.trim();
```
mgol, line 18: @fecore1 As @gibson042 rightly noticed, we can't actually call `.trim()` as that removes all Unicode whitespace:
https://262.ecma-international.org/12.0/#sec-trimstring
https://262.ecma-international.org/12.0/#prod-WhiteSpace
We need to only remove the CSS whitespace for which we already have a regex in [`src/selector/var/whitespace.js`](https://github.com/jquery/jquery/blob/175db73ec7938e774d9e93d3afdfb35a24466b47/src/selector/var/whitespace.js), please use it.

This gotcha is a reason why we don't use `String#trim` at all in jQuery anymore but we only rely on this regex.

test/unit/css.js
mgol: I'd be good to have a test case with some Unicode whitespace that is not CSS whitespace to confirm the new planned behavior.

mgol (changes requested):
One comment about `trim` plus more test cases needed.

src/css/curCSS.js
fecore1: wow, thanks @gibson042 , a very professional opinion, and i will fix it as soon as  possible

src/css/curCSS.js
```diff
@@ -1,16 +1,24 @@
...
+	var ret,
+		isCustomProp = rcustomProp.test( name );
 
 	computed = computed || getStyles( elem );
 
 	// getPropertyValue is needed for `.css('--customProperty')` (gh-3144)
 	if ( computed ) {
 		ret = computed.getPropertyValue( name ) || computed[ name ];
 
+		// trim whitespace for custom property (issue #4926)
```
mgol, line 17: That's a jQuery-specific gotcha:
```suggestion
		// trim whitespace for custom property (issue gh-4926)
```
Since we have lots of references using `#` to our legacy Trac bug tracker at https://bugs.jquery.com/, we use the `gh-XXXX` syntax to make it explicit what comes from GitHub.

We'll need to replace these legacy `#` uses one day to reduce confusion.

mgol (approved):
It's hard to believe but this PR is actually smaller than `main` and by 21 bytes! 😮 Even the non-gzipped minified version is smaller.

Just one tiny comment fix I'll apply myself, the code looks good to me. Thanks for following up with our review remarks!

mgol:
Ouch, I was too quick with merging that last iteration. The `rtrim` regex from the selector module is quite selector-specific and it's more complex than what we need for CSS custom properties values. I'll submit a followup PR with a fix.

mgol:
A followup PR: #4938.

goarblt:
hello
https://www.evil.com

goarblt:
hello
https://www.evil.com

src/css/curCSS.js
```diff
@@ -1,16 +1,24 @@
...
 
 	computed = computed || getStyles( elem );
 
 	// getPropertyValue is needed for `.css('--customProperty')` (gh-3144)
 	if ( computed ) {
 		ret = computed.getPropertyValue( name ) || computed[ name ];
 
+		// trim whitespace for custom property (issue gh-4926)
+		if ( isCustomProp ) {
+			ret = ret.replace( rtrim, "$1" );
```
andersk, line 19: This throws `TypeError: ret is undefined` when the property isn’t set: #5105.

PR #4938, opened 2021-09-24
CSS: Don't special-case backslashes when trimming CSS Custom Properties

mgol:
### Summary ###

The original logic from PR gh-4930 re-used the `rtrim` regex from the selector
module but selectors have special escaping rules which means whitespace
following a backslash character would not get trimmed.

This change splits the logic at the unfortunate but necessary cost of size.

Ref gh-4926
Ref gh-4930

This is at +44 bytes now but together with gh-4930 where it should be included it's at about half that.

When this lands, I'll CP gh-4930 together with this PR to `3.x-stable`.

### Checklist ###

### DIFF
--- a/src/css/curCSS.js
+++ b/src/css/curCSS.js
@@ -2,7 +2,7 @@ import jQuery from "../core.js";
 import isAttached from "../core/isAttached.js";
 import getStyles from "./var/getStyles.js";
 import rcustomProp from "./var/rcustomProp.js";
-import rtrim from "../var/rtrim.js";
+import rtrimAscii from "../var/rtrimAscii.js";
 
 function curCSS( elem, name, computed ) {
 	var ret,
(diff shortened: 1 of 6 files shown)


mgol:
As discussed at the today's meeting, this is not a correct change. [Whitespace is getting trimmed after tokenization](https://www.w3.org/TR/css-syntax-3/#consume-declaration) and [some tokens contain](https://www.w3.org/TR/css-syntax-3/#escaping) a [trailing whitespace](https://www.w3.org/TR/css-syntax-3/#escape-diagram).

We still need to discuss whether the current solution from https://github.com/jquery/jquery/pull/4930 is fine or if it needs more iterations but this PR is not what we need.

goarblt:
hello
https://www.evil.com

goarblt:
hello
https://www.evil.com

PR #4956, opened 2021-10-18
CSS: Justify use of rtrim on CSS property values

gibson042:
CSS does not acknowledge carriage return or form feed characters as whitespace but it does replace them _with_ whitespace, making it acceptable to use `rtrim`.


Mansalol:
G

goarblt:
hello
https://www.evil.com

PR #4986, opened 2021-12-17
Core: Don't rely on splice being present on input

bubbatls:
Without this fix it does that

TypeError: results.splice is not a function
    at Function.jQuery.uniqueSort (https://code.jquery.com/jquery-git.js:664:12)
    at jQuery.fn.init.find (https://code.jquery.com/jquery-git.js:2394:27)
    at gocusihafe.js:3:4

Reproduced here:
https://jsbin.com/gocusihafe/1/edit?html,css,js,console,output

### Checklist ###

### DIFF
--- a/test/unit/selector.js
+++ b/test/unit/selector.js
@@ -1956,8 +1956,9 @@ QUnit.test( "jQuery.uniqueSort", function( assert ) {
 
 	jQuery.each( tests, function( label, test ) {
 		var length = test.length || test.input.length;
-		//we duplicate the test.input, because if not it is modified by the uniqueSort and the second test becomes worthless
-    assert.deepEqual( jQuery.uniqueSort( test.input.slice( 0 ) ).slice( 0, length ), test.expected, label + " (array)" );
+		// We duplicate `test.input` because otherwise it is modified by `uniqueSort`
+		// and the second test becomes worthless.
+		assert.deepEqual( jQuery.uniqueSort( test.input.slice( 0 ) ).slice( 0, length ), test.expected, label + " (array)" );
 		assert.deepEqual( jQuery.uniqueSort( new Arrayish( test.input ) ).sliceForTestOnly( 0, length ), test.expected, label + " (quasi-array)" );
 	} );
 } );
--- a/test/unit/selector.js
+++ b/test/unit/selector.js
@@ -1958,8 +1958,10 @@ QUnit.test( "jQuery.uniqueSort", function( assert ) {
 		var length = test.length || test.input.length;
 		// We duplicate `test.input` because otherwise it is modified by `uniqueSort`
 		// and the second test becomes worthless.
-		assert.deepEqual( jQuery.uniqueSort( test.input.slice( 0 ) ).slice( 0, length ), test.expected, label + " (array)" );
-		assert.deepEqual( jQuery.uniqueSort( new Arrayish( test.input ) ).sliceForTestOnly( 0, length ), test.expected, label + " (quasi-array)" );
+		assert.deepEqual( jQuery.uniqueSort( test.input.slice( 0 ) ).slice( 0, length ),
+			test.expected, label + " (array)" );
+		assert.deepEqual( jQuery.uniqueSort( new Arrayish( test.input ) ).sliceForTestOnly( 0, length ),
+			test.expected, label + " (quasi-array)" );
 	} );
 } );
 
--- a/test/unit/selector.js
+++ b/test/unit/selector.js
@@ -1894,9 +1894,7 @@ QUnit.test( "jQuery.uniqueSort", function( assert ) {
 		}
 	}
 	Arrayish.prototype = {
-		slice: [].slice,
-		sort: [].sort,
-		splice: [].splice
+		sliceForTestOnly: [].slice
 	};
 
 	var i, tests,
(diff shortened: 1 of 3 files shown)


timmywil:
Thanks for opening a pull request! It seems we forgot to change our use of splice when [removing it from the jQuery object](https://github.com/jquery/jquery/pull/4473). We did take care of sort and push, just not splice. I'm surprised our tests didn't catch this, but we're going to need to add a test.

Something like the case in the jsbin will work fine to get duplicates on the jQuery object (i.e. `jQuery('body div').find('div')`). A relatively pointless call to `find` as it won't find anything new, but it does the job here of inserting duplicates. That said, there are a lot of divs on the test page, so it would be good to limit it to smaller selection. Something like `jQuery("#moretests div").find("div")` should work.

Also, splice should be its own var file, just like with [sort](https://github.com/jquery/jquery/blob/main/src/var/sort.js)

Thanks!

mgol:
@bubbatls Hey, are you interested in addressing Timmy's concerns?

mgol:
@bubbatls Ping. This looks like an issue we'd like to see fixed but to land your PR we need to have Timmy's concerns addressed. If you don't have time for this, please let us know and we'll likely fix this by ourselves.

bubbatls:
I'm sorry, I was in holiday and after holidays there is a lot of work to do...
Yes I could fix this using your kind of var files, just let me few days.

Thanks

mgol:
@bubbatls Sure, no problem. Let us know if you hit any roadblocks.

mgol (changes requested):
Code changes look good now but we still need a test. Can you add one?

@timmywil BTW, I think we can have a simpler & more direct test:
```js
jQuery.uniqueSort( { 0: div, 1: div, length: 2 } );
```
where `div` is a test div appended to `#qunit-fixture`. What do you think? Or maybe we can add both test cases.

bubbatls:
yes, the test you write is the most unit one about this issue, and the method name is uniqueSort so, checking the "unique" is completely relevant.

But I agree that another one covering more code by doing the .find('div').find('div') is worth being written for potential future regression.

This test should be written in the test/unit/selector.js file?

mgol:
> This test should be written in the test/unit/selector.js file?

Yes, the both tests should be written in `test/unit/core.js` but in separate `QUnit.test` blocks as they're testing two different things.

mgol:
EDIT: I made a typo and I mentioned~~`test/unit/core.js`~~ instead of the proper `test/unit/selector.js`; I fixed the comment now.

bubbatls:
I'm sorry but I spend some time to try to understand how to run the tests before modifying anything. But I have always the error :
test/data/qunit-fixture.js net::ERR_ABORTED 404 (Not Found)
And I don't understand what I'm doing wrong.

mgol:
How are you running the tests?

You need to build before running the tests as the build process creates a few test files and also builds jQuery and the built version is what gets tested by default. Run `npm run build` before running the tests.

test/unit/selector.js
```diff
@@ -1958,8 +1956,9 @@ QUnit.test( "jQuery.uniqueSort", function( assert ) {
 
 	jQuery.each( tests, function( label, test ) {
 		var length = test.length || test.input.length;
-		assert.deepEqual( jQuery.uniqueSort( test.input ).slice( 0, length ), test.expected, label + " (array)" );
-		assert.deepEqual( jQuery.uniqueSort( new Arrayish( test.input ) ).slice( 0, length ), test.expected, label + " (quasi-array)" );
+		//we duplicate the test.input, because if not it is modified by the uniqueSort and the second test becomes worthless
```
mgol, line 1959:
```suggestion
		// We duplicate `test.input` because otherwise it is modified by `uniqueSort`
		// and the second test becomes worthless.
```

test/unit/selector.js
```diff
@@ -1958,8 +1956,9 @@ QUnit.test( "jQuery.uniqueSort", function( assert ) {
...
+    assert.deepEqual( jQuery.uniqueSort( test.input.slice( 0 ) ).slice( 0, length ), test.expected, label + " (array)" );
```
mgol, line 1960:
```suggestion
		assert.deepEqual( jQuery.uniqueSort( test.input.slice( 0 ) ).slice( 0, length ), test.expected, label + " (array)" );
```

mgol (approved):
Just some style nits but otherwise looks good, thanks! It's nice you found an existing broken test.

timmywil (approved):
Looks good now. Thanks for the tests.

test/unit/selector.js
```diff
@@ -1958,8 +1956,10 @@ QUnit.test( "jQuery.uniqueSort", function( assert ) {
 
 	jQuery.each( tests, function( label, test ) {
 		var length = test.length || test.input.length;
-		assert.deepEqual( jQuery.uniqueSort( test.input ).slice( 0, length ), test.expected, label + " (array)" );
-		assert.deepEqual( jQuery.uniqueSort( new Arrayish( test.input ) ).slice( 0, length ), test.expected, label + " (quasi-array)" );
+		// We duplicate `test.input` because otherwise it is modified by `uniqueSort`
+		// and the second test becomes worthless.
+		assert.deepEqual( jQuery.uniqueSort( test.input.slice( 0 ) ).slice( 0, length ), test.expected, label + " (array)" );
+		assert.deepEqual( jQuery.uniqueSort( new Arrayish( test.input ) ).sliceForTestOnly( 0, length ), test.expected, label + " (quasi-array)" );
```
mgol, line 1962:
```suggestion
		assert.deepEqual( jQuery.uniqueSort( test.input.slice( 0 ) ).slice( 0, length ),
			test.expected, label + " (array)" );
		assert.deepEqual( jQuery.uniqueSort( new Arrayish( test.input ) ).sliceForTestOnly( 0, length ),
			test.expected, label + " (quasi-array)" );
```

bubbatls:
Nice! thanks you all :)

goarblt:
hello
https://www.evil.com

Issue #4981, opened 2021-12-02
Invalid links exist in jquery  


cjchnwscqwsbsy:
jquery/dist/jquery.js：
// https://web.archive.org/web/20100324014747/http://blindsignals.com/index.php/2009/07/jquery-delay/
The 'http://blindsignals.com/index.php/2009/07/jquery-delay/' is a malicious web site

wenz:
The blindsignals site indeed seems to be malicious now, but the link goes to the Web Archive copy (which is not malicious). Still probably worth removing (or replacing) the link.

wenz:
Related issue: https://github.com/jquery/jquery/issues/4980

mgol:
This happens pretty often with old sites - they expire & are bought by folks trying to leverage the domain being saved for many users.

That said, we are referring to a Web Archive snapshot which is fine. There's no other way to still refer to the old content. Should we just delete the link? I'm not sure if it's always a good strategy.

For example, this old jQuery blog post refers to an expired `jquery14.com` domain: https://blog.jquery.com/2010/01/08/14-days-of-jquery-and-the-new-api-browser/. I think we should replace the links with Web Archive ones for posterity but why remove them completely, losing history? Would Wikipedia remove a Web Archive link because the current version of the site is malicious? I'm not sure.

@timmywil @gibson042 @dmethvin what do you think?

timmywil:
We'll remove these links from source, but leave the blog posts for now.

Mansalol:
U

rkshah1988:
The invalid links cannot be seen in the jquery.min.js
So if the application is bundling the minified version, then it should be safe right?

adiGershon:
we are using angular js , which use jqlite. Is this vulnerability exist also in jqlite?

Mansalol:
H

mgol:
@adiGershon There's no vulnerability here, neither in jQuery, nor in jqLite. The issue is just a link in a code comment and comments are ignored when code is run.

mgol:
PR: #4991

adiGershon:
Do you know when supposed to release a new version of JQuery without this issue?

wenz:
Those links do not show up in the production version of jQuery anyway?!

mgol:
We can't provide any dates. We still have a few things to fix for the next release.

scottkuhl:
[...]

PR #4991, opened 2022-01-03
Docs: Remove links to Web Archive from source

mgol:
### Summary ###

Neither of the removed links is crucial; one of them refers to a site that has
since started being malicious; while the Web Archive links remain safe, some
scanners warn about such links. Removing them is the safest thing to do.

Fixes gh-4981

### Checklist ###

### DIFF
--- a/src/attributes/prop.js
+++ b/src/attributes/prop.js
@@ -59,7 +59,6 @@ jQuery.extend( {
 				// Support: IE <=9 - 11+
 				// elem.tabIndex doesn't always return the
 				// correct value when it hasn't been explicitly set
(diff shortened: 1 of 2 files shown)


timmywil (approved):

sileclercq:
any idea when this fix will get in ? which version is this expected to be on ? is there a way to get a 3.6.1 with just this change ?

mgol:
Landed on `main` in e24f2dcf3f6bda1a672502e0233c732065cbbe89 and on `3.x-stable` in 4b0d8900eca0180720636e15a4038e066bc95280.

mgol:
@sileclercq It will be included in jQuery 3.6.1 as can be seen by looking at the Milestone field. We don't plan any special release just with this patch as this is not an actual vulnerability.

goarblt:
hello
https://www.evil.com

Issue #4998, opened 2022-01-06
addClass() called with an array doesn't add all of the classes if one of them is an empty string


MatmaRex:
### Description ###
When `addClass()` is called with an array, and one of the classes in the array is an empty string, it doesn't add all of the classes (it stops processing them after the empty string).

Minimal test case:
```js
$( '<div>' ).addClass( [ 'a', '', 'b' ] ).attr( 'class' )
```
Expected output: `a b`
Actual output: `a`

### Link to test case ###
https://jsbin.com/voxetucoho/edit?js,console

mgol:
Thanks for the report. This is caused by this line:
https://github.com/jquery/jquery/blob/main/src/attributes/classes.js#L41

I wonder if there's any concrete reason why we're not checking for `length` here. I know there have been some issues related to this property in old IE but perhaps not when we're dealing with a regular array.

mgol:
Note, though, that an empty string is not a valid class name and if you pass it to the native `classList.add`, it will throw an error. Array input to `addClass` requires valid classes to be present in the input so the proposed input is invalid.

That said, the current behavior may still looks surprising so I'd be open to changing this.

timmywil:
I'd be open to changing this as well. It makes sense to apply as many valid classes as possible, and I could see a case where a variable is passed that happens to be empty string.

MatmaRex:
> Note, though, that an empty string is not a valid class name and if you pass it to the native `classList.add`, it will throw an error. Array input to `addClass` requires valid classes to be present in the input so the proposed input is invalid.

It is not valid for sure, but jQuery is happy to ignore other invalid class names, such as `' '` (a space).
```js
$( '<div>' ).addClass( [ ' ' ] ).attr( 'class' )
```
Output: `undefined`

It is also happy to clean up and tweak them to make them valid:

```js
$( '<div>' ).addClass( [ 'a b', ' c ' ] ).attr( 'class' )
```
Output: `a b c`

If you'd like these to throw an exception, that'd be fine by me, but I'd expect all invalid class names to behave consistently.




> I'd be open to changing this as well. It makes sense to apply as many valid classes as possible, and I could see a case where a variable is passed that happens to be empty string.

Yes, when I discovered this bug, the empty string was actually generated by a library very far away from the place where it was added as a class.

mgol:
> If you'd like these to throw an exception, that'd be fine by me, but I'd expect all invalid class names to behave consistently.

[...]

PR #5003, opened 2022-01-20
CSS: Skip falsy values in `addClass( array )`, compress code

mgol:
### Summary ###

This change makes jQuery skip falsy values in `addClass( array )`
& `removeClass( array )` instead of stopping iteration when the first falsy
value is detected. This makes code like:
```js
elem.addClass( [ "a", "", "b" ] );
```
add both the `a` & `b` classes.

The code was also optimized for size a bit.

Fixes gh-4998

+0 bytes thanks to the optimizations 🎉; initially it was at +8 bytes. In `toggleClass` we now also avoid needlessly recomputing the array of classes.

I'm yet to check how much can these optimizations be backported to 3.x-stable as we have some deprecated `toggleClass` logic there but I'll check after this lands.

### Checklist ###

### DIFF
--- a/src/attributes/classes.js
+++ b/src/attributes/classes.js
@@ -20,8 +20,7 @@ function classesToArray( value ) {
 
 jQuery.fn.extend( {
 	addClass: function( value ) {
-		var classes, elem, cur, curValue, clazz, j, finalValue,
-			i = 0;
+		var classNames, cur, curValue, className, i, finalValue;
 
 		if ( typeof value === "function" ) {
 			return this.each( function( j ) {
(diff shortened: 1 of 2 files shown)


timmywil (approved):
I like the change to use each.

mgol:
Landed on `main` in a338b407f2479f82df40635055effc163835183f and on `3.x-stable` in a slightly modified version in 9b34bdb1c35c5c2bdb367dd80928825874d3ec3f.

The `3.x-stable` version has fewer modifications in `toggleClass` as it needs to support the deprecated value-less signature but even in that state it's also at +0 bytes.

goarblt:
hello
https://www.evil.com

PR #5024, opened 2022-03-15
Event: Support EventListener interface objects.

### LINKED ISSUE #1735: Allow objects as event handlers
Originally reported by petka_antonov@… at: http://bugs.jquery.com/ticket/12031

Doing OOP with jQuery events requires a lot of $.proxy boilerplate. the .addEventListener interface allows passing
an object as a handler like so:

```js
function SomeWidget( elem ) {
	this.elem = elem;
}

SomeWidget.prototype = {
	constructor: SomeWidget,

	renderTo: function( target ) {
		$(target).append(this.elem);
		this.elem.addEventListener( "click", this );
		this.elem.addEventListener( "mousemove", this );
	},

	mousemove: function( e ) {

	},

	click: function( e ) {

	},

	handleEvent: function( e ) {
		return this[e.type].apply( this, arguments );
	}
};
```

With jQuery you have to do this:

```js
function SomeWidget( elem ) {
	this.elem = elem;
	//Each method needs to be bound to the instance. This also has overhead of creating many functions.
	this.mousemove = $.proxy( this.mousemove, this );
	this.click = $.proxy( this.click, this );
}

SomeWidget.prototype = {
	constructor: SomeWidget,

	renderTo: function( target ) {
		$(target).append(this.elem);
		$(this.elem).on( {
			click: this.click,
			mousemove: this.mousemove
		});
	},

	mousemove: function( e ) {

	},

	click: function( e ) {

	}
};
```

With jQuery supporting objects as event handlers, it would work like:

```js
function SomeWidget( elem ) {
	this.elem = elem;
}

SomeWidget.prototype = {
	constructor: SomeWidget,

	renderTo: function( target ) {
		$(target).append(this.elem);
		$(this.elem).on( "click mousemove", this );
	},

	mousemove: function( e ) {

	},

	click: function( e ) {

	},

	handleEvent: function( e ) {
		return this[e.type].apply( this, arguments );
	}
};
```

And that's it. No $.proxy hacks, no redundant creation of functions just to retain binding.

Issue reported for jQuery 1.7.2

MarkMYoung:
### Summary ###

When an event handler implements the `EventListener` interface (by being an object and having a function property called `handleEvent`), the handler function used internally is replaced with that `handleEvent` function bound to the handler object.
http://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-EventListener
Fixes gh-1735

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -117,6 +117,11 @@ jQuery.event = {
 			selector = handleObjIn.selector;
 		}
 
+		// Support objects implementing the EventListener interface.
+		if ( handler && typeof handler.handleEvent === "function" ) {
+			handler = handler.handleEvent.bind( handler );
+		}
+
 		// Ensure that invalid selectors throw exceptions at attach time
 		// Evaluate against documentElement in case elem is a non-element node (e.g., document)
 		if ( selector ) {
--- a/test/unit/event.js
+++ b/test/unit/event.js
@@ -45,6 +45,28 @@ QUnit.test( "on() with non-null,defined data", function( assert ) {
 
 } );
 
+QUnit.test( "on() with EventListener interface object", function( assert ) {
+
+	assert.expect( 4 );
+
+	var handler = {
+		eventDelegateData: "can be found",
+		handleEvent: function( event, data ) {
+			assert.equal( data, 0, "non-null, defined data (zero) is correctly passed" );
+			assert.equal( this.eventDelegateData, "can be found", "event delegate is accessible via this" );
+		}
+	};
+
+	jQuery( "#foo" ).on( "foo.on", handler );
+	jQuery( "div" ).on( "foo.delegate", "#foo", handler );
+
+	jQuery( "#foo" ).trigger( "foo", 0 );
+
+	jQuery( "#foo" ).off( "foo.on", handler );
+	jQuery( "div" ).off( "foo.delegate", "#foo" );
+
+} );
+
 QUnit.test( "Handler changes and .trigger() order", function( assert ) {
 	assert.expect( 1 );
 


mgol:
Thanks for the PR. Please sign our CLA before we review it.

MarkMYoung:
I have signed it twice today, but it still says that it is missing.

dmethvin:
The discussion in gh-1735 is pretty old. In that ticket the reason I was referring to the EventListener interface as "low-level" was because I thought it might be useful to use it as a foundation for existing jQuery functionality without exposing EventListener directly in the API. If it's exposed as a top-level ability with `.on()` (as in, "an EventListener can be passed instead of a function") then it would probably prevent us from using EventListener for the internal low-level implementation.

I think there may be some other cases and design issues that need to be considered here, especially since we're looking to support new features like passive listeners or the capture phase in jQuery 4. (Passive listeners in particular seem to be something that people want, most likely because Chrome's console complains about them incessantly.) 

Assuming those issues are resolved or not relevant, the spec says:

> When a Node is copied using the cloneNode method the EventListeners attached to the source Node are not attached to the copied Node. If the user wishes the same EventListeners to be added to the newly created copy the user must add them manually.

However, jQuery's `.clone()` method supports cloning with copying data and/or events. I'm not sure if this PR handles that and it would be good to have a test to ensure it does when the handler is an EventListener.

dmethvin:
Also I just realized that the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/EventListener) seem to indicate the object with `handleEvent` is for compatibility, which isn't a good signal for adding new support:

> Note: Due to the need for compatibility with legacy content, EventListener accepts both a function and an object with a handleEvent() property function. This is shown in the [example](https://developer.mozilla.org/en-US/docs/Web/API/EventListener#example) below.

MarkMYoung:
> it would probably prevent us from using EventListener for the internal low-level implementation

Please explain.  This PR does not actually pass around and repeatedly accommodate `EventListener` objects, it supplants it with a bound `handleEvent` function.

> we're looking to support new features like passive listeners or the capture phase in jQuery 4.

This solution only affects the second parameter to `addEventListener`.  Implementating event capturing or passive listeners is not affected by this.  I know this because I had to develop my own `EventTarget` before (taking into account phases, bubbling, capturing, etc.).

> I'm not sure if this PR handles that and it would be good to have a test to ensure it does when the handler is an EventListener object.

Again, this solution is merely a function binding.  I can look into adding a unit test to demonstrate `.clone()` with `withDataAndEvents` set to true, but I don't see how that necessary since a bound member function is just a function.

> compatibility...isn't a good signal for adding new support

`EventListener` is not legacy, it is good design and forethought.  `EventListener` being around for 22 years and being around 11 years before `.on()` was added to jQuery is a good signal for compliance.  That example on Mozilla's website does not even attempt to demonstrate the advantages `EventListener`s have over callback functions (this PR's unit test does a better job of demonstrating how `this` object having contextual members).  For starters, callback event handlers usually make (strong) references to outside identifiers (because they have no context of their own), but `EventListener`s encourage code reuse and facilitate testability (by explicitly encapsulating the event handlers behavior).

I rapid develop using `.on()` with callback event handlers, but after the event/behavior relationships are formalized into `EventListener` event delegates, I either have to drop `.on()` in favor of `.addEventListener` or use something like:
```
function bindEventListener( evtListener ) {
	return evtListener.handleEvent.bind( evtListener );
}
```
but why?

```
// Contrived example.
class VisitedControl {
	constructor( { displayElem } ) {
		this.visitCount = 0;
		this.displayElem = displayElem;
	}
	handleEvent( evt ) {
		++this.visitCount;
		jQuery( this.displayElem ).text( this.visitCount );
	}
}
// Works with the `.on( events [, selector ] [, data ], handler )` signature.
jQuery( "button" ).on( "click", bindEventListener( new VisitedControl( { displayElem: "#button-count" } ) ) );
// Also works with the `.on( events [, selector ] [, data ] )` signature.
jQuery( "input" ).on( "blur", { handler: bindEventListener( new VisitedControl( { displayElem: "#input-count" } ) ) } );
```

mgol:
> That example on Mozilla's website does not even attempt to demonstrate the advantages `EventListener`s have over callback functions (this PR's unit test does a better job of demonstrating how `this` object having contextual members). For starters, callback event handlers usually make (strong) references to outside identifiers

The way it's implemented it in this PR nullifies that advantage as there's a strong reference to the handler, it's the handler that's being saved. To be able to fully utilize this, we'd have to actually store the object, not the handler and this should be confirmed by unit tests checking if swapping `handleEvent` for another callback after registration is respected. We'd probably have to create our own `EventListener` wrappers when a function is provided as a handler to unify the way we store handlers.

> This solution only affects the second parameter to `addEventListener`. Implementating event capturing or passive listeners is not affected by this.

We realize this but when you take above into account, there's a lot more complexity to it. Implementing support for passive events requires us to do a significant refactor of our event module while making sure the test suite passes with as few breaking changes as possible. This is just not a good moment to add new significant features like this one. When the refactor is done, we can return to the subject if it's not already handled by the refactor.

mgol:
As for the CLA, I've heard the error might be caused by the email used for the commit in this PR not being attached to your GitHub account. Can you make sure that's done?

PR #5046, opened 2022-05-23
Tests: Exclude tests based on compilation flags, not API presence

mgol:
### Summary ###

Introduces a new test API, `includesModule`. The method returns whether
a particular module like "ajax" or "deprecated" is included in the current
jQuery build; it handles the slim build as well. The util was created so that
we don't treat presence of particular APIs to decide whether to run a test as
then if we accidentally remove an API, the tests would still not fail.

### Checklist ###


timmywil (approved):
Good improvement

PR #4803, opened 2020-10-27
Build: support browserstack test

ygj6:
### Summary ###

This week I tried to do something for the jQuery test migration, as follows:

I applied for a free browserstack account, and added browserstack configuration to my fork project. It seems that the jQuery test cases run properly in the real browser (Windows 10 Chrome) provided by the browserstack platform. 

Local operation steps:
1. git clone -b browserstack https://github.com/jquery/jquery.git
2. `export BROWSERSTACK_USERNAME=<your browserstack username>`
   `export BROWSERSTACK_ACCESS_KEY=<your browserstack access key>`
3. Run `npm run test:browserstack`

output:
```
       Running "karma:browserstack" (karma) task
       27 10 2020 14:21:24.584:INFO [karma-server]: Karma v5.1.0 server started at http://0.0.0.0:9876/
       27 10 2020 14:21:24.586:INFO [launcher]: Launching browsers bs_chrome_win10 with concurrency 3
       27 10 2020 14:21:24.655:INFO [launcher]: Starting browser chrome 86.0 (Windows 10) on BrowserStack
       27 10 2020 14:22:01.426:INFO [launcher.browserstack]: chrome 86.0 (Windows 10) session at https://automate.browserstack.com/builds/f5e04e82b7db52fab492e6147fb7ce6f44c8e337/sessions/8a87ce5a344ff150f8072beeb0be5d32606ca2fc
       27 10 2020 14:22:07.500:INFO [Chrome 86.0.4240.75 (Windows 10)]: Connected on socket Of1EpjUUAKiRdu4CAAAA with id 43017044
       Chrome 86.0.4240.75 (Windows 10): Executed 435 of 1094 SUCCESS (0 secs / 52.279 secs)
       Chrome 86.0.4240.75 (Windows 10): Executed 493 of 1094 SUCCESS (0 secs / 1 min 2.398 secs)
       Chrome 86.0.4240.75 (Windows 10): Executed 650 of 1094 SUCCESS (0 secs / 1 min 27.918 secs)
       Chrome 86.0.4240.75 (Windows 10): Executed 711 of 1094 SUCCESS (0 secs / 1 min 37.87 secs)
       Chrome 86.0.4240.75 (Windows 10): Executed 1094 of 1094 SUCCESS (4 mins 39.082 secs / 4 mins 2.227 secs)
       TOTAL: 1094 SUCCESS
       
       Done.
```   

Next, I plan to continue to do it, such as integrating into github actions, configuring and running more real browsers, etc.;

[...]

PR #5068, opened 2022-06-27
Performance: improve .trim performance for large strings contains lots of whitespaces

vlsi:
The old implementation took O(N^2) time to trim the string when multiple adjacent spaces
were present.

In my case, the issue impacts JIRA which takes ~20 seconds to display an issue where most of the time is spent in `jQuery.trim`.

For instance, consider the string "A          B" (10 spaces between A and B).
Then old regexp /[\s]+$/ would take 10 steps until it realizes
the regexp does not match at position 1.
Then it would try to match the regexp at position 2, and it would take 9 steps.
Then 8 steps for position 3, ... so it would take 10*10/2 steps until
it figures out the regexp does not match.

The new approach is to require "non-whitespace" char before the whitespace run,
so it spends just one step for each space in the string.


mgol:
Thanks for the PR. This is adding 63 bytes to the gzipped minified build, we consider that a significant number. As I suggested in the other thread, can you try removing the branch that leverages native `trim` when available to decrease the size impact? Are there any issues with this strategy?

To verify the size increase, please run `npm run build` locally on the `3.x-stable` branch and then on your PR rebased over that branch; you should see at the end what's the size of your current build compared to `3.x-stable` - the number 68 I posted above.

vlsi:
Native trim should be faster than any regexp, and I think the recent browsers should not require the regexp.
Frankly speaking, I think going for native trim is worth doing as it would make the webpages faster for the end-user.

Just in case:

```
#baseline
   raw     gz Sizes
289641  85451 dist/jquery.js
 89653  30945 dist/jquery.min.js

# with regexp only (without native trim)
   raw     gz Sizes
290048  85627 dist/jquery.js
 89691  30959 dist/jquery.min.js
  +407   +176 dist/jquery.js
   +38    +14 dist/jquery.min.js

# regexp and native trim
   raw     gz Sizes
290297  85690 dist/jquery.js
 89812  31008 dist/jquery.min.js
  +656   +239 dist/jquery.js
  +159    +63 dist/jquery.min.js
```

vlsi:
Here's a sample benchmark: https://jsbench.me/oml5j9kjfu/1

Note that even the optimized regex still needs to walk over the entire string.

Results on my machine are:

## trim 1000 char text like `__________ __________ __________ __________ __________ __________ __________ __________ __________ __________ __________ __________ ...`

```javascript
let n = 1;
let wordLen = 10;
let numWords = 100;
let spaceAround = 1;
let mid = ('_'.repeat(wordLen) + ' ').repeat(numWords) + '_';
let src = ' '.repeat(spaceAround) + mid + ' '.repeat(spaceAround);
let ltrim = /^[\s\uFEFF\xA0]+/;
let rtrim = /([^\s\uFEFF\xA0])[\s\uFEFF\xA0]+$/;
```

```
native: 16M ops/sec ~ 0.06 ms/trim
regex: 700K ops/sec ~ 1.4 ms/trim (~23 times slower)
```

## trim `' '.repeat(1000) + '_' + ' '.repeat(1000)`

```
native: 750K ops/sec ~ 1.3 ms/trim
regex: 680K ops/sec ~ 1.47 ms/trim (13% slower)
```

## trim `' _ '`

```
native: 85M ops/sec ~ 13 us/trim
regex: 5.5M ops/sec ~ 180 us/trim (~13 times slower)
```

Frankly speaking, I think the results show that performance improvement is worth the added 63 bytes.

WDYT?

src/deprecated.js
```diff
@@ -79,9 +85,21 @@ jQuery.isNumeric = function( obj ) {
 		!isNaN( obj - parseFloat( obj ) );
 };
 
-jQuery.trim = function( text ) {
-	return text == null ?
-		"" :
-		( text + "" ).replace( rtrim, "" );
-};
+if ( String.prototype.trim ) {
+	useDefaultTrim = "_" === " \uFEFF\xA0_\uFEFF\xA0 ".trim();
```
mgol, line 89: We do this type of things using support tests. Please create a file `src/deprecated/support.js` modeled after e.g. `src/event/support.js` and save this support test result as `support.trim`. Then, you'll need to add `./deprecated/support` to AMD dependencies of this file and read `support.trim` in the below `if`.

Also, I expect it will be shorter if you:
* check `"".trim` instead of `String.prototype.trim`
* Incorporate this check into the definition, removing the `if`

Something like:
```js
support.trim = "".trim && "_" === " \uFEFF\xA0_\uFEFF\xA0 ".trim();
```

src/deprecated.js
```diff
@@ -79,9 +85,21 @@ jQuery.isNumeric = function( obj ) {
...
+			"" :
+			( text + "" ).trim();
+	};
+} else {
+	jQuery.trim = function( text ) {
+		return text == null ?
+			"" :
+			( text + "" ).replace( ltrim, "" ).replace( rtrim, "$1" );
+	};
+}
```
mgol, line 104: This is pretty long and has repetitions. I'd incorporate the support check into the function to make it shorter; how about:
```suggestion
jQuery.trim = function( text ) {
	if ( text == null ) {
		return "";
	}
	return support.trim ?
		( text + "" ).trim() :
		( text + "" ).replace( ltrim, "" ).replace( rtrim, "$1" );
};

```
?

src/deprecated.js
```diff
@@ -15,7 +15,13 @@ define( [
 
 // Support: Android <=4.0 only
 // Make sure we trim BOM and NBSP
-var rtrim = /^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g;
+// Require that the "whitespace run" starts from a non-whitespace
+// to avoid O(N^2) behavior when the engine would try matching "\s$" at each space position.
+// It is important that non-whitespace char is mandatory for regexp performance reasons,
+// however, it does not break correctness since whitespace-only string will be trimmed by ltrim above.
+var ltrim = /^[\s\uFEFF\xA0]+/,
+	rtrim = /([^\s\uFEFF\xA0])[\s\uFEFF\xA0]+$/,
```
mgol, line 23: I'm wondering if we really need it if we use native `trim` by default. It'd be much shorter if the fallback logic stayed the same. I understand that'd mean the performance issues still exist in Android <=4.0 but that browser is hardly used at this moment so making everyone download extra bytes to resolve a performance issue that only exists there doesn't seem worth it to me.

mgol (changes requested):
Thanks for the test cases.

I added some comments that should contribute to making this patch much smaller.

It's a bit tricky as this is a deprecated util that should generally not be used at all... The use case presented at https://github.com/jquery/jquery-migrate/pull/461#issuecomment-1167630231 says something about a slow JIRA instance. I wonder: would updating jQuery with this patch help here, considering that JIRA would have to update the included jQuery?

That said, I'd be leaning towards accepting this change with my suggestions as trimming is a pretty common operation and I imagine `$.trim` must be used a lot, especially in older code. My suggestions should make the size increase significantly down.

Thoughts, @timmywil @dmethvin @gibson042?

src/deprecated.js
mgol: Also, when you do this, you'll need to update `test/unit/support.js` with support test results for every listed browser. This should be `true` for all browsers except Android 4.0. Since results for Android 4.0-4.3 are grouped at the moment, you'll need to split that into 4.0 & 4.1-4.3.

src/deprecated.js
vlsi: > I understand that'd mean the performance issues still exist in Android <=4.0 

Performance issues happen in ALL the browsers. For instance, 10 second trim was on my Apple M1 Max notebook.

---

The new jQuery 4 has no `jQuery.trim` at all, so I am not sure if it is worth discussing browser support here.

If you can somehow prove that all the browsers supported by jQuery3 have a good native trim implementation, then we can indeed drop the regular expressions.

However, I just wanted to play safe here: keep backward compatibility (jQuery is used by a LOT of apps!), and improve the performance for well-known cases (native trim, and the one without native trim).

vlsi:
Here's profiling of JIRA issue I run into. I'm running Chrome 103.0.5060.114 on Apple M1 Max. There are two `jquery.trim` executions, and each of them takes 5 seconds.

JIRA freeze

vlsi:
I've merged functions into one:

```
   raw     gz Sizes
290221  85685 dist/jquery.js
 89781  31006 dist/jquery.min.js
  +580   +234 dist/jquery.js
  +128    +61 dist/jquery.min.js
```

vlsi:
I've created `deprecated/support`, and the size is

```
   raw     gz Sizes
290171  85668 dist/jquery.js
 89767  30997 dist/jquery.min.js
  +530   +217 dist/jquery.js
  +114    +52 dist/jquery.min.js
```

src/deprecated.js
vlsi: Just in case, IE9 does not seem to support trim natively: https://caniuse.com/mdn-javascript_builtins_string_trim, so the polyfill it might still be needed.

vlsi:
>I wonder: would updating jQuery with this patch help here, considering that JIRA would have to update the included jQuery?

Of course, it would definitely help.
If I remember well, then the optimized regex improves the performance from 5 seconds to 5 milliseconds. Unfortunately, the input data has sensitive information, and I would like to avoid spending time distilling the test case since the optimization is kind of obvious.

src/deprecated.js
mgol: > > I understand that'd mean the performance issues still exist in Android <=4.0
> 
> Performance issues happen in ALL the browsers. For instance, 10 second trim was on my Apple M1 Max notebook.

I don't think you're running Android 4.0 on your Apple M1 Max MacBook. :) Since native trim is preferred when it works, it would be applied everywhere except for Android 4.0. Therefore, the performance issue would only exist on that very old browser.

> The new jQuery 4 has no `jQuery.trim` at all, so I am not sure if it is worth discussing browser support here.

I don't understand this point. jQuery 3 & 4 have their own browser support matrices and they are of course relevant when discussing changes to land in either of those lines.

> Just in case, IE9 does not seem to support trim natively: [caniuse.com/mdn-javascript_builtins_string_trim](https://caniuse.com/mdn-javascript_builtins_string_trim), so the polyfill it might still be needed.

This info is incorrect. `trim` support in IE started with version 9, not 10.

> If you can somehow prove that all the browsers supported by jQuery3 have a good native trim implementation, then we can indeed drop the regular expressions.

But your PR always uses native trim as long as it doesn't have this Android 4.0 bug. If you found a browser with very slow native trim, your PR would still make it use that very slow trim. What does it have to do with the regular expressions used here?

vlsi: > But your PR always uses native trim as long as it doesn't have this Android 4.0 bug

I've no idea what was Android 4 bug, and the comments in code do not explain it.
So I assume something was wrong, and I just fix the regex so it is identical in the behavior.


>But your PR always uses native trim as long as it doesn't have this Android 4.0 bug. If you found a browser with very slow native trim, your PR would still make it use that very slow trim. What does it have to do with the regular expressions used here?

I assume native trim (if available) should always be faster.
However, I assume it is worth fixing the regex just in case there's a browser that does not support trim natively.

> Therefore, the performance issue would only exist on that very old browser.

I do not know if there are other browsers that fail to support trim natively, and if jQuery cares to support such browsers, then why make the support inherently slow?

src/deprecated.js
mgol: > Is there a framework to launch the test with all the browsers to see if it behaves properly?

You can run them - after building via `grunt` - on your local browsers by passing the comma-separated `BROWSERS` environment variable and setting `GITHUB_ACTION` to anything non-empty:
```
GITHUB_ACTION=1 BROWSERS=Chrome,FirefoxHeadless grunt karma:main
```
(the `GITHUB_ACTION` part shouldn't be necessary, I need to update our Gruntfile, but for now it is what it is)

To test on IE, you need to run Windows; then you should be able to pass `IE`, a proper Karma launcher is installed.

If you want to run it on a specific version of a browser that you don't have installed locally - especially IE - then it's not possible without a service like BrowserStack. We run such tests after merge at https://swarm.jquery.org/project/jquery.

You can also run:
```
grunt karma:chrome-debug
```
and then open the printed URL to a browser you want to test and click DEBUG.

Remember to keep the browser window focused, unless it's a headless browser like with the Firefox example above.

mgol: > a) It fails with obscure error.

These tests check that all support tests in the current browser are:
* listed in the proper results object above and match those values
* fail in at least one browser - i.e. at least in one of the lists the value is not equal to `true`

> The CI seems to verify only Chrome and Firefox, while I have to modify expected outputs for ALL the browsers. I am not really sure what safari vs safari_9_10 is, thus I can just plug "trim": true and pray it would work. I do not see much value in that type of tests.

The CI on GitHub runs only these two browsers but our full test infra at https://swarm.jquery.org/project/jquery runs them in all supported browsers. We just haven't ported all of this to GitHub Actions as there are issues with test flakiness that are easier to deal with in TestSwarm than on GitHub Actions. Those tests run post-merge so yes, we can't verify everything easily before merging other than by running tests on a browser where we expect issues manually.

That said, I expect `trim` to be supported everywhere except for Android 4.0 without issues. We've already checked this in the past which is why the support comment only mentioned Android 4.0.

mgol: BTW, there's a WIP PR (#4803) adding support for running browser tests on BrowserStack on GitHub Actions; without BrowserStack we can only run them on headless Chrome & headless Firefox.

mgol: BTW, there's a WIP PR (#4803) adding support for running browser tests on BrowserStack on GitHub Actions; without BrowserStack we can only run them on headless Chrome & headless Firefox.

However, there has been no movement on the PR recently.

vlsi: Ok, now I got a clean run in CI: https://github.com/vlsi/jquery/actions/runs/2663071906

src/deprecated/support.js
```diff
@@ -0,0 +1,11 @@
+define( [
+	"../var/support"
+], function( support ) {
+
+	"use strict";
+
+	support.trim = "".trim ? "_" === " \uFEFF\xA0_\uFEFF\xA0 ".trim() : false;
+
+	return support;
```
mgol, line 9: We avoid indenting contents of this outer function so that larger files are not completely indented; let's follow that here as well:
```suggestion
"use strict";

support.trim = "".trim ? "_" === " \uFEFF\xA0_\uFEFF\xA0 ".trim() : false;

return support;
```

src/deprecated.js
mgol: The jQuery browser support policy is defined at https://jquery.com/browser-support. We support those browsers and we don't really care about others. If they don't support `String#trim` then most likely they don't support a ton of other stuff that jQuery requires to work.

Krinkle:
Could we save space by adding only the native branch and possibly golf towards a (gzip-aware) compacter way to introduce this branch?

src/deprecated.js
```diff
@@ -6,16 +6,22 @@ define( [
...
 
 "use strict";
 
 // Support: Android <=4.0 only
 // Make sure we trim BOM and NBSP
-var rtrim = /^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g;
+// Require that the "whitespace run" starts from a non-whitespace
+// to avoid O(N^2) behavior when the engine would try matching "\s$" at each space position.
+// It is important that non-whitespace char is mandatory for regexp performance reasons,
+// however, it does not break correctness since whitespace-only string will be trimmed by ltrim above.
```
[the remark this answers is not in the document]
vlsi: Initially I created a comment on top of the variable, so `ltrim` was indeed above. However, the code style seems to forbid comment-per-variable, so I had to collapse all the comments on top of `var`, so `above` got stale.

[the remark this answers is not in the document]
mgol: There’s no rule forbidding comment per variable, each comment just needs to be preceded by an empty line.

mgol:
> Could we save space by adding only the native branch and possibly golf towards a (gzip-aware) compacter way to introduce this branch?

What do you mean by that? I don't follow.

As for your patch, the AMD definition changes are incorrect; the `"./deprecated/support"` dependency needs to stay where it was. Restoring the non-native branch to what's on `3.x-stable` is what I've suggested above as this specific optimization would only help Android 4.0 from all of our supported browsers.

I don't think replacing `_` with `:` in the trim test helps with size, does it? Also, I think you can save a byte by keeping the `rtrim` definition at the top, as it already is on `3.x-stable`.

vlsi:
: is more common in the file, so it might indeed help with gzip-compressed size

mgol:
Yeah, but there's no need to guess, it's enough to run `grunt` and see what it reports. In my local testing replacing `_` with `:` doesn't result in size gains so I'd leave it at `_`.

Krinkle:
> > Could we save space by adding only the native branch and possibly golf towards a (gzip-aware) compacter way to introduce this branch?
> 
> What do you mean by that? I don't follow. […] Restoring the non-native branch to what's on `3.x-stable` is what I've suggested above as this specific optimization would only help Android 4.0 from all of our supported browsers.

I didn't understand that from your comment, but I guess we're proposing the same thing then, as that is indeed what I am suggesting and what my diff does 🙂 

> As for your patch, the AMD definition changes are incorrect; the `"./deprecated/support"` dependency needs to stay where it was.
> 
> I don't think replacing `_` with `:` in the trim test helps with size, does it? Also, I think you can save a byte by keeping the `rtrim` definition at the top, […]

Why does it need to stay where it was? I did not see an implicit dependency here, but maybe I'm missing it? It seems `support.js` defines only the `support.trim` bool, which is not used by other files. I realize I forgot to update the function signature, that is indeed a mistake. I suspect the gains still hold up with that, we can verify that. The re-ordered definitions helped to bring the whitespace escape sequences and rtrm propertys closer together, which gained a few bytes.

I made each of these tweaks after numerous different tries and ideas, as informed by continual size information provided via `npx grunt dev`, keeping only what is beneficial and discarding the rest. The same goes for the colon, and for moving down the `var rtrim` to right above `jQuery.trim` assignment. Maybe a later change I made ended up making it smaller such that it also implicitly grew by a byte and thus making an earlier change redundant. I haven't yet gone back to undo ones that I already verified. But these changes (rm ltrim, change char, move define order, move var position) shrunk it compared to the previous run.

mgol:
> I didn't understand that from your comment, but I guess we're proposing the same thing then, as that is indeed what I am suggesting and what my diff does 🙂

That's good! 🙂 

> > I don't think replacing _ with : in the trim test helps with size, does it? Also, I think you can save a byte by keeping the rtrim definition at the top, […]
>
> Why does it need to stay where it was?

It doesn't _need to_, technically (although our code style uses a single `var` per function placed possibly close to the beginning of the function body), but in my local testing such changes (i.e. restoring the original fallback regex code) resulted in +38 bytes and you reported +39 so I thought moving it up would reduce the byte. I checked now, though, and I see +38 bytes regardless of where the `rtrim` definition is placed. In that case, I'd leave it at the top to make the code style similar to other files.

The size differences we're seeing is cased by another difference; my definition of `support.trim` didn't use a ternary; this will make it result in a `true` or `undefined` value instead of `true` or `false` but that doesn't really matter.

My local diff:
```diff
diff --git a/src/deprecated.js b/src/deprecated.js
index 095c11da..8995a2c5 100644
--- a/src/deprecated.js
+++ b/src/deprecated.js
@@ -16,12 +16,7 @@ define( [
 
 // Support: Android <=4.0 only
 // Make sure we trim BOM and NBSP
-// Require that the "whitespace run" starts from a non-whitespace
-// to avoid O(N^2) behavior when the engine would try matching "\s$" at each space position.
-// It is important that non-whitespace char in rtrim is mandatory for regexp performance reasons,
-// however, it does not break correctness since whitespace-only string will be trimmed by ltrim.
-var ltrim = /^[\s\uFEFF\xA0]+/,
-       rtrim = /([^\s\uFEFF\xA0])[\s\uFEFF\xA0]+$/;
+var rtrim = /^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g;
 
 // Bind a function to a context, optionally partially applying any
 // arguments.
@@ -91,6 +86,7 @@ jQuery.trim = function( text ) {
        }
        return support.trim ?
                ( text + "" ).trim() :
-               ( text + "" ).replace( ltrim, "" ).replace( rtrim, "$1" );
+               ( text + "" ).replace( rtrim, "" );
 };
+
 } );
diff --git a/src/deprecated/support.js b/src/deprecated/support.js
index 4eb7eb1c..028d5993 100644
--- a/src/deprecated/support.js
+++ b/src/deprecated/support.js
@@ -4,7 +4,7 @@ define( [
 
 "use strict";
 
-support.trim = "".trim ? "_" === " \uFEFF\xA0_\uFEFF\xA0 ".trim() : false;
+support.trim = "".trim && "_" === " \uFEFF\xA0_\uFEFF\xA0 ".trim();
 
 return support;
 
```

vlsi:
>his will make it result in a true or undefined value instead of true or false but that doesn't really matter.

It matters for the unit test. Unit test expects `false` there.

mgol:
Where do unit tests expect `false` here?

mgol:
Yes, but these are your tests. Also, Android 4.0 has `String#trim` but buggy so the `false` value would still apply here. Just browsers with no `trim` at all would get `undefined` there.

However, that made me realize we don't actually need for `String#trim` presence. All browsers we care about have it implenented (even Android 2.2 had it), just sometimes buggy. We can remove that part of the check, saving even more size.

vlsi:
```
   raw     gz Compared to 3.x-stable @ 0f6c3d9efc5f7e844bdcf8ef44f9327f414bea77
  +102    +28 dist/jquery.js
   +53    +29 dist/jquery.min.js
```

src/deprecated/support.js
```diff
@@ -0,0 +1,11 @@
+define( [
+	"../var/support"
+], function( support ) {
+
+"use strict";
+
+support.trim = "" === "\uFEFF\xA0".trim();
```
mgol, line 7: Nice optimization! I checked and it still fails on Android 4.0 as expected.

test/unit/support.js
```diff
@@ -284,7 +295,28 @@ testIframe(
...
+				"noCloneChecked": true,
+				"option": true,
+				"optSelected": true,
+				"pixelBoxStyles": false,
+				"pixelPosition": false,
+				"radioValue": true,
+				"reliableMarginLeft": false,
+				"reliableTrDimensions": true,
+				"scrollboxSize": true,
+				"trim": false
```
mgol, line 319: All the other browser entries are put from newest to oldest within the same family so let's put this after the `android` entry for consistency.

mgol (commented):
Just a few minor comments, otherwise looks good! I'm also happy where this landed size-wise.

test/unit/support.js
```diff
@@ -332,7 +369,9 @@ testIframe(
 		expected = expectedMap.firefox_60;
 	} else if ( /firefox/i.test( userAgent ) ) {
 		expected = expectedMap.firefox;
-	} else if ( /android 4\.[0-3]/i.test( userAgent ) ) {
+	} else if ( /android 4\.0/i.test( userAgent ) ) {
+		expected = expectedMap.android_4_0;
```
mgol, line 373: Let's put this after the Android 4.1-3 entry as well.

gibson042:
29 bytes seems like too much to me, but I would be comfortable with still improving pathological-input performance by orders of magnitude for the earlier 6-byte increase:
```patch
@@ -15,7 +15,7 @@ define( [
 
 // Support: Android <=4.0 only
 // Make sure we trim BOM and NBSP
-var rtrim = /^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g;
+var rtrim = /^[\s\uFEFF\xA0]+|([^\s\uFEFF\xA0])[\s\uFEFF\xA0]+$/g;
 
 // Bind a function to a context, optionally partially applying any
 // arguments.
@@ -82,6 +82,6 @@ jQuery.isNumeric = function( obj ) {
 jQuery.trim = function( text ) {
        return text == null ?
                "" :
-               ( text + "" ).replace( rtrim, "" );
+               ( text + "" ).replace( rtrim, "$1" );
 };
 } );
```

src/deprecated/support.js
gibson042: To be clear, I am arguing against the introduction of `support.trim` at +29 bytes (and really, even at +10).

gibson042, line 7: To be clear, I am arguing against the introduction of `support.trim` at +29 bytes (and really, even at +10).

gibson042 (changes requested):

src/deprecated/support.js
mgol: To me personally the biggest issue with size increases is they creep in, adding up to a meaningful increase over time. In this case, this danger does not exist as this is deprecated code that's already removed in `4.0.0-pre`. That makes +29 bytes more acceptable to me.

BTW, we could still make it smaller by dropping `\xA0` - this is actually trimmed properly by Android 4.0, `\uFEFF` is not, I just checked. That lowers the size increase to +24 bytes.

gibson042: Are you claiming that a real-life website, let alone a significant number of them, are spending so much time in `jQuery.trim` that an improvement from O(n²) to O(n) (with real effect of 240x for input with hundreds of medial whitespace characters and 23000x for input with tens of thousands, taking the OP 20 seconds down to about 1 ms) is not sufficient to address the overall experience? I'm not denying that native `String.prototype.trim` is astronomically faster still, but how would such sites even _get_ the input data fast enough for that further improvement to make a material difference?

That said, I'm not going to panic over 18 bytes in a legacy branch. If @mgol is in favor, that's fine with me.

Krinkle: A real site, can and should, of course, call the native `trim()` method directly if its product requirements allows that, and could do so, at least for the handful of rare cases where this is known to make a significant difference.

While sites may be embedding a particular jQuery version of a variety of different reasons (e.g. blocked on migration of code, or offer jQuery API to other plugins, or browser support for _one_ of the legacy browsers, such as the case for Wikipedia where the JS payload supports IE11 and Android 5 but not IE9 or Android 4). It's quite likely a consumer could even adopt String-trim unconditionally in its code, if all its clients support ES5, even if jQuery itself maintains wider support.

timmywil:
Thanks for your contribution @vlsi! We discussed this in the meeting and we'd prefer to go with [@gibson042's shorter solution](https://github.com/jquery/jquery/pull/5068#issuecomment-1184653387) and not add `support.trim` to 3.x this late in the game. If you can make that change, we'd be happy to get this into 3.6.1.

vlsi:
For reference, I've captured the problematic text (~900KiB).

Old regexp takes ~5 seconds.
The updated one takes 3ms.

For trivial cases, the updated regexp executes with a rate of ~11M/sec (the native executes at a rate of 130M/sec).
In other words, 10K trims would consume 1ms of CPU time.

Krinkle (approved):

timmywil (approved):

gibson042 (approved):

mgol:
Landed, thanks! This change will be included in the upcoming `3.6.1` release.

PR #5070, opened 2022-06-28
feat: option to include unselected values in form serialize

### LINKED ISSUE #5033: add a flag to serialize() to include empty <select multiple> values
### Description ###

if we use .serialize() on a <form> that contains a <select multiple> element, that happens to have no selections, then it is not included in the serialized output.

I propose that .serialize() have an optional parameter .serialize(includeAll) that if defined and true would include such fields, with an empty value.

e.g.

someField=test**_&emptySelect=_**&anotherfield=123

it is sometimes useful for a backend script to know that the user emptied a select.

mikeirvingweb:
### Summary ###
New optional parameter `includeUnseleted` in **Form Serialize** - `serialize: function( includeUnselected )` that specifies that we'd like to output **unselected** `input[type=radio]`, `input[type=checkbox]` and `select[multiple]` elements to the QueryString as empty values, e.g. _myRadioButton=&myCheckbox=&myMultiSelect=_

this commit resolves #5033

### Checklist ###

### DIFF
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -92,8 +92,25 @@ jQuery.param = function( a, traditional ) {
 };
 
 jQuery.fn.extend( {
-	serialize: function() {
-		return jQuery.param( this.serializeArray() );
+	serialize: function( includeUnselected ) {
+		var serialized = jQuery.param( this.serializeArray() ), empty = "";
+
+		this.each( function() {
+			if ( includeUnselected ) {
+				jQuery( this ).find(
+					"input[name][type=checkbox]:not(:checked),input[name][type=radio]:not(:checked)"
+				).each( function() {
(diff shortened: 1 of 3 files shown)


timmywil:
Thank you for your contribution. We discussed this in the meeting, but considering it's been this way for so long and considering the size of this PR, we don't feel it's worth it. `serialize` has never fit all use cases, just the most common ones. Anything beyond that is usually done in a plugin or custom code.

mikeirvingweb:
@timmywil yes, a Plug-in was my Plan B. I may progress with that soon. Thanks for considering my PR though 👍

mikeirvingweb:
I have just published a modified version as a plugin at https://github.com/mikeirvingweb/jquery-serialize-all

Issue #4418, opened 2019-06-07
jQuery does not respond to events on shadow root


paul-f:
### Description ###
Running Chrome 73.0.3683.86 on Windows 7 with jQuery 3.4.1.

When you use jQuery's on() function on the shadow root of a custom element, it does not receive events. Using plain JS addEventListener() works as expected.

This occurs because a shadow root is implemented as a document fragment, and jQuery specifically does not attach data to non-element nodes; see line 4047 call to acceptData().

### Link to test case ###
https://jsbin.com/ketocucaku/edit?html,js,output

timmywil:
Duplicate of #4317

Issue #4317, opened 2019-03-09
Is not possible to listen events on ShadowRoot


blikblum:
### Description ###

When trying to listen events on a ShadowRoot element using `on`, the handler is not fired.
The native addEventListener works

Minimal example:
```javascript
const el = $("#test")[0];
el.attachShadow({ mode: "open" });
el.shadowRoot.innerHTML = `<div class=".inner">Inner content</div>`;
$(el.shadowRoot).on("click", ".inner", () => {
  console.log("jquery - shadow inner clicked");
});
```

The issue is that the element data returned by dataPriv is always empty because of [this check](https://github.com/jquery/jquery/blob/master/src/data/var/acceptData.js#L16)

ShadowRoot elements have nodeType === 11

The [issue](https://bugs.jquery.com/ticket/8335) [pointed](https://github.com/jquery/jquery/blob/e743cbd28553267f955f71ea7248377915613fd9/src/data/Data.js#L28) as reason for the check seems not valid for ShadowRoot elements which are supported only on modern browsers

So, changing the check for 
```javascript
return owner.nodeType === 1 || owner.nodeType === 9 || owner.nodeType === 11 || !( +owner.nodeType );
```

should work

### Link to test case ###
https://codepen.io/blikblum/pen/QogVgO?editors=1111

dmethvin:
I agree with the bug but not the solution. We shouldn't be adding event listeners to any `nodeType===11` which includes `DocumentFragment`. For example, if a jQuery event handler was attached to a `DocumentFragment` and the DF was later inserted in to the document, we wouldn't clean up internal data structures.

blikblum:
BTW: the underlying issue is not specific for events. It also affects [setting data](https://codepen.io/blikblum/pen/zbMpaW?editors=1111).

AFAICS the memory leak that prevented attaching data to non elements does not occurs with current implementation of jQuery and >= IE8, so should not be a problem.

If attaching to `DocumentFragment` is still a concern, should be acceptable a more specific check to restrict to `ShadowRoot`?

Something like:
```javascript
return owner.nodeType === 1 || owner.nodeType === 9 || (owner.nodeType === 11 && owner.host) || !( +owner.nodeType );
```

dmethvin:
Yeah that might work!

extempl:
Bump?

dmethvin:
Hi @extempl, a proposed solution is above. Would you like to submit a pull request?

Kevinjohn:
[...]

PR #5076, opened 2022-07-26
Data: ShadowRoot support listen events & data get/set

fecore1:
### Summary ###

JQuery adds constraints, do not allow add data and add event listen to non-elements, see [bugs-8335](https://bugs.jquery.com/ticket/8335)

And ShadowRoot elements have `nodeType === 11`

But ShadowRoot elements are supported only on modern browsers, so this fix (owner.nodeType === 11 && owner.host) can be discussed here

Fixes [gh-4317](https://github.com/jquery/jquery/issues/4317) [gh-4418](https://github.com/jquery/jquery/issues/4418)

### Checklist ###

### DIFF
--- a/src/data/var/acceptData.js
+++ b/src/data/var/acceptData.js
@@ -7,7 +7,7 @@ export default function( owner ) {
 	//  - Node
 	//    - Node.ELEMENT_NODE
 	//    - Node.DOCUMENT_NODE
-	//    - Node.SHADOWROOT_NODE
+	//    - Node.DOCUMENT_FRAGMENT_NODE
 	//  - Object
 	//    - Any
 	return owner.nodeType === 1 || owner.nodeType === 9 ||
--- a/src/data/var/acceptData.js
+++ b/src/data/var/acceptData.js
@@ -7,7 +7,9 @@ export default function( owner ) {
 	//  - Node
 	//    - Node.ELEMENT_NODE
 	//    - Node.DOCUMENT_NODE
+	//    - Node.SHADOWROOT_NODE
 	//  - Object
 	//    - Any
-	return owner.nodeType === 1 || owner.nodeType === 9 || !( +owner.nodeType );
+	return owner.nodeType === 1 || owner.nodeType === 9 ||
+		( owner.nodeType === 11 && owner.host ) || !( +owner.nodeType );
 }
--- a/test/unit/data.js
+++ b/test/unit/data.js
@@ -29,6 +29,55 @@ QUnit.test( "jQuery.data & removeData, expected returns", function( assert ) {
 
 } );
 
+QUnit[ document.body.attachShadow && document.body.getRootNode ?
+	"test" :
+	"skip"
+]( "jQuery.data & removeData, expected returns, shadow dom", function( assert ) {
+
+	assert.expect( 4 );
+
+	jQuery( "<div id='shadowHost'></div>" ).appendTo( "#qunit-fixture" );
+	var shadowHost = document.querySelector( "#shadowHost" );
+	var shadowRoot$ = jQuery( shadowHost.attachShadow( { mode: "open" } ) );
+
+	shadowRoot$.data( "hello", "world" );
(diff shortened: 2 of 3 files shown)


src/data/var/acceptData.js
```diff
@@ -7,7 +7,9 @@ export default function( owner ) {
 	//  - Node
 	//    - Node.ELEMENT_NODE
 	//    - Node.DOCUMENT_NODE
+	//    - Node.SHADOWROOT_NODE
```
mgol, line 10: There's no such API as `Node.SHADOWROOT_NODE`. `nodeType === 11` represents `Node.DOCUMENT_FRAGMENT_NODE` and is shared by all document fragments.

src/data/var/acceptData.js
```diff
@@ -7,7 +7,9 @@ export default function( owner ) {
...
 	//  - Object
 	//    - Any
-	return owner.nodeType === 1 || owner.nodeType === 9 || !( +owner.nodeType );
+	return owner.nodeType === 1 || owner.nodeType === 9 ||
+		( owner.nodeType === 11 && owner.host ) || !( +owner.nodeType );
```
mgol, line 14: @jquery/core do you remember the reasons we don't allow data on document fragments? Or in general, what are the reasons to maintain any restrictions? In the past there were issues with host objects, browser plugins, etc., but plugins went away and many of these problems got smoothed out.

mgol (commented):
Thanks for the PR. We'll review it and provide further feedback.

src/data/var/acceptData.js
fecore1: > nodeType === 11 represents Node.DOCUMENT_FRAGMENT_NODE and is shared by all document fragments.

Yes, So the condition `owner.host` was added to limit the [shadowroot](https://dom.spec.whatwg.org/#shadowroot)

Or Maybe `toString.call( obj ) === "[object ShadowRoot]"` better?

mgol: My comment was mostly about the comment. :) These is not just a description, `Node.DOCUMENT_NODE` etc. are real APIs; you can type them in the browser console and you'll see they resolve to the underlying numbers. We just don't use these APIs but use numbers directly to save on size.

Meanwhile, there's no such thing as `Node.SHADOWROOT_NODE` - if you type it in the browser console, it will print `undefined` in all browsers.

See https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType

fecore1: got it :)

I just want to make it clear here that `nodeType === 11 && owner.node` limit the [shadowroot](https://dom.spec.whatwg.org/#shadowroot), and I'll change to `Node.DOCUMENT_FRAGMENT_NODE` first.

sujand: > > nodeType === 11 represents Node.DOCUMENT_FRAGMENT_NODE and is shared by all document fragments.
> 
> 
> 
> Yes, So the condition `owner.host` was added to limit the [shadowroot](https://dom.spec.whatwg.org/#shadowroot)
> 
> 
> 
> Or Maybe `toString.call( obj ) === "[object ShadowRoot]"` better?

mgol (changes requested):
I found why we're restricting `nodeTypes` in `acceptData`; it started with https://bugs.jquery.com/ticket/8335: jQuery doesn't properly clean up data for non-elements as most jQuery APIs only work with elements.

Document Fragments hit the same issue and you can easily reproduce it after changing `acceptData` implementation to always return true: attaching data to the shadow root and then calling `.remove()` on its parent still leaves the data attached to the shadow root element.

This may include jQuery private data, including event handlers. For regular data it might be a smaller issue but various projects may depend on `jQuery.cleanData()` being called for all data-holding nodes, even if they are document fragments.

I think we need to resolve this issue first before expanding the list.

mgol:
@fecore1 Can I ask you why you need to attach data to shadow roots? What's the use case?

mgol:
@fecore1 ping

mgol:
Closing for lack of response. We can reopen if we get an answer to the question.

PR #5077, opened 2022-07-26
Core: Drop support for IE (all versions)

mgol:
### Summary ###

~~This PR is to show what dropping IE support would look like in code. If we make a decision to drop IE in v4, this PR can be merged, guarded on possible review remarks.~~ EDIT: we've decided to land this in `5.0.0`.

Of course, dropping IE would open a lot of other doors for various refactors but those would be considered separately.

Size difference:
```
main @d6fb53bce28d597ac309ec81efe761bd410bdf41
   raw     gz     br Filename
 -1968   -621   -521 dist/jquery.min.js
 -1762   -580   -498 dist/jquery.slim.min.js
 -1968   -621   -582 dist-module/jquery.module.min.js
 -1762   -583   -506 dist-module/jquery.slim.module.min.js
```

### Checklist ###


src/event.js
```diff
@@ -511,10 +498,10 @@ jQuery.event = {
 // synthetic events by interrupting progress until reinvoked in response to
 // *native* events that it fires directly, ensuring that state changes have
 // already occurred before other listeners are invoked.
-function leverageNative( el, type, expectSync ) {
+function leverageNative( el, type, isSetup ) {
```
mgol, line 501: @gibson042 This is a tricky function, I'd feel safer if you reviewed those changes.

src/event.js
```diff
@@ -363,8 +352,6 @@ jQuery.event = {
 			// Support: Firefox <=42 - 66+
 			// Suppress spec-violating clicks indicating a non-primary pointer button (trac-3861)
 			// https://www.w3.org/TR/DOM-Level-3-Events/#event-type-click
-			// Support: IE 11+
-			// ...but not arrow key "clicks" of radio inputs, which can have `button` -1 (gh-2343)
```
mgol, line 367: Firefox doesn't have the behavior from the above comment (`click` events are not fired for non-primary mouse buttons) so perhaps this all could be removed. But one of our tests checks this condition explicitly so I'd rather leave it for a separate PR.

src/event.js
mgol: It’s much easier to understand the logic when asynchronous handling is removed from here, though. 😅

src/selector.js
```diff
@@ -209,8 +193,7 @@ function find( selector, context, results, seed ) {
 			}
 
 			// Take advantage of querySelectorAll
-			if ( !nonnativeSelectorCache[ selector + " " ] &&
-				( !rbuggyQSA || !rbuggyQSA.test( selector ) ) ) {
```
mgol, line 213: At this point, all buggy qSA handling applied to IE only. With this gone, perhaps we wouldn’t need rewriting of buggy selectors in 4.0. We’d still need parsing to inject `:scope` or a temporary ID and to support custom pseudos, though.

test/delegatetest.html
```diff
@@ -172,30 +172,20 @@ <h2>Submit Tests</h2>
...
+		if ( api === "onX" ) {
 			$(this).find("input, button, select, textarea").each(function(){
 				this["on"+type] = function(e){ e = $.event.fix(e||event); e.data = $cell; blinker.call(this, e); };
 			});
-		} else if ( api == "bind" ) {
-			$(this).find("input, button, select, textarea").bind(type, $cell, blinker);
+		} else if ( api === "bind" ) {
+			$(this).find("input, button, select, textarea").on(type, $cell, blinker);
 		} else {
-			$(id+" input,"+id+" button,"+id+" select,"+id+" textarea").live(type, $cell, blinker);
```
mgol, line 182: This file has been broken for a long time due to the use of `live`… But it looks potentially useful so I fixed it.

dmethvin:
Seems like the right move at this point. Things clean up nicely!

ollie-iterators:
The blog about jquery 3.7.0: https://blog.jquery.com/2023/05/11/jquery-3-7-0-released-staying-in-order/ says that "support for IE will be dropped in jQuery 4.0"

mgol:
@ollie-iterators Thanks for the info; that should say 5.0; we’ll fix it.

src/event.js
mgol: This code landed separately, as part of making focus events sync in IE.

src/selector.js
mgol: This code is back here as we had to handle `:has`.

vrubleg:
It's 2023, IE11 is 10 years old, half of the web already doesn't work in this browser. IE11 was introduced with Windows 8.1 that is already not supported by a lot of software (including Firefox and Chrome) and libraries (like Qt6+ or Python 3.9+). It makes sense to drop IE earlier.

mgol:
@vrubleg I understand the sentiment and I admit more than a year has passes since I raised this PR but I really don't want to make any changes to our 4.0.0 plans at this point as any further change is just going to delay the whole thing, and it's already been delayed a lot.

I want to get the beta out with only our current blockers resolved. Once we have that, once we also have the respective Migrate release and the upgrade guide prepared, we can talk about it once again.

nagyimre1980:
IE drop 4.1.0 ?

mgol:
@nagyimre1980 No; we treat reducing browser support policy as a breaking change so it has to land in a major version update.

vrubleg:
It's 2025. IE11 is not supported by Microsoft for years. I just checked stats in my service that still supports IE11. Among 16751 unique users over last 4 weeks there were just 2 with IE11. While it's just one service and it doesn't represent the whole internet, the number is not surprising. Just by looking on browser versions of my users, I can tell that other already dropped old versions of other browsers are many times more popular than IE11.

Just merge this PR and cross out the sentence about IE11 support in the announcement. Further refactoring that dropping IE11 allows can be done later when there is more time for it. It's 2025, it's easy to justify that the plan to not drop it has changed.

Whoever needs obsolete browser support, they'll continue to use older jQuery versions anyway.

mgol:
@vrubleg at this point, we're trying to get an RC and then a 4.0.0 stable release ASAP without more changes to the library. 4.x will definitely be the last line supporting any version of IE.

Also, note that it's not just IE 11; we now also test on Edge in IE mode which is still supported by Microsoft without any announced EOL date, even on Windows 11, and that uses the IE 11 engine. I get that it's still pretty niche, but we feel it's worth allowing apps depending on IE mode to be able to update to a supported version of jQuery even in a few years.

IRainman:
Why only in 5.0? WebView on Windows 7 is much more functional and convenient and also too outdated as IE11… IE11 isn't legacy; it's dead completely. You can use many improvements in code generation with let and const as basic features, and many more if we drop support for IE completely.

IRainman:
By removing IE support, you also remove many crutches and difficult logic that aren't needed in the modern browsers.

IRainman:
Furthermore, please check [this](https://en.wikipedia.org/wiki/Usage_share_of_web_browsers) — IE isn't used at all.

IRainman:
Moreover, [Firefox ESR 115 still supports Windows 7 and plans to support it until 2026](https://whattrainisitnow.com/release/?version=esr). You can completely remove support for IE without any issues. 

We decided to [extend support](https://blog.mozilla.org/futurereleases/2025/09/04/firefox-esr-115-support-for-windows-7-8-and-8-1-and-old-mac-os-versions-extended-until-march-2026/) for ESR 115 only on Windows 7-8.1 and macOS 10.12-10.14 up to March 2026.
We will re-evaluate this decision in February 2026 and announce any updates on ESR 115's end-of-life then.

nagyimre1980:
Agree. IE11 has been dead for years! Considering the current release pace, in my opinion, 5.0 will take at least 4-5 years to appear!My main issue: Speed: It would be noticeably faster if we removed all those IE11 branches!The final release of 4.0 is still many months away (I don't think it'll even come out this year)

PR #5085, opened 2022-08-10
Selector: Re-introduce selector-native.js

mgol:
### Summary ###

Due to the change of scope for `4.0.0`, this re-introduces the `selector-native` module (removed in gh-4395) known from jQuery 3.x.

I wasn't sure what to name it as we shouldn't use the name Sizzle anymore so I opted for `selector-full`. Suggestions for better naming welcome.

~~There are two selector tests that I still need to fix.~~ Fixed

Ref gh-4395

### Checklist ###


test/unit/selector.js
```diff
@@ -1,7 +1,7 @@
 QUnit.module( "selector", {
 	beforeEach: function() {
 		this.safari = /\bsafari\b/i.test( navigator.userAgent ) &&
-			!/\bchrome\b/i.test( navigator.userAgent );
+			!/\b(?:headless)?chrome\b/i.test( navigator.userAgent );
```
mgol, line 4: Tests in CI are running on Chrome Headless which has a different user agent, making `this.safari === true`. This is what caused the original test failures; it was just a test issue.

mgol:
CI is green now.

build/tasks/build.js
```diff
@@ -128,11 +128,11 @@ module.exports = function( grunt ) {
 			 * Adds the specified module to the excluded or included list, depending on the flag
 			 * @param {String} flag A module path relative to
 			 *  the src directory starting with + or - to indicate
-			 *  whether it should included or excluded
+			 *  whether it should be included or excluded
 			 */
 			const excluder = flag => {
 				let additional;
-				const m = /^(\+|\-|)([\w\/-]+)$/.exec( flag );
+				const m = /^(\+|-|)([\w\/-]+)$/.exec( flag );
```
mgol, line 135: This was just WebStorm complaining that the escape was not necessary.

README.md
```diff
@@ -101,6 +101,12 @@ Some example modules that can be excluded are:
 - **exports/global**: Exclude the attachment of global jQuery variables ($ and jQuery) to the window.
 - **exports/amd**: Exclude the AMD definition.
 
+As a special case, you may also replace the full jQuery `selector` module by using a special flag `grunt custom:-selector`.
```
gibson042, line 104: Is this true, or is it necessary to use `-selector-full`?

mgol: That was an omission on my part. It doesn't work right now, it will actually print a warning. But I'll update the PR since we now want to use this instead of `custom:-selector-full`.

mgol: Done. The renaming is done in the latest commit.

src/selector-native.js
```diff
@@ -0,0 +1,88 @@
...
+ * :has(selector)
+ * :not(complex selector)
+ * custom selectors via jQuery extensions
+ * Leading combinators (e.g., $collection.find("> *"))
+ * Reliable functionality on XML fragments
+ * Requiring all parts of a selector to match elements under context
+ *   (e.g., $div.find("div > *") now matches children of $div)
+ * Matching against non-elements
+ * Reliable sorting of disconnected nodes
+ * querySelectorAll bug fixes (e.g., unreliable :focus on WebKit)
```
gibson042, line 20: Can we turn these lines into bullets?
```suggestion
 * * Attribute not equal selector (!=)
 * * Positional selectors (:first; :eq(n); :odd; etc.)
 * * Type selectors (:input; :checkbox; :button; etc.)
 * * State-based selectors (:animated; :visible; :hidden; etc.)
 * * :has(selector)
 * * :not(complex selector)
 * * custom selectors via jQuery extensions
 * * Leading combinators (e.g., $collection.find("> *"))
 * * Reliable functionality on XML fragments
 * * Requiring all parts of a selector to match elements under context
 *     (e.g., $div.find("div > *") now matches children of $div)
 * * Matching against non-elements
 * * Reliable sorting of disconnected nodes
 * * querySelectorAll bug fixes (e.g., unreliable :focus on WebKit)
```

src/selector-native.js
```diff
@@ -0,0 +1,88 @@
...
+ * Positional selectors (:first; :eq(n); :odd; etc.)
+ * Type selectors (:input; :checkbox; :button; etc.)
+ * State-based selectors (:animated; :visible; :hidden; etc.)
+ * :has(selector)
+ * :not(complex selector)
+ * custom selectors via jQuery extensions
+ * Leading combinators (e.g., $collection.find("> *"))
+ * Reliable functionality on XML fragments
+ * Requiring all parts of a selector to match elements under context
+ *   (e.g., $div.find("div > *") now matches children of $div)
```
gibson042, line 17: I wonder if this a step too far... would it be worthwhile to preserve that jQuery behavior by filtering `querySelectorAll` results like e.g. `elem => jQuery.find.matchesSelector( elem, "#" + ephemeralId + " *" )` or `elem => jQuery.contains( context, elem )`?

There's also a larger but more performant version of the former using rudimentary selector transformation, e.g.
```js
// Prefix each item in the `selector` list with a scope prefix.
var match,
	tokenReplacements = [],
	functionReplacements = [],
	// Helper here for clarity, inlining at its two call sites probably compresses better.
	restoreFunctions = s => s.replace( /"(\d+)"/g, (_, i) => functionReplacements[ i - 1 ] );

// Temporarily replace backslash escapes and strings,
// which could otherwise disrupt identification of the list's constituent selectors.
selector = selector.replace(
	/\\[^]|(['"])(?:\\[^]|(?!\1)[^\\])*\1/g,
	s => "'" + tokenReplacements.push( s ) + "'"
);

// Likewise for functional pseudo-classes, but recursively.
while ( (match = /([^]*?)(\([^()]*,[^()]*\))([^]*)/.exec(selector)) ) {
	selector = match[ 1 ] +
		'"' + functionReplacements.push( restoreFunctions( match[ 2 ] ) ) + '"' +
		match[ 3 ];
}

// Prefix and restore replacements in each selector of the list.
selector = selector.replace(
	/[^,]+/g,
	sel => scopePrefix +
		restoreFunctions(sel).replace( /'(\d+)'/g, (_, i) => tokenReplacements[ i - 1 ] ) )
);
```
(where `scopePrefix` is either `":scope "` or `"#" + ephemeralId + " "`)

src/selector-native.js
mgol: Sure. We should backport it to `3.x-stable` as well.

src/selector-native.js
mgol: @gibson042 The issue is not that the results are not children of the context but that not the whole selector matches the contents part. It seems to me that just checking `elem => jQuery.find.matchesSelector( elem, "#" + ephemeralId + " *" )` or `elem => jQuery.contains( context, elem )` doesn't find those issues?

To make sure the whole selector matches within the context, we'd need to first split the selector list into individual selectors which requires some parsing; that would make `selector-native` much larger. Am I missing anything?

mgol: OK, would you like any further changes to this module for `4.0.0` then?

mgol: Yeah, it's a big change. But it's been like that in the `3.x-stable` version as well and avoiding that would require making this module much bigger. If you see a quick & reliable way to just split the selector list to individual selectors that can be prepended with `:scope`, that'd be great; otherwise, I don't see a great way to resolve this.

mgol: I’d have to try porting it; I’m not that fluent with selector code.

timmywil (approved):
LGTM

src/selector-native.js
mgol: @gibson042 I'm checking this right now but would you mind merging this PR as-is and exploring such an extension in a separate one? This PR largely restores the current 3.x version of `selector-native` so I feel it'd be beneficial to separate extending it.

mgol: My preliminary checks seem to indicate resolving this issue would cost about 1000 bytes minified gzipped; it'd still be 3300 bytes minified gzipped smaller than the full build compared to 4300 bytes minified gzipped without this change.

@timmywil @gibson042 what do you think? Is it a good idea? I can prepare a POC PR. I'd still like to keep it separate from this one, though.

gibson042: I agree with separating it, since selector-native has always had such a limitation.

mgol: I created #5185 to track this issue.

Issue #5105, opened 2022-09-03
.css("--custom") throws TypeError: ret is undefined (3.6.0 → 3.6.1 regression)


andersk:
### Description ###

In jQuery 3.6.0, `$element.css("--custom")` correctly returned `undefined` for a custom property that isn’t set. But in jQuery 3.6.1, it crashes with `TypeError: ret is undefined`.

This breaks the Zulip settings dialog.

This was presumably introduced by #4930 (cc @fecore1 @mgol).

### Link to test case ###

https://codepen.io/anderskaseorg/pen/rNvNZWW

mgol:
Thanks for the report. That said, I'm not sure the `undefined` return value was correct either. We are using `.getPropertyValue()` for custom property values and it returns an empty string for unknown CSS custom properties. We just then fallback to `computed[ value ]` which will be undefined. The logic is at:
https://github.com/jquery/jquery/blob/3.6.1/src/css/curCSS.js#L31

That said, for regular unknown properties that is the logic we are using. I guess we should discuss what's our goal here.

andersk:
Yeah, jQuery has given `undefined` here for standard properties since 1.8.0, and for custom properties since they were first supported in 3.2.0. “Correct” or not, it seems to me that a patch release isn’t a good time to break this compatibility.

mgol:
I agree with that but there’s also a case with a custom property set to an empty string or a whitespace-only string. IMO then we should return an empty string, not undefined. I’ll need to check if this is possible with the Web APIs we have.

mgol:
Fix landed on `main` in ed306c0261ab63746040e5d58bb4477c3069a427 and on `3.x-stable` in c0db6d70320ad6fb29a6a89aae3811a81b5adedf.

Issue #5098, opened 2022-08-31
Selector `:contains` within `:has` seems broken on Chrome


Rinzwind:
Using the following example, the console shows `1,1,1,1` on Safari (15.6.1) and Firefox (104.0.1), but `1,1,0,0` on Chrome (105.0.5195.52). I would expect the result to also be `1,1,1,1` on Chrome. Based on our usage in [Selenium](https://www.selenium.dev), I think the result actually was also `1,1,1,1` on previous versions of Chrome (≤ 104).

```html
<html>
<head>
	<script src="https://code.jquery.com/jquery-3.6.1.min.js"></script>
</head>
<body>
	<ul>
		<li>Item</li>
	</ul>
	<script>
		console.log("" + [
			$("li:contains('Item')").length,
			$("ul:has(li)").length,
			$("ul:has(li:contains('Item'))").length,
			$("ul:has(> li:contains('Item'))").length ]);
	</script>
</body>
</html>
```

mgol:
Thanks for the report!

The way the jQuery selector engine works, the selector is tried against `querySelectorAll` with minor modifications and if it fails, it goes through the internal selector engine. It's either-or.

Chrome 105 doesn't support `:contains` as it's a non-native selector but it does support `:has`. I'd expect that to still fail the internal check but, surprisingly, `document.querySelectorAll("ul:has(li:contains('Item'))")` no longer throws an error in Chrome 105. It looks like it must be doing some lax internal parsing?

This, indeed, sounds problematic. I wonder if we should detect `:contains` presence in the selector and default to the internal jQuery engine in such a case? An interesting issue.

@miketaylr @paulirish could you have a look? This change in Chrome 105 has potential compatibility issues.

mgol:
cc @gibson042 for ideas on how to potentially handle it on jQuery side. That won't resolve compatibility issues in jQuery versions in the wild, though.

mgol:
Worth noting is that this change makes Chrome fail both the jQuery test suite on `main`:
<img width="1170" alt="Screen Shot 2022-08-31 at 11 48 42" src="https://user-images.githubusercontent.com/1758366/187650776-6b911a27-324b-44c2-bd23-6df3a494fb49.png">

and the Sizzle one:
<img width="962" alt="Screen Shot 2022-08-31 at 11 48 52" src="https://user-images.githubusercontent.com/1758366/187650811-6f1aeb59-f1ad-4b57-93ad-4bee66356549.png">

[...]

PR #5106, opened 2022-09-03
CSS: Don’t trim whitespace of undefined custom property

andersk:
### Summary ###

Fixes gh-5105

### Checklist ###

### DIFF
--- a/test/unit/css.js
+++ b/test/unit/css.js
@@ -1748,6 +1748,7 @@ QUnit.testUnlessIE( "css(--customProperty)", function( assert ) {
 		"        --prop10:\f\r\n\t val10 \f\r\n\t;\n" +
 		"        --prop11:\u000C\u000D\u000A\u0009\u0020val11\u0020\u0009\u000A\u000D\u000C;\n" +
 		"        --prop12:\u000Bval12\u000B;\n" +
+		"        --empty:;\n" +
 		"    }\n" +
 		"</style>"
 	);
@@ -1756,7 +1757,7 @@ QUnit.testUnlessIE( "css(--customProperty)", function( assert ) {
 		$elem = jQuery( "<div>" ).addClass( "test__customProperties" )
 			.appendTo( "#qunit-fixture" ),
 		webkitOrBlink = /\bsafari\b/i.test( navigator.userAgent ),
-		expected = 17;
+		expected = 19;
 
 	if ( webkitOrBlink ) {
 		expected -= 2;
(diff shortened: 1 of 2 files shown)


mgol:
Thanks for the PR with a test!

We’ll discuss what approach we want to take here and we’ll get back to you.

test/unit/css.js
```diff
@@ -1802,6 +1802,7 @@ QUnit.testUnlessIE( "css(--customProperty)", function( assert ) {
 	assert.equal( $elem.css( "--prop10" ), "val10", "Multiple preceding and following escaped unicode whitespace trimmed" );
 	assert.equal( $elem.css( "--prop11" ), "val11", "Multiple preceding and following unicode whitespace trimmed" );
 	assert.equal( $elem.css( "--prop12" ), "\u000Bval12\u000B", "Multiple preceding and following non-CSS whitespace reserved" );
+	assert.equal( $elem.css( "--nonexistent" ), undefined );
```
mgol, line 1805: Please also add a test for a CSS variable that is defined but with no meaningful value: e.g. `--empty: ;`. This should also return `undefined`. I wish we were able to return an empty string there but I don't see a way so let's at least test the current behavior.

mgol (changes requested):
Just one comment, otherwise it looks good to me. Thanks!

test/unit/css.js
andersk: Added.

mgol (changes requested):
One of the tests failed in Chrome; can you have a look?

Disregard the `selector` `:has` failures, they are caused by #5098 and will be fixed when #5107 lands.

andersk:
This is a cross-browser difference with `--empty: ;`. In Firefox, `.getPropertyValue("--empty")` returns the empty string `""`, while in Chrome, it returns a space `" "`.

I’ve changed the test to `--empty:;` without the space to avoid this issue. We can consider further normalizing this cross-browser difference, but since that’s neither a crash nor a regression, I think we should consider it separately.

mgol (approved):
Makes sense, thanks! I'll need to ask you to rebase the PR once #5107 lands so that the build is green here. Other than that, LGTM.

andersk:
Rebased.

andersk:
Backported to `3.x-stable` as #5118.

mgol:
Landed on `main` in ed306c0261ab63746040e5d58bb4477c3069a427 and on `3.x-stable` in c0db6d70320ad6fb29a6a89aae3811a81b5adedf. Thanks!

mgol:
> I’ve changed the test to `--empty:;` without the space to avoid this issue. We can consider further normalizing this cross-browser difference, but since that’s neither a crash nor a regression, I think we should consider it separately.

I submitted #5120 to handle this case.

PR #5107, opened 2022-09-05
Selector: Use jQuery :has if `CSS.supports(selector(...))` non-compliant

mgol:
### Summary ###

jQuery has followed the following logic for selector handling for ages:
1. Modify the selector to adhere to scoping rules jQuery mandates.
2. Try `qSA` on the modified selector. If it succeeds, use the results.
3. If `qSA` threw an error, run the jQuery custom traversal instead.

It worked fine so far but now CSS has a concept of forgiving selector lists that
some selectors like `:is()` & `:has()` use. That means providing unrecognized
selectors as parameters to `:is()` & `:has()` no longer throws an error, it will
just return no results. That made browsers with native `:has()` support break
selectors using jQuery extensions inside, e.g. `:has(:contains("Item"))`.

Detecting support for selectors can also be done via:

```js
CSS.supports( "selector(SELECTOR_TO_BE_TESTED)" )
```
which returns a boolean. There was a recent spec change requiring this API to
always use non-forgiving parsing:
https://github.com/w3c/csswg-drafts/issues/7280#issuecomment-1143852187
However, no browsers have implemented this change so far.

To solve this, two changes are being made:
1. In browsers supports the new spec change to `CSS.supports( "selector()" )`,
   use it before trying `qSA`.
2. Otherwise, add `:has` to the buggy selectors list.

Fixes gh-5098

+117 bytes but +91 out of that is this new mechanism using `CSS.supports( "selector()" )` which we have to add anyway due to what I described above. Of course, suggestions on how to reduce that welcome.

A backport of that to Sizzle will be required once this PR is accepted.

### Checklist ###


mgol:
Chrome 105 is not available in GitHub Actions yet but I checked locally that all tests are passing there.

mgol:
Oh, tests already run on Chrome 105 in this PR. That confirms the fix works.

mgol:
Sizzle version of this PR, meant for jQuery 3.6.2: https://github.com/jquery/sizzle/pull/486

src/selector.js
```diff
@@ -252,6 +253,27 @@ function find( selector, context, results, seed ) {
...
+					// https://drafts.csswg.org/selectors/#relational
+					// `CSS.supports` is still expected to return `false` then:
+					// https://drafts.csswg.org/css-conditional-4/#typedef-supports-selector-fn
+					// https://drafts.csswg.org/css-conditional-4/#dfn-support-selector
+					if ( support.cssSupportsSelector &&
+
+						// eslint-disable-next-line no-undef
+						!CSS.supports( "selector(" + newSelector + ")" ) ) {
+
+						// Support: IE 11+
```
mgol, line 269: This is actually not just targeted at IE but at all browsers where CSS.supports is buggy. I should update the comment.

timmywil (approved):
LGTM!

PR #5114, opened 2022-09-10
Deferred: Deprecate setTimeout() usage under latest browsers.

diegocr:
### Summary ###

Hi there, as you may know, Chrome/ium has been lately aggressively throttling background tabs, this may result in those setTimeout()-based promises/deferreds to never resolve until the tab becomes active again.

This PR is an attempt to prevent that, by using [queueMicrotask](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask) under browsers supporting it. Also, the logic has been changed to dispatch those processes in bulk instead of individually, since the main reason for that throttling seems to be by having too many concurrent setTimeout()s running.

This implements a new `jQuery.asap()` function that will use either `queueMicrotask` or `setTimeout` for older browsers.

I've observed that around this whole codebase there are several other uses of that `window.setTimeout( func )` pattern so perhaps I should move this `jQuery.asap()` elsewhere to be used in place of those as well (as part of another PR) ?

Thanks!

### Checklist ###

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -40,6 +40,8 @@ function adoptValue( value, resolve, reject, noValue ) {
 	}
 }
 
+var asap = window.queueMicrotask || window.setTimeout;
+
 jQuery.extend( {
 
 	Deferred: function( func ) {
@@ -221,7 +223,7 @@ jQuery.extend( {
 								if ( jQuery.Deferred.getStackHook ) {
 									process.stackTrace = jQuery.Deferred.getStackHook();
 								}
-								window.setTimeout( process );
+								asap( process );
 							}
 						};
 					}
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -1522,7 +1522,37 @@ testIframe(
 	}
 );
 
-QUnit[ includesModule( "deferred" ) ? "test" : "skip" ]( "jQuery.readyException (original)", function( assert ) {
+QUnit[ includesModule( "deferred" ) && window.queueMicrotask ? "test" : "skip" ]( "jQuery.readyException (microtask)", function( assert ) {
+	assert.expect( 2 );
+
+	var done = assert.async();
+	var expected = "Error in jQuery ready #" + Math.random().toString( 36 ).slice( -7 );
+
+	this.sandbox.stub( jQuery, "readyException" ).callsFake( function( error ) {
+		assert.strictEqual( error.message, expected, "jQuery.readyException called with unexpected error" );
+		window.setTimeout( function() {
+			throw error;
+		} );
+	} );
+
+	this.sandbox.stub( window, "setTimeout" ).callsFake( function( fn ) {
+		var message;
+		try {
+			fn();
+		} catch ( error ) {
+			message = error.message;
+		}
+
+		assert.strictEqual( message, expected, "The error should have been caught" );
+		done();
+	} );
+
+	jQuery( function() {
+		throw new Error( expected );
+	} );
+} );
+
+QUnit[ includesModule( "deferred" ) && !window.queueMicrotask ? "test" : "skip" ]( "jQuery.readyException (timer)", function( assert ) {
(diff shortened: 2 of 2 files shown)


src/deferred.js
```diff
@@ -40,7 +40,23 @@ function adoptValue( value, resolve, reject, noValue ) {
...
+				var queue = asap$queue,
+				    i = 0,
+				    length = queue.length;
+				asap$queue = [];
+				for ( ; i < length; ++i ) {
+					queue[i]();
+				}
+			});
+		}
+	}
```
diegocr, line 59:
```suggestion
	},
```

mgol:
Thanks for the PR. Please fix the reported ESLint violations. Also, we don't want to expose a new method, please keep it internal. Once that's done, we should see the actual test suite running.

Also, do we need a manual queue when using `queueMicrotask`? Won't the browser manage such a queue by itself?

diegocr:
Hi, thank you, pleased you guys are positive about those kind of changes.

Well, the manual queue for queueMicrotask can be seen as subjective yes, i think it could help the browser but probably we won't prevent much overhead there. Also, i was expecting it was fine exposing a method so that for example i could then replace it with my own scheduler. If these are no-go then we can simplify this a lot, will now proceed to do so.

test/unit/core.js
```diff
@@ -1523,26 +1523,33 @@ testIframe(
...
+
+	this.sandbox.stub( jQuery, "readyException" ).callsFake( function( error ) {
+		assert.strictEqual( error.message, expected, "jQuery.readyException called with unexpected error" );
+		window.setTimeout( function() {
+			throw error;
+		} );
+	} );
 
 	this.sandbox.stub( window, "setTimeout" ).callsFake( function( fn ) {
+		var message;
```
diegocr, line 1539: Hmm, i am now realizing this test still won't work properly if the tests are then run under older browsers, perhaps we could deprecate this `"jQuery.readyException (original)"` test given looks like `"jQuery.readyException (custom)"` will ultimately do the same.

diegocr: Well, did split them in the last commit, let me know if this is ok.

Btw, locally i got the below failed tests, but also over the `main` branch:

```
Chrome Headless 105.0.5195.102 (Windows 8.1) selector pseudo - has FAILED
Chrome Headless 105.0.5195.102 (Windows 8.1) selector pseudo - misc FAILED
```

mgol:
@diegocr you need to rebase the PR to avoid the issues on `main` as they have recently been fixed.

diegocr:
Cool, done 🚀

test/unit/core.js
```diff
@@ -1522,7 +1522,37 @@ testIframe(
...
+
+	var done = assert.async();
+	var expected = "Error in jQuery ready #" + Math.random().toString( 36 ).slice( -7 );
+
+	this.sandbox.stub( jQuery, "readyException" ).callsFake( function( error ) {
+		assert.strictEqual( error.message, expected, "jQuery.readyException called with unexpected error" );
+		window.setTimeout( function() {
+			throw error;
+		} );
+	} );
```
mgol, line 1536: Why mocking `jQuery.readyException`? Can't this test be done in a similar way as the `setTimeout` one? In that case, maybe we could just keep a single test?

mgol (changes requested):
Sorry for a late review. I have one doubt about the testing strategy.

mgol:
@mvorisek thanks for the comment, that’s worth discussing.

Treating this as breaking would mean `4.0.0` could be a good release for this, but we’d still have the issue that the fallback has different behavior from the main code path and the fallback would effectively be applied in IE 11 only. I wonder if that’s fine.

timmywil:
4.0 is entering rc, so I'd like to not introduce anymore breaking changes, and I think we need to treat this as a breaking change and push to 5.0.

PR #5120, opened 2022-09-21
CSS: Return `undefined` for whitespace-only CSS variable values

mgol:
### Summary ###

The spec requires that CSS variable values are trimmed. In browsers that do this
- mainly, Safari, but also Firefox if the value only has leading whitespace
- we currently return `undefined`; in other browsers, we return an empty string as the logic to fall back to `undefined` happens before trimming.

This commit adds another explicit callback to `undefined`.

Also, more explicit comments about behaviors we need to work around in various browsers have been added.

Ref gh-5106

+3 bytes

I verified that IE 11 passes all the tests with these changes.

### Checklist ###


timmywil (approved):
LGTM. Makes it consistent.

gibson042 (approved):

mgol:
Landed in 7eb0019640a5856c42b451551eb7f995d913eba9 on `main` & in 8bea1dec18da3f3a02751dc226d51b9d0546b49e on `3.x-stable`.

Issue #4592, opened 2020-01-15
ESM distribution / `exports` in `package.json`


brettz9:
### Description ###

With your ESM work on source, will you provide an ESM distribution file as well? An ESM distribution can be handy for (ESM-based) projects that don't have a compilation step but which wish to avoid polluting HTML with their dependencies (i.e., by using `import jquery from './node_modules/jquery/dist/jquery-esm.js';` or such).

And can you point to a non-bundled entry point with ~`module`~ `exports` in `package.json` so bundlers like Rollup can discover it, so that `import jquery from 'jquery'` will work for such projects out of the box, e.g., without need for adding additional Rollup plugins or having to dig through source to find the entry point?

~There are various references about `module` as links at https://stackoverflow.com/questions/42708484/what-is-the-module-package-json-field-for~ *For native ESM that works with Node, see https://nodejs.org/api/packages.html *

mgol:
Thanks for the report. FWIW, you wouldn't ever need:
```js
import jquery from './node_modules/jquery/dist/jquery-esm.js';
```
In fact, it could be quite fragile as npm dependencies may be hoisted. This is enough if we generate the `jquery-esm.js` file:
```js
import jquery from 'jquery/dist/jquery-esm.js';
```

brettz9:
While yes, there is that fragility, it can still be useful. For example, I like to use it in repo demos and tests where it is just meant for development and can safely be used as a direct dependency. It should also be safe for applications which are known to be the root.

With such paths, there is no need for bundling as one needs when pointing to `jquery/dist/jquery-esm.js` (at least until import maps [may support](https://github.com/WICG/import-maps/issues/7) such mapping of `node_modules`).

mgol:
Ah, right, I was thinking about a bundler use case and you mean native browser import, got it. If there's no need for bundling then we could just point to `src/jquery.js` as that's our main source file. However, that would load quite a lot of files so while it's fine for development, for production you'd usually want at least partial bundling. HTTP/2 may solve this issue partially but it's probably still slower than loading a few bigger files.

brettz9:
Regarding bundling, Rollup allows "esm" as a `format` for distribution, so one could just add and inform users in the docs of the presence of such a pre-bundled `dist` file, while pointing `module` to `src/jquery.js` for the sake of bundlers (assuming you have no dependencies and your files all have extensions, as both appear to be the case).

florealcab:
[...]

PR #5122, opened 2022-09-21
WIP (example) Core: Define the `exports` & `module` fields in package.json

mgol:
### Summary ###

`module` is needed for Parcel which still doesn't support `exports`: https://github.com/parcel-bundler/parcel/issues/4155

I prepared this after reading https://nodejs.org/api/packages.html and https://webpack.js.org/guides/package-exports/. Important information is:
* jQuery can be loaded & used both in the browser and in Node.js
* jQuery is stateful so we need to make sure it's not loaded twice by accident (vide the [Dual package hazard](https://nodejs.org/api/packages.html#dual-package-hazard) section from Node.js docs)
* Bundling using popular bundlers (Webpack, Parcel, Vite, Browserify, etc.) should be possible

This draft contains files as they'd exist in a final package. Of course, the final PR would be different, files would be generated during the build and only included in published packages but I included them in the PR so that it can be experimented with.

Ref gh-4592

### Checklist ###


mgol:
An example script showing how it works in Node.js:
```js
'use strict';

// Get a `window` implementation from jsdom and assign it to the global.
import jsdom from 'jsdom';
const { JSDOM } = jsdom;
const { window } = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`);
globalThis.window = window;

// Log the jQuery slim version
const jQuerySlim = (await import('jquery/slim')).default;
console.log('jQuery slim version:', jQuerySlim?.fn?.jquery);

// Log the contents of the internal cssExpand array.
const cssExpand = (await import('jquery/src/css/var/cssExpand.js')).default;
console.log('jQuery cssExpand array:', cssExpand);
```

package.json
```diff
@@ -3,7 +3,29 @@
...
+        "import": "./dist/jquery-node-wrapper.slim.mjs",
+        "default": "./dist/jquery.slim.js"
+      },
+      "browser": "./dist/jquery.slim.mjs"
+    },
+    "./src/*.js": "./src/*.js",
+    "./amd/*.js": "./amd/*.js"
+  },
   "main": "dist/jquery.js",
+  "module": "dist/jquery.mjs",
```
mgol, line 28: Parcel doesn't support `package.json` `exports` so this field is still needed:
https://github.com/parcel-bundler/parcel/issues/4155

package.json
```diff
@@ -3,7 +3,29 @@
...
+    "./slim": {
+      "node": {
+        "module": "./dist/jquery.slim.mjs",
+        "import": "./dist/jquery-node-wrapper.slim.mjs",
+        "default": "./dist/jquery.slim.js"
+      },
+      "browser": "./dist/jquery.slim.mjs"
+    },
+    "./src/*.js": "./src/*.js",
+    "./amd/*.js": "./amd/*.js"
```
mgol, line 25: These entries are tricky. Technically, we don't guarantee anything about files in those directories, they can come and go even in patch releases. On the other hand, we've been telling people that for advanced cases they can use our modules, taking into account these restrictions.

`src` contains jQuery source written in ES modules, `amd` those modules transpiled to AMD as jQuery 3.x & older used AMD as the source format and we wanted to still make it available to people.

The modules in `src` are compatible with direct usage in a browser without any bundler as well:
```html
<script type="module">
	import jQuery from "./src/jquery.js";
    /* do someting with jQuery */
</script>
```

package.json
```diff
@@ -3,7 +3,29 @@
...
+      "browser": "./dist/jquery.mjs"
+    },
+    "./slim": {
+      "node": {
+        "module": "./dist/jquery.slim.mjs",
+        "import": "./dist/jquery-node-wrapper.slim.mjs",
+        "default": "./dist/jquery.slim.js"
+      },
+      "browser": "./dist/jquery.slim.mjs"
+    },
```
mgol, line 23: This part mostly corresponds with a recommendation from Webpack from:
https://webpack.js.org/guides/package-exports/#providing-commonjs-and-esm-version-stateful

mgol:
To run the above Node.js example, save it as `test.js` in an npm project, install `jsdom` as a dependency and link `node_modules/jquery` to a clone of this branch.

mgol:
Oh, and, of course, run `node test.js` to actually run the example.

mgol:
@guybedford @jkrems @mylesborins could you have a look?

package.json
```diff
@@ -3,7 +3,29 @@
...
   "version": "4.0.0-pre",
+  "type": "commonjs",
+  "exports": {
+    ".": {
+      "node": {
+        "module": "./dist/jquery.mjs",
+        "import": "./dist/jquery-node-wrapper.mjs",
+        "default": "./dist/jquery.js"
+      },
+      "browser": "./dist/jquery.mjs"
```
aduh95, line 14: You probably want to add the `require` and `default` in case, that should cover most cases. I'm not sure if `"browser"` is needed if you already have `default`, but also can't hurt.
```suggestion
      "require": "./dist/jquery.js",
      "browser": "./dist/jquery.mjs",
      "default": "./dist/jquery.mjs"
```

package.json
```diff
@@ -3,7 +3,29 @@
...
+    },
+    "./slim": {
+      "node": {
+        "module": "./dist/jquery.slim.mjs",
+        "import": "./dist/jquery-node-wrapper.slim.mjs",
+        "default": "./dist/jquery.slim.js"
+      },
+      "browser": "./dist/jquery.slim.mjs"
```
aduh95, line 22: Same here.
```suggestion
      "require": "./dist/jquery.slim.js",
      "browser": "./dist/jquery.slim.mjs",
      "default": "./dist/jquery.slim.mjs"
```

package.json
GeoffreyBooth: I would also recommend flipping around the file extensions. Use `.cjs` for the CommonJS version and `.js` for all other versions, and add `"type": "module"` to the `package.json`. I think it’s still the case that many servers/CDNs don’t serve the proper MIME type for `.mjs` files, so the browser version of jQuery (even the browser ESM version) should absolutely have a `.js` file extension for maximum compatibility.

aduh95: > I think it’s still the case that many servers/CDNs don’t serve the proper MIME type for `.mjs` files

It's been years since I've since this issue, not sure if it's worth it. Note that the `.js` files are not CJS but AMD FWIW.

GeoffreyBooth: > It’s been years since I’ve since this issue, not sure if it’s worth it.

I think jQuery’s enormous scope justifies being careful. There’s no _benefit_ to `.mjs`/`.js` over `.cjs`/`.js`. Node 10 doesn’t support `"exports"` at all; if you want to explicitly provide a Node 10-and-below ESM build, an `.mjs` file somewhere just for those consumers could be included. I would still save all the others as `.js` files as `.js` works everywhere.

mgol: @GeoffreyBooth I'd like to understand where every entry comes from. Why do we need `require`, `browser` & `default` for non-Node.js cases?

aduh95: > Why do we need `require` […] for non-Node.js cases?

E.g. if someone uses `require()` to write browser focused code. If you set `browser` to point to the ESM file, depending on the builder it may or may not work (i.e. it would work only if the bundler supports requiring ES modules).

ollie-iterators:
Shouldn't this be put in the  4.0.0 milestone?

PR #5137, opened 2022-10-07
Selector: Drop support for legacy pseudos, test custom pseudos

mgol:
### Summary ###

This backports custom pseudos tests from Sizzle; they were missed in original test backports. Also, the support for legacy custom pseudos has been dropped.

### Checklist ###

### DIFF
--- a/test/unit/selector.js
+++ b/test/unit/selector.js
@@ -2142,3 +2142,59 @@ QUnit.test( "jQuery.escapeSelector", function( assert ) {
 	assert.equal( jQuery.escapeSelector( "\uDF06" ), "\uDF06", "Doesn't escape lone low surrogate" );
 	assert.equal( jQuery.escapeSelector( "\uD834" ), "\uD834", "Doesn't escape lone high surrogate" );
 } );
+
+QUnit.test( "custom pseudos", function( assert ) {
+	assert.expect( 6 );
+
+	jQuery.expr.filters.foundation = jQuery.expr.filters.root;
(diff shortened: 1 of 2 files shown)


test/unit/selector.js
```diff
@@ -2142,3 +2142,59 @@ QUnit.test( "jQuery.escapeSelector", function( assert ) {
 	assert.equal( jQuery.escapeSelector( "\uDF06" ), "\uDF06", "Doesn't escape lone low surrogate" );
 	assert.equal( jQuery.escapeSelector( "\uD834" ), "\uD834", "Doesn't escape lone high surrogate" );
 } );
+
+QUnit.test( "custom pseudos", function( assert ) {
+	assert.expect( 6 );
+
+	jQuery.expr.filters.foundation = jQuery.expr.filters.root;
+	assert.deepEqual( jQuery.find( ":foundation" ), [ document.documentElement ], "Copy element filter with new name" );
+	delete jQuery.expr.filters.foundation;
```
gibson042, line 2151: For good hygiene, I think we should wrap this in a `try` block move the cleanup into `finally` (or anything else that accomplishes the same isolation).

gibson042 (approved):

test/unit/selector.js
mgol: @gibson042 I used `try-finally` for each part. Does that look OK now?

gibson042 (approved):

test/unit/selector.js
gibson042: 👍

Issue #4989, opened 2021-12-25
Null value in jQuery.get doesn't allow use of dataType


jcubic:
### Description ###
According to [jQuery.get documentation](https://api.jquery.com/jQuery.get/) you can use `null` as success function to be able to use `dataType`. But it seems that it's not working. The only way to get text from JSON response is to use `$.noop`. So this need to be fixed or documentation needs to be updated.

I use jQuery 3.6.0.

### Link to test case ###

```javascript
await $.get("https://httpbin.org/json", null, "text")
await $.get("https://httpbin.org/json", $.noop, "text")
```

https://jsbin.com/safuliduje/edit?html,css,js,output

mgol:
Thanks for the report. Can you provide a test case on [JS Bin](https://jsbin.com/qawicop/edit?html,css,js,output) as the issue template requested? Thanks!

jcubic:
@mgol Added link.

mgol:
OK, it looks like the current implementation allows you to provide `null` as a placeholder for the callback but doesn't allow you to skip the `data` parameter then. This means the following:
```js
await $.get("https://httpbin.org/json", null, "text");
```
doesn't work as expected but the following:
```js
await $.get("https://httpbin.org/json", null, null, "text");
```
does. With hindsight, all of these short forms of `jQuery.ajax` with their positional parameters were a mistake as there's always a lot of parameter juggling. Perhaps we should update the doc that if you provide `null` for the callback then you also need to provide the `data` parameter, even if `null` or `undefined`.

We could also update the implementation to handle this case as well but that wouldn't fix all those past versions anyway so I'm not sure if it's worth it; we'd need to update the docs as well anyway.

Thoughts, @jquery/core?

goarblt:
hello
https://www.evil.com

goarblt:
hello
https://www.evil.com

gibson042:
Even though parameter hockey isn't great and we don't want to encourage the pattern, I think it's worth fixing code to match documentation. The interesting cases are those with parameter gaps, and since `dataType` is documented to require `success` (excluding `jQuery.get(url, data, dataType)` and `jQuery.get(url, dataType)`), those are limited to missing `data` (i.e., `jQuery.get(url, fnOrNull, strOrUndef)`).

[...]

PR #5139, opened 2022-10-11
Ajax: Support `null` as success functions in `jQuery.get`

mgol:
### Summary ###

According to the docs, one can use `null` as a success function in `jQuery.get`
of `jQuery.post` so the following:

```js
await jQuery.get( "https://httpbin.org/json", null, "text" )
```

should get the text result. However, this shortcut hasn't been working so far.

Fixes gh-4989

### Checklist ###

* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com - API PR: https://github.com/jquery/api.jquery.com/pull/1208

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -846,8 +846,9 @@ jQuery.extend( {
 jQuery.each( [ "get", "post" ], function( _i, method ) {
 	jQuery[ method ] = function( url, data, callback, type ) {
 
-		// Shift arguments if data argument was omitted
-		if ( typeof data === "function" ) {
+		// Shift arguments if data argument was omitted.
+		// Handle the null callback placeholder.
+		if ( typeof data === "function" || data === null ) {
 			type = type || callback;
 			callback = data;
 			data = undefined;
--- a/src/ajax/xhr.js
+++ b/src/ajax/xhr.js
@@ -15,7 +15,6 @@ var xhrSuccessStatus = {
 jQuery.ajaxTransport( function( options ) {
 	var callback;
 
-	// Cross domain only allowed if supported through XMLHttpRequest
 	return {
 		send: function( headers, complete ) {
 			var i,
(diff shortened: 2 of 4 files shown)


mgol:
I added this to the discussion list to discuss whether we want this in `3.x` or not. This has a chance of breaking some code; on the other hand, documentation states the behavior from this PR.

src/ajax.js
```diff
@@ -846,8 +846,9 @@ jQuery.extend( {
 jQuery.each( [ "get", "post" ], function( _i, method ) {
 	jQuery[ method ] = function( url, data, callback, type ) {
 
-		// Shift arguments if data argument was omitted
-		if ( typeof data === "function" ) {
+		// Shift arguments if data argument was omitted.
+		// Handle the null callback placeholder.
+		if ( typeof data === "function" || data === null ) {
```
mgol, line 851: @gibson042 allowing `undefined` here would break the 4-param signature with the second parameter (`data`) being `undefined` which is e.g. used in the `$.getScript` definition. We could add more checks to make it work but I thought in that case it may be better to just allow `null`. `null` is not an allowed `data` value so we don’t have a conflict.

src/ajax/xhr.js
```diff
@@ -15,7 +15,6 @@ var xhrSuccessStatus = {
 jQuery.ajaxTransport( function( options ) {
 	var callback;
 
-	// Cross domain only allowed if supported through XMLHttpRequest
```
mgol, line 18: This is unrelated but this comment has long been invalid… We only have support checks on `3.x-stable` here.

src/ajax.js
gibson042: I concur.

gibson042 (approved):

timmywil:
Since this has been like this for so long, we're going to leave 3.x unchanged and fix this for 4.0.

mgol:
I submitted the API PR at https://github.com/jquery/api.jquery.com/pull/1208.

PR #5148, opened 2022-10-24
Docs: remove stale badge

timmywil:
### Summary ###
This removes the stale badge on the README from Fossa. It may be a temporary loss of service, but I'd prefer not to have badges that don't load.

### Checklist ###


mgol (approved):
Something like that would be better as a CI check than a badge that's easy to overlook anyway. LGTM.

Issue #5147, opened 2022-10-22
Safari 16 crashes on in-template DOM manipulation


jded76:
### Description ###
After the new IOS 16, Safari changed the behavior of  `document.adoptNode` when called on a template. It's not changing the ownerDocument any more. It leaves this on an empty document (about:blank).

This causing an error at the function contains at this [line](https://github.com/jquery/jquery/blob/d0ce00cdfa680f1f0c38460bc51ea14079ae8b07/src/selector/contains.js#L5), because the documentElement property is null on an empty document (about:blank).

If you change this [line](https://github.com/jquery/jquery/blob/d0ce00cdfa680f1f0c38460bc51ea14079ae8b07/src/selector/contains.js#L5) from
`var adown = a.nodeType === 9 ? a.documentElement : a, bup = b && b.parentNode;`
to
`var adown = a.nodeType === 9 && a.documentElement ? a.documentElement : a, bup = b && b.parentNode;`
it's working again.

More details can be found [here ](https://discourse.aurelia.io/t/any-issues-with-safari-16-0/4923/9) and aurelia/framework#1003.

### Link to test case ###
You can see the problem [in this fiddler](https://jsfiddle.net/oxn8t2uh/) if you try it with Safari 16+.
If you press "add" and then "remove" you get the error.

jded76:
I made a [repository](https://github.com/jded76/Safari16-vanilla-js) if you like to test it without fiddler.

If you change this [line ](https://github.com/jded76/Safari16-vanilla-js/blob/299a5779791a32a7ed226b3423ac42ec2ac71e98/index.html#L7) to use the local jquery file `<script src="jquery-3.5.1.js"></script>` the error is gone.

mgol:
Reopening until the fix lands in Sizzle and - then - on `3.x-stable`.

mgol:
Landed on `main` in 3299236c898136dc1aa57dc5148811203e931895.

The `3.x-stable` part of the fix needs to first happen in Sizzle.

Sizzle PR: https://github.com/jquery/sizzle/pull/490
`3.x-stable` PR (will fail until the Sizzle upgrade is incorporated into it): #5159

mgol:
[...]

PR #5158, opened 2022-11-14
Selector:Manipulation: Fix DOM manip within template contents

mgol:
### Summary ###

The `<template/>` element `contents` property is a document fragment that may
have a `null` `documentElement`. In Safari 16 this happens in more cases due
to recent spec changes - in particular, even if that document fragment is
explicitly adopted into an outer document. We're testing both of those cases
now.

The crash used to happen in `jQuery.contains`. As it turns out, we don't need
to query the supposed container `documentElement` if it has the
`Node.DOCUMENT_NODE` (9) `nodeType`; we can call `.contains()` directly on
the `document`. That avoids the crash.

Fixes gh-5147

-18 bytes

### Checklist ###

### DIFF
--- a/src/selector/contains.js
+++ b/src/selector/contains.js
@@ -2,15 +2,14 @@ import jQuery from "../core.js";
 
 // Note: an element does not contain itself
 jQuery.contains = function( a, b ) {
-	var adown = a.nodeType === 9 ? a.documentElement : a,
-		bup = b && b.parentNode;
+	var bup = b && b.parentNode;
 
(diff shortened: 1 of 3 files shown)


gibson042 (approved):

mgol:
-18 bytes

timmywil (approved):

mgol:
Landed on `main` in 3299236c898136dc1aa57dc5148811203e931895. The `3.x-stable` part of the fix needs to first happen in Sizzle.

PR #5159, opened 2022-11-15
Selector:Manipulation: Fix DOM manip within template contents

mgol:
### Summary ###

The `<template/>` element `contents` property is a document fragment that may have a `null` `documentElement`. In Safari 16 this happens in more cases due to recent spec changes - in particular, even if that document fragment is explicitly adopted into an outer document. We're testing both of those cases now.

The crash used to happen in `jQuery.contains` which is an alias for `Sizzle.contains` in jQuery 3.x.

The Sizzle fix is at jquery/sizzle#490. This PR pulls the new Sizzle version into jQuery's `3.x-stable` branch and backports one test from gh-5158.

Fixes gh-5147
Ref jquery/sizzle/pull/490
Ref gh-5158

### Checklist ###

* [ ] Update Sizzle to a version including PR: https://github.com/jquery/sizzle/pull/490


mgol:
@timmywil I also had to apply a fix to `selector-native` to pass this new test. I opted to skip `documentElement` there, similarly to the fix on jQuery master since `documentElement` only mattered in IE <9.

PR #5167, opened 2022-11-24
Core:Selector: Move jQuery.contains from the selector to the core module

mgol:
### Summary ###

The `jQuery.contains` method is quite simple in jQuery 4+. On the other side, it's a dependency of the core `isAttached` util which is not ideal; moving it from the `selector` the `core` module resolves the issue.

-1 byte

### Checklist ###


mgol:
Actually, it's -3 bytes, not +3 bytes

gibson042 (approved):

timmywil (approved):

mgol:
It's now at -1 byte, not -3 bytes.

Issue #5177, opened 2022-12-16
unsupported pseudo: valid with selector list


antti-manninen-vmv:
**Description**

After upgrading to jQuery 3.6.2 pseudo selector `:valid` does not work in selector list with Firefox and Safari. Everything works with Chrome. 

This worked in 3.6.1

```
jQuery('body').find('div:valid, span')
```

In  3.6.2 it gives following error:

```
Uncaught Error: Syntax error, unrecognized expression: unsupported pseudo: valid
```

This still  works with old and new versions with Firefox and Safari:

```
jQuery('body').find('div:valid')
```

mgol:
Thanks for the report!

jQuery has a mechanism of - simplifying a bit - trying a provided selector against `querySelectorAll` in a `try-catch` and, if it fails, falling back to its own traversal mechanism inherited from [Sizzle](https://github.com/jquery/sizzle).

In jQuery 3.6.2 we started using `CSS.supports( "selector(SELECTOR)" )` (for a proper `SELECTOR`) before using `querySelectorAll` on the selector. This was to solve #5098 - some selectors, like `:is()`, now have their parameters parsed in a forgiving way, meaning that `:is(:fakepseudo)` no longer throws but just returns 0 results, breaking that jQuery mechanism.

[A recent spec change](https://github.com/w3c/csswg-drafts/issues/7280) made `CSS.supports( "selector(SELECTOR)" )` always use non-forgiving parsing, allowing us to use this API for what we've used `try-catch` before. This change has only arrived in Firefox stable so far, though.

Now, **apparently, our implementation has a bug** - in `CSS.supports( "selector(SELECTOR)" )`, [`SELECTOR`](https://w3c.github.io/csswg-drafts/css-conditional-4/#at-supports-ext)  needs to be a [`<complex-selector>`](https://w3c.github.io/csswg-drafts/selectors-4/#typedef-complex-selector) and not a [`<complex-selector-list>`](https://w3c.github.io/csswg-drafts/selectors-4/#typedef-complex-selector-list). Which means:
```js
CSS.supports("selector(div:valid, span)"); // false
CSS.supports("selector(div:valid)"); // true
CSS.supports("selector(span)"); // true
```
and explains your observations. See the live test case: https://jsbin.com/vesureh/1/edit?html,js,console

[...]

PR #5178, opened 2022-12-17
Selector: Make selector lists work with `qSA` again

mgol:
### Summary ###

jQuery 3.6.2 started using `CSS.supports( "selector(SELECTOR)" )` before using `querySelectorAll` on the selector. This was to solve gh-5098 - some selectors, like `:has()`, now have their parameters parsed in a forgiving way, meaning that `:has(:fakepseudo)` no longer throws but just returns 0 results, breaking that jQuery mechanism.

A recent spec change]() made `CSS.supports( "selector(SELECTOR)" )` always use non-forgiving parsing, allowing us to use this API for what we've used `try-catch` before.

To solve the issue on the spec side for older jQuery versions, `:has()` parameters are no longer using forgiving parsing in the latest spec update but our new mechanism is more future-proof anyway.

However, the jQuery implementation has a bug - in
`CSS.supports( "selector(SELECTOR)" )`, `SELECTOR` needs to be a `<complex-selector>` and not a `<complex-selector-list>`. Which means that selector lists now skip `qSA` and go to the jQuery custom traversal:
```js
CSS.supports("selector(div:valid, span)"); // false
CSS.supports("selector(div:valid)"); // true
CSS.supports("selector(span)"); // true
```

To solve this, this commit wraps the selector list passed to `CSS.supports( "selector(:is(SELECTOR))" )` with `:is`, making it a single selector again.

See:
* https://w3c.github.io/csswg-drafts/css-conditional-4/#at-supports-ext
* https://w3c.github.io/csswg-drafts/selectors-4/#typedef-complex-selector
* https://w3c.github.io/csswg-drafts/selectors-4/#typedef-complex-selector-list

Fixes gh-5177
Ref w3c/csswg-drafts#7280

+2 bytes

### Checklist ###

### DIFF
--- a/src/selector/rbuggyQSA.js
+++ b/src/selector/rbuggyQSA.js
@@ -27,9 +27,8 @@ if ( !support.cssSupportsSelector ) {
 	// `:has()` uses a forgiving selector list as an argument so our regular
 	// `try-catch` mechanism fails to catch `:has()` with arguments not supported
 	// natively like `:has(:contains("Foo"))`. Where supported & spec-compliant,
-	// we now use `CSS.supports("selector(:is(SELECTOR_TO_BE_TESTED))")` but,
-	// outside that, let's mark `:has` as buggy to always use jQuery traversal
-	// for `:has()`.
+		// we now use `CSS.supports("selector(:is(SELECTOR_TO_BE_TESTED))")`, but
+		// outside that we mark `:has` as buggy.
 	rbuggyQSA.push( ":has" );
 }
 
--- a/src/selector/rbuggyQSA.js
+++ b/src/selector/rbuggyQSA.js
@@ -27,9 +27,9 @@ if ( !support.cssSupportsSelector ) {
 	// `:has()` uses a forgiving selector list as an argument so our regular
 	// `try-catch` mechanism fails to catch `:has()` with arguments not supported
 	// natively like `:has(:contains("Foo"))`. Where supported & spec-compliant,
-	// we now use `CSS.supports("selector(SELECTOR_TO_BE_TESTED)")` but outside
-	// that, let's mark `:has` as buggy to always use jQuery traversal for
-	// `:has()`.
+	// we now use `CSS.supports("selector(:is(SELECTOR_TO_BE_TESTED))")` but,
+	// outside that, let's mark `:has` as buggy to always use jQuery traversal
+	// for `:has()`.
 	rbuggyQSA.push( ":has" );
 }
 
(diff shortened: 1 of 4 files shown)


mgol:
+2 bytes

timmywil (approved):

src/selector/rbuggyQSA.js
```diff
@@ -27,9 +27,9 @@ if ( !support.cssSupportsSelector ) {
 	// `:has()` uses a forgiving selector list as an argument so our regular
 	// `try-catch` mechanism fails to catch `:has()` with arguments not supported
 	// natively like `:has(:contains("Foo"))`. Where supported & spec-compliant,
-	// we now use `CSS.supports("selector(SELECTOR_TO_BE_TESTED)")` but outside
-	// that, let's mark `:has` as buggy to always use jQuery traversal for
-	// `:has()`.
+	// we now use `CSS.supports("selector(:is(SELECTOR_TO_BE_TESTED))")` but,
+	// outside that, let's mark `:has` as buggy to always use jQuery traversal
+	// for `:has()`.
```
mgol, line 32:
```suggestion
		// we now use `CSS.supports("selector(:is(SELECTOR_TO_BE_TESTED))")`, but
		// outside that we mark `:has` as buggy.
```

src/selector/rbuggyQSA.js
```diff
@@ -27,9 +27,8 @@ if ( !support.cssSupportsSelector ) {
 	// `:has()` uses a forgiving selector list as an argument so our regular
 	// `try-catch` mechanism fails to catch `:has()` with arguments not supported
 	// natively like `:has(:contains("Foo"))`. Where supported & spec-compliant,
-	// we now use `CSS.supports("selector(SELECTOR_TO_BE_TESTED)")` but outside
-	// that, let's mark `:has` as buggy to always use jQuery traversal for
-	// `:has()`.
+		// we now use `CSS.supports("selector(:is(SELECTOR_TO_BE_TESTED))")`, but
+		// outside that we mark `:has` as buggy.
```
mgol, line 31:
```suggestion
	// we now use `CSS.supports("selector(:is(SELECTOR_TO_BE_TESTED))")`, but
	// outside that we mark `:has` as buggy.
```

mgol:
Landed on `main` (future jQuery `4.0.0`) in 09d988b774e7ff4acfb69c0cde2dab373559aaca and on `3.x-stable` (future jQuery `3.7.0`) in 848de625425c6b08ec9d8ad9a4bcab7e913c2477.

Issue #5185, opened 2022-12-22
Selector: Fix scoping logic of selector-native


mgol:
### Description ###

As discussed in https://github.com/jquery/jquery/pull/5085#discussion_r991562552, it would be worthwhile to try to fix the long-standing `selector-native` issue of not requiring all selector parts to match element under context, e.g. so that:
```js
$div.find("div > *")
```
no longer matches children of `$div`.

My preliminary research indicates resolving this issue would cost about 1000 bytes minified gzipped; it'd still be 3300 bytes minified gzipped smaller than the full build compared to 4300 bytes minified gzipped without this change.

### Link to test case ###

No test case since we don't upload selector-native versions anywhere. But the example from the description is pretty straightforward.

mgol:
PR: #5186

PR #5186, opened 2022-12-22
Selector: Backport jQuery selection context logic to selector-native

mgol:
### Summary ###

Backport jQuery selection context logic to `selector-native`.

This makes:
```js
$div.find("div > *")
```
no longer matching children of `$div`.

Also, leading combinators now work, e.g.:
```js
$div.find( "> *" );
```
returns children of `$div`.

As a result of that, a number of tests are no longer skipped in the `selector-native` mode.

Fixes gh-5185
Ref gh-5085

The full build is at `+19 bytes` now. I'm sure this can be reduced a bit but some may be unavoidable unless we want to duplicate a lot of code between `selector` & `selector-native` which I'd like to avoid.

The `selector-native` build is at `-3274 bytes` compared to the full `main` (c66d4700dcf98efccb04061d575e242d28741223) build and at `+1056 bytes` compared to the `selector-native` build on that same `main` build.

It's quite likely more size optimizations are possible in the `selector-native` build, I'd appreciate some ideas. One thing I know about is the `tokenize` & `toSelector` functions could be simplified slightly as it's enough to collect string parts without all the other context information - instead of:
```js
tokens.push( {
	value: matched,
	type: type,
	matches: match
} );
```
it'd be enough to:
```js
tokens.push( matched )
```
and `toSelector` would be simplified from:
```js
function toSelector( tokens ) {
	var i = 0,
		len = tokens.length,
		selector = "";
	for ( ; i < len; i++ ) {
		selector += tokens[ i ].value;
	}
	return selector;
}
```
to:
```js
function toSelector( tokens ) {
	return tokens.join( "" );
}
```

That would save `37 bytes` in the `selector-native` build but would require duplicating those two functions and especially the `tokenize` one is quite large so I opted against that for now.

### Checklist ###


Pointy:
A recent [Stack Overflow question](https://stackoverflow.com/questions/75038754/jquery-find-use-with-not-selector-is-missing-some-elements-the-selector/75040800?noredirect=1#) is relevant, in that it's complaining about the fact that a `.find()` with a selector like `:not(body *)` will find nothing. That is, the selector inside the `:not` thing is evaluated without respecting the "starting point" of the `.find()`.  Whether that's good or bad for jQuery I can't say.

src/selector-native.js
```diff
@@ -62,18 +74,59 @@ jQuery.extend( {
...
+
+			// qSA considers elements outside a scoping root when evaluating child or
+			// descendant combinators, which is not what we want.
+			// In such cases, we work around the behavior by prefixing every selector in the
+			// list with an ID selector referencing the scope context.
+			// The technique has to be used as well when a leading combinator is used
+			// as such selectors are not recognized by querySelectorAll.
+			// Thanks to Andrew Dupont for this technique.
+			if ( nodeType === 1 &&
+				( rdescend.test( selector ) || rcombinators.test( selector ) ) ) {
```
gibson042, line 89: `rcombinators` is a much less clear name than e.g. `rleadingCombinator`, but I'm weighing the costs of gratuitous divergence from Sizzle. I think it would be best to rename in both.

src/selector-native.js
```diff
@@ -62,18 +74,59 @@ jQuery.extend( {
...
+
+				// Outside of IE, if we're not changing the context we can
+				// use :scope instead of an ID.
+				if ( newContext !== context || isIE ) {
+
+					// Capture the context ID, setting it first if necessary
+					if ( ( nid = context.getAttribute( "id" ) ) ) {
+						nid = jQuery.escapeSelector( nid );
+					} else {
+						context.setAttribute( "id", ( nid = jQuery.expando ) );
```
gibson042, line 104: Where is this id attribute removed?

gibson042 (approved):
I'm pretty happy with this, modulo some nits. Thanks @mgol!

src/selector-native.js
mgol: Makes sense to me.

src/selector-native.js
mgol: Nice catch. I removed the `try-catch` since we don't do custom traversal here but I should have instead changed it into `try-finally` to keep this `finally` block:
https://github.com/jquery/jquery/blob/b02a257f98688aa890e06a85672cd1a54c3ffa3a/src/selector.js#L288-L290

src/selector-native.js
mgol: ...with Sizzle going away soon, though, perhaps it'd be fine to just rename it on `main` & `3.x-stable`?

src/selector-native.js
mgol: @gibson042 The `finally` block added.

src/selector-native.js
mgol: Renamed here; I can submit a separate PR just renaming this on `3.x-stable`.

mgol: Oh, well, this is a simple change so I submitted a PR to Sizzle as well:
* jQuery `3.x-stable`: #5208
* Sizzle `main`: https://github.com/jquery/sizzle/pull/494

timmywil (approved):

PR #5190, opened 2023-01-03
Build: Run GitHub Action browser tests on Playwright WebKit

mgol:
### Summary ###

So far, we've been running browser tests on GitHub Actions in Chrome and Firefox. Regular Safari is not available in GitHub Actions but Playwright WebKit comes close to a dev version of Safari.

With this change, our GitHub CI (& local test runs) will invoke tests on all actively developed browser engines on all PRs.

Also, switch our GitHub Actions browser tests to use Node.js 18.

In addition, we now allow to pass custom browsers when one needs it; e.g., to run the tests in all three engines on Linux/macOS, run:
```
grunt && BROWSERS=ChromeHeadless,FirefoxHeadless,WebkitHeadless grunt karma:main
```

I'd like to prepare a version for `3.x` as well but that will be a bit more complex since the common `AppleWebKit/605.1.15` token identifying Safari is only frozen in this version since Safari 13.

### Checklist ###


.github/workflows/node.js.yml
```diff
@@ -71,6 +71,10 @@ jobs:
       run: |
         npm install
 
+    - name: Install Playwright dependencies
+      run: npx playwright install-deps
```
mgol, line 75: I think I should change that to `playwright-webkit` so that it uses a local version.

mgol:
The `karma-webkit-launcher` issues relevant for us, https://github.com/google/karma-webkit-launcher/issues/1 & https://github.com/google/karma-webkit-launcher/issues/3, have been fixed upstream so I re-enabled those tests even locally. 🎉

mgol:
I also submitted https://github.com/google/karma-webkit-launcher/issues/2 to be able to modify the user agent of the Playwright Webkit browser. That would help us with detecting the browser as we could attach some Playwright-identifying token.

mgol:
@dmethvin could you check if this works fine on Windows?

timmywil (approved):

mgol:
@dmethvin I went ahead & merged it; we can sort out any potential Windows issues separately later.

mgol:
Landed on `main` in b02a257f98688aa890e06a85672cd1a54c3ffa3a and on `3.x-stable` (with slight modifications) in da7057e9b0d48afdf5666597279f0a24e9828bc8.

El-o-heka (approved):
Never surrender

Issue #5142, opened 2022-10-18
jQuery.ajax doesn’t send headers when ‘dataType‘ is ‘script’ and ‘crossDomain’ is true


Rinzwind:
In the following example, while the same `headers` setting is used for all four requests, the header is only sent in the first three requests, it is not sent in the fourth request:

```javascript
const page = 
`<html>
<head>
	<script src='https://code.jquery.com/jquery-3.6.1.min.js'></script>
</head>
<body>
	<script>
		var headers = { 'X-Test-Header': 'Test-Header-Value' };
		$.ajax('Request_1', { dataType: 'html', headers: headers });
		$.ajax('Request_2', { dataType: 'html', crossDomain: true, headers: headers });
		$.ajax('Request_3', { dataType: 'script', headers: headers });
		$.ajax('Request_4', { dataType: 'script', crossDomain: true, headers: headers });
	</script>
</body>
</html>`;
const express = require('express')();
express.get('/', (request, response) => { response.send(page); });
express.get(/^\/Request_[1-4]$/, (request, response) => {
	console.log(`${request.path}: ${request.get('X-Test-Header')}`);
	response.send(''); });
express.listen(9000);
```

The server logs:

```
/Request_1: Test-Header-Value
/Request_2: Test-Header-Value
/Request_3: Test-Header-Value
/Request_4: undefined
```

Shouldn’t the header also be sent in the fourth request?

mgol:
Thanks for the report. The `script` transport has two modes - using a script tag with `src` or fetching the contents and inserting an inline script (via `jQuery.globalEval`). The second one requires CORS so the first one is preferred when cross domain. Here are the current conditions:
https://github.com/jquery/jquery/blob/3.6.1/src/ajax/script.js#L47
The code on `main` (future jQuery 4.0) is a bit more involved:
https://github.com/jquery/jquery/blob/main/src/ajax/script.js#L6-L20
We should investigate adding `headers` to the list.

I wonder if this would be a breaking change for 3.x? We'll discuss this at a team's meeting.

mgol:
PR: #5193

mgol:
For jQuery 4.0, this will be fixed via #5193. For jQuery 3.x, you can workaround by setting `crossDomain` to `false`, even if the request is cross-domain.

idiamant:
[...]

PR #5193, opened 2023-01-12
Ajax: Support `headers` for script transport even when cross-domain

mgol:
### Summary ###

The AJAX script transport has two versions: XHR + `jQuery.globalEval` or appending a script tag (note that `jQuery.globalEval` also appends a script tag now, but inline). The former cannot support the `headers` option which has so far not been taken into account.

For jQuery 3.x, the main consequence was the option not being respected for cross-domain requests. Since in 4.x we use the latter way more often, the option was being ignored in more cases.

The transport now checks whether the `headers` option is specified and uses the XHR way unless `scriptAttrs` are specified as well.

Fixes gh-5142

+5 bytes

### Checklist ###

### DIFF
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -6,20 +6,28 @@ import "../ajax.js";
 function canUseScriptTag( s ) {
 
 	// A script tag can only be used for async, cross domain or forced-by-attrs requests.
+	// Requests with headers cannot use a script tag. However, when both `scriptAttrs` &
+	// `headers` options are specified, both are impossible to satisfy together; we
+	// prefer `scriptAttrs` then.
 	// Sync requests remain handled differently to preserve strict script ordering.
-	return s.crossDomain || s.scriptAttrs ||
+	return s.scriptAttrs || (
+		!s.headers &&
+		(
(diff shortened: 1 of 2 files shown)


src/ajax/script.js
```diff
@@ -6,20 +6,28 @@ import "../ajax.js";
...
+			//   * have `scriptAttrs` set as that's a script-only functionality
+			// Note that this means JSONP requests violate strict CSP script-src settings.
+			// A proper solution is to migrate from using JSONP to a CORS setup.
+			( s.async && jQuery.inArray( "json", s.dataTypes ) < 0 )
+		)
+	);
 }
 
-// Install script dataType. Don't specify `content.script` so that an explicit
+// Install script dataType. Don't specify `contents.script` so that an explicit
```
mgol, line 30: This is not related but since I was already modifying the nearby code... The comment was added in 025da4dd343e6734f3d3c1b4785b1548498115d8 and had the typo from the start.

timmywil (approved):
Agreed, just go the XHR route in this case

mgol:
+5 bytes

PR #5196, opened 2023-01-16
Build: Migrate middleware-mockserver to modern JS

mgol:
### Summary ###

The `test/middleware-mockserver.js` file used to have the same ESLint settings applied as other test files that are directly run in tested browsers. Now it shares settings of other Node.js files.

The file is now also written using modern JS, leveraging ES2018.

### Checklist ###

* ~~New tests have been added to show the fix or feature works~~
* [ ] Grunt build and unit tests pass locally with these changes
* ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~

Issue #4150, opened 2018-08-06
Make ajax understand other acceptable data 


jimmywarting:
How to send a file or formdata with jQuery is brought up a lot at StackOverflow

I think you should do things better by trying to understand what kind of data they are sending without having to do `processData = false` and `contentType = false`

that is why i propose this should be possible

```js
jQuery.ajax({
  url: url, 
  method: 'POST',
  data: FormData || typed arrays || Blob || File || UrlSearchParams

  // without the need of this:
  // processData: false,
  // contentType: false
});

// and also 
jQuery.post(url, FormData || typed arrays || Blob || File || UrlSearchParams)
```

What i usually do is propose using the fetch api instead of jquery's ajax, this is something that you don't make any easier for the developers, it's quite the opposite. It makes it harder and more confusing to why things don't work

timmywil:
Thank you for the issue. However, jQuery automatically processes data to query strings that can be sent in GET requests. This is likely still the most common use case. Besides, changing the defaults of these options would break a lot of code.

jimmywarting:
Sorry, i fail to see how something like that could break when it don't even work in the first place.

I didn't mean you should change the default settings of processData & contentType to false. I only meant that you should check whatever if data is a `instanceof FormData` and if it's do the correct thing.

timmywil:
The default would have to be changed if you don't want to set `contentType` at all since the default is `application/x-www-form-urlencoded`. However, rather than removing the Content-Type header completely, you could set `contentType` to the more appropriate `"multipart/form-data"`. As for `processData`, the default behavior is to try to convert the data to a string. It's been like this a long time. We could start adding exceptions to this (it probably wouldn't stop with FormData), and that is where breaking existing code comes in. Plugins or custom methods that handle the correct `$.ajax` options for each of these types of requests would be simpler (no added magic needed in $.ajax) and prettier `sendFormData()`.

jimmywarting:
> you could set contentType to the more appropriate "multipart/form-data"

You shouldn't have to set the request `contentType` header yourself to `multipart/form-data` since the boundary has to be added (something of which xhr and fetch do for you automatically) 
```
Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryAtoyumhGsqZNfA8E
```
just setting the content to `multipart/form-data` would break the backend code to not know how to read the boundary. I think you should not set the request content type at all if the data is a FormData. just let the XHR handle it

[...]

PR #5197, opened 2023-01-16
Ajax: Support binary data (including FormData)

mgol:
### Summary ###

~~**NOTE:** Please don't review the first commit; that one is #5196, it was just easier for me to depend on it.~~ (that PR landed)

Two changes have been applied:
* prefilters are now applied before data is converted to a string;
  this allows prefilters to disable such a conversion
* a prefilter for binary data is added; it disables data conversion
  for non-string non-plain-object `data`; for `FormData` bodies, it
  removes manually-set `Content-Type` header - this is required
  as browsers need to append their own boundary to the header

Ref gh-4150

`+39 bytes`. We should discuss whether we want this fully in core or just the small change of running prefilters early enough for this to be possible. It is quite generic, though, with the small exception of special `FormData` treatment - which, I think, is not necessarily desired in all cases? Unless it is - then the PR overhead would be `+20` bytes instead of `+39`.

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -562,13 +562,13 @@ jQuery.extend( {
 			}
 		}
 
+		// Apply prefilters
+		inspectPrefiltersOrTransports( prefilters, s, options, jqXHR );
+
 		// Convert data if not already a string
 		if ( s.data && s.processData && typeof s.data !== "string" ) {
 			s.data = jQuery.param( s.data, s.traditional );
 		}
-
-		// Apply prefilters
-		inspectPrefiltersOrTransports( prefilters, s, options, jqXHR );
 
 		// If request was aborted inside a prefilter, stop there
 		if ( completed ) {
--- a/src/ajax/binary.js
+++ b/src/ajax/binary.js
@@ -1,0 +1,17 @@
+import jQuery from "../core.js";
+
+import "../ajax.js";
+
+jQuery.ajaxPrefilter( function( s ) {
+
+	// Binary data needs to be passed to XHR as-is without stringification.
+	if ( typeof s.data !== "string" && !jQuery.isPlainObject( s.data ) ) {
+		s.processData = false;
+	}
+
+	// `Content-Type` for requests with `FormData` bodies needs to be set
+	// by the browser as it needs to append the `boundary` it generated.
+	if ( s.data instanceof window.FormData ) {
+		s.contentType = false;
+	}
+} );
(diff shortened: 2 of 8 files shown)


mgol:
If we also clear `contentType` for all binary data cases, the PR overhead would be `+20` bytes instead of `+39`.

src/ajax.js
```diff
@@ -562,14 +562,14 @@ jQuery.extend( {
...
+		// Apply prefilters
+		inspectPrefiltersOrTransports( prefilters, s, options, jqXHR );
+
 		// Convert data if not already a string
 		if ( s.data && s.processData && typeof s.data !== "string" ) {
 			s.data = jQuery.param( s.data, s.traditional );
 		}
 
-		// Apply prefilters
-		inspectPrefiltersOrTransports( prefilters, s, options, jqXHR );
```
gibson042, line 571: I think moving prefilters ahead of string coercion is a breaking change with respect to custom prefilters, cf. https://api.jquery.com/jQuery.ajaxPrefilter/ (emphasis mine):
> <code>originalOptions</code> are the options as provided to the <code>$.ajax()</code> method, **unmodified** and, thus, without defaults from <code>ajaxSettings</code>

Is there a way to instead establish something like prefilter phases?

mgol: The line that now happens after the prefilters is only setting `data`, not options. Is there any change that I'm missing?

BTW, I would also consider this a breaking change regardless of the above, that's why I'm targeting it at v4.

timmywil (approved):
Good for 4.0. Might need a note about the prefilter going before stringifying `s.data`, if that was ever documented.

src/ajax/binary.js
```diff
@@ -0,0 +1,17 @@
...
+
+	// Binary data needs to be passed to XHR as-is without stringification.
+	if ( typeof s.data !== "string" && !jQuery.isPlainObject( s.data ) ) {
+		s.processData = false;
+	}
+
+	// `Content-Type` for requests with `FormData` bodies needs to be set
+	// by the browser as it needs to append the `boundary` it generated.
+	if ( s.data instanceof window.FormData ) {
+		s.contentType = false;
```
mgol, line 15: Note to self: check if this could be overridden in options if this is set to `false` for all binary data, not just `FormData`.

mgol:
I missed handling array `data` values; a fix at #5203.

src/ajax/binary.js
mgol: OK, this cannot be changed via settings as of now. :/ I'll need another patch, I think.

mgol: PR: #5205

Issue #5194, opened 2023-01-13
jQuery 3.6.3 throws on some invalid selectors accepted in <=3.6.1 in Firefox (it works in Chrome & Safari)


mgol:
_Originally posted by @khetaramsb in https://github.com/jquery/jquery/issues/5177#issuecomment-1381649144_

After upgrading jQuery from 3.6.0 to3.6.3 pseudo selector :before does not work in selector list with Firefox. Everything works with Chrome.

This worked in 3.6.0
`jQuery(".inline-dialog-content:before").css({'left': 40});`

In 3.6.3 it gives following error:

Uncaught Error: Syntax error, unrecognized expression: unsupported pseudo: before

mgol:
@khetaramsb despite using a single colon here, `:before` (and `::before`) are really pseudo-elements, not pseudo-classes. They can neither be selected via jQuery nor via `querySelectorAll`.

I guess there's a behavior difference here because while a selector like `.inline-dialog-content:before` doesn't match anything, it also doesn't throw an error; it just returns an empty collection. The code you pasted was previously effectively a noop and now it throws. The reason for that is `CSS.supports('selector(:before')` returns `true` but `CSS.supports('selector(:is(:before)')` returns `false` as `:is()` doesn't support pseudo-elements inside.

That said, this is invalid input so I don't think we're going to change anything here; I don't see an easy way anyway. I've added it to the items to discuss with the team, though.

mgol:
@timmywil @gibson042 I found what's up with this `:before` handling that we discussed at the team meeting. So, `$('.foo:before,bar')` always throws but `$('.foo:before')`throws only if `$('.foo')` doesn't match anything. The difference is caused by an optimization in:
https://github.com/jquery/jquery/blob/c66d4700dcf98efccb04061d575e242d28741223/src/selector.js#L1575
where 1-elem selector lists are treated differently. Now, `.foo:before` looks like a class + a pseudo-class (just because of those legacy pseudo-elements allowing a single colon as well) and since `.foo` doesn't match anything, there's a bail out at:
https://github.com/jquery/jquery/blob/c66d4700dcf98efccb04061d575e242d28741223/src/selector.js#L1615-L1621
On the other hand, `.foo::before` (with double colons) fails during tokenizing so `$('.foo:before')` does throw regardless of whether `$('.foo')` would match anything or not.

mgol:
[...]

PR #5206, opened 2023-02-09
Selector: Stop relying on CSS.supports( "selector(...)" )

mgol:
### Summary ###

`CSS.supports( "selector(...)" )` has different semantics than selectors passed to `querySelectorAll`. Apart from the fact that the former returns `false` for unrecognized selectors and the latter throws, `qSA` is more forgiving and accepts some invalid selectors, auto-correcting them where needed - for example, mismatched brackers are auto-closed. This behavior difference is breaking for many users.

To add to that, a recent CSSWG resolution made `:is()` & `:where()` the only pseudos with forgiving parsing; browsers are in the process of making `:has()` parsing unforgiving.

Taking all that into account, we go back to our previous try-catch approach without relying on `CSS.supports( "selector(...)" )`. The only difference is we detect forgiving parsing in `:has()` and mark the selector as buggy.

The PR also updates `playwright-webkit` so that we test against a version of WebKit that already has non-forgiving `:has()`.

Fixes gh-5194
Ref gh-5098
Ref gh-5107
Ref w3c/csswg-drafts#7676

-61 bytes

### Checklist ###

### DIFF
--- a/src/selector/support.js
+++ b/src/selector/support.js
@@ -2,14 +2,14 @@ import document from "../var/document.js";
 import support from "../var/support.js";
 
 // Support: Chrome 105 - 110+, Safari 15.4 - 16.3+
-// Make sure forgiving mode is not used in `:has()`.
-// `*` is needed as Safari & newer Chrome implemented something in between
-// for `:has()` - it throws in `qSA` if it only contains an unsupported
-// argument but multiple ones, one of which is supported, are fine.
-// We want to play safe in case `:is()` gets the same treatment.
-//
-// Note that we don't need to detect the complete lack of support for `:has()`
-// as then the `qSA` path will throw and fall back to jQuery traversal anyway.
+// Make sure the the `:has()` argument is parsed unforgivingly.
+// We include `*` in the test to detect buggy implementations that are
+// _selectively_ forgiving (specifically when the list includes at least
+// one valid selector).
+// Note that we treat complete lack of support for `:has()` as if it were
+// spec-compliant support, which is fine because use of `:has()` in such
+// environments will fail in the qSA path and fall back to jQuery traversal
+// anyway.
 try {
 	document.querySelector( ":has(*,:jqfake)" );
 	support.cssHas = false;
--- a/src/selector/support.js
+++ b/src/selector/support.js
@@ -1,27 +1,20 @@
+import document from "../var/document.js";
 import support from "../var/support.js";
 
-// Support: IE 11+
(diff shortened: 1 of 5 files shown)


mgol:
I didn't write any new unit tests. I was wondering whether to do that but while being able to leverage the more forgiving native `qSA` parsing was our goal, I'm not sure if we want to guarantee such lax parsing in our unit tests. What do you think?

mgol:
I still need to prepare a version for `3.x-stable` (it won't be an exact cherry-pick) and Sizzle.

mgol:
-61 bytes

mgol:
The `3.x` version of this PR: #5207.

mgol:
Sizzle PR: https://github.com/jquery/sizzle/pull/493

timmywil (approved):

test/unit/support.js
```diff
@@ -59,39 +59,39 @@ testIframe(
...
 				reliableTrDimensions: false
 			},
 			firefox: {
-				cssSupportsSelector: true,
+				cssHas: true,
 				reliableTrDimensions: false
 			},
 			ios: {
-				cssSupportsSelector: false,
+				cssHas: false,
```
mgol, line 86: Note to self - iOS 14 has to report `true` here; this entry needs to be split just like in #5207.

src/selector/support.js
```diff
@@ -1,27 +1,20 @@
...
-// in this support test.
+// Support: Chrome 105 - 110+, Safari 15.4 - 16.3+
+// Make sure forgiving mode is not used in `:has()`.
+// `*` is needed as Safari & newer Chrome implemented something in between
+// for `:has()` - it throws in `qSA` if it only contains an unsupported
+// argument but multiple ones, one of which is supported, are fine.
+// We want to play safe in case `:is()` gets the same treatment.
+//
+// Note that we don't need to detect the complete lack of support for `:has()`
+// as then the `qSA` path will throw and fall back to jQuery traversal anyway.
```
gibson042, line 12:
```suggestion
// Make sure the the `:has()` argument is parsed unforgivingly.
// We include `*` in the test to detect buggy implementations that are
// _selectively_ forgiving (specifically when the list includes at least
// one valid selector).
// Note that we treat complete lack of support for `:has()` as if it were
// spec-compliant support, which is fine because use of `:has()` in such
// environments will fail in the qSA path and fall back to jQuery traversal
// anyway.
```

gibson042 (approved):
LGTM! I left some comment rewording suggestions.

PR #5207, opened 2023-02-09
Selector: Stop relying on CSS.supports( "selector(...)" )

mgol:
### Summary ###

This is a `3.x-stable` version of #5206

`CSS.supports( "selector(...)" )` has different semantics than selectors passed to `querySelectorAll`. Apart from the fact that the former returns `false` for unrecognized selectors and the latter throws, `qSA` is more forgiving and accepts some invalid selectors, auto-correcting them where needed - for example, mismatched brackers are auto-closed. This behavior difference is breaking for many users.

To add to that, a recent CSSWG resolution made `:is()` & `:where()` the only pseudos with forgiving parsing; browsers are in the process of making `:has()` parsing unforgiving.

Taking all that into account, we go back to our previous try-catch approach without relying on `CSS.supports( "selector(...)" )`. The only difference is we detect forgiving parsing in `:has()` and mark the selector as buggy.

The PR also updates `playwright-webkit` so that we test against a version of WebKit that already has non-forgiving `:has()`.

Fixes gh-5194
Ref gh-5098
Ref gh-5107
Ref w3c/csswg-drafts#7676

-46 bytes

### Checklist ###


mgol:
Sizzle PR: https://github.com/jquery/sizzle/pull/493

timmywil (approved):

PR #5213, opened 2023-02-16
Update jquery-1.9.1.js

mattbrundage:
Fixed a couple of typos


mgol:
Thanks, but this is just an older version of jQuery kept in the repository for testing purposes; it generally shouldn't be modified.

mattbrundage:
I only bothered to make a commit because I noticed that there was a recent change in the commit history.

mgol:
This was na exception to remove malicious links that were being flagged by some security scanners.

PR #5218, opened 2023-03-07
Release: add support for md5 sums in windows

timmywil:
### Summary ###

I wanted to see if the release could be run on Windows. To my surprise, the md5 hash was the only thing that didn't work.

This checks the OS and switches up the command used for the md5 sums and formats it the same way as default mac/linux.

### Checklist ###

### DIFF
--- a/build/release/cdn.js
+++ b/build/release/cdn.js
@@ -69,9 +69,7 @@ function makeArchives( Release, callback ) {
 
 		console.log( "Creating production archive for " + cdn );
 
-		var i,
-			sum,
-			result,
+		var i, sum, result,
 			archiver = require( "archiver" )( "zip" ),
 			md5file = cdnFolder + "/" + cdn + "-md5.txt",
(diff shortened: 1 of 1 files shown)


build/release/cdn.js
```diff
@@ -68,12 +69,15 @@ function makeArchives( Release, callback ) {
 
 		console.log( "Creating production archive for " + cdn );
 
-		var sum,
+		var i,
+			sum,
+			result,
```
mgol, line 74: The style usually used in other places is to put all uninitialized variable declarations in one line:
```suggestion
		var i, sum, result,
```

mgol (approved):
LGTM with one style suggestion. I assume this is to be cherry-picked to `main` & `3.x-stable` as well.

build/release/cdn.js
Kristianko1995: build/release/cdn.js

Issue #4950, opened 2021-10-11
You cannot programmatically focus an element after you have attempted to focus it while it was hidden (display:none)


JoolsCaesar:
### Description ###

Possibly related to #4867, I have replicated this in version 3.5.1 and 3.6. The linked example below uses 3.6. I haven't tested Firefox, but both Chrome and Edge display this issue.

See the following example. There's a focusable div at the top, which has a handler to report (at the bottom of the example) the number of times it is focused. When you click the div, it is only ever focused programmatically, because a handler calls `preventDefault()` on the `mousedown`. 

There is a button to toggle the visibility (`display`) of the focusable element, but before showing the element, it attempts to programmatically focus it. 

Focusing a `display:none` element is not expected to work, but the issue is what follows.

Once the element is shown, no future attempt to focus the element will work.

HTML:
```
  <div id="hideable">
    <div id="focusable" tabindex="1">
      I am focusable
    </div>
  </div>
  <button id="button">
    Toggle visibility
  </button>
  <div id="output">
    Click "I am focusable" to focus. Click "Toggle visibility" to hide/show the focusable element
  </div>
```
JS:
```
const $button = $('#button');
const $focusable = $('#focusable');
const $hideable = $('#hideable');
const $output = $('#output');

let focusinCount = 0;
$hideable.on('focusin', function() {
  focusinCount++;
  $output.text('I felt that ' + focusinCount);
});

let visible = true;
$button.click(function() {
  if (visible) {
    $hideable.hide();
  } else {
    $focusable.focus(); // Comment out this to fix
    $hideable.show();
  }
  visible = !visible;
});

$focusable.on('mousedown', function(event) {
  event.preventDefault();
  $focusable.focus();
})
```

### Link to test case ###

https://jsfiddle.net/JoolsCaesar/2o089jga/

mgol:
Thanks for the report. This looks like a valid issue.

jwmerrill:
I have bisected this problem to 753d591aea698e57d6db58c9 using the provided test case.

jwmerrill:
It appears that commenting out this line https://github.com/jquery/jquery/blob/4250b628783d7bfa92ec6c5550c6e4b22fab6034/src/event.js#L577

fixes the test case, but presumably that would cause other problems.

fullermetric:
Thanks for this Joels.  
The issue became more apparent after adding #focusable:focus { outline: 3px solid green; } to your test case.

[...]

Issue #4859, opened 2021-03-15
jQuery v3.6.0 breaks Bootstrap's test suite in IE


XhmikosR:
*Opening the issue after quickly chatting with mgol on IRC*

I made a branch to update jQuery to v3.6.0 in https://github.com/twbs/bootstrap/pull/33243. I notice our BrowserStack tests fail consistently on Internet Explorer 10 and 11. v3.5.1 works fine.

```sh
git clone https://github.com/twbs/bootstrap.git -b v4-dev-xmr-update-jquery
npm i
npm run js-test-cloud #assumes you have BrowserStack credentials set up
```

The failures seem to point to https://github.com/twbs/bootstrap/blob/v4-dev/js/tests/unit/dropdown.js#L860 and https://github.com/twbs/bootstrap/blob/v4-dev/js/tests/unit/dropdown.js#L870

EDIT: using jQuery v3.3.1 also works fine.

mgol:
An isolated test case is basicaly:
```js
input1.trigger('focus');
input2.trigger('focus');
input1.trigger('focus');
```

See versions with jQuery:
* 3.6.0: https://output.jsbin.com/wunumej/1
* 3.5.1: https://output.jsbin.com/vuhonam/1
* 3.3.1: https://output.jsbin.com/giduboh/1

The 3.6.0 version fails in IE, the older ones succeed.

jQuery 3.4 should be similar to 3.5 as there weren't major changes in relevant areas.

I think this is the same bug that makes some tests fail in jQuery UI only with jQuery 3.6.0+:
* 3.6.0: http://swarm.jquery.org/job/10970
* 3.5.1: http://swarm.jquery.org/job/10969

Also related to #4856, although that issue manifests itself in all 3.4+ versions.

timmywil:
We'll address this in a patch. Related: dbcffb396c2db61ff96edc4162602e850797d61f

gibson042:
@mgol The idea discussed in today's weekly meeting:
```diff
 // Ensure the presence of an event listener that handles manually-triggered
 // synthetic events by interrupting progress until reinvoked in response to
 // *native* events that it fires directly, ensuring that state changes have
 // already occurred before other listeners are invoked.
 function leverageNative( el, type, expectSync ) {
+	var result = {}, queue = [];
 
 	// Missing expectSync indicates a trigger call, which must force setup through jQuery.event.add
 	if ( !expectSync ) {
 		if ( dataPriv.get( el, type ) === undefined ) {
[...]

Issue #4856, opened 2021-03-10
In jQuery 3.4.0+ in IE a focus-blur-focus sequence doesn't leave the element focused


mgol:
### Description ###

Invoking:
```js
$( inputNode )
	.trigger( "focus" )
	.trigger( "blur" )
	.trigger( "focus" );
```
leaves input focused in all browsers except IE.

I discovered it while working to fix the remaining jQuery UI issues when tested against Core 3.6.0+. Interestingly, the relevant UI issue doesn't manifest itself in 3.4 & 3.5, I'm not sure why.

A relevant UI test run: http://swarm.jquery.org/result/3963629. That's the "dialog: core: focus tabbable" test.

### Link to test case ###

https://jsbin.com/zoyowez/1/edit?html,js,output

Changing the jQuery version to 3.3.1 makes it work again.

mgol:
OK, the relevant UI test (dialog: core: focus tabbable):
https://builds.jenkins.jquery.com/jquery-ui/9ea690a0d9e645280dbb448579d2ac3babb510ed/tests/unit/dialog/dialog.html?nojshint=true&jquery=3.4.1&testId=a84fb0fe
does fail with jQuery 3.4+, it's just that in 3.4 & 3.5 it doesn't fail when run together with the rest of the dialog test suite.

mgol:
That's the assertion code that's failing: https://github.com/jquery/jquery-ui/blob/e31cf579a0f3e64250af2e7269eef2e6c6e3d8fb/tests/unit/dialog/core.js#L121-L129

timmywil:
We'll look at addressing focus issues in 3.6.1.

mgol:
It looks like we can fix this via simulating the async `focus` & `blur` in IE via their bubbling but also synchronous equivalents: `focusin` & `focusout`. In other browsers, all those events are synchronous.

PRs for:
* jQuery 4.0: #5223
* jQuery 3.x: #5224

mgol:
Fixed on `main` (future `4.x`) via #5223 and on `3.x-stable` via #5224.

PR #5223, opened 2023-03-13
Event: Simulate focus/blur in IE via focusin/focusout

mgol:
### Summary ###

In IE (all versions), `focus` & `blur` handlers are fired asynchronously
but `focusin` & `focusout` are run synchronously. In other browsers, all
those handlers are fired synchronously.

Simulate `focus` via `focusin` & `blur` via `focusout` in IE to
avoid these issues.

Fixes gh-4856
Fixes gh-4859
Fixes gh-4950

+30 bytes

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -1,5 +1,4 @@
 import jQuery from "./core.js";
-import document from "./var/document.js";
 import documentElement from "./var/documentElement.js";
 import rnothtmlwhite from "./var/rnothtmlwhite.js";
 import rcheckableType from "./var/rcheckableType.js";
@@ -20,16 +19,6 @@ function returnTrue() {
 
 function returnFalse() {
 	return false;
-}
-
-// Support: IE <=9 - 11+
-// focus() and blur() are asynchronous, except when they are no-op.
-// So expect focus to be synchronous when the element is already active,
-// and blur to be synchronous when the element is not already active.
-// (focus and blur are always synchronous in other supported browsers,
-// this just defines when we can count on it).
-function expectSync( elem, type ) {
-	return ( elem === document.activeElement ) === ( type === "focus" );
 }
 
 function on( elem, types, selector, data, fn, one ) {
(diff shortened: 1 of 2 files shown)


timmywil (approved):
Looks good for 4.0. We will not be guaranteeing order anymore, but https://github.com/jquery/jquery/issues/4300 points out that everyone has opinions on what the order should be and modern browsers don't even follow the spec.

mgol:
+42 bytes

mgol:
Fixes gh-4856
Fixes gh-4859

mgol:
The `3.x-stable` version of the PR: #5224.

mgol:
Reduced to +34 bytes

mgol:
@timmywil I added a few commits on top of the original PR so I'm re-requesting a review from you.

src/event.js
```diff
@@ -582,16 +559,14 @@ function leverageNative( el, type, expectSync ) {
...
-					value: jQuery.event.trigger(
-
-						// Support: IE <=9 - 11+
-						// Extend with the prototype to reset the above stopImmediatePropagation()
-						jQuery.extend( saved[ 0 ], jQuery.Event.prototype ),
-						saved.slice( 1 ),
-						this
-					)
-				} );
+				dataPriv.set( this, type, jQuery.event.trigger(
```
mgol, line 562: @gibson042 One thing to note - while #4350 which added this wrapping was mainly aimed at IE with its async `focus` handlers, there has been a report about this also affecting Chrome somehow:
https://github.com/jquery/jquery/issues/4350#issuecomment-482292263

I'm not sure if we can do anything about it due to the lack of data, though.

mgol:
Reduced to +28 bytes

mgol:
It looks like this also fixes #4950; I added a test for this in a separate commit & verified it fails on `main`.

dmethvin (commented):
👏 👏 👏

PR #5224, opened 2023-03-13
Event: Simulate focus/blur in IE via focusin/focusout

mgol:
### Summary ###

In IE (all versions), `focus` & `blur` handlers are fired asynchronously
but `focusin` & `focusout` are run synchronouslu. In other browsers, all
those handlers are fired synchronously.

Simulate `focus` via `focusin` & `blur` via `focusout` in IE to
avoid these issues.

Also, simplify `leverageNative` - With IE now using `focusin` to simulate
`focus` and `focusout` to simulate `blur`, we don't have to deal with
async events in `leverageNative`.

Fixes gh-4856
Fixes gh-4859

All tests are passing in IE 9-11, I verified manually.

### Checklist ###


mgol:
The `main` version of the PR: #5223.

mgol:
+71 bytes

mgol:
Reduced to +66 bytes

mgol:
Reduced to +58 bytes

mgol:
It looks like this also fixes #4950; I added a test for this in a separate commit & verified it fails on `3.x-stable`.

mgol:
This also seems to fix one heavily upvoted select2 issue: https://github.com/select2/select2/issues/5993.

PR #5230, opened 2023-03-27
Update README.md

Poofmoyoo:
### Summary ###
I have updated the readme file because it was not windows friendly.


mgol (changes requested):
Thanks for the PR. Let’s not single out Windows but just split it for everyone.

Also, we need you to sign our CLA.

Issue #3982, opened 2018-02-19
outerHeight(true) does not handle negative margins properly (starting with 3.3.x release)


bmurmistro:
Hello,

outerHeight(true) does not handle negative margins properly (starting with 3.3.x release). This was working fine in 3.2.x.

3.2.1: https://jsfiddle.net/b1mtv4ts/12/
3.3.1: https://jsfiddle.net/b1mtv4ts/13/

gibson042:
The problem seems to be that `offsetHeight` and `offsetWidth` aren't affected by negative margins. We should be able to account for that in `boxModelAdjustment`.

mgol:
PR: #5234

PR #5234, opened 2023-03-30
CSS: Make `offsetHeight( true )`, etc. include negative margins

mgol:
### Summary ###

This regressed in gh-3656 as the added logic to include scroll gutters in `.innerWidth()` / `.innerHeight()` didn't take negative margins into account. This broke handling of negative margins in `.offsetHeight( true )` and `.offsetWidth( true )`. To fix it, calculate margin delta separately and only add it after the scroll gutter adjustment logic.

Fixes gh-3982
Ref gh-3656

+4 bytes

### Checklist ###

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -46,7 +46,8 @@ function setPositiveNumber( _elem, value, subtract ) {
 function boxModelAdjustment( elem, dimension, box, isBorderBox, styles, computedVal ) {
 	var i = dimension === "width" ? 1 : 0,
 		extra = 0,
-		delta = 0;
+		delta = 0,
+		marginDelta = 0;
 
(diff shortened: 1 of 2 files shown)


mgol:
+4 bytes

mgol:
@gibson042 I'd like to start asking some projects to test against `3.x-git` as a way to verify we didn't break anything with some of the more risky changes so I'll land this. If you have different ideas on how to address it, we can address them separately later.

mgol:
Landed on `main` in bce13b72c1753e16cc0db53ebf0f0456bdcf6b48 and on `3.x-stable` in 7bb48a0290a20594ea2a5a7b5772e0410a67164c.

PR #5235, opened 2023-03-30
Event: Avoid collisions between jQuery.event.special & Object.prototype

mgol:
### Summary ###

This is a follow-up to similar changes to data & event storages from gh-4603.

This is a breaking change as there may be code like:
```js
jQuery.event.special.hasOwnProperty( "something" )
```
in the wild.

Ref gh-4603

+3 bytes

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -429,7 +429,7 @@ jQuery.event = {
 			new jQuery.Event( originalEvent );
 	},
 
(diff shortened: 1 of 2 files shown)


mgol:
+3 bytes

gibson042 (approved):

PR #5236, opened 2023-03-30
Event: Simplify the check for saved data in leverageNative

mgol:
### Summary ###

Previously, when `leverageNative` handled async events, there was a case where an empty placeholder object was set as a result. Covering both such an object and `false` required a `length` check. However, this is not necessary since gh-5223 and the check was already simplified in other places; this one was missed.

Ref gh-5223

-1 byte

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -556,7 +556,7 @@ function leverageNative( el, type, isSetup ) {
 
 			// If this is a native event triggered above, everything is now in order
 			// Fire an inner synthetic event with the original arguments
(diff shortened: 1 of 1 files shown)


gibson042 (approved):
LGTM

mgol:
Landed on `main` in dfe212d5a1eed6b4a67d1cbd04ece09bbac33699 and on `3.x-stable` in 9ab26aa508c6cca6afa9c6247ee6d50eaed2da77.

PR #5258, opened 2023-05-24
Build: Switch the minifier from UglifyJS to Terser

mgol:
### Summary ###

UglifyJS is ES5-only, while Terser supports newer ECMAScript versions. jQuery is authored in ES5 but jQuery 4.x will also have an ESM build that cannot be minified using UglifyJS directly.

We could strip the `export` statement, minify via UglifyJS and re-add one but that increases complexity & may not fully play nice with source maps.

On the other hand, switching to Terser increases the minfied size by just 324 bytes and the minified gzipped one by just 70 bytes. Such differences largely disappear among bigger size gains from the `3.x-stable` line - around 2.7 KB minified gzipped as of now.

### Checklist ###

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -329,25 +329,26 @@ module.exports = function( grunt ) {
 			files: [ "<%= eslint.dev.src %>" ],
 			tasks: [ "dev" ]
 		},
-		uglify: {
+		terser: {
 			all: {
 				files: {
 					"dist/<%= grunt.option('filename').replace('.js', '.min.js') %>":
 						"dist/<%= grunt.option('filename') %>"
 				},
 				options: {
-					preserveComments: false,
-					sourceMap: true,
-					sourceMapName:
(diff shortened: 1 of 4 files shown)


mgol:
This is a pre-requisite to finalizing the `exports` PR (#5255).

gibson042 (approved):
I'm in favor of this, but it supports my suspicion that Terser is not as good as Uglify w.r.t. output compressibility.

mgol:
@gibson042 Yes, although it got better in the past year. Last time I was testing with Terser `5.14.2` (released on 2022-07-13), the numbers were +620 minified, +99 minified gzipped.

I also think it may be a good idea to compare the output of UglifyJS & Terser in more detail and work with Terser maintainers to improve it, getting closer to UglifyJS. But I wouldn't treat this as a blocker for this PR, just ground for possible future optimizations.

Issue #5271, opened 2023-06-19
Selector: Firefox 114 fails the ":lang respects escaped backslashes" selector test


mgol:
### Description ###

Firefox 114 fails the ":lang respects escaped backslashes" selector test. We should fix it before releasing `3.7.1`.

### Link to test case ###

https://swarm.jquery.org/result/4756651

<img width="509" alt="Screenshot 2023-06-20 at 00 12 52" src="https://github.com/jquery/jquery/assets/1758366/4dd1e9b3-393e-4692-8d4a-47dc216ae068">

ewolfman:
hi @mgol - general question not necessarily related to this issue - when is jQuery 3.7.1 scheduled to be released? (under the 3.7.1 it says "No due date").

mgol:
@ewolfman We don't promise any dates, it'll be released when we're ready. In particular, all issues under the milestone must be closed.

mgol:
I submitted bug to Mozilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1839747

mgol:
PR disabling the problematic test: #5277

PR #5277, opened 2023-06-26
Tests: Disable the ":lang respects escaped backslashes" test

mgol:
### Summary ###

Firefox 114+ no longer match on backslashes in `:lang()`, even when escaped. It is an intentional change as `:lang()` parameters are supposed to be valid BCP 47 strings. Therefore, we won't attempt to patch it. We'll keep this test here until other browsers match the behavior.

Fixes gh-5271
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1839747#c1 Ref https://github.com/w3c/csswg-drafts/issues/8720#issuecomment-1509242961

### Checklist ###

### DIFF
--- a/test/unit/selector.js
+++ b/test/unit/selector.js
@@ -1474,7 +1474,7 @@ QUnit.test( "pseudo - :target and :root", function( assert ) {
 } );
 
 QUnit.test( "pseudo - :lang", function( assert ) {
-	assert.expect( QUnit.jQuerySelectors ? 105 : 55 );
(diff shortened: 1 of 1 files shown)


timmywil (approved):

mgol:
Landed on `main` in 62b9a2583460c2384f6de1d0a6aeaf05e51d523b and on `3.x-stable` in 5aa7d93acf61b165b12c4e4cc2440536a5fa94af.

Issue #5281, opened 2023-06-29
Iframes Memory Leak with jQuery and Sizzle


vamsikrishnach90:
### Description ###
We have a situation where there are 3 levels of nested iframes. The 3rd iframe initializes few heavy size objects, but this iframe is discarded and rebuilt in the DOM whenever user navigates across the application.

A memory leak is noticed where heap size of the application keeps on increasing whenever a new iframe at 3rd level is discarded and constructed.

### Link to test case ###
Goto https://vamsikrishnach90.github.io/ in Chrome. Open developer tools (Hit F12) and navigate to Memory Tab. Pay attention to the JS heap size which is arppox 5 MB.

1. Click on 'Load Child Page' button. A child page is loaded inside the iframe of parent page.
2. Click on 'Bombard the page!' button on the child page. This will initialize 1 global variable of custom type with a huge json. Open devtools and check for the heap which is now 75 MB.
3. Click on 'Remove Child Page' button. The child page (iframe) is removed from the DOM. 

Repeat steps 1 - 3 for about 2 more times. After step 2, note down the heap size. You'll notice that the heap keeps on increasing every time we add a child iframe and bombard it. Even though at step 3, we discard the frame, the heap is not claimed by the garbage collector.
[Usecase demo help (Picture)](https://i.stack.imgur.com/5rUI0.png)

## Observations ##
After repeating above 1-3 steps for 3x times, take a heap snapshot. Notice that the 3 instances of custom class 'Vamsi' that were created every time we added the child page and bombarded it still exist in the heap, even though the corresponding iframes are deleted from the DOM.

Notice in the object graph, that an unloadHandler() has been bound to the Window object belonging to the discarded iframes. [Heap Memory Snapshot](https://i.stack.imgur.com/je59l.png)

## Root Cause ##
The child page has a jquery lookup statement in the parent page:
`$('div.msg-banner', window.parent.document).show();`

This one line statement is causing the issue for memory leak.

> $('div.msg-banner', window.parent.document)

[...]

PR #5282, opened 2023-06-29
Selector: Only attach the unload handler in IE & Edge Legacy

mgol:
### Summary ###

Both IE & Edge Legacy need the workaround of calling `setDocument()` in an `unload` handler to avoid "permission denied" errors. However, due to not being possible to feature-detect this issue, the handler has been applied in all browsers for windows different than the one in which jQuery was loaded.

jQuery 4.0, which drops Edge Legacy support, guards this workaround with a `document.documentMode` check. This won't work in the 3.x line due to still supporting Edge Legacy but we can check for
`document.documentElement.msMatchesSelector` instead as that API is supported in IE 9+ and all Edge Legacy versions.

Fixes gh-5281
Ref gh-4792

+6 bytes

### Checklist ###

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -523,12 +523,17 @@ function setDocument( node ) {
 		documentElement.msMatchesSelector;
 
 	// Support: IE 9 - 11+, Edge 12 - 18+
-	// Accessing iframe documents after unload throws "permission denied" errors (see trac-13936)
-	// Support: IE 11+, Edge 17 - 18+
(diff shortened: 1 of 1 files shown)


mgol:
Test case showing all supported aliases of `matches`: https://output.jsbin.com/lawacad/1. In Edge 15-18 this includes `matches`, `webkitMatchesSelector` and `msMatchesSelector`.

timmywil (approved):

Issue #5285, opened 2023-07-06
The map file incorrectly contains the `dist/` prefix in & `file` `sources` fields


mgol:
### Description ###

The `dist/jquery.min.js.map` file generated by Terser contains two fields with incorrect values:
```json
"file":"dist/jquery.min.map",
"sources":["dist/jquery.js"]
```
The proper values are:
```json
"file":"jquery.min.js",
"sources":["jquery.js"]
```
and that's what we have on the `3.x-stable` line with UglifyJS.

Note that not only the `dist/` prefix should go away but also the Terser's source map incorrectly points to itself in the `file` property.

We need to fix it before the final `4.0.0` release.

mgol:
Note that not only the `dist/` prefix should go away but also the Terser's source map incorrectly points to itself in the `file` property. I updated the original description:

The `dist/jquery.min.js.map` file generated by Terser contains two fields with incorrect values:
```json
"file":"dist/jquery.min.map",
"sources":["dist/jquery.js"]
```
The proper values are:
```json
"file":"jquery.min.js",
"sources":["jquery.js"]
```
and that's what we have on the `3.x-stable` line with UglifyJS.

mgol:
Switching to SWC and fixing this as well in #5286.

PR #5286, opened 2023-07-06
Build: Switch form Terser to SWC for JS minification

mgol:
### Summary ###

Also, as part of this, fix the `file` & `sources` properties of the source map file.

Fixes gh-5285
Ref gh-5258

-18 bytes

### Checklist ###

### DIFF
--- a/build/tasks/minify.js
+++ b/build/tasks/minify.js
@@ -1,0 +1,56 @@
+/**
+ * Minify JavaScript using SWC.
+ */
+
+"use strict";
+
+module.exports = ( grunt ) => {
+	const swc = require( "@swc/core" );
+
+	grunt.registerMultiTask(
+		"minify",
+		"Minify JavaScript using SWC",
+		async function() {
+			const done = this.async();
+			const options = this.options();
+			const sourceMapFilename = options.sourceMap && options.sourceMap.filename;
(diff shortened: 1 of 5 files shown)


mgol:
-18 bytes

build/tasks/minify.js
```diff
@@ -0,0 +1,56 @@
...
+
+				if ( sourceMapFilename ) {
+
+					// Apply map overrides if needed. See the task config description
+					// for more details.
+					const mapObject = {
+						...JSON.parse( incompleteMap ),
+						...sourceMapOverrides
+					};
+					const map = JSON.stringify( mapObject );
```
mgol, line 47: I wish we could avoid this parse + fix + stringify back behavior, but it's still quick enough (I actually cannot measure a meaningful time difference with & without this code) and there's not much we can do until the issues are fixed or I learn which parameters to use to get what we need out of the box.

timmywil (approved):
I forget now why we were reticent to switch to swc, but this looks good.

mgol:
> I forget now why we were reticent to switch to swc, but this looks good.

It used to not support the preamble option - and inserting it manually would require source map adjustments which is not something I'd like to do on our end. They added support after my report in https://github.com/swc-project/swc/issues/7475.

PR #5290, opened 2023-07-10
Build: Make sure `*.cjs` & `*.mjs` files use UNIX line endings as well

mgol:
### Summary ###

We've had this rule for `*.js` files so far but we now have two new JS extensions.

Another option is to set this for all text files, not just JS. Is there any downside? AFAIK most Windows editors/IDEs should handle LF line endings just fine nowadays?

### Checklist ###

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,4 +2,7 @@
 * text=auto
 
 # JS files must always use LF for tools to work
+# JS files may have mjs or cjs extensions now as well
 *.js eol=lf
+*.cjs eol=lf
+*.mjs eol=lf


timmywil (approved):

mgol:
Landed on `main` in 198b41c8c2cd726b875615023b2b37b213040ad3 and on `3.x-stable` in 3c18c1f33cfc69e1e1bd1410ab5176b2abc5fe3a.

jose8109:
stop

jose8109 (changes requested):
stop

PR #5293, opened 2023-07-10
Core: Move the factory to separate exports

mgol:
### Summary ###

Since versions 1.11.0/2.1.0, jQuery has used a module wrapper with one strange
addition - in CommonJS environments, if a global `window` with a `document` was
not present, jQuery exported a factory accepting a `window` implementation and
returning jQuery.

This approach created a number of problems:
1. Properly typing jQuery would be a nightmare as the exported value depends on
   the environment. In practice, typing definitions ignored the factory case.
2. Since we now use named exports for the jQuery module version, it felt weird
   to have `jQuery` and `$` pointing to the factory instead of real jQuery.

Instead, for jQuery 4.0 we leverage the just added `exports` field in
`package.json` to expose completely separate factory entry points: one for the
full build, one for the slim one.

Exports definitions for `./factory` & `./factory-slim` are simpler than for `.`
and `./slim` - this is because it's a new entry point, we only expose a named
export and so there's no issue with just pointing Node.js to the CommonJS
version (we cannot use the module version for `import` from Node.js to avoid
double package hazard). The factory entry points are also not meant for the Web
browser which always has a proper `window` - and they'd be unfit for an
inclusion in a regular script tag anyway. Because of that, we also don't
generate minified versions of these entry points.

The factory files are not pushed to the CDN since they are mostly aimed
at Node.js.

### Checklist ###


mgol:
Since the main builds no longer include the factory code in the wrapper, there's a small size reduction - bigger for ESM builds:
```
   raw     gz Compared to main @ 8be4c0e4f89d6c8f780e5937a0534921d8c7815e      
   -36     -8 dist/jquery.min.js                                               
   -36    -11 dist/jquery.slim.min.js                                          
  -139    -34 dist-module/jquery.module.min.js                                 
  -139    -34 dist-module/jquery.slim.module.min.js                            
```

build/tasks/node_smoke_tests.js
```diff
@@ -17,33 +18,116 @@ if ( !verifyNodeVersion() ) {
...
+
+	if ( !testFiles.length ) {
+		throw new Error( `No test files found for ${
+			libraryType
+		} ${ sourceType } "${ module }"` );
+	}
+
 	await Promise.all(
 		testFiles.map( ( testFile ) =>
-			exec( `node "${dir}/${testFile.name}" "${module}"` )
```
mgol, line 31: We should have spaces inside of interpolations in template literals as it's consistent with the jQuery code style. We should update our ESLint config to account for that.

build/tasks/node_smoke_tests.js
```diff
@@ -17,33 +18,116 @@ if ( !verifyNodeVersion() ) {
...
-	const dir = `./test/node_smoke_tests/${sourceType}`;
+	const dir = `./test/node_smoke_tests/${ sourceType }/${ libraryType }`;
 	const files = await fs.promises.readdir( dir, { withFileTypes: true } );
 	const testFiles = files.filter( ( testFilePath ) => testFilePath.isFile() );
+
+	if ( !testFiles.length ) {
+		throw new Error( `No test files found for ${
+			libraryType
+		} ${ sourceType } "${ module }"` );
+	}
```
mgol, line 36: When making the changes after the rebase, I noticed tests are finishing a bit too quickly. Then I realized I forgot to update the `dir` definition. I added some validation that anything got run.

It'd be a bit safer to also print a list of files with tests that are executed as then if wrong ones were run, I'd notice immediately. This would make the output here a bit longer, though, and we don't have a good mechanism for verbose mode so I didn't add anything like that.

@timmywil what do you think? Should we print the list of test files?

build/tasks/node_smoke_tests.js
timmywil: TBF, the jQuery eslint plugin doesn't specify this. And, while our style guide does designate spaces inside curly braces, template literals could be considered separate, and it seems eslint agrees.

build/tasks/node_smoke_tests.js
timmywil: I don't think so. We probably won't touch this file much in the future and I'd prefer less noise. We could add more checks if it becomes an issue.

README.md
```diff
@@ -143,6 +143,20 @@ By default, jQuery generates a regular script JavaScript file. You can also gene
...
+```bash
+grunt custom --factory
+```
+
+The default mode doesn't generate a factory, but you can also pass it explicitly via `--no-factory`:
+
+```bash
+grunt custom --no-factory
+```
+
```
timmywil, line 159: Need to update the build/command.js and add the boolean argument `factory`. The equivalent of `--no-factory` would be `--factory=false`. Then update docs here.

build/tasks/node_smoke_tests.js
mgol: I know it does not specify this but since the jQuery style guide spirit is basically "spaces everywhere", it makes sense to cover template literals with this rule as well for consistency.

It doesn't touch on this now because for a long time the style guide was only catering to ES5.

build/tasks/node_smoke_tests.js
mgol: Cool, so what we have here should be enough then.

README.md
mgol: I think I'll skip the `false` versions since we don't cover it for other options either.

eslint.config.js
```diff
@@ -136,6 +136,56 @@ export default [
...
+					ignoredNodes: [
+						"Program > FunctionDeclaration > *"
+					]
+				}
+			],
+			"import/no-unused-modules": "off"
+		}
+	},
+
+
```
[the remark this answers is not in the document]
mgol: There is but there are tiny differences so grouping wouldn't work, we would still need to add some overrides. I felt it's simpler to just have an entry per file in that case.

README.md
mgol: Updated.

mgol:
@timmywil PR updated

build/tasks/node_smoke_tests.js
mgol: There are many spacing rules in ESLint and yet we don't describe each of them separately as they all form a bigger style guide for us (some rules we've had to enable: `space-before-blocks`, `space-in-parens`, `space-infix-ops`, `space-before-function-paren`, `space-unary-ops`). ESLint rules are just often quite granular.

We've actually gravitated towards consistent spacing within any kind of brackets over the years (old jQuery didn't space out the ending `});`, for example). This makes it easy to remember the rules - always put a space after `(`, `{`, `[` and before `}`, `)`, `]`, etc. - regardless of what they're used for.

Also, while the jQuery style results overall in longer lines, it has its advantages - spaces in all kinds of brackets make it easier to distinguish tokens when parsing the code with our own eyes. For the same reason, I prefer to have spaces in template literal interpolations; it makes it easier to read them for me. Especially when I'm used to this rich spacing in jQuery projects, having this one being a sole case of any opening paren not being followed by a space makes it look inconsistent to me - and it doesn't really matter that in ESLint this is a separate rule.

eslint.config.js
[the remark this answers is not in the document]
timmywil: Now that there are 4 files, it's hard to see the differences. Here's a pass that makes the differences more obvious:

```js
{
	files: [
		"src/wrapper.js",
		"src/wrapper-esm.js",
		"src/wrapper-factory.js",
		"src/wrapper-factory-esm.js"
	],
	languageOptions: {
		globals: {
			jQuery: false
		}
	},
	rules: {
		"no-unused-vars": "off",
		indent: [
			"error",
			"tab",
			{

				// This makes it so code within the wrapper is not indented.
				ignoredNodes: [
					"Program > FunctionDeclaration > *"
				]
			}
		]
	}
},

{
	files: [
		"src/wrapper.js",
		"src/wrapper-factory.js"
	],
	languageOptions: {
		sourceType: "script",
		globals: {
			module: false
		}
	}
},

{
	files: [ "src/wrapper.js" ],
	rules: {
		indent: [
			"error",
			"tab",
			{

				// This makes it so code within the wrapper is not indented.
				ignoredNodes: [
					"Program > ExpressionStatement > CallExpression > :last-child > *"
				]
			}
		]
	}
},

```

timmywil (changes requested):
See my comment on eslint config.

eslint.config.js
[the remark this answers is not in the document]
mgol: OK, that works.

build/tasks/node_smoke_tests.js
mgol: Yes, I'd expect the change to be done in the plugin.

mgol:
PR updated

timmywil (approved):

PR #5297, opened 2023-07-11
Migrate linting off of grunt

timmywil:
### Summary ###

I'm beginning the migration off of grunt, and I'm starting with the lint task. Lint is run during the browserless tests, which will run on all PRs. The pre-commit hook (with updated husky) uses a lint task with a cache for faster commits. I've also changed the eslint config files to JS (`.cjs` to be exact) so we have more flexibility and comments are valid.

### Checklist ###


dist/.eslintrc.cjs
```diff
@@ -0,0 +1,46 @@
...
+
+				// When custom compilation is used, the version string
+				// can get large. Accept that in the built version.
+				"max-len": "off",
+				"one-var": "off"
+			}
+		},
+
+		{
+			files: ".eslintrc.cjs",
```
mgol, line 42: Hmm, so we are linting this file. Why doesn’t it complain about sloppy mode?

package.json
```diff
@@ -30,6 +30,30 @@
     "./src/*.js": "./src/*.js"
   },
   "main": "dist/jquery.js",
+  "scripts": {
+    "build": "npm install && npm run build-all-variants",
```
mgol, line 34: Is there a good reason to move it all up? That breaks git history for selection.

package.json
```diff
@@ -30,6 +30,30 @@
...
+    "build-all-variants": "grunt custom:slim --esm --filename=jquery.slim.module.js && grunt custom --esm --filename=jquery.module.js && grunt custom:slim --filename=jquery.slim.js && grunt custom",
+    "jenkins": "npm run test:browserless",
+    "lint:dev": "eslint .",
+    "lint:dist": "eslint --no-ignore dist/jquery.js dist/jquery.slim.js dist-module/jquery.module.js dist-module/jquery.slim.module.js",
+    "lint:json": "jsonlint --quiet package.json",
+    "lint:cache": "concurrently -r \"npm:lint:dev -- --cache\" \"npm:lint:json\"",
+    "lint": "concurrently -r \"npm:lint:dev\" \"npm:lint:json\" && npm run lint:dist",
+    "prepare": "husky install",
+    "start": "grunt watch",
+    "test:browserless": "grunt && npm run lint && npm run test:node_smoke_tests && grunt test:slow",
```
mgol, line 44: This is equivalent to what we have now so it’s fine for now but on GitHub Actions we should only run lint on a single Mode version. Otherwise, if a PR has a linting error, an inline comment will be added by GitHub to each violation the number of times we run the lint. This is not a pleasant view. 😄

mgol (changes requested):
Nice start!

I added a few comments.

Also, you need to reimplement the logic skipping ESLint on old Node for the purpose of Node 10; CI is failing on your PR.

dist/.eslintrc.cjs
timmywil: Yeah, I'm not sure what's wrong. It's linted, but it doesn't always complain about no 'use strict'; I'm looking into it.

package.json
timmywil: Not a *good* reason, but maybe an okay reason. We're going to be adding more scripts and working with them more often so I figured it'd be nice to have it closer to the top.

package.json
mgol: Yeah, that's a good idea.

If you do that, please make sure it's still executed on `npm test`. `npm test` should run all the checks & tests we have. We don't run it in CI directly anywhere but I sometimes use it to verify my changes locally before creating a PR.

package.json
```diff
@@ -30,6 +30,30 @@
     "./src/*.js": "./src/*.js"
   },
   "main": "dist/jquery.js",
+  "scripts": {
+    "build": "npm install && npm run build-all-variants",
+    "build-all-variants": "grunt custom:slim --esm --filename=jquery.slim.module.js && grunt custom --esm --filename=jquery.module.js && grunt custom:slim --filename=jquery.slim.js && grunt custom",
+    "jenkins": "npm run test:browserless",
+    "lint:dev": "eslint .",
```
mgol, line 37: I think you need to specify extensions here via the `--ext` option or it will only check `*.js`.

dist/.eslintrc.cjs
mgol: It is possible it would only complain if you have at least a single function in the file.

Or maybe this is the reason: https://github.com/jquery/jquery/pull/5297/files#r1259812907

package.json
mgol: Hmm, I see. I guess it's a good moment then as this PR doesn't change that much in scripts and further changes may be easier to review.

package.json
timmywil: `js, cjs, and mjs` are the default extensions. But, what do you think of moving to the new flat config? Especially since we'll only run on latest node https://eslint.org/docs/latest/use/configure/configuration-files. The old config pattern will be deprecated in version 9.

timmywil: I think all that's needed is for eslint-config-jquery to use JS instead of JSON.

PR #5113, opened 2022-09-09
Selector: Inline Sizzle into the selector module: 3.x version

mgol:
### Summary ###

The first commit is a cherry-pick of #4406.

The next commits are cherry-picks of the branch from which I landed #4395; see that PR for more details about this part. The last commit from this batch is named "Selector: Reduce selector-native.js". NOTE: not all of the changes were backported as on `3.x-stable` we need to maintain backwards compatibility. The end of `src/selector.js` consists of a lot of assignments to `jQuery.find` to keep the old Sizzle interface intact.

All the subsequent commits are largely backporting changes to `src/selector.js` & `src/selector/**/*` that landed after #4395 and changes from Sizzle after the backport was done.

Some last commits are fixing minor browser issues.

This PR passes all jQuery tests in latest Firefox, Chrome & IE 9-11. All Sizzle tests also pass with `Sizzle = jQuery.find`. There are minor issues in iOS 7 (just 2 tests failing that I'll fix) and I still need to verify old Android and generally all remaining browsers supported by 3.x. But considering IE is good and old iOS has only minor issues, I'm optimistic this can work.

I'd love some preliminary reviews of this.

TODO for me:
* [ ] Run the full TestSwarm run on browsers supported by 3.x from this branch
* [ ] Manually run tests on browsers out of TestSwarm, like old iOS & Android Browser
* [ ] Fix remaining browser issues, like the iOS 7 ones.

### Checklist ###

* [x] New tests have been added to show the fix or feature works
* [ ] Grunt build and unit tests pass locally with these changes
* [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #5306, opened 2023-08-08
Release: migrate release process to release-it

timmywil:
### Summary ###

*Authors*
- Checking and updating authors has been migrated to a custom script in the repo

*Changelog*
- changelogplease is no longer maintained
- generate changelog in markdown for GitHub releases
- generate changelog in HTML for blog posts

*dist*
- clone dist repo, copy files, and commit/push
- commit tag with dist files on the 3.x-stable branch; remove dist files from main branch after release
  * Intentionally committed on the branch rather than creating a detached tag, which triggers warnings on GitHub. It's debatable whether we need to commit the dist files on the tag commit, but since we are, they need to be removed in a subsequent commit when doing it on the branch.
  * Sort of related to the subsequent commit, I lean towards not updating the version back to a `-pre` version. It's a practice we invented and have been doing a long time, but I'm not sure is actually necessary. It simplifies some release-it steps to not do it, which bases it's automatic version detection on the assumption that the version in package.json is the last version released. Besides, our `-pre` version is often wrong anyway (for instance, when the next version is going to be a minor or major instead of a patch). If jquery-git.js is a concern, it does already include the SHA in the header and I think we could remove the version there completely.

*cdn*
- clone cdn repo, copy files, and commit/push
- create versioned and unversioned copies in cdn/
  * Intentionally moved out of the dist/ folder. It never really needed to be there and it makes even less sense on the main branch with dist-module/ being a thing. The separation also simplifies some scripts.
- generate md5 sums and archives for Google and MSFT

*GH actions*
- add GitHub workflow for running releases in CI
- releases can be triggered from GitHub; args can be added for pre releases
- npm publish can now run with provenance

Fixes jquery/jquery-release#114

### Checklist ###


AUTHORS.txt
```diff
@@ -1,3 +1,5 @@
+Authors ordered by first contribution.
```
timmywil, line 1: This line was accidentally removed at some point. The `main` branch still has it.

AUTHORS.txt
```diff
@@ -366,3 +337,6 @@ Simon Legner <Simon.Legner@gmail.com>
 Vladimir Sitnikov <sitnikov.vladimir@gmail.com>
 Anders Kaseorg <andersk@mit.edu>
 Alex <aleksandrosansan@gmail.com>
+Timo Tijhof <krinkle@fastmail.com>
+Gabriela Gutierrez <gabigutierrez@google.com>
+Dimitri Papadopoulos Orfanos <3234522+DimitriPapadopoulos@users.noreply.github.com>
```
timmywil, line 342: I know the number of changes looks odd, but I've confirmed that `grunt authors` has the same result. My guess is this has something to do with privacy rules being respected more accurately in git logs.

build/release/dist.mjs
```diff
@@ -0,0 +1,93 @@
+import { argv } from "process";
+import fs from "fs";
+import util from "util";
+import { exec as nativeExec } from "child_process";
+import { rimraf } from "rimraf";
+import inquirer from "inquirer";
```
timmywil, line 6: This file is `.mjs`  because of `inquirer` being esm only.

AUTHORS.txt
mgol: No, and I actually made these changes long ago on purpose... This was to certify that contributions to Sizzle matter as much to jQuery as contributions to the main jQuery repo. I changed the file so that the entries are ordered by the first contribution to any of those two repositories. I'd like to preserve that.

Cf. https://github.com/jquery/jquery/pull/4395/files#diff-b897336dda0579c40f3623b68eb6652ded79e206f4cf1af7ceb24b683fd46ff1 & https://github.com/jquery/jquery/pull/5113/files#diff-b897336dda0579c40f3623b68eb6652ded79e206f4cf1af7ceb24b683fd46ff1

timmywil: I see. Let's incorporate that into the authors script then.

timmywil: I've updated the authors script to clone the sizzle repo, combine the authors, sort by their first contribution, ensure they are unique by name, and then run cleanup so the sizzle repo doesn't stick around.

One case that I found interesting is Timo's entry. I noticed that his old email is still in the authors, whereas both emails were in there before. I think we only need the one and the email will automatically update when Timo updates the mailmap.

build/release/dist.mjs
timmywil: Removed inquirer and passed the blog URL from a workflow input.

package.json
```diff
@@ -2,12 +2,12 @@
   "name": "jquery",
   "title": "jQuery",
   "description": "JavaScript library for DOM operations",
-  "version": "3.7.1-pre",
+  "version": "3.7.0",
```
mgol, line 5: I am not a fan of this change. With it, the version in package.json can no longer be universally trusted. Also, unless this was addressed somewhere else, Git builds will contain an incorrect version in the comment header and in the `jquery` property.

mgol: Judging from the discussion at https://github.com/release-it/release-it/issues/946, it should be possible to keep our versioning strategy.

timmywil: > the version in package.json can no longer be universally trusted

It never really could with `-pre` either. It was often wrong about what version was coming next. `-pre` is a convention we invented that I don't think has a place anymore. The concern was that people would mistake some state of the repo with the exact state used to release the last version. But, the practice of leaving the version and continuing to develop has become so common I don't think anyone actually makes that mistake anymore. It's part of the language of maintaining semantically versioned repos.

In fact, because what we do is so foreign, it has gotten increasingly confusing. Keep in mind that changing to `-pre` obfuscates what the last version actually was. And the advantages of leaving the version alone have only been increasing. 

If `-pre` had caught on and become a standard, it might have been nice, but now we're the outliers.

> Git builds will contain an incorrect version in the comment header and in the jquery property.

Again, I don't the `-pre` version made sense either. As I explained in the issue description, git jQuery already has the SHA. All we'd need to do is remove the version from the header.

> Judging from the discussion at https://github.com/release-it/release-it/issues/946, it should be possible to keep our versioning strategy.

I'm not sure that issue is related. It's talking about alpha and beta releases, which actually work better when the version in the package.json doesn't have a custom suffix. release-it has a much easier time determining the next version (and give you alpha and beta versions options) when the package.json version is left alone, which I think helps illustrate my point about the strangeness of what we do.

mgol: OK, I see your points about `package.json`.

I still feel a bit uneasy about the value of the `jquery` property; it's true we add the commit hash to the Git version but AFAIK that doesn't happen for local builds from commits. Maybe we could tweak the build process to always read the current Git hash (instead of relying on the environment variable) and always appending it on non-tags?

mgol:
@timmywil are you targeting it at 4.0 or 3.x? I may not have time to do the full review before September 4 (it may happen but no promises).

timmywil:
@mgol I mostly wanted to see how hard it was so I hadn't decided yet, but I think I should release 3.7.1 and we can keep churning on this.

package.json
timmywil: I was actually thinking that as well. I'm also preparing a PR to migrate the build process off of grunt and I also like the idea of defaulting to adding the SHA.

timmywil:
Closing in favor of #5522

PR #5318, opened 2023-09-12
Migrate most grunt tasks off of grunt

timmywil:
### Summary ###

Migrates most grunt tasks to npm scripts.

While there are a lot of changes, the only change to the built jQuery file is the version string. See below for full details. 

I've tried to mostly keep the commits to one per task, but there are some early commits migrating the lint task in steps and some later commits to combine certain tasks (i.e. build, minify, dist, and compare_size) and simplify them. Here are some important things to notes:

- Added a new build command using [yargs](https://yargs.js.org/) to parse command line arguments and provide a pretty help command.
- minify and dist always run with a build.
- With `grunt npmcopy` removed, I committed the package-lock.json. We could actually commit a yarn.lock as well if we want. npm now updates both.
- The new watch task takes advantage of rollup's watcher directly. I've removed the copying functionality from the dist task, which now only fulfills its original purpose of making sure output is ascii-only and doesn't contain carriage returns. I think the copying was added very early on to copy builds to a dropbox folder to test things on jsfiddle. Fortunately, if that's still needed, the build command-and therefore the watch task-can write to any directory and multiple can be run at once.
- I've written a simplified version of compare_size that now lives in the jQuery repo and has no non-node dependencies other than `chalk` for pretty terminal colors. It uses [Node's zlib](https://nodejs.org/dist/latest-v18.x/docs/api/zlib.html) directly instead of gzip-js or another npm package. gzip-js hasn't been touched in years. Fortunately, zlib is easy-to-use. Sizes are a bit larger than before using the default options, but that should be okay as we're looking for comparisons, not completely accurate size predictions.
- I've moved the tests that don't require karma off of grunt. This includes the promises tests and node smoke tests.
- I removed the necessary grunt tasks all together in the last commit
- As I mentioned in the meeting, I've temporarily disabled an eslint-plugin-import rule (no-unused-imports) until the plugin [supports the new flat config format](https://github.com/import-js/eslint-plugin-import/issues/2556).
- the authors task, to replace grunt-git-authors, is already migrated in another PR: https://github.com/jquery/jquery/pull/5306

**Version string**

The version string now includes a short SHA when the build is not a release build (e.g. `4.0.0-pre+SHA`). Also, I made the version follow semver syntax. So, the slim version would now be `4.0.0+slim`. In a non-release build, that would look like `4.0.0-pre+slim.SHA`. Custom builds still add excluded/included modules to the version string (when not the slim build), but I didn't bother adding those in a way that follows semver syntax (e.g. `4.0.0-pre+a85a6aa2 -ajax,-ajax/binary,etc`). I'd like to remove `-pre` as well, but that change has already been made in my release PR (https://github.com/jquery/jquery/pull/5306).

### Checklist ###

## TODO

- [x] ~I see that yargs throws an error on load in Node 10.x. I'll need to think on the best way to fix this.~ Should be good now. Turns out `concurrently` has `yargs` as a dependency. yargs requires Node >=12, so it can't be used in the pretest script on Node 10.


package.json
```diff
@@ -30,6 +30,31 @@
...
+    "pretest": "npm run qunit-fixture && npm run babel:tests",
+    "qunit-fixture": "node build/tasks/qunit-fixture.js",
+    "start": "node -e \"require('./build/tasks/build.js').buildDefaultFiles({ watch: true })\"",
+    "test:browserless": "npm run test:node_smoke_tests && npm run test:promises_aplus && npm run test:jsdom",
+    "test:browser": "grunt karma:main",
+    "test:esmodules": "grunt karma:esmodules",
+    "test:jsdom": "grunt test:jsdom",
+    "test:no-deprecated": "npm run build -- -e deprecated && npm run test:browser",
+    "test:selector-native": "npm run build -- -e selector && npm run test:browser",
+    "test:slim": "npm run build -- --slim && npm run test:browser",
```
mgol, line 51: It's a bit weird that some of the `test:*` scripts do run the build and others do not. Previously it was more consistent - all of them run the build. Perhaps you can rename some of those to make it less confusing?

BTW, it's a common pattern that `npm it` (or, in its longer form, `npm install && npm test`) should work and test the package so if tests need a build, `npm test` should also build.

eslint.config.js
```diff
@@ -0,0 +1,316 @@
...
+			"Gruntfile.cjs",
+			"test/node_smoke_tests/commonjs/**",
+			"test/node_smoke_tests/module/**",
+			"test/promises_aplus_adapters/**",
+			"test/middleware-mockserver.cjs"
+		],
+		languageOptions: {
+			globals: {
+				...globals.node,
+				...globals.es2021
```
mgol, line 28: Can we do that? We still run CI on Node 10 on our rusty Jenkins...

This applies to multiple places.

eslint.config.js
```diff
@@ -0,0 +1,316 @@
...
+			globals: {
+				window: false
+			}
+		},
+		rules: {
+			...jqueryConfig.rules,
+			"import/extensions": [ "error", "always" ],
+			"import/no-cycle": "error",
+
+			// "import/no-unused-modules": [
```
mgol, line 53: Can you add a comment above this explaining shortly why it's commented out, maybe even linking to the upstream issue?

eslint.config.js
```diff
@@ -0,0 +1,316 @@
...
+		},
+		rules: {
+			...jqueryConfig.rules,
+			strict: [ "error", "global" ]
+		}
+	},
+
+	// Source
+	{
+		files: [ "src/**" ],
```
mgol, line 39: `main` has this comment for browser files above an empty `env`:
```js
	// The browser env is not enabled on purpose so that code takes
	// all browser-only globals from window instead of assuming
	// they're available as globals. This makes it possible to use
	// jQuery with tools like jsdom which provide a custom window
	// implementation.
    env: {},
```

I think it's worth preserving it somewhere here.

eslint.config.js
```diff
@@ -0,0 +1,316 @@
...
+			}
+		},
+		rules: {
+			"no-unused-vars": "off",
+			indent: [
+				"error",
+				"tab",
+				{
+
+					// Unlike other codes, "wrapper.js" is implemented in UMD.
```
mgol, line 123: It's already on `main` like that but I noticed this comment is not correct; `wrapper-esm.js` is not UMD (and the wrong file is referenced here anyway). It's just a wrapper which necessitates different rules but it's definitely not UMD (`wrapper.js` is not really a true UMD either, BTW 😉).

eslint.config.js
```diff
@@ -0,0 +1,316 @@
...
+			strict: [ "error", "global" ]
+		}
+	},
+
+	{
+		files: [
+			"dist/jquery.js",
+			"dist/jquery.slim.js",
+			"dist-module/jquery.module.js",
+			"dist-module/jquery.slim.module.js"
```
mgol, line 272: These two are repeated below. Also, I don't think you can group `dist` & dist-module` files since the latter ones are modules. The fact that it works suggests to me something's wrong about the config. Shouldn't we explicitly declare `sourceType` for all the file groups?

test/data/testinit-jsdom.js
```diff
@@ -1,4 +1,4 @@
-"use strict";
```
mgol, line 1: Why remove strict mode from this file?

test/data/testinit.js
```diff
@@ -1,6 +1,5 @@
 /* eslint no-multi-str: "off" */
 
-"use strict";
```
mgol, line 3: Why remove strict mode from this file?

test/data/testrunner.js
```diff
@@ -1,6 +1,5 @@
 ( function() {
 
-"use strict";
```
mgol, line 3: Why remove strict mode from this file?

eslint.config.js
```diff
@@ -0,0 +1,316 @@
...
+			"test/data/json_obj.js"
+		]
+	},
+
+	{
+		files: [
+			"eslint.config.js",
+			"Gruntfile.cjs",
+			"test/node_smoke_tests/commonjs/**",
+			"test/node_smoke_tests/module/**",
```
mgol, line 21: These also shouldn't be grouped, I think.

test/unit/callbacks.js
```diff
@@ -20,15 +20,15 @@ var output,
...
 	tests = {
-		"":                   "XABC   X     XABCABCC  X  XBB X   XABA  X   XX",
-		"once":               "XABC   X     X         X  X   X   XABA  X   XX",
-		"memory":             "XABC   XABC  XABCABCCC XA XBB XB  XABA  XC  XX",
-		"unique":             "XABC   X     XABCA     X  XBB X   XAB   X   X",
-		"stopOnFalse":        "XABC   X     XABCABCC  X  XBB X   XA    X   XX",
-		"once memory":        "XABC   XABC  X         XA X   XA  XABA  XC  XX",
-		"once unique":        "XABC   X     X         X  X   X   XAB   X   X",
-		"once stopOnFalse":   "XABC   X     X         X  X   X   XA    X   XX",
-		"memory unique":      "XABC   XA    XABCA     XA XBB XB  XAB   XC  X",
```
mgol, line 31: This used to be ASCII-art and is now gibberish. Can we make an exception here?

test/unit/core.js
```diff
@@ -38,19 +38,21 @@ QUnit.test( "jQuery()", function( assert ) {
 	// few here but beware of modular builds where these methods may be excluded.
 	if ( includesModule( "deprecated" ) ) {
 		expected++;
-		attrObj[ "click" ] = function() { assert.ok( exec, "Click executed." ); };
+		attrObj.click = function() {
+ assert.ok( exec, "Click executed." );
+};
```
mgol, line 43: Formatting here doesn't look right. A similar mistake is repeated in many, many places, please fix them all.

test/unit/core.js
```diff
@@ -584,20 +586,20 @@ QUnit.test( "inArray()", function( assert ) {
...
 	},
 	tests = {
-		p:    { elem: jQuery( "#ap" )[ 0 ],           index: 2 },
-		em:   { elem: jQuery( "#siblingfirst" )[ 0 ], index: 1 },
-		div:  { elem: jQuery( "#nothiddendiv" )[ 0 ], index: 1 },
-		a:    { elem: jQuery( "#simon1" )[ 0 ],       index: 3 }
+		p: { elem: jQuery( "#ap" )[ 0 ],           index: 2 },
+		em: { elem: jQuery( "#siblingfirst" )[ 0 ], index: 1 },
+		div: { elem: jQuery( "#nothiddendiv" )[ 0 ], index: 1 },
+		a: { elem: jQuery( "#simon1" )[ 0 ],       index: 3 }
```
mgol, line 599: Let's remove the extra spaces before `index`

test/unit/offset.js
```diff
@@ -165,9 +165,9 @@ testIframe( "absolute", "offset/absolute.html", function( assert, $ ) {
 
 	// get offset tests
 	tests = [
-		{ "id": "#absolute-1",     "top":  1, "left":  1 },
-		{ "id": "#absolute-1-1",   "top":  5, "left":  5 },
-		{ "id": "#absolute-1-1-1", "top":  9, "left":  9 },
+		{ "id": "#absolute-1",     "top": 1, "left": 1 },
+		{ "id": "#absolute-1-1",   "top": 5, "left": 5 },
+		{ "id": "#absolute-1-1-1", "top": 9, "left": 9 },
```
mgol, line 170: If you're removing extra spaces here, let's be consistent and remove them before `"top"` as well. A similar remark applies to many places below as well.

test/data/core/jquery-iterability-transpiled.html
```diff
@@ -3,7 +3,7 @@
 <head>
 	<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
 	<title>jQuery objects transpiled iterability test page</title>
-	<script src="../../../external/core-js-bundle/core-js-bundle.js"></script>
+	<script src="../../../node_modules/core-js-bundle/minified.js"></script>
```
mgol, line 6: [IMPORTANT] We cannot do this. I've actually made a similar changes some time ago and I had to back them out. We do not copy files from `node_modules` to TestSwarm, doing that is a security risk. We have to limit the included files to just the ones we need which means we do need the `external` folder and something similar to `grunt-npmcopy`.

Most of this commit needs to be reversed and we need a pure JS implementation of the `npmcopy` Grunt task.

Since further commits build upon this one, we need to audit the final state of the PR to make sure no `node_modules` references were left in in files accessed by TestSwarm.

test/middleware-mockserver.cjs
```diff
@@ -365,6 +385,7 @@ function MockserverMiddlewareFactory() {
 			return;
 		}
 
+		console.log( "Mock handling", req.method, parsed.href );
```
mgol, line 388: Should this be removed? It looks like it can get quite noisy.

test/server.js
```diff
@@ -0,0 +1,17 @@
...
+
+const nameHTML = fs.readFileSync( "./test/data/name.html", "utf8" );
+const app = express();
+
+app.use( mockServer() );
+
+app.post( "/test/data/name.html", function( _req, res ) {
+	res.send( nameHTML );
+} );
+app.use( express.static( "." ) );
```
mgol, line 13: Is there a way to limit this to a few directories we need to avoid the CodeQL warning?

build/tasks/lib/verifyNodeVersion.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+const { version } = require( "process" );
+const nodeV16OrNewer = !/^v1[0-5]\./.test( version );
+
+module.exports = function verifyNodeVersion() {
```
mgol, line 6: Too bad we still need to run on Node 10 or we could write ESM even for Node.js...

build/tasks/node_smoke_tests.js
```diff
@@ -0,0 +1,46 @@
...
+
+async function runTests() {
+	const promises = args.map( async( arg ) => {
+		const [ moduleType, module ] = arg.split( ":" );
+		if ( !allowedModules.includes( moduleType ) ) {
+			throw new Error(
+				`Usage: \`node_smoke_tests [${allowedModules.join( "|" )}]:JQUERY\``
+			);
+		}
+		const files = await fs.promises.readdir(
```
mgol: `fs.promises` is long deprecated, we should use `fs/promises` or `node:fs/promises`. If you check for Node version early enough, this should be fine.

test/node_smoke_tests/module/lib/jquery-module-specifier.js
```diff
@@ -5,6 +5,11 @@ const dirname = path.dirname( fileURLToPath( import.meta.url ) );
 
 const ROOT_DIR = path.resolve( dirname, "..", "..", "..", ".." );
 
+// import does not work with Windows-style paths
+function ensureUnixPath( path ) {
+	return path.replace( /^[a-z]:/i, "" ).split( "\\" ).join( "/" );
```
mgol, line 10: [Suggestion] You can do another replace to avoid a split & join:
```js
return path.replace( /^[a-z]:/i, "" ).replace( /\\\\/g, "/" );
```

Gruntfile.cjs
```diff
@@ -343,7 +342,7 @@ module.exports = function( grunt ) {
 	] );
 
 	grunt.registerTask( "test", [
-		"test:slow"
+		"test:jsom"
```
mgol, line 221: jsom?

build/command.js
```diff
@@ -0,0 +1,68 @@
+"use strict";
+
+const { build } = require( "./tasks/build" );
+const yargs = require( "yargs/yargs" );
+const slimExclude = require( "./tasks/lib/slim-exclude" );
+
+const argv = yargs( process.argv.slice( 2 ) )
```
mgol, line 7: Too bad we can't just use the native [`util.parseArgs` Node.js API](https://nodejs.org/api/util.html#utilparseargsconfig)...

mgol (changes requested):
Publishing what I have so far. Tomorrow, I'll resume with commit [`386acf4` (#5318)](https://github.com/jquery/jquery/pull/5318/commits/386acf4f639aee989c30c94d3d85f8dcc28e0d61)

test/node_smoke_tests/.eslintrc.cjs
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+module.exports = {
+	root: true,
+
+	extends: "../../.eslintrc-node.cjs",
+
+	parserOptions: {
+		ecmaVersion: 6
```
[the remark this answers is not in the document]
timmywil: I think you're looking at an old commit. eslint was migrated in stages. First off of grunt, but using the legacy config format. The next commit removed this file in favor of eslint.config.js. However, I'll use 2015 if that's equivalent.

.husky/commit-msg
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
timmywil: This is how it gets generated using the husky API. I think we should leave it and let husky manage it.

package.json
timmywil: I didn't like that `npm test` was building the same thing multiple times, so I've removed all the duplicates. However, I agree that `npm install && npm test` should work without any other commands, but I think it already does. Here's the order of operations for `npm it` as I've tried to write it:

1. npm install
2. npm test triggers the `pretest` script automatically, which builds the qunit fixture and transpiles the babel script.
3. npm test is run, which does the following scripts in order

1. `npm run build-all-variants` - builds all the normal jQuery files, including the new module.js files 
2. `npm run lint` - lints the repo, including the just-built dist files
3. `npm run test:browserless` - runs node smoke tests, promises tests, and jsdom tests, which requires all built files
4. `npm run test:browser` - main karma tests, which requires files already built with `build-all-variants`
5. `npm run test:esmodules` - same as (4), but the modules files instead
6. `npm run test:slim` - runs main karma tests against the slim build, which is built in the `test:slim` command so that jquery.[min].js is the slim build (not just jquery.slim.js)
7. `npm run test:no-deprecated` - runs main karma tests against a custom build of jquery.[min].js that removes the deprecated module
8. `npm run test:selector-native` - runs main karma tests against a custom build that replaces the selector engine with selector-native

I think this all works fine without further building.

eslint.config.js
timmywil: I can at least confirm that the build works and the tests pass on Node 10. If we need to fix any issues on the Jenkins server, perhaps we can address that after the merge.

eslint.config.js
timmywil: I've updated to a simple comment about the purpose of the line, so that it does not indent code within the wrapper.

eslint.config.js
timmywil: I considered that, but learned that the sourceType is inferred from the package.json. Specifically, `"type": "module"` set all js files to sourceType "module". So, we only need to specify `sourceType` when we want it to differ from the default. I actually meant to remove the repeated lines below this.

test/data/testinit-jsdom.js
timmywil: Thanks for catching that. I ran `eslint --fix` at some point and some accidental changes got past me.

eslint.config.js
timmywil: Again, I think we're good because of the package.json files.

test/unit/callbacks.js
timmywil: good point. I just need to figure out how to make that exception.

timmywil: got it. It's key-spacing.

test/unit/core.js
timmywil: I thought I got them all, but I think I only fixed them on my 3.x branch.

test/unit/offset.js
timmywil: Technically, the eslint rule only applies to space *after* the key.

test/data/core/jquery-iterability-transpiled.html
timmywil: Ok, I didn't know the Testswarm restriction. Unfortunate, but shouldn't take me long to make an npmcopy equivalent.

test/server.js
timmywil: I was thinking about that too. I think we can do each directory we need.

build/tasks/node_smoke_tests.js
timmywil: I actually didn't know that. Just thought it was an alias.

build/command.js
timmywil: That would be nice, but we would have recreate the help command.

test/unit/offset.js
timmywil: I removed the extra spacing on index from the above comment. It didn't seem to serve a purpose anyway as nothing was lined up. These I don't mind as much. More importantly, we don't have a rule set to catch this, so I'm reticent to change it without eslint verification.

build/tasks/node_smoke_tests.js
timmywil: Bah, apparently `fs/promises` is not supported in Node 10.

timmywil: Out of curiosity, can you point me to the deprecation notice? `fs.promises` is no longer in the docs, but I also can't find a notice saying it's deprecated (https://github.com/nodejs/node/blob/main/doc/api/deprecations.md).

test/server.js
```diff
@@ -0,0 +1,21 @@
...
+app.use( mockServer() );
+
+app.post( "/test/data/name.html", function( _req, res ) {
+	res.send( nameHTML );
+} );
+
+app.use( "/dist", express.static( "dist" ) );
+app.use( "/src", express.static( "src" ) );
+app.use( "/test", express.static( "test" ) );
+app.use( "/node_modules", express.static( "node_modules" ) );
```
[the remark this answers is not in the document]
timmywil: This should be fixed when I add npmcopy.

.husky/commit-msg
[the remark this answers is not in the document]
mgol: If it's generated this way, then OK

package.json
mgol: With the proposed setup, `npm run test:browser && npm run test:slim` may run tests against the regular build & the slim build... or it may not use the regular build - who knows what's in the built file at that point. If you run `npm run test:slim && npm run test:browser` then you'll run tests against the slim build twice.

I don't like this inconsistency, it makes it hard to remember what does what and we will need to run these npm scripts manually during local development a lot.

I know `npm test` builds jQuery multiple times on `main` but I left it that way to avoid the above issues. The full `npm test` is not run on CI anywhere anyway so this was just for running tests locally. The build, even repeated a few times, is still a tiny fraction of time compared to all the tests run.

I'm fine with removing these multiple re-builds but then I'd definitely want more consistency in npm scripts, the current proposed setup doesn't fulfill that desire. The mental overhead is real, I'd rather not have to remember all these inconsistencies.

eslint.config.js
mgol: We may not have issues now simply because we don't use any of these new globals. If we did, it would surely break on Jenkins. Can we lower that to `es2018` for now? That's what Node 10 supports.

eslint.config.js
mgol: Nice!

build/tasks/node_smoke_tests.js
mgol: They may not have had to deprecate formally as `fs.promises` existed while the API was experimental so they were free to do any changes. I wouldn't depend on undocumented APIs which it is.

It shouldn't matter that `fs/promises` is not supported in Node 10 as we already don't run Node smoke tests on Node 10 on `main`. From my comment above:
> If you check for Node version early enough, this should be fine.

One of the advantages of CommonJS vs. ESM. 😉

package.json
timmywil: I see your point. That's just not really been my experience. IMO, `npm test` is the only command we need to worry about when it comes to running tests out of context. The others are used individually when debugging their respective builds, but one at a time. In other words, I don't see a practical reason for needing `npm run test:slim && npm run test:browser` to work. I'm either debugging the slim build or the main one, and I'd say 99% of the time it's just the main one with a watch task, which is why I find the extra builds superfluous.

In fact, before this, I have found myself editing the build out of npm scripts locally. I found it easier than trying to remember which grunt test command it was. I'm not saying I can't be won over, but I can say I think the previous scripts have wasted more time than just the extra time the builds take during the main `test` script.

That said, if your personal experience has led you to wanting to add the build to the test scripts, let's braintstorm.

eslint.config.js
timmywil: `es2018` is not on the of the options. There's `es2015`, `es2017`, and `es2020`. But removing this line actually works here.

build/tasks/node_smoke_tests.js
timmywil: Yes, but I'm also using `fs.promises` in the new build script, which does run on Node 10. I'd say let's switch all of them once we can drop Node 10 (and Jenkins), which hopefully will be soon.

test/data/core/jquery-iterability-transpiled.html
timmywil: New npmcopy script is ready now.

package.json
timmywil: I thought about this more and tested how long the build takes. It's fine. We'll just have some extra builds. I think I prefer that to more npm scripts, which was the only other option I could think of. 

Also, to that end, I've consolidated the `node_smoke_tests` scripts into one and moved the file combinations to JS. I think that will be better for your upcoming PR as well.

build/tasks/minify.js
```diff
@@ -0,0 +1,65 @@
...
+	);
+
+	const minFilename = filename.replace( rjs, ".min.js" );
+	const mapFilename = filename.replace( rjs, ".min.map" );
+
+	const map = JSON.stringify( {
+		...JSON.parse( incompleteMap ),
+		file: minFilename,
+		sources: [ filename ]
+	} );
```
mgol, line 44: Please re-add the comment from:
https://github.com/jquery/jquery/blob/f75daab09102a4dd5107deadb55d4a169f86254a/Gruntfile.cjs#L349-L351
here.

build/tasks/compare_size.mjs
```diff
@@ -0,0 +1,132 @@
...
+	} catch ( err ) {
+		return {};
+	}
+}
+
+function saveCache( loc, cache ) {
+	return fs.promises.writeFile( loc, JSON.stringify( cache ) );
+}
+
+async function cleanWorkingDir() {
```
mgol, line 28: This suggests an action; maybe name it `isWorkingDirClean`?

build/tasks/compare_size.mjs
```diff
@@ -0,0 +1,132 @@
+import chalk from "chalk";
+import fs from "fs";
+import { promisify } from "util";
+import zlib from "node:zlib";
+import { exec as nodeExec } from "child_process";
```
mgol, line 5: If we already rely on modern features here, let's use the `node:` prefix for all core Node modules consistently (there should be an ESLint rule for this!):
```js
import fs from "node:fs";
import { promisify } from "node:util";
import zlib from "node:zlib";
import { exec as nodeExec } from "node:child_process";
```

build/tasks/compare_size.mjs
```diff
@@ -0,0 +1,132 @@
...
+	const branch = await getBranchName();
+	const sizeCache = await getCache( cache );
+
+	let rawPadLength = 0;
+	let gzPadLength = 0;
+	const results = await Promise.all(
+		files.map( async function( filename ) {
+			const contents = await fs.promises.readFile( filename, "utf8" );
+			const size = Buffer.byteLength( contents, "utf8" );
+			const gzippedSize = ( await gzip( contents ) ).length;
```
mgol, line 58: I like that we're using Node's native zlib here. I did try that in the past but got discouraged by different numbers reported than from our previous Grunt task. But you're right we should care more about relative comparisons than keeping extra low absolute numbers.

build/tasks/compare_size.mjs
```diff
@@ -0,0 +1,132 @@
...
+					} ) || {};
+					return (
+						compareSizes( result.raw, compareResult.raw, rawPadLength ) +
+						" " +
+						compareSizes( result.gz, compareResult.gz, gzPadLength ) +
+						" " +
+						filename
+					);
+				} ).join( "\n" );
+		} ),
```
mgol, line 100: I think it'd be easier to read if you extracted some of the inner computations to separate variables.

Also, I'd prefer to use template literals instead of concatenation.

eslint.config.js
```diff
@@ -262,6 +262,15 @@ export default [
 		}
 	},
 
+	{
+		files: [
+			"build/**/*.mjs"
+		],
+		languageOptions: {
+			sourceType: "module"
+		}
```
mgol, line 271: Is this necessary? Doesn't ESLint infer it from the extension?

README.md
```diff
@@ -47,41 +47,35 @@ How to build your own jQuery
 
 First, [clone the jQuery git repo](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/cloning-a-repository).
 
-Then, enter the jquery directory and run the build script:
+Then, enter the jquery directory, install dependencies, and run the build script:
+
 ```bash
-cd jquery && npm run build
+cd jquery
+npm run install
```
mgol, line 54: `npm install` is enough and more idiomatic.

README.md
```diff
@@ -47,41 +47,35 @@ How to build your own jQuery
...
+npm run build
 ```
-The built version of jQuery will be put in the `dist/` subdirectory, along with the minified copy and associated map file.
 
-If you want to create custom build or help with jQuery development, it would be better to install [grunt command line interface](https://github.com/gruntjs/grunt-cli) as a global package:
+The built version of jQuery will be placed in the `dist/` directory, along with a minified copy and associated map file.
```
mgol, line 58: It'd be better to point out here now that we generate multiple files now and some of them are in `dist-module`.

build/tasks/dist.js
```diff
@@ -1,42 +1,32 @@
...
-module.exports = async function dist( {
-	filename = "jquery.js",
-	folder = "dist"
-} = {} ) {
-	const distPaths = [
-		`${folder}/${filename}`,
-		`${folder}/${filename.replace( ".js", ".min.js" )}`,
-		`${folder}/${filename.replace( ".js", ".min.map" )}`
-	];
+// optionall copy them to another dir.
```
mgol, line 4: Typo: `optionally`

build/tasks/node_smoke_tests.js
```diff
@@ -19,26 +17,34 @@ if ( !verifyNodeVersion() ) {
...
+		runTests( "commonjs", "jquery" ),
+		runTests( "commonjs", "jquery/slim" ),
+		runTests( "commonjs", "./dist/jquery.js" ),
+		runTests( "commonjs", "./dist/jquery.slim.js" ),
+		runTests( "module", "jquery" ),
+		runTests( "module", "jquery/slim" ),
+		runTests( "module", "./dist-module/jquery.module.js" ),
+		runTests( "module", "./dist-module/jquery.slim.module.js" )
+	] );
+}
```
mgol, line 48: I like it. It reduces the noise in `package.json` and those tests are quick anwyay.

build/tasks/node_smoke_tests.js
mgol: OK, fair enough.

build/tasks/build.js
```diff
@@ -6,350 +6,349 @@
...
+			withFileTypes: true
+		} );
+	} catch ( e ) {
+		return all;
+	}
+	for ( const file of files ) {
+		if ( file.isDirectory() ) {
+			all = all.concat(
+				await readdirRecursive( path.join( dir, file.name ) )
+			);
```
mgol, line 65: Array spread is easier to read than `concat` and avoids `concat` issues with array detection changing the behavior:
```js
			all = [
				...all,
				...( await readdirRecursive( path.join( dir, file.name ) ) ),
			];
```

build/tasks/build.js
```diff
@@ -6,350 +6,349 @@
...
+		} else {
+			all.push( moduleName( path.join( dir, file.name ) ) );
```
mgol, line 67: This is modifying the input parameter, while in the `if` above we're not. I'd align here to avoid surprises; either always create a new one or push to the existing one.

I think I'd push for perf reasons as there may be a lot of file involved. That'd change the above recommendation from using array spread to `all.push( ...sth )`.

mgol (commented):
I reviewed the whole thing now. Nice work! I like a lot of the simplifications.

I haven't had a chance to test it locally, I'll do that next.

build/tasks/build.js
```diff
@@ -0,0 +1,278 @@
...
+	version
+} = {} ) {
+
+	// Add the short commit hash to the version string
+	// when the version is not for a release.
+	if ( !version ) {
+		const { stdout } = await exec( "git rev-parse --short HEAD" );
+
+		// "+[slim.]SHA" is semantically correct
+		version = `${pkg.version}+${slim ? "slim." : ""}${stdout.trim()}`;
```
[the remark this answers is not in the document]
timmywil: Good point. I've seen before some using a build identifier for a dirty state. We could do the same here and add `.dirty`.

build/tasks/compare_size.mjs
timmywil: I also normally prefer template literals, unless the line is very long like this one. Then you're only option for new lines is within the `${}`, which I don't think reads as easily. For instance, the lines above might look like this...

```js
return `${compareSizes( result.raw, compareResult.raw, rawPadLength )} ${
	compareSizes( result.gz, compareResult.gz, gzPadLength )
} ${filename}`;
```

And I don't think that looks as each expression getting its own line. It also takes longer to tell how much space is between each one.

timmywil: Ok, but if each of those expressions is in a variable, the one line wouldn't be too long.

eslint.config.js
timmywil: Yes and no. The above `build/**/*` overwrites the default to be commonjs, which all `.js` files are in the build folder unless they have the `.mjs` extension. The other way I was thinking to write this was to only apply the above fules only to `.js` and `.cjs` files, but then I'd have to copy all the other rules to `.mjs` in a separate rule. That said, I think I know a better way, which is to not set sourceType above, but change this rule to set `.js` files to `commonjs`.

README.md
timmywil: `npm run build` only generates one file and it is always placed in the `dist/` directory. `npm run build-all-variants` builds all of our release files, which passes `dir: 'dist-module'` to the build script for the module files.

That said, I've added a section on how to use `npm run build-all-variants` and its purpose.

timmywil:
I think I've covered all the remaining feedback. However, how would you feel about changing the name of `build-all-variants` to something shorter, like `build-all`?

timmywil:
I noticed an inconsistency between multiple compare_size runs and realized it's probably the version string varying by SHA. To fix it, compare_size now removes the banner for comparisons.

mgol:
> I think I've covered all the remaining feedback. However, how would you feel about changing the name of `build-all-variants` to something shorter, like `build:all`? I've added a separate commit to the end to see what it would look like.

Naming is hard. 🙂 I like `build:all`!

build/tasks/compare_size.mjs
```diff
@@ -0,0 +1,128 @@
...
+	const sizeCache = await getCache( cache );
+
+	let rawPadLength = 0;
+	let gzPadLength = 0;
+	const results = await Promise.all(
+		files.map( async function( filename ) {
+
+			let contents = await fs.promises.readFile( filename, "utf8" );
+
+			// Remove the banner for size comparisons.
```
mgol, line 55: It’s not just the banner, the `jquery` property varies as well.

However, I don’t think we should remove it. It’s what’s in the file and that’s what is getting served to the users; we should take that into account.

build/tasks/npmcopy.js
```diff
@@ -0,0 +1,40 @@
...
+	"qunit/LICENSE.txt": "qunit/LICENSE.txt",
+
+	"requirejs/require.js": "requirejs/require.js",
+
+	"sinon/sinon.js": "sinon/pkg/sinon.js",
+	"sinon/LICENSE.txt": "sinon/LICENSE"
+};
+
+async function npmcopy() {
+	await fs.promises.mkdir( path.resolve( __dirname, "..", "..", "external" ), {
```
mgol, line 27: It’d be easier to read & more portable if you defined root dir at the top in a similar way and then use it here & below instead of `__dirname` repeatedly. Especially once we migrate Node scripts to ESM as there’s no `__dirname` there.

README.md
```diff
@@ -97,97 +103,90 @@ Some example modules that can be excluded are:
...
-There is also a special alias to generate a build with the same configuration as the official jQuery Slim build is generated:
+*Non-official custom builds are not regularly tested. Use them at your own risk.*
+
+### Building All Default jQuery Files
+
+During a release, jQuery builds `jquery.js`, `jquery.slim.js`, `jquery.module.js`, and `jquery.slim.module.js` by default. To build all four files, run the following:
+
 ```bash
-grunt custom:slim
+npm run build:all
```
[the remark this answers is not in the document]
timmywil: I forgot I already documented it.

timmywil:
Remaining feedback addressed. As soon as we merge this, I'll start preparing the 3.x PR. I had already made 3.x branch, but I need to copy the changes made from PR feedback.

mgol (approved):
I need to test this locally but let's merge, any feedback for how things work we can address separately.

Issue #5327, opened 2023-09-18
New build system: re-add commit hashes in compare_size


mgol:
### Description ###

The new code for comparing size from #5318 removed commit hashes from being recorded. I found them useful in the past to be able to quickly verify whether particular size data is up to date; I'd like to have them back.

Now, since we've changed the size cache format anyway, we can think about the best one here. Right now, it's:
```json
{
  "main": {
    "dist/jquery.min.js": {
      "raw": 78901,
      "gz": 27648
    },
    "dist/jquery.slim.min.js": {
      "raw": 56222,
      "gz": 19605
    },
    "dist-module/jquery.module.min.js": {
      "raw": 78947,
      "gz": 27669
    },
    "dist-module/jquery.slim.module.min.js": {
      "raw": 56268,
      "gz": 19626
    }
  },
  /* ... */
}
```
The old format had something like the following at the top:
```json
{
  "": {
    "version": 0.4,
    "tips": {
      "main": "2bdecf8b7bd10864e5337a4e24e39476c78cf23a",
      /* ... */
    }
  },
  /* ... */
}
```
That made commit hashes show up far away from the data which also makes it harder to do manual cleanups. How about something like:
```json
{
  "main": {
    "meta": {
      "commit": "2bdecf8b7bd10864e5337a4e24e39476c78cf23a"
    },
    "data": {
      "dist/jquery.min.js": {
        "raw": 78901,
        "gz": 27648
      },
      "dist/jquery.slim.min.js": {
        "raw": 56222,
        "gz": 19605
      },
      "dist-module/jquery.module.min.js": {
        "raw": 78947,
        "gz": 27669
      },
      "dist-module/jquery.slim.module.min.js": {
        "raw": 56268,
        "gz": 19626
      }
    }
  },
  /* ... */
}
```
?

### Link to test case ###

N/A

mgol:
@timmywil another feature that was dropped in the rewrite that I found useful - always recording the last run data and printing the comparison to it at the end. Helps immensely when testing size reductions locally before pushing. Can you re-add it as well?

timmywil:
Sure, that should be easy enough.

timmywil:
I like your format suggestion, but I'd replace "data" with "files" or "sizes".

mgol:
@timmywil this has been handled by 09972bcc680e89e38f56d83043bb368eb7fbda91, hasn't it?

timmywil:
@mgol Yup!

PR #5329, opened 2023-09-20
Build: add commit SHAs and last runs to comparisons

timmywil:
Fixes gh-5327

### Summary ###

- [BREAKING] changes the format of .sizecache.json. Delete your local .sizecache.json before running comparisons.

```json
{
  "branch": {
    "meta": {
      "commit": "87de182dada4df850a64f3f062fe656b7d589ed9"
    },
    "files": {
      "dist/jquery.min.js": {
        "raw": 79053,
        "gz": 27701
      }
    }
  }
}
```

- The commit SHA is now saved for branches and printed with comparisons in gray.
- every run is now saved as `" last run"`. The leading space is intentional to avoid any possible collisions.
- compare_size was removed the banner for comparisons, but @mgol made me realize all the exclusions can make a noticeable difference in the size so should be included. However, the short SHA varies by commit and we'd like to be able to compare exactly across commits. Also, `.dirty` is removed for comparisons because we'd like to be able to compare local changes (i.e. dirty git directory") to all branches and last runs that were clean.
- Added a version number check that will automatically rewrite the cache on version mismatch.

I can make another PR if preferred, but I tacked on another change here to add the factory files that were just added to the main branch to the dist eslint.

### Checklist ###

### DIFF
--- a/build/tasks/compare_size.mjs
+++ b/build/tasks/compare_size.mjs
@@ -52,9 +52,9 @@ export async function compareSize( { cache = ".sizecache.json", files } = {} ) {
 
 			let contents = await fs.promises.readFile( filename, "utf8" );
 
-			// Remove the banner for size comparisons.
-			// The version string can vary widely by short SHA.
-			contents = contents.replace( /\/\*\! jQuery[^\n]+/, "" );
+			// Remove the short SHA and .dirty from comparisons
+			const sha = /\/\*\! jQuery v\d+.\d+.\d+(?:-\w+)?\+(?:slim.)?(\w{8}(?:\.dirty)?)/.exec( contents )[ 1 ];
+			contents = contents.replace( new RegExp( sha, "g" ), "" );
 
 			const size = Buffer.byteLength( contents, "utf8" );
 			const gzippedSize = ( await gzip( contents ) ).length;
--- a/build/tasks/compare_size.mjs
+++ b/build/tasks/compare_size.mjs
@@ -13,6 +13,11 @@ async function getBranchName() {
 	return stdout.trim();
 }
 
+async function getCommitHash() {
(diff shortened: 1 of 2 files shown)


build/tasks/compare_size.mjs
```diff
@@ -22,6 +37,17 @@ async function getCache( loc ) {
 	}
 }
 
+function cacheResults( results ) {
+	const files = {};
```
mgol, line 41: It’s a bit safer to do:
```suggestion
	const files = Object.create( null );
```
to avoid weird behavior with a possible `__proto__` branch.

mgol (commented):
One remark

timmywil:
Rather than making us manually delete the cache if the format changes, I've added a version number check, which will automatically rewrite the cache on version mismatch.

build/tasks/compare_size.mjs
```diff
@@ -142,7 +152,9 @@ export async function compareSize( { cache = ".sizecache.json", files } = {} ) {
 	console.log( output );
 
 	// Always save the last run
+	// Save version under last run
 	sizeCache[ " last run" ] = {
+		meta: { version: VERSION },
```
mgol, line 157: [Suggestion] It's a bit weird that the version of the whole schema is under the ` last run` key instead of in a global `meta` section. You may consider moving that.

mgol (approved):
One suggestion. LGTM overall, though.

build/tasks/compare_size.mjs
timmywil: I considered it, but last run always gets saved and this is just a little smaller, saving another object. Practically, I think the difference is minimal.

mgol: OK, this was just a suggestion, I don't insist. 🙂

PR #5332, opened 2023-09-20
Build: fix inconsistent builds in Node 20

timmywil:
### Summary ###

- one fileOverrides per build
- only run the lint build when running lint

### Checklist ###

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -70,12 +70,8 @@ jobs:
       run: npx playwright-webkit install-deps
       if: matrix.NPM_SCRIPT == 'test:browser' && contains(matrix.BROWSERS, 'WebkitHeadless')
 
-    - name: Build jQuery for Lint
-      run: npm run build:all
-      if: matrix.NPM_SCRIPT == 'test:browserless'
-
     - name: Lint code
-      run: npm run lint
+      run: npm run build:all && npm run lint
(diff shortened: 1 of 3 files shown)


build/tasks/build.js
```diff
@@ -326,6 +321,11 @@ async function build( {
 
 		return watcher;
 	} else {
+		const bundle = await rollup.rollup( {
+			...inputOptions,
+			plugins: [ rollupFileOverrides( fileOverrides ) ]
+		} );
+
```
timmywil, line 328: I moved this here as it's not needed if watching.

.github/workflows/node.js.yml
```diff
@@ -70,12 +70,8 @@ jobs:
       run: npx playwright-webkit install-deps
       if: matrix.NPM_SCRIPT == 'test:browser' && contains(matrix.BROWSERS, 'WebkitHeadless')
 
-    - name: Build jQuery for Lint
-      run: npm run build:all
-      if: matrix.NPM_SCRIPT == 'test:browserless'
-
     - name: Lint code
-      run: npm run lint
+      run: npm run build:all && npm run lint
```
timmywil, line 74: The removed build above was only the first build on Node 20, but I think any of them could have failed. But, I realized lint only runs on Node 18.x, so I moved it there.

mgol (approved):

PR #5333, opened 2023-09-21
Build: sort branches in compare_size; last run last

timmywil:
### Summary ###

Sorts the branches and keep "last run" last.

### Checklist ###

### DIFF
--- a/build/tasks/compare_size.mjs
+++ b/build/tasks/compare_size.mjs
@@ -81,7 +81,13 @@ function sortBranches( a, b ) {
 	if ( b === lastRunBranch ) {
 		return -1;
 	}
-	return a.localeCompare( b );
+	if ( a < b ) {
+		return -1;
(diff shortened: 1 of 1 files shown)


build/tasks/compare_size.mjs
```diff
@@ -73,6 +74,16 @@ function compareSizes( existing, current, padLength ) {
...
 }
 
+function sortBranches( a, b ) {
+	if ( a === lastRunBranch ) {
+		return 1;
+	}
+	if ( b === lastRunBranch ) {
+		return -1;
+	}
+	return a.localeCompare( b );
```
mgol, line 84: `localeCompare` may not be stable across environments. Adding two more checks for `>` & `<` would be a bit safer.

mgol (approved):
LGTM but I do have one suggestion.

PR #5335, opened 2023-09-21
refactor: an identical and smaller installation of 4.0.0-pre

jxshin:
### Summary ###

I developed a tool which can generate a new `package-lock.json` file base on the original one.
With this new file, you can have an identical and smaller installation.

The equivalence can be verified by the output of `npm ls --all` for the installations with the original and new files.

For jquery 4.0.0-pre, the number of installed package-versions decreases from 704 to 673, and the size of `node_modules`  folder decreases from 275MB to 272MB.

I am going to release this tool after obtaining some feedbacks. Thank you!

### Checklist ###

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -114,18 +114,6 @@
         "url": "https://github.com/sponsors/isaacs"
       }
     },
-    "node_modules/@babel/cli/node_modules/minimatch": {
-      "version": "3.1.2",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.2.tgz",
-      "integrity": "sha512-J7p63hRiAjw1NDEww1W7i37+ByIrOWO5XQQAzZ3VOcL0PNybwpfmV/N05zFAzwQ9USyEcX6t3UO+K5aqBQOIHw==",
-      "dev": true,
-      "dependencies": {
-        "brace-expansion": "^1.1.7"
-      },
-      "engines": {
-        "node": "*"
-      }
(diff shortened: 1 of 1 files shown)


mgol:
Thanks for the PR. However:
1. We added a lockfile for better security of our installations. I prefer it comes from npm, not from a community tool we have no control over; that reduces security of the whole thing.
2. The gains are not that substantial, certainly not to counter the previous point.

I don't think we want to do this.

mgol:
Thanks for the PR. However:
1. We added a lockfile for better security of our installations. I prefer it comes from npm, not from a community tool we have no control over; that reduces security of the whole thing.
2. The gains are not that substantial, certainly not to counter the previous point.
3. We're updating dependencies from time to time; sometimes manually, sometimes via dependabot PRs. Anything that requires frequent manual calls of an external tool would just make this whole process unnecessarily complex.

I don't think we want to do this.

jxshin:
> Thanks for the PR. However:
> 
> 1. We added a lockfile for better security of our installations. I prefer it comes from npm, not from a community tool we have no control over; that reduces security of the whole thing.
> 2. The gains are not that substantial, certainly not to counter the previous point.
> 3. We're updating dependencies from time to time; sometimes manually, sometimes via dependabot PRs. Anything that requires frequent manual calls of an external tool would just make this whole process unnecessarily complex.
> 
> I don't think we want to do this.

@mgol  Thanks for your constructive feedback.

PR #5338, opened 2023-09-27
Build: Run pretest before test:* npm scripts

mgol:
### Summary ###

Build was already happening in scripts like `test:browser` but those scripts were missing `pretest`, meaning that running `npm install && npm test:browser` may have failed if `pretest` wasn't run before or if its results were out of date.

Even worse, with such stale data some tests may erroneously succeed.

This also removes a separate `pretest` step from GitHub Actions as it's no longer needed.

### Checklist ###


timmywil (approved):
I suppose we were always moving in this direction.

mgol:
Landed on `main` in 1ad66aeb6d7d94f8e4c8e2286569722ca41f9868 and on `3.x-stable` in cd27f306621d88882df003a9bd32391a3bb95ef3.

PR #5364, opened 2023-11-13
Build: Reformat GitHub workflow Yaml files

mgol:
### Summary ###

Use Prettier 3.1.0 to reformat the Yaml files. This makes their format identical to the one used on `3.x-stable`, making for much easier cherry-picks.

The main difference is the list under `steps:` was not indented while all other lists were.

Running the same version of Prettier on `3.x-stable` Yaml files results in no changes.

### Checklist ###


timmywil (approved):
I like it. I think my editor did this in my release PR.

PR #5383, opened 2024-01-03
Selector: Eliminate `selector.js` depenencies from various modules

mgol:
### Summary ###

There are two main reasons for why some of those dependencies are no longer needed:
1. `jQuery.contains` which is now a part of `core`.
2. `jQuery.find.attr` no longer exists, native `getAttribute` is used instead.

### Checklist ###


gibson042 (approved):

mgol:
This change should not be controversial so I'm going to merge it with 1 approval.

Issue #5379, opened 2023-12-29
Strange error behavior of event delegation when customizing jQuery


ajiho:
### Description ###
My build script is as follows:

```
npm run build -- --esm --include=event --include=traversing
```
When I use event delegation

html
```
<button data-test1="test1" data-test2="test2">test</button>
```

```
$(document).on('click', '[data-test1][data-test2]', function () {
    //error: Uncaught TypeError: jQuery.attr is not a function
});
```

According to the description of readme.md,**selector: The full jQuery selector engine. When this module is excluded, it is replaced with a rudimentary selector engine based on the browser's querySelectorAll**,However, querySelectorAll supports this type of selector
I would like to ask if this design is intentional or is it a flaw?

By the way, I have already tested it in the full version and there is no such issue

Thank you.

mgol:
Thanks for the report. This is on the `main` branch, isn't it? That's code for future jQuery 4.0 which is not stable yet. When building for production use, it's recommended to check out the latest tag and build there.

In jQuery 3.x, the `ATTR` filter uses `find.attr`:
https://github.com/jquery/jquery/blob/87467a6f62b5fbd820ab387836e2a6fb186cbc1b/src/selector.js#L1057
defined in the same file:
https://github.com/jquery/jquery/blob/87467a6f62b5fbd820ab387836e2a6fb186cbc1b/src/selector.js#L853-L876

On the `main` branch, I removed the selector version and started using `jQuery.attr`:
https://github.com/jquery/jquery/blob/c98597eaf5e144ee5e549cb41984687cd1033068/src/selector.js#L454
which lies outside of this module and this dependency was not made explicit.

Depending on `attr.js` is a bit hard because that module depends on `selector` itself:
https://github.com/jquery/jquery/blob/c98597eaf5e144ee5e549cb41984687cd1033068/src/attributes/attr.js#L7
but that's mostly an artifact from jQuery 3.x previously depending on `jQuery.find.attr` from the selector module here:
https://github.com/jquery/jquery/blob/87467a6f62b5fbd820ab387836e2a6fb186cbc1b/src/attributes/attr.js#L68
while on the `main` branch a native `getAttribute` is used:
https://github.com/jquery/jquery/blob/c98597eaf5e144ee5e549cb41984687cd1033068/src/attributes/attr.js#L61

We should remove the dependency on `selector.js` from `attr.js`. As for fixing the selector issue, we have then two options:
[...]

PR #5384, opened 2024-01-04
Selector: Make `selector.js` module depend on `attributes/attr.js`

mgol:
### Summary ###

This fixes custom builds using the `--include` switch that don't include the `attributes` module.

Fixes gh-5379

### Checklist ###

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -97,6 +97,11 @@ if ( isIE ) {
 	};
 }
 
+// HTML boolean attributes have special behavior:
+// we consider the lowercase name to be the only valid value, so
+// getting (if the attribute is present) normalizes to that, as does
+// setting to any non-`false` value (and setting to `false` removes the attribute).
+// See https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes
 jQuery.each( (
 	"checked selected async autofocus autoplay controls defer disabled " +
 	"hidden ismap loop multiple open readonly required scoped"
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -97,7 +97,10 @@ if ( isIE ) {
 	};
 }
 
-jQuery.each( jQuery.expr.match.bool.source.match( /\w+/g ), function( _i, name ) {
+jQuery.each( (
+	"checked selected async autofocus autoplay controls defer disabled " +
+	"hidden ismap loop multiple open readonly required scoped"
+).split( " " ), function( _i, name ) {
 	jQuery.attrHooks[ name ] = {
 		get: function( elem ) {
 			var ret,
--- a/src/selector-native.js
+++ b/src/selector-native.js
@@ -25,7 +25,6 @@ import { jQuery } from "./core.js";
 import { document } from "./var/document.js";
 import { whitespace } from "./var/whitespace.js";
 import { isIE } from "./var/isIE.js";
-import { booleans } from "./selector/var/booleans.js";
 import { rleadingCombinator } from "./selector/var/rleadingCombinator.js";
 import { rdescend } from "./selector/var/rdescend.js";
 import { rsibling } from "./selector/var/rsibling.js";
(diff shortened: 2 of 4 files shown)


mgol:
The size report for this is pretty surprising...
```
main @e8b7db4b0f1e1b8e08578641b30a92b955ccc4ec
   raw     gz Filename
   -51    +51 dist/jquery.min.js
   -47     +8 dist/jquery.slim.min.js
   -51    +51 dist-module/jquery.module.min.js
   -47    +10 dist-module/jquery.slim.module.min.js
```
I could use some gzip magic from @gibson042 here. 😉

mgol:
After the latest changes:
```
main @e8b7db4b0f1e1b8e08578641b30a92b955ccc4ec
   raw     gz Filename
   -65    +39 dist/jquery.min.js
   -61     -8 dist/jquery.slim.min.js
   -65    +39 dist-module/jquery.module.min.js
   -61     -7 dist-module/jquery.slim.module.min.js
```
`+39 bytes` is still significant for something that in theory should remove bytes.

mgol:
After the latest changes:
```
main @e8b7db4b0f1e1b8e08578641b30a92b955ccc4ec
   raw     gz Filename
   -68    +37 dist/jquery.min.js
   -64    -10 dist/jquery.slim.min.js
   -68    +37 dist-module/jquery.module.min.js
   -64     -8 dist-module/jquery.slim.module.min.js
```
`+37 bytes` is still significant for something that in theory should remove bytes.

mgol:
OK, without the addition of a dependency on `attributes/attr.js` to `selector.js` the size difference is as follows:
```
main @e8b7db4b0f1e1b8e08578641b30a92b955ccc4ec
   raw     gz Filename
   -51    -13 dist/jquery.min.js
   -51    -25 dist/jquery.slim.min.js
   -51    -16 dist-module/jquery.module.min.js
   -51    -25 dist-module/jquery.slim.module.min.js
```
so it looks this part is to blame.

timmywil (approved):
LGTM

src/attributes/attr.js
```diff
@@ -97,7 +97,10 @@ if ( isIE ) {
 	};
 }
 
-jQuery.each( jQuery.expr.match.bool.source.match( /\w+/g ), function( _i, name ) {
+jQuery.each( (
+	"checked selected async autofocus autoplay controls defer disabled " +
+	"hidden ismap loop multiple open readonly required scoped"
+).split( " " ), function( _i, name ) {
```
gibson042, line 103:
```suggestion
// HTML boolean attributes have special behavior:
// we consider the lowercase name to be the only valid value, so
// getting (if the attribute is present) normalizes to that, as does
// setting to any non-`false` value (and setting to `false` removes the attribute).
// See https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes
jQuery.each( (
	"checked selected async autofocus autoplay controls defer disabled " +
	"hidden ismap loop multiple open readonly required scoped"
).split( " " ), function( _i, name ) {
```

Note also that our list is outdated; it is missing several boolean attributes from [HTML Attributes](https://html.spec.whatwg.org/multipage/indices.html#attributes-3), includes a `scoped` that isn't even present there, and treats [`hidden`](https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute) as boolean despite the expansion of that attribute to be an enumeration including "until-found" by https://github.com/whatwg/html/commit/f5def65bda4d01a6a1e105fcf13fe45201b4f48a .

gibson042 (approved):

src/attributes/attr.js
mgol: Oh, interesting. Perhaps we should update this list for 4.0 as this can be considered a breaking change.

Or we should get rid of this manual normalization, depending on users doing the right thing here. The biggest issue are possibly falsy checks on values since an empty string is also a valid value. If we're worried about compat for such things, my guess would be we could limit this list to just a few entries like `checked` & `selected`. Thoughts, @timmywil @gibson042?

mgol: I extracted the issue about the list itself into https://github.com/jquery/jquery/issues/5388; let's limit this PR to just the issue at hand and keep the list the same here.

mgol:
Note: as of this change, `jQuery.expr.match.bool` is no longer defined.

PR #5387, opened 2024-01-09
Update attr.js

Siss3l:
fix: Added few protections against prototype pollution

### Summary ###
Checking the variable types used in the `attr` method that could lead possibly to some problematics.

### Checklist ###
* [ ] Better  Sanitization would be needed

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -18,8 +18,7 @@ jQuery.fn.extend( {
 
 jQuery.extend( {
 	attr: function( elem, name, value ) {
-		var ret, hooks,
-			nType = elem.nodeType;
+		var ret, hooks, nType = elem.nodeType;
 
 		// Don't get/set attributes on text, comment and attribute nodes
 		if ( nType === 3 || nType === 8 || nType === 2 ) {
@@ -43,9 +42,17 @@ jQuery.extend( {
 				return;
 			}
 
-			if ( hooks && "set" in hooks &&
-				( ret = hooks.set( elem, value, name ) ) !== undefined ) {
+			if ( hooks && "set" in hooks && ( ret = hooks.set( elem, value, name ) ) !== undefined ) {
 				return ret;
+			}
+
+			if ( typeof name !== "string" || typeof name === "undefined" ) { // Optimizable
+				return;
+			}
+
+			// Better to be sanitized (Content Security Policy followed) with DOMPurify or HTML Sanitizer API (experimental technology) somewhere
+			if ( hooks instanceof Object || value instanceof Object ) { // To prevent some prototype pollution on "srcdoc" etc
+				return;
 			}
 
 			elem.setAttribute( name, value );


src/attributes/attr.js
```diff
@@ -43,11 +42,19 @@ jQuery.extend( {
...
+			if ( hooks && "set" in hooks && ( ret = hooks.set( elem, value, name ) ) !== undefined ) {
 				return ret;
 			}
 
+			if ( typeof name !== "string" || typeof name === "undefined" ) { // Optimizable
+				return;
+			}
+
+			// Better to be sanitized (Content Security Policy followed) with DOMPurify or HTML Sanitizer API (experimental technology) somewhere
+			if ( hooks instanceof Object || value instanceof Object ) { // To prevent some prototype pollution on "srcdoc" etc
```
mgol, line 54: I don't understand the comment, can you elaborate?

mgol (changes requested):
Thanks for the PR.

I don't think being careful about members of `jQuery.attrHooks` is needed; this is API open for extensions by the users and you can do all sorts of bad things there without the need to mess with prototypes.

We should perhaps make `jQuery.attrHooks` itself initialized as `Object.create( null )` instead of `{}` but that's a separate issue.

Also, all changes like that need tests.

src/attributes/attr.js
Siss3l: Hello and thanks for your consideration,
It is a bit difficult to reproduce it, so I have tried to provide an atypical proof-of-concept script to make it clearer:

```py
from flask import Flask, Response
app = Flask(__name__)

@app.route("/", methods=["GET"]) # CORS Headers
def home():
    htm = """<!DOCTYPE html><html lang="en">
          <body>
            <!-- With "__proto__.srcdoc" => Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'set' in <svg/onload=alert(document.cookie)> -->
            <form id=poc><input name="__proto__.srcdoc[]" value="<svg/onload=alert(document.cookie)>"></form>
            <iframe id="test"></iframe>
            <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.7.1/jquery.js"></script>
            <script src="https://cdnjs.cloudflare.com/ajax/libs/axios/1.6.3/axios.js"></script>
            <script>(() => {
                axios.post("http://localhost:3000/poc",
                  $("#poc").get(0), {"headers":{"Content-Type":"application/json"}}).then((k) => {
                    $("#test").attr({"src": k.data.test});
                });
            })();</script>
          </body>
        </html>""".strip()
    return Response(htm, content_type="text/html")

@app.route("/poc", methods=["POST"])
def poc():
    return Response("{}", content_type="application/json")

if __name__ == "__main__": app.run(debug=0, host="localhost", port=3000)
```

When making an `Axios` POST request (with/out sanitization), the way `jQuery` handles `get/attr` permit to add some problematic attribute causing a TypeError of `Cannot use 'in' operator to search for 'set'` if we use something like `__proto__.srcdoc` value instead.

You are indeed right that this isn't directly linked to `attrHooks` maybe, but in my honest opinion (reading the code) I thought it would be useful in most cases to use more additional (type, typeof, instanceof, `__proto__`, etc.) checks here?

Even more as you mentioned to possibly change the current initialization behavior which perhaps should have been better to be an issue first!

I will make the unit tests as soon as possible,
Best regards.

mgol:
This got a bit stale, I'm closing the PR for now to get a queue smaller.

@Siss3l if you want to proceed, please submit an issue explaining the problem with a concrete code example, ideally one we can run on a platform like JS Bin. Once we look at it & confirm it's something worth fixing, you'll be able to proceed with the PR - after adding tests.

PR #5393, opened 2024-01-20
Build: add --reset-compare and --clear-compare options to build command

timmywil:
### Summary ###

This adds 2 options to the build command.

My compare-size cache was getting a little big with all my branches. Rather than manually editing or deleting the file, I would prefer a command that kept the `main` branch and `last run` in there (`npm run build -- --reset-compare`).

I also added `--clear-compare` in case there's ever a need for starting over completely. It just deletes the cache file.

### Checklist ###

### DIFF
--- a/build/command.js
+++ b/build/command.js
@@ -13,8 +13,7 @@ const argv = yargs( process.argv.slice( 2 ) )
 	.option( "filename", {
 		alias: "f",
 		type: "string",
-		description:
-			"Set the filename of the built file. Defaults to jquery.js."
+		description: "Set the filename of the built file. Defaults to jquery.js."
 	} )
 	.option( "dir", {
 		alias: "d",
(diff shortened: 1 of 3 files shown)


build/command.js
```diff
@@ -61,21 +59,30 @@ const argv = yargs( process.argv.slice( 2 ) )
...
-			"By default, a UMD bundle is built."
+			"Build the factory bundle. By default, a UMD bundle is built."
 	} )
 	.option( "slim", {
 		alias: "s",
 		type: "boolean",
-		description: "Build a slim bundle, which excludes " +
-			slimExclude.join( ", " )
+		description:
+			"Build a slim bundle, which excludes " + slimExclude.join( ", " )
```
timmywil, line 68: The changes above were from running prettier over the file once, which I do sometimes to get line length right.

timmywil:
Sure, that can be configured

timmywil:
Looking at this again, I think I'll just write to the cache in prettier JSON. Then we can easily edit the cache manually if it gets out of hand.

PR #5395, opened 2024-01-22
Build: migrate grunt authors to a custom script

timmywil:
- the new script pulls all authors from the Sizzle repo
- added temporary grunt task for releases

### Summary ###

I started putting together the release and ran into a stale authors list. I then ran `grunt update-authors`, but forgot that we want to continue including Sizzle authors and I already wrote a script for that. This pulls that script from #5306.

I've also added a grunt task that uses the updated script. Since we're not using the updated release process just yet, we need that for the jquery-release script. The grunt task logs all authors to console and jquery-release does its own check.

### Checklist ###

### DIFF
--- a/Gruntfile.cjs
+++ b/Gruntfile.cjs
@@ -209,6 +209,14 @@ module.exports = function( grunt ) {
 		grunt.log.writeln( "Node.js 17 or newer detected, skipping jsdom tests..." );
 	} );
 
+	grunt.registerTask( "authors", async function() {
+		const done = this.async();
+		const { getAuthors } = require( "./build/release/authors.js" );
+		const authors = await getAuthors();
+		console.log( authors.join( "\n" ) );
+		done();
+	} );
+
 	grunt.registerTask( "test:jsdom", [
 
 		// Support: Node.js 17+
--- a/build/release/authors.js
+++ b/build/release/authors.js
@@ -1,0 +1,105 @@
+"use strict";
+
+const fs = require( "node:fs" );
+const util = require( "node:util" );
+const exec = util.promisify( require( "node:child_process" ).exec );
+const rnewline = /\r?\n/;
+const rdate = /^\[(\d+)\] /;
+
+const ignore = [
+	/dependabot\[bot\]/
+];
+
+function compareAuthors( a, b ) {
+	const aName = a.replace( rdate, "" ).replace( / <.*>/, "" );
+	const bName = b.replace( rdate, "" ).replace( / <.*>/, "" );
+	return aName === bName;
+}
+
+function uniq( arr ) {
+	const unique = [];
+	for ( const item of arr ) {
+		if ( ignore.some( re => re.test( item ) ) ) {
+			continue;
+		}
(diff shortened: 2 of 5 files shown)


Gruntfile.cjs
```diff
@@ -209,6 +209,14 @@ module.exports = function( grunt ) {
 		grunt.log.writeln( "Node.js 17 or newer detected, skipping jsdom tests..." );
 	} );
 
+	grunt.registerTask( "authors", async function() {
+		const done = this.async();
+		const { getAuthors } = require( "./build/release/authors.js" );
+		const authors = await getAuthors();
+		console.log( authors.join( "\n" ) );
+		done();
```
mgol, line 217: I like this pattern of using an async function and calling `done()` in a way that looks synchronous. 🙂

build/release/authors.js
```diff
@@ -0,0 +1,105 @@
...
+	return exec( "npx rimraf .sizzle" );
+}
+
+function cloneSizzle() {
+	console.log( "Cloning Sizzle..." );
+	return exec( "git clone https://github.com/jquery/sizzle .sizzle" );
+}
+
+async function getLastAuthor() {
+	const authorsTxt = await fs.promises.readFile( "AUTHORS.txt", "utf8" );
```
mgol, line 43: `fs.promises` is deprecated, the new way to import it is via `node:fs/promises`. I guess the problem is the latter doesn't work on Node 10 but we are not going to run this task on our rusty old Jenkins, are we? Let's use the official API then.

build/release/authors.js
```diff
@@ -0,0 +1,105 @@
...
+	return authorsTxt.trim().split( rnewline ).pop();
+}
+
+async function logAuthors( preCommand ) {
+	let command = "git log --pretty=format:\"[%at] %aN <%aE>\"";
+	if ( preCommand ) {
+		command = preCommand + " && " + command;
```
mgol, line 50: [Suggestion] I prefer template literals:
```suggestion
		command = `${ preCommand } && ${ command }`;
```

build/release/authors.js
```diff
@@ -0,0 +1,105 @@
...
+	await cloneSizzle();
+	const authors = await logAuthors( "cd .sizzle" );
+	await cleanupSizzle();
+	return authors;
+}
+
+function sortAuthors( a, b ) {
+	const [ , aDate ] = rdate.exec( a );
+	const [ , bDate ] = rdate.exec( b );
+	return parseInt( aDate ) - parseInt( bDate );
```
mgol, line 66: Since these are supposed to be numbers, `Number` would be better instead of `parseInt`:
```suggestion
	return Number( aDate ) - Number( bDate );
```

mgol (changes requested):

build/release/authors.js
timmywil: I think we talked about that, but had decided to use `fs.promises` for now and then when we drop node 10, do a find/replace on all of them. Then we don't have to keep straight which are used in Node 10 and which aren't.

mgol (approved):
LGTM!

PR #5397, opened 2024-01-29
Tests: Fix Karma tests on Node.js 20

mgol:
### Summary ###

Node.js 20 started throwing errors when `writeHead` is called twice on a response. This might have already been invalid before but it wasn't throwing on Node.js 18.

Compute the headers object and call `writeHead` once to avoid the issue.

### Checklist ###

### DIFF
--- a/test/middleware-mockserver.cjs
+++ b/test/middleware-mockserver.cjs
@@ -83,9 +83,7 @@ const mocks = {
 		if ( resp.set ) {
 			resp.set( headers );
 		} else {
-			for ( const key in headers ) {
-				resp.writeHead( 200, { [ key ]: headers[ key ] } );
-			}
+			resp.writeHead( 200, headers );
 		}
 
 		if ( req.query.callback ) {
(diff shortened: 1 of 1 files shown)


mgol:
Since the fix is in `middleware-mockserver.js`, I expect it to be valid even after Grunt & Karma removal.

test/middleware-mockserver.cjs
```diff
@@ -83,9 +83,7 @@ const mocks = {
 		if ( resp.set ) {
 			resp.set( headers );
 		} else {
-			for ( const key in headers ) {
-				resp.writeHead( 200, { [ key ]: headers[ key ] } );
-			}
+			resp.writeHead( 200, headers );
```
mgol, line 86: @timmywil What's the reason for the above `resp.set` check? Is it to cater to Express as well? If that's the case, why wasn't this needed in other functions using `writeHead`? I see this has been added in #5318.

timmywil: Indeed it was to cater to Express, but I don't remember exactly why the others didn't need it. But, now that I think about it, it could be that I was seeing `writeHead` wipe out previous calls and I compensated by using an express method. You *may* be able to remove the `.set` and `.append` calls, but you can double-check with `node test/server.js` and `http://localhost:3000/test/`.

mgol: The `.append(` call needs to stay due to different handling of duplicate header names but the `.set(` one is redundant now; removed.

timmywil (approved):

mgol:
`3.x-stable` version at #5413.

PR #5404, opened 2024-02-05
Release: remove scripts and dev deps from dist package.json

timmywil:
- this became necessary due to the addition of the prepare script
- scripts aren't needed and don't work in the dist repo

### Summary ###

Hit this snag on the last step testing npm publish in the dist repo, which tried to run the new `prepare` script, which of course failed.

### Checklist ###

* [x] New tests have been added to show the fix or feature works
* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #5408, opened 2024-02-05
Release: use buildDefaultFiles directly and pass version

timmywil:
- also add the ability to pass VERSION in env to test final builds
- adjust sha regex to account for lack of shas
- set the version on the dist package.json

### Checklist ###

### DIFF
--- a/build/release.js
+++ b/build/release.js
@@ -24,6 +24,7 @@ module.exports = function( Release ) {
 	];
 	const cdn = require( "./release/cdn" );
 	const dist = require( "./release/dist" );
+	const { buildDefaultFiles } = require( "./tasks/build" );
 
 	const npmTags = Release.npmTags;
 
@@ -53,8 +54,8 @@ module.exports = function( Release ) {
 		 * committed before creating the tag.
 		 * @param {Function} callback
 		 */
-		generateArtifacts: function( callback ) {
-			Release.exec( "npm run build:all" );
+		generateArtifacts: async function( callback ) {
+			await buildDefaultFiles( { version: Release.newVersion } );
 
 			cdn.makeReleaseCopies( Release );
 			Release._setSrcVersion();
@@ -77,7 +78,7 @@ module.exports = function( Release ) {
 		 * Publish to distribution repo and npm
 		 * @param {Function} callback
 		 */
-		dist: async callback => {
+		dist: async function( callback ) {
 			await cdn.makeArchives( Release );
 			dist( Release, distFiles, callback );
 		}
--- a/build/release/dist.js
+++ b/build/release/dist.js
@@ -119,6 +119,7 @@ module.exports = function( Release, files, complete ) {
 		delete packageJson.devDependencies;
 		delete packageJson.dependencies;
 		delete packageJson.commitplease;
+		packageJson.version = Release.newVersion;
 		await fs.writeFile(
 			`${ Release.dir.dist }/package.json`,
 			JSON.stringify( packageJson, null, 2 )
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -353,6 +353,10 @@ async function build( {
 }
 
 async function buildDefaultFiles( { version, watch } = {} ) {
+	if ( !version ) {
+		version = process.env.VERSION;
+	}
+
 	await Promise.all( [
 		build( { version, watch } ),
 		build( { filename: "jquery.slim.js", slim: true, version, watch } ),
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -352,11 +352,10 @@ async function build( {
 	}
 }
 
-async function buildDefaultFiles( { version, watch } = {} ) {
-	if ( !version ) {
-		version = process.env.VERSION;
-	}
-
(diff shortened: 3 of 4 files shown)


build/release.js
```diff
@@ -77,7 +78,7 @@ module.exports = function( Release ) {
 		 * Publish to distribution repo and npm
 		 * @param {Function} callback
 		 */
-		dist: async callback => {
+		dist: async function( callback ) {
```
timmywil, line 81: This was changed to match the others

build/release.js
```diff
@@ -53,8 +54,8 @@ module.exports = function( Release ) {
 		 * committed before creating the tag.
 		 * @param {Function} callback
 		 */
-		generateArtifacts: function( callback ) {
-			Release.exec( "npm run build:all" );
+		generateArtifacts: async function( callback ) {
+			await buildDefaultFiles( { version: Release.newVersion } );
```
timmywil, line 58: Passes the version directly from jquery-release, which overrides any version creation in our builder.

build/release/dist.js
```diff
@@ -119,6 +119,7 @@ module.exports = function( Release, files, complete ) {
 		delete packageJson.devDependencies;
 		delete packageJson.dependencies;
 		delete packageJson.commitplease;
+		packageJson.version = Release.newVersion;
```
timmywil, line 122: Something I missed from an earlier PR

build/tasks/build.js
```diff
@@ -353,6 +353,10 @@ async function build( {
 }
 
 async function buildDefaultFiles( { version, watch } = {} ) {
+	if ( !version ) {
+		version = process.env.VERSION;
+	}
```
timmywil, line 358: Allows me to test version setting from the command line.

build/release/dist.js
mgol: How was this process working for previous releases then? jQuery `3.7.1` etc. has a correct version in its `package.json`.

build/tasks/compare_size.mjs
```diff
@@ -109,7 +109,7 @@ export async function compareSize( { cache = ".sizecache.json", files } = {} ) {
 			// Remove the short SHA and .dirty from comparisons.
 			// The short SHA so commits can be compared against each other
 			// and .dirty to compare with the existing branch during development.
-			const sha = /jQuery v\d+.\d+.\d+(?:-\w+)?\+(?:slim.)?([^ \.]+(?:\.dirty)?)/.exec( contents )[ 1 ];
+			const sha = /jQuery v\d+.\d+.\d+(?:-\w+)?(?:\+|\+slim\.)?([^ \.]+(?:\.dirty)?)?/.exec( contents )[ 1 ];
```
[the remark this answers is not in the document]
timmywil: There are a few things fixed here, but, mainly, the regex failed to match on `4.0.0-beta`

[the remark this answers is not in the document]
timmywil: - The `+` should not have been required, but can be followed by `slim.` or not.
- The dot after slim is a period, not any  character
- The sha is optional (the last `?`)

build/release/dist.js
timmywil: This was a regression from [today's PR](https://github.com/jquery/jquery/pull/5404) removing `scripts`.

mgol: But why did it work previously? Before that PR, we weren't writing `package.json` to the disk at all in this script. Was this happening earlier by `jquery-release` itself and was the regression that we were overwriting a version already provided by `jquery-release`?

timmywil: It worked because jquery-release does change the version in the `package.json` and before that PR, the repo's package.json was copied to `dist`. Note that `pkg` here is read before the version is changed.

build/tasks/compare_size.mjs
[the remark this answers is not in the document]
mgol: It seems to me there's no need to tie `+` with `slim` in the first part, avoiding repetition and making them more independent to the human eye:
```suggestion
			const sha = /jQuery v\d+.\d+.\d+(?:-\w+)?\+?(?:slim\.)?([^ \.]+(?:\.dirty)?)?/.exec( contents )[ 1 ];
```

build/release/dist.js
mgol: Where was this copying happening? I don't see any code for that removed in https://github.com/jquery/jquery/pull/5404.

build/tasks/compare_size.mjs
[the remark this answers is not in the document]
mgol: Ah, but if `slim.` is included you do want the `+`, don't you? Now I get it.

mgol (approved):
One non-blocking question about the logic which is more me trying to understand the code flow here & before. LGTM!

build/release/dist.js
timmywil: It's only one line in the `extras` array.

mgol: Ah, right… 🤦🏼‍♂️

mgol (approved):
LGTM!

PR #5418, opened 2024-02-16
Migrate testing infrastructure to minimal dependencies

timmywil:
### Summary ###

This is a complete rework of our testing infrastructure. The main goal is to modernize and drop deprecated or undermaintained dependencies (specifically, grunt, karma, and testswarm). We've achieved that by limiting our dependency list to ones that are unlikely to drop support any time soon. The new dependency list includes:

- `qunit` (our trusty unit testing library)
- `selenium-webdriver` (for spinning up local browsers)
- `express` (for starting a test server and adding middleware)
  - express middleware includes uses of `body-parser` and `raw-body`
- `yargs` (for constructing a CLI with pretty help text)
- BrowserStack (for running each of our QUnit modules separately in all of our supported browsers)
  - `browserstack-local` (for opening a local tunnel. This is the same package still currently used in the new Browserstack SDK)
  - We are not using any other BrowserStack library. The newest BrowserStack SDK does not fit our needs (and isn't open source). Existing libraries, such as `node-browserstack` or `browserstack-runner`, either do not quite fit our needs, are under-maintained and out-of-date, or are not robust enough to meet all of our requirements. We instead call the [BrowserStack REST API](https://github.com/browserstack/api) directly.

## BrowserStack Runner
- automatically retries individual modules in case of test failure(s)
- automatically attempts to re-establish broken tunnels
- automatically refreshes the page in case a test run has stalled
- runs all browsers concurrently and uses as many sessions as are available under the BrowserStack plan. It will wait for available sessions if there are none.
- supports filtering the available list of browsers by browser name, browser version, device, OS, and OS version (see `npm run test:unit -- --list-browsers` for more info). It will retrieve the latest matching browser available if any of those parameters are not specified.
- cleans up after itself (closes the local tunnel, stops the test server, etc.)
- Requires `BROWSERSTACK_USERNAME` and `BROWSERSTACK_ACCESS_KEY` environment variables.

## Selenium Runner
- supports running any local browser as long as the driver is installed, including support for headless mode in Chrome, FF, and Edge
- supports running `basic` tests on the latest [jsdom](https://github.com/jsdom/jsdom#readme), which can be seen in action in this PR (see `test:browserless`)
- Node tests will run as before in PRs and all non-dependabot branches, but now includes tests on real Safari in a GH actions macos image instead of playwright-webkit.
- can run multiple browsers and multiple modules concurrently

Other notes:
- Stale dependencies have been removed and all remaining dependencies have been upgraded to their latest versions with a few exceptions:
  - `sinon`: stopped supporting IE in version 10. But, `sinon` has been updated to 9.x.
  - `husky`: latest does not support Node 10 and runs on `npm install`. Needed for now until git builds are migrated to GitHub Actions.
  - `rollup`: latest does not support Node 10. Needed for now until git builds are migrated to GitHub Actions.
- BrowserStack tests are set to run on each `main` branch commit
- `debug` mode leaves Selenium browsers open whether they pass or fail and leaves browsers with test failures open on BrowserStack. The latter is to avoid leaving open too many sessions.
- This PR includes a workflow to dispatch BrowserStack runs on-demand
- The Node version used for most workflow tests has been upgraded to 20.x
- updated supportjQuery to 3.7.1

Run `npm run test:unit -- --help` for CLI documentation or view the [command.js file](https://github.com/jquery/jquery/blob/c0297908223d1d305bc820e2b8fd02c06cd8f9cd/test/runner/command.js) directly.

### Checklist ###


test/runner/run.js
```diff
@@ -0,0 +1,248 @@
...
+
+		await cleanupAllWorkers( verbose );
+	}
+
+	asyncExitHook(
+		async() => {
+			await stopServer();
+			await cleanup();
+		},
+		{ wait: 60000 }
```
mgol, line 170: Could we move that to a constant declared at the top of the file?

test/runner/run.js
```diff
@@ -0,0 +1,306 @@
...
+
+	if ( verbose ) {
+		console.log( browserstack ? "Running in BrowserStack." : "Running locally." );
+	}
+
+	const errorMessages = [];
+	const pendingErrors = {};
+
+	// Convert browsers to browser objects
+	browsers = browsers.map( ( b ) => ( { browser: b } ) );
```
mgol, line 53: Same remark as somewhere above - I'd prefer a different variable when the shape of the data changes; even if we're not using TypeScript, it makes it easier to reason about the code. Plus, we have an option to one day experiment with an JSDoc-based TS type checking which would require such a separation.

test/runner/selenium/createDriver.js
```diff
@@ -0,0 +1,49 @@
+import { Builder, Capabilities } from "selenium-webdriver";
+import Chrome from "selenium-webdriver/chrome.js";
```
mgol, line 2: BTW, we might want to eventually consider switching to Chrome for Testing here: https://developer.chrome.com/blog/chrome-for-testing

test/runner/selenium/createDriver.js
```diff
@@ -0,0 +1,49 @@
...
+	const chromeOptions = new Chrome.Options();
+	chromeOptions.addArguments( "--enable-chrome-browser-cloud-management" );
+
+	// Alter the chrome binary path if
+	// the CHROME_BIN environment variable is set
+	if ( process.env.CHROME_BIN ) {
+		chromeOptions.setChromeBinaryPath( process.env.CHROME_BIN );
+	}
+
+	const firefoxOptions = new Firefox.Options();
```
mgol, line 27: Why is `FIREFOX_BIN` not supported as opposed to Chrome/Edge?

test/runner/selenium/createDriver.js
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	driver = await driver
+		.setChromeOptions( chromeOptions )
+		.setFirefoxOptions( firefoxOptions )
+		.setEdgeOptions( edgeOptions )
+		.build();
+
+	// Increase script timeout to 10min
+	await driver.manage().setTimeouts( { script: 60000 * 10 } );
```
mgol, line 74: Please move the number to a constant declared at the top.

test/runner/selenium/runSelenium.js
```diff
@@ -0,0 +1,32 @@
...
+	url,
+	{ browser },
+	{
+		debug,
+		headless
+	} = {}
+) {
+	if ( debug ) {
+		if ( headless ) {
+			console.warn( chalk.italic( "Cannot debug in headless mode." ) );
```
mgol, line 15: Just warn? Don't we want to change the flag or quit with an error here?

mgol (commented):
I managed to only review the first commit so far but I'm posting it for review so that you don't need to wait for me to finish with the rest.

.github/workflows/node.js.yml
```diff
@@ -3,7 +3,9 @@ name: CI
 on:
   pull_request:
   push:
-    branches-ignore: "dependabot/**"
+    branches-ignore:
+      - main
```
mgol: Why ignore `main`? And what about `3.x-stable`?

test/runner/browserstack/workers.js
```diff
@@ -27,18 +29,29 @@ export async function cleanupWorker( reportId, verbose ) {
...
 }
 
-export function acknowledgeWorker( reportId ) {
-	touchWorker( reportId );
+export function retryTest( reportId, retries ) {
 	const worker = workers[ reportId ];
 	if ( worker ) {
-		worker.ack = true;
+		worker.retries ||= 0;
+		if ( ++worker.retries <= retries ) {
```
mgol, line 50: Can we move the incrementation to the previous line?

test/runner/run.js
```diff
@@ -67,15 +63,47 @@ export async function run( {
...
-				failedCount += reportEnd( message.data, reports[ message.id ] );
-				await cleanupWorker( message.id, verbose );
+				const failedCount = reportEnd(
+					message.data,
+					message.id,
+					reports[ message.id ]
+				);
+
+				if ( !failedCount ) {
+					if ( pendingErrors[ message.id ] ) {
```
mgol: What are pending errors? How they are used to detect flaky tests?

.github/workflows/node.js.yml
```diff
@@ -1,87 +1,105 @@
...
-      - main
-      - "dependabot/**"
+    branches-ignore: "dependabot/**"
 
 permissions:
   contents: read # to fetch code (actions/checkout)
 
 jobs:
-  build:
+  build-and-test:
```
mgol, line 12: What about the Git builds? They are not handled in this PR yet, are they? As long as they're not, our Jenkins builds are still useful. But we should implement it in GA as well to completely get rid of the Jenkins dependency.

In the meantime, we still need Node 10 builds to work (tests not so much but at least the build).

package.json
```diff
@@ -54,18 +54,19 @@
...
+    "test:jsdom": "npm run pretest && npm run build:main && npm run test:unit -- -b jsdom -m basic",
     "test:node_smoke_tests": "npm run pretest && npm run build:all && node build/tasks/node_smoke_tests.js",
     "test:promises_aplus": "npm run build:main && node build/tasks/promises_aplus_tests.js",
+    "test:firefox": "npm run pretest && npm run build:main && npm run test:unit -- -v -b firefox -h --no-isolate",
+    "test:safari": "npm run pretest && npm run build:main && npm run test:unit -- -b safari --no-isolate",
     "test:server": "node test/runner/server.js",
+    "test:esm": "npm run pretest && npm run build:main && npm run test:unit -- --esm --no-isolate -h",
+    "test:no-deprecated": "npm run pretest && npm run build -- -e deprecated && npm run test:unit -- --no-isolate -h",
+    "test:selector-native": "npm run pretest && npm run build -- -e selector && npm run test:unit -- --no-isolate -h",
+    "test:slim": "npm run pretest && npm run build -- --slim && npm run test:unit -- --no-isolate -h",
```
mgol, line 68: Can we unify the flags passed to all these? The `-h` parameter is placed in a different place in different scripts, the Safari one doesn't have it at all, etc. The more similar they all are, the better.

test/index.html
```diff
@@ -26,19 +26,20 @@
...
-		// This is in a different script tag to ensure that
-		// jQuery is on the page when the testrunner executes
-		// QUnit.config is populated from QUnit.urlParams but only at the beginning
-		// of the test run. We need to read both.
-		var esmodules = QUnit.config.esmodules || QUnit.urlParams.esmodules;
+		( function () {
+			// Load tests if they have not been loaded
+			// This is in a different script tag to ensure that
+			// jQuery is on the page when the testrunner executes
+			// QUnit.config is populated from QUnit.urlParams but only at the beginning
```
mgol, line 33: Can you move the word `beginning` to the next line? It got a bit long (GitHub UI breaks it to a new line for me) and we have space in the next line.

test/runner/selenium/runSelenium.js
```diff
@@ -15,7 +16,7 @@ export async function runSelenium(
 		}
 	}
 
-	const driver = await createDriver( { browser, headless } );
+	const driver = await createDriver( { browserName: browser, headless, verbose } );
```
mgol, line 19: This line got a bit long; can you break the object into multiple lines?

test/unit/ajax.js
```diff
@@ -1834,13 +1834,7 @@ QUnit.module( "ajax", {
...
 					"Etag": {
 						url: "mock.php?action=etag",
-
-						// Support: TestSwarm
-						// TestSwarm is now proxied via Cloudflare which cuts out
-						// headers relevant for ETag tests, failing them. We're still
-						// running those tests in Karma on Chrome & Firefox (including
-						// Firefox ESR).
-						qunitMethod: QUnit.isSwarm ? "skip" : "test"
+						qunitMethod: "test"
```
mgol, line 1837: You can just remove `qunitMethod` from here & above and use `QUnit.test` directly below. Also, with a single entry in the objects, you can now map to URLs directly, simplifying the logic below a bit.

test/unit/event.js
```diff
@@ -2242,14 +2242,13 @@ QUnit.test( "focusin bubbles", function( assert ) {
 		assert.equal( 0, order++, "focusin on the element first" );
 	} );
 
-// Removed since DOM focus is unreliable on test swarm
 	// DOM focus method
-//	input[ 0 ].focus();
+	input[ 0 ].focus();
```
mgol, line 2246: :tada:

build/tasks/minify.js
```diff
@@ -28,8 +28,9 @@ module.exports = async function minify( { filename, dir, esm } ) {
 					" | (c) OpenJS Foundation and other contributors" +
 					" | jquery.org/license */\n"
 			},
-			mangle: true,
 			inlineSourcesContent: false,
+			mangle: true,
+			module: esm,
```
mgol, line 33: Where is this option coming from? I don't see anything in the docs: module, Ignored. Currently, all files are treated as module.

test/runner/browserstack/api.js
```diff
@@ -118,6 +118,14 @@ export async function getBrowsers( flat ) {
 	return browsers.sort( sortBrowsers );
 }
 
+function matchVersion( browserVersion, version ) {
+	if ( !version ) {
+		return false;
+	}
+	const regex = new RegExp( `^${ version.replace( /\./g, "\\." ) }\\b`, "i" );
+	return regex.test( browserVersion );
```
mgol: Wouldn't it be enough to return `browserVersion.startsWith( version )`? What do we gain by the regex here?

test/data/testinit.js
```diff
@@ -1,12 +1,12 @@
 /* eslint no-multi-str: "off" */
 "use strict";
 
-var parentUrl = window.location.href.replace( /\/test\/.*$/, "/" ),
+var parentUrl = window.location.origin,
```
mgol, line 4: Some versions of IE 11 didn't support `location.origin`. I'm not sure if this is still an issue but I remember this pain...

mgol (changes requested):
OK, I managed to review the whole thing in the end; the first commit contained the majority of the work. Nice job! :tada:

For now I looked at the code, I haven't played with it locally yet.

test/runner/browsers.js
```diff
@@ -0,0 +1,30 @@
...
+import { getBrowsers } from "./browserstack/api.js";
+
+export const browsers = [
+	"chrome",
+	"ie",
+	"firefox",
+	"edge",
+	"safari",
+	"opera",
+	"yandex",
```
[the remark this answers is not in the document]
timmywil: As we discussed, this is just how it comes back from browserstack. We can lower case them all, but lean towards just leaving it.

test/runner/browserstack/api.js
```diff
@@ -0,0 +1,260 @@
...
+import { createAuthHeader } from "./createAuthHeader.js";
+
+const browserstackApi = "https://api.browserstack.com";
+const apiVersion = 5;
+
+const username = process.env.BROWSERSTACK_USERNAME;
+const accessKey = process.env.BROWSERSTACK_ACCESS_KEY;
+
+// iOS has null for version numbers,
+// but the list is already sorted by OS version number.
```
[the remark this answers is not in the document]
timmywil: Point being it doesn't need to check for beta OS versions when determining the latest.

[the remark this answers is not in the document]
mgol: Can you add that to the comment?

.github/workflows/node.js.yml
timmywil: That is next on my list, but you're right. I'll add back Node 10 for now.

build/tasks/minify.js
timmywil: This was required to get the latest version working. Open the type declarations and see `TerserCompressOptions` in the [minify function](https://swc.rs/docs/usage/core#minify)

package.json
timmywil: We can do that, but the Safari one should not have it.

timmywil: Looks like I was consistent except for one. Differences exist in `test:safari` (no `-h`) and `test:firefox`. For the latter, I purposely add `-v` so that we can double confirm that the Firefox version matches ESR (verbose mode logs more browser info when spinning up the driver).

test/data/testinit.js
timmywil: What do you think of removing `parentUrl` altogether and using absolute URLs? It works for selenium, browserstack, and jsdom and might be more future proof when it comes time to try this on 3.x. I think it was only necessary on testswarm?

I'll do a test run with it, but we can make changes if you have another reason for including the domain.

test/runner/browserstack/api.js
```diff
@@ -0,0 +1,260 @@
+/**
+ * Browserstack API is documented at
+ * https://github.com/browserstack/api
+ */
+
+import { createAuthHeader } from "./createAuthHeader.js";
+
+const browserstackApi = "https://api.browserstack.com";
+const apiVersion = 5;
```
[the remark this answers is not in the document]
timmywil: I believe we discussed this. You still need the REST API to create workers. I don't know if one is preferred in cases where they are equivalent, but the REST API is required. The only equivalent it didn't have was the plan info.

test/runner/browserstack/workers.js
```diff
@@ -0,0 +1,271 @@
+import chalk from "chalk";
+import { getBrowserString } from "../lib/getBrowserString.js";
+import { changeUrl, createWorker, deleteWorker, getWorker } from "./api.js";
+
+const workers = {};
```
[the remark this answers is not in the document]
timmywil: The keys are always hashes, but I'm fine with it.

test/runner/browserstack/workers.js
```diff
@@ -0,0 +1,209 @@
...
+						ACKNOWLEDGE_WORKER_TIMEOUT / 1000
+					}s.`
+				)
+			);
+		}, ACKNOWLEDGE_WORKER_TIMEOUT );
+	} );
+}
+
+/**
+ * All arguments are required or the request will fail
```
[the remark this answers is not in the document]
timmywil: Good point. `restarts` was added later.

test/runner/command.js
```diff
@@ -0,0 +1,118 @@
...
+	.option( "no-isolate", {
+		type: "boolean",
+		description: "Run all modules in the same browser instance."
+	} )
+	.option( "browserstack", {
+		type: "array",
+		description:
+			"Run tests in BrowserStack.\nRequires BROWSERSTACK_USERNAME and BROWSERSTACK_ACCESS_KEY environment variables.\n" +
+			"The value can be empty for the default configuration, or a string in the format of\n" +
+			"\"browser_[browserVersion | :device]_os_osVersion\".\n" +
```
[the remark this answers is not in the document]
timmywil: It's all optional

test/runner/createTestServer.js
```diff
@@ -0,0 +1,60 @@
+import bodyParser from "body-parser";
+import express from "express";
+import bodyParserErrorHandler from "express-body-parser-error-handler";
+import fs from "fs";
```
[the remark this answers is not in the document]
timmywil: Heh, you comment this in a lot of my recent PRs. But, if you recall, I'd like to use `"fs"` everywhere for now and when we drop Node 10, I'll do a find/replace on all of them so we don't have to keep track of which need to run in Node 10 and which don't.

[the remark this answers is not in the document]
timmywil: I wasn't 100% sure we *could* drop Node 10 in this PR, but as you pointed out we can't until we put together a GA for git builds.

test/runner/createTestServer.js
```diff
@@ -0,0 +1,60 @@
...
+
+	// Redirect home to test page
+	app.get( "/", ( _req, res ) => {
+		res.redirect( "/test/" );
+	} );
+
+	// Redirect to trailing slash
+	app.use( ( req, res, next ) => {
+		if ( req.path === "/test" ) {
+			const query = req.url.slice( req.path.length );
```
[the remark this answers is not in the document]
timmywil: `req.query` parses the query string into an object. I just want the string here.

test/runner/queue.js
```diff
@@ -0,0 +1,61 @@
...
+		// Keep checking max sessions for browserstack
+		if (
+			promises.length >=
+			( browserstack ?
+				Math.min( concurrency, await getMaxSessions() ) :
+				concurrency )
+		) {
+
+			// Wait until at least one promise resolves
+			await Promise.any( promises );
```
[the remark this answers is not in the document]
timmywil: yeah, I think so

test/runner/selenium/createDriver.js
timmywil: The chrome driver works with chrome for testing and is the default if you don't have chrome already installed. I don't think it's necessary in GitHub actions and you can force it here by setting `CHROME_BIN`.

.github/workflows/browserstack.yml
```diff
@@ -0,0 +1,64 @@
+name: Browserstack
+
+on:
+  push:
+    branches:
+      - migrate-test
```
mgol, line 6: Just a comment to not forget to change it back to `main` before merging (& also `3.x-stable`? Or is that happening separately?)

test/runner/browserstack/api.js
```diff
@@ -0,0 +1,260 @@
...
+		} )
+	} );
+}
+
+/**
+ * Stop all workers
+ */
+export async function stopWorkers() {
+	const workers = await getWorkers();
+	await Promise.all( workers.map( ( worker ) => deleteWorker( worker.id, true ) ) );
```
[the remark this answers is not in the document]
timmywil: We discussed live, but I've added a warning to the command.

test/runner/browserstack/listeners.js
```diff
@@ -0,0 +1,86 @@
...
+				!( value instanceof RegExp ) &&
+				!( value instanceof String )
+			) {
+
+				// Return a string early for elements
+				if ( value.nodeType ) {
+					return value.toString();
+				}
+
+				if ( objects.indexOf( value ) > -1 ) {
```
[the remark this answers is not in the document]
timmywil: I tried Set and ran into issues on IE11. I didn't spend a lot of time debugging, but as you said, it'd be nice to not to have to make too many changes for 3.x. When we no longer test 3.x and drop support for IE11 in v5, we can change this whole function.

test/runner/browserstack/listeners.js
```diff
@@ -0,0 +1,86 @@
...
+
+					// If it is an array, replicate the array.
+					return value.map( derez );
+				} else {
+
+					// If it is an object, replicate the object.
+					var nu = {};
+					Object.keys( value ).forEach( function( name ) {
+						nu[ name ] = derez( value[ name ] );
+					} );
```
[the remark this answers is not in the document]
timmywil: I don't think that's an issue in this case. We're using an already-existing object.

test/runner/browserstack/listeners.js
```diff
@@ -0,0 +1,86 @@
...
+			JSON.stringify( {
+				id: id,
+				type: type,
+				data: data ? decycle( data ) : undefined
+			} )
+		);
+	}
+
+	// Send acknowledgement to the server.
+	send( "ack" );
```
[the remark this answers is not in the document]
timmywil: This is our own reporting call, sent to our testing server at our `/api/report` route. See `createTestServer.js`

test/runner/browserstack/listeners.js
```diff
@@ -0,0 +1,86 @@
...
+
+	QUnit.on( "testEnd", function( data ) {
+		send( "testEnd", data );
+	} );
+
+	QUnit.on( "runEnd", function( data ) {
+
+		// Reduce the payload size.
+		// childSuites is large and unused.
```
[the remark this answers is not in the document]
timmywil: It's undocumented, but it contains all test suites and assertions.

test/runner/browserstack/reporter.js
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	console.log(
+		`\n\nTest ${ test.status } on ${ chalk.yellow(
+			getBrowserString( browser, headless )
+		) }.`
+	);
+	console.log( chalk.bold( `${ test.suiteName }: ${ test.name }` ) );
+
+	// Prefer failed assertions over error messages
```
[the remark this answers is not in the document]
timmywil: Assertions are more specific and usually have the same info

test/runner/browserstack/workers.js
```diff
@@ -0,0 +1,209 @@
...
+ */
+export async function runWorker(
+	url,
+	browser,
+	{ reportId, runId, verbose },
+	retries = 0
```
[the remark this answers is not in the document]
timmywil: I think it should stay a separate parameter. It's only used internally by the function to track calls and should not leak into the options object passed originally. As for the name, `refreshes`, `restarts` (previously the `retries` we are looking at now, but I renamed so as to differentiate it from test retries), and `worker.retries` all count upwards so it's consistent once you know that.

test/runner/browserstack/workers.js
```diff
@@ -0,0 +1,209 @@
...
+		const interval = setInterval( async() => {
+			const worker = workers[ reportId ];
+
+			if ( !worker ) {
+				clearInterval( interval );
+				return resolve();
+			}
+
+			const fetchedWorker = await getWorker( worker.id );
+			if ( !fetchedWorker || ( fetchedWorker.status !== "running" && fetchedWorker.status !== "queue" ) ) {
```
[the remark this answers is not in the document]
timmywil: Yes, these are the only two statuses. This if statement *should* fail on the first part when the worker has stopped, but I don't trust BrowserStack enough not to check.

test/runner/run.js
```diff
@@ -0,0 +1,248 @@
...
+} from "./browserstack/workers.js";
+import { createTestServer } from "./createTestServer.js";
+import { buildTestUrl } from "./lib/buildTestUrl.js";
+import { generateHash, printModuleHashes } from "./lib/generateHash.js";
+import { getBrowserString } from "./lib/getBrowserString.js";
+import { addRun, runAll } from "./queue.js";
+
+// Limit concurrency to 8 by default
+// More than this will log MaxListenersExceededWarning
+const MAX_CONCURRENCY = 8;
```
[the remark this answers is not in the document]
timmywil: I know, but I didn't want to have to change the user's environment. And to your point, I agree it's enough on local. It's pretty fast.

test/runner/run.js
```diff
@@ -0,0 +1,248 @@
...
+		browsers.join( ":" )
+	);
+
+	// Create the test app and
+	// hook it up to the reporter
+	const reports = {};
+	const app = await createTestServer( async( message ) => {
+		switch ( message.type ) {
+			case "testEnd":
+				touchWorker( message.id );
```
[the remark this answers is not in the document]
timmywil: I believe we discussed live, but "touching" sets the last time a request was received related to the worker. I've added a comment.

.github/workflows/node.js.yml
timmywil: This was temporary. No longer ignoring main.

test/runner/run.js
timmywil: Pending errors are cleared if a retry passes.

test/runner/browserstack/api.js
timmywil: Notice the `\b`.

test/data/testinit.js
timmywil: Seems to work. Let me know if there's an issue with this.

timmywil:
That manipulation test in firefox might be flakey.

.github/workflows/browserstack.yml
timmywil: 3.x-stable will be another PR. Since most of the test code runs on Node, I'm hoping it won't be too different.

timmywil:
Passing BrowserStack run after all the changes: 

I'd like to get this in and iterate more afterwards as needed, unless there's anything else glaring that needs changing. Then I can start on a 3.x PR and add a workflow for updating the git versions. Then, we can finally decommission old Jenkins & Testswarm servers.

test/runner/browserstack/listeners.js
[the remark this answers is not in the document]
mgol: Using an already existing object does not resolve an issue. Take the following example:
```js
var value = JSON.parse( '{"__proto__": []}' );
console.log( value.__proto__ ); // []
console.log( Object.getProtypeOf( value ) ); // Object.prototype

// `__proto__` is not magical here:
console.log( Object.getProtypeOf( value ) === o.__proto__ ); // false

var nu = {};
Object.keys( value ).forEach( function( name ) {
    nu[ name ] = value[ name ];
} );

console.log( nu.__proto__ ); // []

// Oops! The prototype changed:
console.log( Object.getProtypeOf( nu ) ); // []
console.log( Object.getProtypeOf( nu ) === nu.__proto__ ); // true

test/data/testinit.js
mgol: This may have been required because of local PHP testing (also used by TestSwarm) & Karma using different URLs. Perhaps irrelevant with your changes.

test/runner/browserstack/listeners.js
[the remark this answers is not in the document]
timmywil: Right, but that wouldn’t matter across the wire because it doesn’t stringify the prototype, right?

mgol (approved):
One last comment in:
https://github.com/jquery/jquery/pull/5418#discussion_r1501710738
Otherwise, LGTM. Great work!

test/runner/browserstack/listeners.js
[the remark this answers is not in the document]
mgol: That's a valid point - plus, this data probably comes from QUnit which we trust - but I find it mentally easier if I don't need to look into how the function is used to determine it's secure, especially that the change is pretty easy to do here.

That said, I don't consider this a critical issue so while I'd prefer to replace this definition with `Object.create( null )`, I approved the PR regardless.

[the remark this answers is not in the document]
timmywil: I was more playing devil's advocate because I was curious. I'll switch to `Object.create`

PR #5423, opened 2024-02-28
Tests: add support for latest and latest-\d+ in test runner

timmywil:
Adds support for filtering browser version by "latest" and "latest-n", which we can now use in the browserstack.yml.

I also made it so `npm run test:unit -- --list-browsers=_latest` lists the latest version of all browsers, rather than matching on the last browser in the sorted list.

I considered adding support for "latest" to OS versions, but didn't want to spend too much time on it since OS versions don't get updated nearly as often as browser versions.

I looked into setting browser versions using the environment, but workflows do not support environment variables at the matrix level, only in steps. There were a couple ways around that, but they were convoluted.

- also fix an issue with retries. It was starting each retry the first time, but closing the worker immediately afterward, so any failures were getting ignored. Fortunately, retries were not actually common and we still have all green: https://github.com/timmywil/jquery/actions/runs/8075030746. In the process, I edited the CLI documentation for `--retries` to make it clear it only applies on BrowserStack runs.

### Checklist ###

### DIFF
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -21,19 +21,19 @@ jobs:
       matrix:
         BROWSER:
           - 'IE_11'
-          - 'Safari_17'
-          - 'Safari_16'
-          - 'Chrome_120'
-          - 'Chrome_119'
-          - 'Edge_120'
-          - 'Edge_119'
-          - 'Firefox_121'
-          - 'Firefox_120'
+          - 'Safari_latest'
+          - 'Safari_latest-1'
+          - 'Chrome_latest'
+          - 'Chrome_latest-1'
+          - 'Edge_latest'
+          - 'Edge_latest-1'
+          - 'Firefox_latest'
+          - 'Firefox_latest-1'
+          - 'Opera_latest'
           - 'Firefox_115'
           - '__iOS_17'
           - '__iOS_16'
           - '__iOS_15'
-          - 'Opera_106'
     steps:
       - name: Checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
--- a/test/runner/browserstack/api.js
+++ b/test/runner/browserstack/api.js
@@ -14,6 +14,7 @@ const accessKey = process.env.BROWSERSTACK_ACCESS_KEY;
 // iOS has null for version numbers,
 // and we do not need a similar check for OS versions.
 const rfinalVersion = /(?:^[0-9\.]+$)|(?:^null$)/;
+const rlatest = /^latest-(\d+)$/;
 
 const rnonDigits = /(?:[^\d\.]+)|(?:20\d{2})/g;
 
@@ -137,6 +138,15 @@ function matchVersion( browserVersion, version ) {
 	return regex.test( browserVersion );
 }
 
+function findLatest( browsers ) {
+
+	// The list is sorted in ascending order,
+	// so the last item is the latest.
+	return browsers.findLast( ( browser ) =>
+		rfinalVersion.test( browser.browser_version )
+	);
+}
+
 export async function filterBrowsers( filter ) {
 	const browsers = await getBrowsers( { flat: true } );
 	if ( !filter ) {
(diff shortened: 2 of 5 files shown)


test/runner/browserstack/api.js
```diff
@@ -148,17 +158,54 @@ export async function filterBrowsers( filter ) {
...
+
+	if ( !filterVersion ) {
+		return filteredWithoutVersion;
+	}
+
+	if ( filterVersion.startsWith( "latest" ) ) {
+		const groupedByName = filteredWithoutVersion
+			.filter( ( b ) => rfinalVersion.test( b.browser_version ) )
+			.reduce( ( acc, browser ) => {
+				acc[ browser.browser ] = acc[ browser.browser ] || [];
```
mgol, line 178:
```suggestion
				acc[ browser.browser ] = acc[ browser.browser ] ?? [];
```

test/runner/browserstack/api.js
```diff
@@ -148,17 +158,54 @@ export async function filterBrowsers( filter ) {
...
+				acc[ browser.browser ].push( browser );
+				return acc;
+			}, {} );
```
mgol, line 181:
```suggestion
			}, Object.create( null ) );
```

test/runner/browserstack/api.js
```diff
@@ -148,17 +158,54 @@ export async function filterBrowsers( filter ) {
...
 			( !filterDevice || filterDevice === ( browser.device || "" ).toLowerCase() )
 		);
 	} );
+
+	if ( !filterVersion ) {
+		return filteredWithoutVersion;
+	}
+
+	if ( filterVersion.startsWith( "latest" ) ) {
+		const groupedByName = filteredWithoutVersion
```
mgol, line 175: Is it worth to create such a structure when we always need the browser name specified? It doesn't make sense to support "latest" without the browser name so we could just filter by name, right?

test/runner/browserstack/api.js
```diff
@@ -148,17 +158,54 @@ export async function filterBrowsers( filter ) {
...
+
+		const filtered = [];
+		for ( const group of Object.values( groupedByName ) ) {
+			const num = rlatest.exec( filterVersion );
+			const latest = group[ group.length - 1 ];
+
+			if ( !latest.browser_version ) {
+				filtered.push( latest );
+				continue;
+			}
```
mgol, line 191: Can you elaborate this block?

test/runner/browserstack/api.js
```diff
@@ -148,17 +158,54 @@ export async function filterBrowsers( filter ) {
...
+
+			// Get the latest version and subtract the number from the filter,
+			// ignoring any patch versions, which may differ between major versions.
```
mgol, line 194: For a given major, we likely want the latest version matching that major - why skip the patch part then?

test/runner/browserstack/api.js
```diff
@@ -148,17 +158,54 @@ export async function filterBrowsers( filter ) {
...
+			const version = parseInt( latest.browser_version ) - ( num ? num[ 1 ] : 0 );
+			const match = group.findLast( ( browser ) => {
+				return matchVersion( browser.browser_version, version.toString() );
+			} );
+			if ( match ) {
+				filtered.push( match );
+			}
+		}
+		return filtered;
```
mgol, line 203: Where is the browser name checked here?

mgol (commented):
A few comments - but I like the overall idea!

.github/workflows/browserstack.yml
```diff
@@ -21,19 +21,19 @@ jobs:
...
-          - 'Firefox_120'
+          - 'Safari_latest'
+          - 'Safari_latest-1'
+          - 'Chrome_latest'
+          - 'Chrome_latest-1'
+          - 'Edge_latest'
+          - 'Edge_latest-1'
+          - 'Firefox_latest'
+          - 'Firefox_latest-1'
+          - 'Opera_latest'
```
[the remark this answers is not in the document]
timmywil: There is only one version of Opera. I moved it next to the other browsers using "latest" as their version number.

test/runner/browserstack/api.js
timmywil: I'm not sure what you're asking. The list API will return all matching browsers even when there's no name.

test/runner/browserstack/api.js
timmywil: I'll add a comment. Mobile devices do not have browser version, so it skips the version check, but includes the latest in the list if it made it through filtering. That way, the latest of all browsers is shown. But, `latest-1` doesn't work as there's no version to compare to.

test/runner/browserstack/api.js
timmywil: In other words, `9.3` may exist, but `8.3` may not. Instead, get the latest version of `8`.

test/runner/browserstack/api.js
timmywil: It was already checked above (see `filteredWithoutVersion`).

.github/workflows/browserstack.yml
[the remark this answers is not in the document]
mgol: Yeah, but why group `latest-x` browsers? For me it's more intuitive to keep various versions of the same browser next to each other.

test/runner/browserstack/api.js
mgol: Right, but you're creating an object mapping browser names to an array of matching browser object. Since we know the browser name we want, why not just filter the browser we want without any grouping instead of transforming everything?

test/runner/browserstack/api.js
mgol: For mobile browsers, can't we we use the `lastest` logic to compare OS versions?

test/runner/browserstack/api.js
mgol: Ah, I understand now, thanks.

.github/workflows/browserstack.yml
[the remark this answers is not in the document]
timmywil: again, there's only one Opera here. They are all still grouped by browser.

test/runner/browserstack/api.js
timmywil: I talk about that in the PR description. It's not bad for iOS, but OS versions do not even always contain numbers. Also, I didn't think it worth the extra effort given that OS versions don't change nearly as often.

test/runner/browserstack/api.js
timmywil: Actually, you can't assume we know the browser name. And the point is to ensure there is only one per browser name returned.

.github/workflows/browserstack.yml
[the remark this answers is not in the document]
timmywil: Oh, I see, but I put 2 between firefoxes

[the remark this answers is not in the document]
mgol: Yes, my point was about separating Firefox ESR.

test/runner/browserstack/api.js
mgol: OK, makes sense.

timmywil:
Closing in favor of another upcoming PR that brings in all the improvements from the 3.x PR. I should have the PR feedback from this one in both of those PRs.

PR #5427, opened 2024-03-02
Migrate testing infrastructure on 3.x-stable

timmywil:
### Summary ###

This is the 3.x version of #5418. There are some changes in this PR that I plan on copying to the `main` branch.

- I've refactored the browserstack workers logic and queue to reuse browser instances. This has reduced the testing time of all 3.x browsers from 35-45min to 5-10min! It uses our built-in report API to set the browser URL when a module completes, rather than trying to use the browserstack REST API to change the URL, which had constant errors for little to no reason. Even if it did work though, it was significantly slower.
- linting is more accurate now for files in `test/runner`. I didn't have the order right in eslint.config.js
- sinon 7.x is the latest version that supports IE9

Sample browserstack runs on my repo:

https://github.com/timmywil/jquery/actions/runs/8118427324
https://github.com/timmywil/jquery/actions/runs/8124843564

Testing time should be around 5-10min for all 3.x browsers!

When reviewing, the first 2 commits are almost identical to the main branch versions. It's the following commits that implement changes for 3.x and new improvements.

### Checklist ###


test/runner/browserstack/browsers.js
```diff
@@ -0,0 +1,196 @@
...
+
+	const worker = await createWorker( {
+		...browser,
+		url: encodeURI( url ),
+		project: "jquery",
+		build: `Run ${ runId }`,
+
+		// Set the max here, so that we can
+		// control the timeout
+		timeout: 1800,
```
mgol, line 113: Please move it to a constant defined at the top of the file.

test/runner/browserstack/queue.js
```diff
@@ -0,0 +1,81 @@
...
+	const previousTest = queue[ index ];
+	queue.splice( index, 1 );
+	for ( const test of queue.slice( index ) ) {
+		if ( test.fullBrowser === previousTest.fullBrowser ) {
+			setBrowserWorkerUrl( test.browser, test.url );
+			test.running = true;
+			return { url: test.url };
+		}
+	}
+}
```
mgol, line 23: It'd be good to have some comments for this function. Especially about this browser reuse, why we look at the tests after the one with `reportId`, etc.

test/runner/browserstack/queue.js
```diff
@@ -0,0 +1,81 @@
...
+export async function runAllBrowserStack( options ) {
+	return new Promise( async( resolve, reject )=> {
+		while ( queue.length ) {
+			try {
+				await checkLastTouches( options );
+			} catch ( error ) {
+				reject( error );
+			}
+
+			// Run one test per browser at a time
```
mgol, line 61: Is it really granular to the level of `QUnit.test`? Or just a test file (which, in our case, equals to modules)?

test/runner/run.js
```diff
@@ -4,38 +4,40 @@ import { getLatestBrowser } from "./browserstack/api.js";
...
+		browsers: browserNames,
+		browserstack,
+		debug,
+		headless,
+		isolate,
+		modules = [],
+		retries = 0,
+		runId,
+		verbose
+	} = options;
```
mgol, line 39: This method is risky IMO. The defaults will only apply to destructured variables like `modules` or `retries` but the `options` object contains them as they were passed. I suspect this may be a problem since you pass `options` to a few APIs as-is.

It'd be safer to keep destructuring directly in parameters. If you then need the `options` object to be fully passed in some places, you can always define it here instead of destructuring:
```js
	const options = {
		amd,
		browsers: browserNames,
		/* ... */
	};
```
The main drawback is each parameter appears twice but the risk above disappears.

test/runner/selenium/queue.js
```diff
@@ -3,33 +3,27 @@
...
 	queue.push( { url, browser, options } );
 }
 
-export async function runFullQueue( {
-	browserstack,
-	concurrency: defaultConcurrency,
-	verbose
-} ) {
+// TODO: Possibly move this to selenium-only
+// and have a queue per browser
```
mgol, line 25: Can you elaborate on this comment?

mgol (commented):
Awesome results!

I left some comments.

As for:

> It uses our built-in report API to set the browser URL when a module completes, rather than trying to use the browserstack REST API to change the URL, which had constant errors for little to no reason.

Can you explain it in a bit more detail to me?

test/runner/run.js
timmywil: Ok, why don't I just go back to what it was before. I'll pass all params to the run functions manually again.

timmywil:
> It uses our built-in report API to set the browser URL when a module completes, rather than trying to use the browserstack REST API to change the URL, which had constant errors for little to no reason.
>> Can you explain it in a bit more detail to me?

Sure, if you look at [this line](https://github.com/jquery/jquery/pull/5427/files#diff-756a39f3d63de0834ea371be2a28bb02a1350f18cb463a4522dc5a31bbbc9247R38), notice that any response from the report function is returned in the report API response. The only times we return responses are in response to the `runEnd` request:

- for retrying the same test: https://github.com/jquery/jquery/pull/5427/files#diff-9d5febf25eeb90f58a863b318706a4363188314e8ca1e85ad06a187bbf0dc174R108
- for starting the next test on failure: https://github.com/jquery/jquery/pull/5427/files#diff-9d5febf25eeb90f58a863b318706a4363188314e8ca1e85ad06a187bbf0dc174R111
- for starting the next test on pass: https://github.com/jquery/jquery/pull/5427/files#diff-9d5febf25eeb90f58a863b318706a4363188314e8ca1e85ad06a187bbf0dc174R125

listeners.js handles the response by [taking the test URL and setting the window location](https://github.com/jquery/jquery/pull/5427/files#diff-6b5959117808dddadde2cf9eed96698af799aa40e08e6648ccdc71531c901e2bR89).

Trying to use BrowserStack's REST API to change the URL on the worker was slower and incredibly error-prone.

test/data/testinit.js
```diff
@@ -458,26 +454,14 @@ this.loadTests = function() {
 				}
 
 			} else {
-				if ( window.__karma__ && window.__karma__.start ) {
-					window.__karma__.start();
-				} else {
-					QUnit.load();
-				}
+				QUnit.load();
```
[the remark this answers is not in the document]
timmywil: Looks like the migration is just to remove it.

test/runner/browserstack/browsers.js
timmywil: This one is a little different than the rest. Setting in a const implies it could be changed in the future to something else. This is the maximum allowed timeout by the browserstack API, and not something that we would likely change unless the API changes. If they ever changed that, I think it could still be hard-coded in worker creation rather than set at the top of the file like other timeouts.

test/runner/browserstack/queue.js
timmywil: I see what you mean, but I was using the word `test` here to mean tests in the BrowserStack queue, which in this case are indeed modules. Perhaps I can clarify this by simply adding the word URL `Run one test URL per browser at a time`

test/runner/selenium/queue.js
timmywil: Oops, this was written before I started this refactor. I've since done that refactor and can remove this comment.

timmywil:
I've incorporate the feedback in #5428  as well.

test/runner/browserstack/browsers.js
mgol: Hmm, can you mention that this is the max allowed by BrowserStack? If there's any link to their doc with this info, this could be useful as well.

test/runner/run.js
```diff
@@ -0,0 +1,337 @@
...
+	// hook it up to the reporter
+	const reports = Object.create( null );
+	const app = await createTestServer( ( message ) => {
+		switch ( message.type ) {
+			case "testEnd": {
+				const reportId = message.id;
+				const report = reports[ reportId ];
+				touchBrowser( report.browser );
+				const errors = reportTest( message.data, reportId, report );
+				pendingErrors[ reportId ] ||= {};
```
mgol, line 82: How about two changes here:
```suggestion
				pendingErrors[ reportId ] ??= Object.create( null );
```
Since keys are possibly arbitrary from the perspective of this code: `message.data.name` - `Object.create( null )` seems safer.

test/runner/selenium/queue.js
```diff
@@ -0,0 +1,70 @@
...
+		// In selenium, this helps avoid undici connect timeout errors.
+		await new Promise( ( resolve ) => setTimeout( resolve, SELENIUM_WAIT_TIME ) );
+
+		if ( verbose ) {
+			console.log( `\nTests remaining: ${ queue.length + 1 }.` );
+		}
+
+		let promise;
+		if ( browser.browser === "jsdom" ) {
+			promise = runJSDOM( url, options );
```
mgol, line 46: I guess it's still not all Selenium in the end in this function. 😉

test/runner/browserstack/queue.js
mgol: Sounds good!

mgol (approved):
Just a few comments (please have a look) but nothing blocking so I'm approving as well.

test/runner/selenium/queue.js
timmywil: True, jsdom is an exception, but can use the same queue logic.

mgol (approved):
No further comments, looks good!

Issue #5416, opened 2024-02-11
`require( "jquery" )` returns a module object when used with Webpack & jQuery 4.0.0-beta


mgol:
The fact that we started ~~exporting more than a single thing from jQuery - apart from the `default` export we have `$` & `jQuery` named tokens pointing to the same thing -~~ providing an ESM version of jQuery via the `module` field in `package.json`’s `exports` makes Webpack treat `require( "jquery" )` as a way to export the full module instead of just the jQuery object...

_Originally posted by @mgol in https://github.com/jquery/jquery/issues/5414#issuecomment-1937378557_

_Issue originally reported by @cocco111 in https://github.com/jquery/jquery/issues/5414_

mgol:
From @cocco111 in https://github.com/jquery/jquery/pull/5414#issuecomment-1937344478, we have a test case:

> https://github.com/cocco111/jQuery4Test
> If you prefer you can simply open `/dist/index.html` in a browser and see result in console.

mgol:
@GeoffreyBooth considering you were a strong advocate of us starting to expose named exports in addition to the default one - do you perhaps know how to fix this issue without stopping to expose those named tokens? Making `require( "jquery" )` not return the jQuery object but instead a transpiled ESM module object when bundled with Webpack is going to break way too much code...

GeoffreyBooth:
CommonJS doesn’t have a way to represent the “default plus named exports” feature of ES modules. The `__esModule` pattern that Babel and TypeScript use is meant to provide a workaround, for tools that understand the hint to treat the CommonJS export differently. You should probably ask a Webpack maintainer, though, as they would be the experts to advise how to output for maximum compatibility with their tool.

mgol:
@sokra @vankop I'd appreciate some Webpack expertise from you here to help with setting up jQeury. We need the following to still work in jQuery 4.x, whether standalone or when bundled with Webpack or others:
```js
const jQuery = require( "jquery" );
jQuery( "<div>" ).appendTo( "body" );
```
and:
```js
import jQuery from "jquery";
jQuery( "<div>" ).appendTo( "body" );
```

At the same time, we cannot just serve two different versions of jQuery - one to `require`, one to `import` as there's lots of internal state that needs to be shared. Is this possible at all to achieve?

[...]

PR #5429, opened 2024-03-02
Core: Fix the exports setup to make bundlers work with ESM & CommonJS

mgol:
### Summary ###

We cannot pass a single file via the `module` condition as then
`require( "jquery" )` will not return jQuery but instead the module object with
`default`, `$` & `jQuery` as keys. Instead:

1. For Node.js, detected via the `node` condition:
    1. Expose a regular CommonJS version to `require`
    2. Expose a tiny wrapper over CommonJS to `import`
2. For bundlers, detected via the `module` condition:
    1. Expose a regular ESM version to `import`
    2. Expose a tiny wrapper over ESM to `require`
3. If neither Node.js nor bundlers are detected (no `node` or `module`
   conditions`):
    1. Expose a regular CommonJS version to `require`
    2. Expose a regular ESM version to `import`

The reasons for such definitions are as follows:
1. In Node.js, one can synchronously import from a CommonJS file inside of
   an ESM one but not vice-versa. To use an ESM file in a CommonJS one,
   a dynamic import is required and that forces asynchronicity.
2. In some bundlers CommonJS is not necessarily enabled - e.g. in Rollup without
   the CommonJS plugin. Therefore, the ESM version needs to be pure ESM.
   However, bundlers allow synchronously calling `require` on an ESM file. This
   is possible since bundlers merge the files before they are passed to
   the browser to execute and the final bundles no longer contain async import
   code.
3. Bare ESM & CommonJS versions are provided to non-Node non-bundler
   environments where we cannot assume interoperability between ESM & CommonJS
   is supported.
4. Bare versions cannot be supplied to Node or bundlers as projects using both
   ESM & CommonJS to fetch jQuery would result in duplicate jQuery instances,
   leading to increased JS size and disjoint data storage.

In addition to the above changes, the `script` condition has been dropped. Only
Webpack documents this condition and it's not clear when exactly it's triggered.
Adding support for a new condition can be added later without a breaking change;
removing is not so easy.

The `production` & `development` conditions have been removed as well. They were
not really applied correctly; we'd need to provide both of them to each current
leaf which would double the size of the definition for the `.` & `./slim` entry
points. In jQuery, the only difference between development & production builds
is minification; there are no logic changes so we can pass unminified versions
to all the tooling, expecting minification down the line.

For `./factory` & `./factory-slim` entry points, only the ESM version is now
supported. This was done to simplify the setup for this niche use case.

Tests with Rollup (both pure ESM & with the CommonJS plugin) & Webpack have
been added.

I added an explainer document to the wiki at https://github.com/jquery/jquery/wiki/jQuery-4-exports-explainer that should help understand the whole `exports` definition. It can be helpful especially if we need to make more changes.

Fixes gh-5416

### Checklist ###


mgol:
CI is failing because `pretest` now de facto depends on `build:all`. This means core tests now require a build. I need to think how to approach this.

test/node_smoke_tests/commonjs/factory/document_missing.cjs
```diff
@@ -1,15 +1,19 @@
 "use strict";
 
-const assert = require( "node:assert" );
+const main = async() => {
```
timmywil, line 3: So far, we have been using function declarations.

```suggestion
async function main() {
```

package.json
```diff
@@ -51,21 +48,21 @@
     "lint": "concurrently -r \"npm:lint:dev\" \"npm:lint:json\"",
     "npmcopy": "node build/tasks/npmcopy.js",
     "prepare": "husky install",
-    "pretest": "npm run qunit-fixture && npm run babel:tests && npm run npmcopy",
+    "pretest": "npm run npmcopy && npm run qunit-fixture && npm run babel:tests && npm run bundlers:tests",
```
timmywil, line 51: I'd like to avoid running any tests in the pretest script. It should only needs to run once. But, if it does turn out to be something we want to run on every build, it could instead be added to the build, similar to `dist()` and `compareSize()`.

test/unit/core.js
```diff
@@ -1532,6 +1532,38 @@ testIframe(
 	}
 );
 
+testIframe(
```
timmywil, line 1535: I see you'd like to run these tests in the browser as well as node. But, it'd be ideal if the files required for these tests were only built if core tests were run.

timmywil (commented):
I know it's not ready for review yet, but I went ahead and reviewed anyway.

package.json
```diff
@@ -119,9 +118,11 @@
     "qunit": "2.20.1",
     "raw-body": "2.5.2",
     "requirejs": "2.3.6",
-    "rollup": "2.79.1",
+    "rollup": "4.12.0",
```
timmywil, line 121: iirc, this won't work on Node 10. I'm working on migrating git builds to GH actions, so maybe that will need to come first.

package.json
mgol: These are not tests, these are bundler runs for later testing as part of core unit tests.

mgol: The main complication is bundlers… well, bundle, so this process does depend on the build being done beforehand.

test/unit/core.js
mgol: The Node ones test the Node usage, the browser ones test that bundlers work correctly. Bundlers generate files for browsers so I need a browser here.

package.json
mgol: Yes, possibly. Otherwise, I’ll need to split it, creating a separate aliased version for tests - and I’m not even sure if that’s going to work given binaries will still conflict.

package.json
timmywil: I think my suggestion still stands. It can be separate from the pretest script and only build files specific to the bundler tests when testing the bundle.

test/unit/core.js
timmywil: Yeah, I realized after that it still needs to build in Node, but it *could* only do that if the `core` module is being tested.

package.json
mgol: The problem is without the bundler runs core unit tests now fail as of this PR. So it kind of is a pre-requisite to running unit tests.

I agree this setup is not ideal. I was thinking about creating a separate bundlers QUnit module just for these tests but I’d need to organize it somehow so that this module is only run when requested explicitly.

Another option would be trying to simulate a browser via jsdom, then this would be another set of Node-only tests.

I’ll think about it.

> only build files specific to the bundler tests when testing the bundle

These are `exports` tests, it’s the safest to build everything before running them.

test/node_smoke_tests/commonjs/factory/document_missing.cjs
mgol: This function no longer exists.

package.json
mgol: I migrated the tests to run on Node + JSDOM so this issue no longer exists.

test/unit/core.js
mgol: These tests no longer exists since I moved them to Node + JSDOM.

package.json
mgol: This change no longer exists as the version was updated on `main`.

mgol:
@timmywil the PR is ready for review now.

package.json
```diff
@@ -53,8 +59,9 @@
     "pretest": "npm run qunit-fixture && npm run babel:tests && npm run npmcopy",
     "qunit-fixture": "node build/tasks/qunit-fixture.js",
     "start": "node -e \"require('./build/tasks/build.js').buildDefaultFiles({ watch: true })\"",
+    "test:bundlers": "node -e \"require('./test/bundler_smoke_tests/utils.cjs').cleanTmpBundlersDir()\" && npm run build:all && rollup -c test/bundler_smoke_tests/rollup-pure-esm.config.js && rollup -c test/bundler_smoke_tests/rollup-commonjs.config.js && webpack -c test/bundler_smoke_tests/webpack.config.cjs && node test/bundler_smoke_tests/run-jsdom-tests.js",
```
timmywil, line 62: What do you say we clean up tmp afterwards? It probably doesn't need to stick around.

timmywil:
Very nice! I only have suggestions on the npm scripts

build/tasks/node_smoke_tests.js
```diff
@@ -5,8 +5,8 @@ const util = require( "util" );
 const exec = util.promisify( require( "child_process" ).exec );
 const verifyNodeVersion = require( "./lib/verifyNodeVersion" );
 
-const allowedLibraryTypes = [ "regular", "factory" ];
-const allowedSourceTypes = [ "commonjs", "module" ];
+const allowedLibraryTypes = new Set( [ "regular", "factory" ] );
+const allowedSourceTypes = new Set( [ "commonjs", "module", "dual" ] );
```
timmywil, line 9: A wild `Set` appears. Here's to dropping Node 10!

package.json
mgol: I'll look into tweaking the scripts. As for the `tmp` directory itself, it may be sometimes easier to analyze  issues if built files stick around. I can always temporarily remove the removal if I have such a need, though.

I would not remove the folder from ignore files, though. A lot can go wrong during a test and it may not complete, leaving the dirty state. I prefer the state to always be clean.

We can optionally change entries in ignore files/arrays to match all directories named `tmp`, regardless of location - then, we won't need to add more entries if we need another directory.

mgol:
@timmywil I haven't addressed your feedback yet but I pushed 3 new commits that I had in progress (the first one is titled "Docs: Mention that the CommonJS module doesn't expose named exports"). The first one is a small docs update, the other two are adding more bundler/Node tests for our exports setup - for the slim build & both factory ones.

I contemplated simplifying the factory definitions by only exposing the ESM versions but I ultimately decided it's too breaking. People depending on factory behavior can update to our new entry points but they may not be able to switch to ESM. Funny enough, I hit it myself here as our Promises/A+ tests adapter uses the factory build but it's a CommonJS file and needs to expose the API synchronously...

The factory setup is still way simpler because there is no default export in ESM and the CommonJS version exports a similar object instead of just the API - and this means adapters are not needed, we can just expose the CommonJS version to Node and the ESM one to bundlers.

mgol:
@timmywil I addressed your feedback now, please re-review.

test/bundler_smoke_tests/run-jsdom-tests.js
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env node
```
mgol, line 1: @timmywil This line, together with `755` permissions, allows to run this file directly, without prepending it with `node `. Does that also work in Windows? If so, this could shorten the npm scripts a bit if we omit the `node` prefix where possible.

timmywil: That does work in Windows. I knew that and left it out, but I don't have a strong preference. Would you rather have that line at the top of the script or use `node` in the npm script? I chose the latter, but I don't mind either way.

timmywil (approved):
Well done!

test/bundler_smoke_tests/run-jsdom-tests.js
mgol: It's not a big deal & I could go either way but I prefer to have the hashbang line for a few reasons:
1. It clearly states at the top of the file that the file is supposed to be run directly and not just exports some APIs to be used elsewhere
2. 1-2 extra lines in a file don't contribute much to the file length. OTH, the `node ` prefix on long npm scripts does influence readability negatively, especially when repeated for multiple scripts.

mgol:
@timmywil FYI, I added an explainer document to the wiki at https://github.com/jquery/jquery/wiki/jQuery-4-%60exports%60-explainer that should help understand the whole `exports` definition. It can be helpful especially if we need to make more changes.

test/bundler_smoke_tests/run-jsdom-tests.js
timmywil: > It clearly states at the top of the file that the file is supposed to be run directly and not just exports some APIs to be used elsewhere

Many of the modules we use do the latter. I wouldn't want to change those, but I'm fine changing the ones we only use as executables.

> the node  prefix on long npm scripts does influence readability negatively, especially when repeated for multiple scripts.

It does make the npm script longer, but I don't think it's all negative. It helps readability to make it clear it's a JS file running in Node.

mgol:
For those following via email, I changed the link to the above Wiki page to https://github.com/jquery/jquery/wiki/jQuery-4-exports-explainer.

test/bundler_smoke_tests/run-jsdom-tests.js
mgol: We discussed this at the meeting. I'll update my file to align with other since it's easier that way; if we ever want to rely on hashbangs, it will be a separate global PR later.

mgol: Done.

PR #5431, opened 2024-03-05
Tests: fix flakey message logs; ignore delete worker failures

timmywil:
### Summary ###

Add some minor improvements after watching the recent first run on 3.x-stable, including:

- fix the flakey message for when a test fails and then passes on retry.
- ignore worker delete failures. We hit this in one browser. The tests all passed and the worker had stopped, but BrowserStack returned a 502.

### Checklist ###

### DIFF
--- a/test/runner/run.js
+++ b/test/runner/run.js
@@ -83,7 +83,19 @@ export async function run( {
 				if ( errors ) {
 					pendingErrors[ reportId ][ message.data.name ] = errors;
 				} else {
-					delete pendingErrors[ reportId ][ message.data.name ];
+					const existing = pendingErrors[ reportId ][ message.data.name ];
+
+					// Show a message for flakey tests
+					if ( existing ) {
+						console.log();
+						console.warn(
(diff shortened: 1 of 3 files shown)


test/runner/run.js
```diff
@@ -83,7 +83,19 @@ export async function run( {
 				if ( errors ) {
 					pendingErrors[ reportId ][ message.data.name ] = errors;
 				} else {
-					delete pendingErrors[ reportId ][ message.data.name ];
+					const existing = pendingErrors[ reportId ][ message.data.name ];
+
+					// Show a message for flakey tests
+					if ( existing ) {
+						console.log();
```
mgol-sumo, line 90: What's the reason for these empty `console.log`s?

mgol-sumo (approved):
One comment, LGTM otherwise

test/runner/run.js
timmywil: I'm going to assume you're okay with that. We can always change in a subsequent PR.

PR #5434, opened 2024-03-08
Build: add GitHub Actions workflow to update Filestash

timmywil:
### Summary ###

Replaces the jenkins workflow to update git versions of jQuery on the Filestash server. The filestash environment is already set up in the jQuery repo settings and includes the necessary SSH private key and filestash server URL.

### Checklist ###

### DIFF
--- a/.github/workflows/filestash.yml
+++ b/.github/workflows/filestash.yml
@@ -1,0 +1,64 @@
+name: Filestash
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    environment: filestash
+    env:
+      NODE_VERSION: 20.x
+    name: Update Filestash
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Use Node.js ${{ env.NODE_VERSION }}
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Cache
+        uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
+        with:
+          path: ~/.npm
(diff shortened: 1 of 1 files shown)


Krinkle:
The following look great:
* Expose private key as "secret" rather than as "environment var", that way we control when it becomes visible, thus not making it visible during earlier steps.
* Install ssh key not until after install/build steps.
* Further limit secrets to a declared group (confusingly called "enviornment"), and the job to a given branch.

The following could tighten it further:
* Configure the secret in the repo setting by branch (i.e. protected-only, or list "main" there). This because as-is the restriction is in the file itself, which feels insufficient. I.e. if it is modified to allow branch X and pushed to X, or if another job is modified to enable the same "filestash" group secrets, that would bypass it. Putting the restriction also/instead in the repo settings would harden this.
* Consider removing use of [actions/setup-node](https://github.com/actions/setup-node) when the purpose of a build isn't to test specific or multiple Node.js versions. This action is pretty convoluted and heavy. The ubuntu image already comes with an LTS version of nodejs pre-installed. For Ubuntu 22 (ubuntu-22, ubuntu-latest) that is currently Node.js 18. https://github.com/actions/runner-images/blob/ubuntu22/20240218.1/images/ubuntu/Ubuntu2204-Readme.md. I imagine Ubuntu 24 will come soon with Node.js 20 (assuming it follows <https://packages.debian.org/experimental/nodejs>).

Neither a big deals. LGTM as-is.

timmywil:
The first suggestions sounds good to me, but I prefer to have control and consistency over the Node version everywhere we build. I've learned never to trust whatever Node version comes preinstalled in the runner image. FWIW, that step usually takes less than a second.

mgol (approved):
LGTM. Let me know when you're ready to merge so that I disable the Jenkins job.

One thing - with the Jenkins job being disabled before merging this, we no longer need to build on Node.js 10. You can remove it here if you want. That said, dropping Node 10 allows us to introduce a few improvements like upgrade Rollup (which I need for the `exports` improvements PR), switch to `node:`-prefixed Node.js core modules, using `node:fs/promises` instead of `fs.promises` from `node:fs`, dropping the `jenkins` npm script and maybe something else. If you'd like, you can group it all in a followup PR. I'd just prefer not to wait too long with dropping the build on Node 10.

Krinkle (approved):

timmywil:
Checked all files and they all updated as expected.

PR #5436, opened 2024-03-08
Drop support for Node 10

timmywil:
### Summary ###

- remove all the things we had in place so the build worked in Node 10
- fix an issue with test cleanup if the server ever stalled

### Checklist ###


build/tasks/build.js
```diff
@@ -6,10 +6,10 @@
 
 "use strict";
 
-const fs = require( "fs" );
-const path = require( "path" );
-const util = require( "util" );
-const exec = util.promisify( require( "child_process" ).exec );
+const fs = require( "node:fs/promises" );
```
mgol, line 9: Not for this PR but maybe we could eventually make all our Node build files ESM instead of CommonJS?

mgol (approved):
LGTM

package.json
```diff
@@ -103,14 +102,14 @@
...
+    "eslint": "8.57.0",
     "eslint-config-jquery": "3.0.2",
     "eslint-plugin-import": "2.29.1",
     "exit-hook": "4.0.0",
-    "express": "4.18.2",
+    "express": "4.18.3",
     "express-body-parser-error-handler": "1.0.7",
     "globals": "14.0.0",
-    "husky": "8.0.3",
+    "husky": "9.0.11",
```
mgol, line 112: On the 3.x PR, you also changed `husky install` to `husky`. Isn't that required here as well?

mgol (commented):
Actually, one comment

package.json
timmywil: Thanks, I meant to do that. Both technically work, but the first is deprecated.

build/tasks/build.js
timmywil: Maybe that's something I could tackle in the release PR.

PR #5438, opened 2024-03-10
Tests: add --hard-retries option to test runner

timmywil:
### Summary ###

Add the ability to retry by restarting the worker and getting a different browser instance, after all normal retries have been exhausted. This can sometimes be successful when a refresh is not.

### Checklist ###

* [x] New tests have been added to show the fix or feature works
* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #5440, opened 2024-03-10
Some build refactors/improvements

timmywil:
### Summary ###

- migrate more uses of fs.promises; use node: protocol. I had missed a lot because my vs code search index was off. I used git grep instead.
- fix size comparison for slim files. I fixed this by preferring `+slim.` to `+` in the regex. It was only matching on the latter even for the slim file, which made it so comparisons between a dirty build and the saved branch were always off by 5-6 bytes.
- make compare size cache readable for manual edits. This is in lieu of more complicated options added to the build command. It's not hard to just edit the JSON, but it might as well be readable.
- Fix worker restarts for when a browser fails the acknowledgement test (I just broke this in #5438)

### Checklist ###


build/release/cdn.js
```diff
@@ -49,7 +49,7 @@ function makeReleaseCopies( Release ) {
 	].forEach( ( { filesMap, cdnFolder } ) => {
 		shell.mkdir( "-p", cdnFolder );
 
-		Object.keys( filesMap ).forEach( key => {
+		Object.keys( filesMap ).forEach( ( key ) => {
```
timmywil, line 52: Ran prettier on this file

build/release/cdn.js
```diff
@@ -60,28 +60,33 @@ function makeReleaseCopies( Release ) {
...
 async function makeArchives( Release ) {
-
 	Release.chdir( Release.dir.repo );
 
 	async function makeArchive( { cdn, filesMap, cdnFolder } ) {
 		return new Promise( ( resolve, reject ) => {
 			if ( Release.preRelease ) {
-				console.log( "Skipping archive creation for " + cdn + "; this is a beta release." );
+				console.log(
+					"Skipping archive creation for " + cdn + "; this is a beta release."
```
mgol, line 88: Let's use template literals for such things:
```suggestion
					`Skipping archive creation for ${ cdn }; this is a beta release.`
```

build/tasks/compare_size.mjs
```diff
@@ -60,7 +60,9 @@ function cacheResults( results ) {
 }
 
 function saveCache( loc, cache ) {
-	return fs.promises.writeFile( loc, JSON.stringify( cache ) );
+
+	// Keep cache readable for manual edits
+	return fs.writeFile( loc, JSON.stringify( cache, null, "  " ) );
```
mgol, line 65: Let's also add a trailing new line

build/release/cdn.js
```diff
@@ -60,28 +60,33 @@ function makeReleaseCopies( Release ) {
...
-				text = fs.readFileSync( builtFile, "utf8" )
-					.replace( /"file":"([^"]+)"/,
-						"\"file\":\"" + unpathedFile.replace( /\.min\.map/, ".min.js\"" ) )
-					.replace( /"sources":\["([^"]+)"\]/,
-						"\"sources\":[\"" + unpathedFile.replace( /\.min\.map/, ".js" ) + "\"]" );
+				text = fs
+					.readFileSync( builtFile, "utf8" )
+					.replace(
+						/"file":"([^"]+)"/,
+						"\"file\":\"" + unpathedFile.replace( /\.min\.map/, ".min.js\"" )
```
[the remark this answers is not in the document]
timmywil: I'm fine making this a template literal, but I worry about changing the logic here and would require testing releases. How about we change that with the release PR?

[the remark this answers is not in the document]
mgol: Sure, that works!

mgol (approved):
LGTM. The comment:
https://github.com/jquery/jquery/pull/5440#discussion_r1519002065
is yet to be addressed but that's going to happen in a separate release PR.

EricMujjona:
#

EricMujjona:
```
``
- - [ ] # > ````~~@mmonteleone ~~`

PR #5439, opened 2024-03-10
Tests: add --hard-retries option to test runner

timmywil:
### Summary ###

3.x version of #5438 

Add the ability to retry by restarting the worker and getting a different browser instance, after all normal retries have been exhausted. This can sometimes be successful for flakey tests when a refresh is not.

### Checklist ###

* [x] New tests have been added to show the fix or feature works
* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #5441, opened 2024-03-10
Some build refactors/improvements

timmywil:
### Summary ###

3.x version of #5440 

- migrate more uses of fs.promises; use node: protocol. I had missed a lot because my vs code search index was off.
- fix size comparison for slim files when the branch is dirty. I fixed this by preferring `+slim.` to `+` in the regex. It was only matching on the latter even for the slim file.
- make compare size cache readable for manual edits. This is in lieu of more complicated options added to the build command. It's not hard to just edit the JSON, but it might as well be readable.
- Fix worker restarts for when a browser fails the acknowledgement test (I just broke this in #5439 )

### Checklist ###


mgol (approved):
LGTM with the same comment as at https://github.com/jquery/jquery/pull/5440#pullrequestreview-1928441721

PR #5443, opened 2024-03-14
Tests: add actual and expected messages to test reporter

timmywil:
### Summary ###

Add the actual and expected messages to the reporter.

Also, I realized that `test.errors` contained all the failed assertions and `test.assertions` contained only passed ones, so I was able to clean up that code a bit.

### Checklist ###

### DIFF
--- a/test/runner/reporter.js
+++ b/test/runner/reporter.js
@@ -10,28 +10,28 @@ export function reportTest( test, reportId, { browser, headless } ) {
 		return;
 	}
 
-	let message = `Test ${ test.status } on ${ chalk.yellow(
+	let message = `${ chalk.bold( `${ test.suiteName }: ${ test.name }` ) }`;
+	message += `\nTest ${ test.status } on ${ chalk.yellow(
 		getBrowserString( browser, headless )
 	) } (${ chalk.bold( reportId ) }).`;
-	message += `\n${ chalk.bold( `${ test.suiteName }: ${ test.name }` ) }`;
 
-	// Prefer failed assertions over error messages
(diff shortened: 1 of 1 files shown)


test/runner/reporter.js
```diff
@@ -10,28 +10,28 @@ export function reportTest( test, reportId, { browser, headless } ) {
...
+	// test.errors contains all failed asssertions
+	if ( test.errors.length ) {
 		for ( const error of test.errors ) {
-			message += `\n${ chalk.red( error.message ) }`;
+			message += "\n";
+			message += `\n${ error.message }`;
 			message += `\n${ chalk.gray( error.stack ) }`;
+			if ( error.expected && error.actual ) {
+				message += `\nexpected: ${ JSON.stringify( error.expected ) }`;
+				message += `\nactual: ${ chalk.red( JSON.stringify( error.actual ) ) }`;
```
mgol, line 27: Could we also do a diff? It can be helpful when comparing objects. It's quite easy to do with the `diff` package; a colored example: https://www.npmjs.com/package/diff#basic-example-in-node

mgol (approved):
Approving but I'd love to see proper diffing in the console; sometimes it can really help.

test/runner/reporter.js
mgol: BTW, feel free to tackle this in a separate PR; it may even be cleaner to have the diffing feature in a separate commit on `main`.

timmywil: Sure, we'll do that in another PR.

PR #5444, opened 2024-03-14
Tests: add actual and expected messages to test reporter

timmywil:
### Summary ###

3.x version of #5443 

Add the actual and expected messages to the reporter.

Also, I realized that `test.errors` contained all the failed assertions and `test.assertions` contained only passed ones, so I was able to clean up that code a bit.

### Checklist ###


mgol (approved):
Approving with the same comment as in https://github.com/jquery/jquery/pull/5443#pullrequestreview-1936079159

PR #5451, opened 2024-03-18
Attributes: Only convert true, false & an empty string

mgol:
### Summary ###

The HTML spec defines boolean attributes:
https://html.spec.whatwg.org/#boolean-attributes
that often correlate with boolean properties. If the attribute is missing, it correlates with the `false` property value, if it's present - the `true` property value. The only valid values are an empty string or the attribute name.

jQuery tried to be helpful here and treated boolean attributes in a special way in the `.attr()` API:
1. For the getter, as long as the attribute was present, it was returning the attribute name lowercased, ignoring the value.
2. For the setter, it was removing the attribute when `false` was passed; otherwise, it was ignoring the passed value and set the attribute - interestingly, in jQuery `>=3` not lowercased anymore.

The problem is the spec occasionally converts boolean attributes into ones with additional attribute values with special behavior - one such example is the new `"until-found"` value for the `hidden` attribute. Our setter normalization means passing those values is impossible with jQuery.

This patch takes a conservative approach, with backwards compatibility in mind, and makes the following changes:
1. It explicitly only applies the logic to the few most popular attribute names.
2. The getter only normalizes the empty string to the lowercased name, returning other attribute values unchanged - it seems unlikely that the empty string value will ever mean anything else than the attribute name repeated.
3. The setter continues to treat `false` as mean attribute removal and `true` as setting the value to the attribute name. All the other inputs are aplied unchanged.

With this change, all existing jQuery unit tests were already passing.

### Checklist ###

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -97,18 +97,27 @@ if ( isIE ) {
 	};
 }
 
-// HTML boolean attributes have special behavior:
-// we consider the lowercase name to be the only valid value, so
-// getting (if the attribute is present) normalizes to that, as does
-// setting to any non-`false` value (and setting to `false` removes the attribute).
+// HTML boolean attributes have special behavior - the attribute presence
+// corresponds to the `true` value and the lack of it - to `false`. The only
+// officially valid values are an empty string or the attribute name.
 // See https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes
+//
+// Because of that, jQuery used to normalize the value to the name in both
+// the getter & setter. However, the spec occasionally adds more values, making
+// the attribute non-boolean anymore.
+//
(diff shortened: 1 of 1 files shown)


mgol:
Size comparison:
```
main @f80e78ef3e7ded1fc693465d02dfb07510ded0ab
   raw     gz Filename
   -30    -29 dist/jquery.min.js
   -30    -26 dist/jquery.slim.min.js
   -30    -25 dist-module/jquery.module.min.js
   -30    -25 dist-module/jquery.slim.module.min.js
```

The PR - if this approach is accepted - still needs unit tests to confirm the new behavior. It is a bit reassuring that it didn't break any of our existing tests.

mgol:
We've decided against this approach; closing.

Issue #5388, opened 2024-01-10
Update the rules for boolean attributes for 4.0


mgol:
Our list of boolean attributes is outdated; it is missing several boolean attributes from [HTML Attributes](https://html.spec.whatwg.org/multipage/indices.html#attributes-3), includes a `scoped` that isn't even present there, and treats [`hidden`](https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute) as boolean despite the expansion of that attribute to be an enumeration including "until-found" by https://github.com/whatwg/html/commit/f5def65bda4d01a6a1e105fcf13fe45201b4f48a.

_Originally posted by @gibson042 in https://github.com/jquery/jquery/pull/5384#discussion_r1445103967_

mgol:
Oh, interesting. Perhaps we should update this list for 4.0 as this can be considered a breaking change.

Or we should get rid of this manual normalization, depending on users doing the right thing here. The biggest issue are possibly falsy checks on values since an empty string is also a valid value. This was also raised by @timmywil in a past issue where I proposed a similar simlification: https://github.com/jquery/jquery/issues/2946.

If we're worried about compat for such things, my guess would be we could limit this list to just a few entries like `checked` & `selected`. I feel keeping the list open-ended leads us to the same problems as the logic for auto-appending `px` - the list is never complete and modifying this is both a maintenance burden as well as a potential source of application breakage. (the `px` behavior was change to have a closed list in https://github.com/jquery/jquery/pull/4055)

Our list right now is:
* `checked`
* `selected`
* `async`
* `autofocus`
* `autoplay`
* `controls`
* `defer`
* `disabled`
* `hidden`
* `ismap`
* `loop`
* `multiple`
* `open`
* `readonly`
* `required`
* `scoped`

As Richard mentioned, `hidden` is no longer even correct here but removing it from the list is a compat break so it'd be good to fit it into 4.0.

If we want to keep the logic for at least some attributes, I could see us doing so just for `checked` & `selected`; maybe `disabled`. That said, I'd expect `hidden` to also be a commonly used one and a good case where our behavior causes issues for future standards changes. If another name from the list gets similar treatment in standards, we won't be able to adjust without another breaking change. It might be more future-proof to give up on this normalization in 4.0 in the end.

Thoughts, @timmywil @gibson042?

mgol:
Tentatively adding it to `4.0.0` blockers so that we don't release before discussing this first.

mgol:
[...]

PR #5452, opened 2024-03-18
Attributes: Make `.attr( name, false )` remove for all non-ARIA attrs

mgol:
### Summary ###

The HTML spec defines boolean attributes:
https://html.spec.whatwg.org/#boolean-attributes
that often correlate with boolean properties. If the attribute is missing, it correlates with the `false` property value, if it's present - the `true` property value. The only valid values are an empty string or the attribute name.

jQuery tried to be helpful here and treated boolean attributes in a special way in the `.attr()` API:
1. For the getter, as long as the attribute was present, it was returning the attribute name lowercased, ignoring the value.
2. For the setter, it was removing the attribute when `false` was passed; otherwise, it was ignoring the passed value and set the attribute - interestingly, in jQuery `>=3` not lowercased anymore.

The problem is the spec occasionally converts boolean attributes into ones with additional attribute values with special behavior - one such example is the new `"until-found"` value for the `hidden` attribute. Our setter normalization means passing those values is impossible with jQuery. Also, new boolean attributes are introduced occasionally and jQuery cannot easily add them to the list without incurring breaking changes.

This patch removes any special handling of boolean attributes - the getter returns the value as-is and the setter sets the provided value.

To provide better backwards compatibility with the very frequent `false` value provided to remove the attribute, this patch makes `false` trigger attribute removal for ALL non-ARIA attributes. ARIA attributes are exempt from the rule since many of them recognize `"false"` as a valid value with semantics different than the attribute missing. To remove an ARIA attribute, use `.removeAttr()` or pass `null` as the value to `.attr()` which doesn't have this exception.

Fixes gh-5388

Richard Gibson is added as a co-author.

### Checklist ###

* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com: https://github.com/jquery/api.jquery.com/issues/1243
* [x] If needed, a Migrate issue/PR was created at https://github.com/jquery/jquery-migrate: https://github.com/jquery/jquery-migrate/issues/504

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -38,7 +38,14 @@ jQuery.extend( {
 		}
 
 		if ( value !== undefined ) {
-			if ( value === null ) {
+			if ( value === null ||
+
+				// For compat with previous handling of boolean attributes,
+				// remove when `false` passed. For ARIA attributes -
+				// many of which recognize a `"false"` value - continue to
+				// set the `"false"` value as jQuery <4 did.
+				( value === false && name.toLowerCase().indexOf( "aria-" ) !== 0 ) ) {
+
 				jQuery.removeAttr( elem, name );
 				return;
 			}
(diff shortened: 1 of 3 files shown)


mgol:
Size differences:
```
main @f80e78ef3e7ded1fc693465d02dfb07510ded0ab
   raw     gz Filename
  -273   -109 dist/jquery.min.js
  -273   -102 dist/jquery.slim.min.js
  -273   -104 dist-module/jquery.module.min.js
  -273   -100 dist-module/jquery.slim.module.min.js
```

mgol:
Issues created:
* https://github.com/jquery/api.jquery.com/issues/1243
* https://github.com/jquery/jquery-migrate/issues/504

gibson042 (approved):
Thanks!

timmywil (approved):
Incredible how little the tests needed to change!

PR #5454, opened 2024-03-22
Tests: improve diffing for values of different types

timmywil:
### Summary ###

Differences are:
- use `diffChars` when both expected and actual are strings
- use `diffWords` for everything after that

Objects, arrays, and numbers are diffed as before.

### Checklist ###

### DIFF
--- a/test/runner/listeners.js
+++ b/test/runner/listeners.js
@@ -57,6 +57,7 @@
 			// Serialize Symbols as string representations so they are
 			// sent over the wire after being stringified.
 			if ( typeof value === "symbol" ) {
+
 				// We can *describe* unique symbols, but note that their identity
 				// (e.g., `Symbol() !== Symbol()`) is lost
 				var ctor = Symbol.keyFor( value ) !== undefined ? "Symbol.for" : "Symbol";
--- a/test/runner/listeners.js
+++ b/test/runner/listeners.js
@@ -53,6 +53,14 @@
 					return nu;
 				}
 			}
+
+			// Serialize Symbols as string representations so they are
+			// sent over the wire after being stringified.
+			if ( typeof value === "symbol" ) {
+				var ctor = Symbol.keyFor( value ) !== undefined ? "Symbol.for" : "Symbol";
+				return ctor + "(" + JSON.stringify( value.description ) + ")";
+			}
+
 			return value;
 		}
 		return derez( object );
--- a/test/runner/listeners.js
+++ b/test/runner/listeners.js
@@ -57,6 +57,8 @@
 			// Serialize Symbols as string representations so they are
 			// sent over the wire after being stringified.
 			if ( typeof value === "symbol" ) {
+				// We can *describe* unique symbols, but note that their identity
+				// (e.g., `Symbol() !== Symbol()`) is lost
 				var ctor = Symbol.keyFor( value ) !== undefined ? "Symbol.for" : "Symbol";
 				return ctor + "(" + JSON.stringify( value.description ) + ")";
 			}
--- a/test/runner/reporter.js
+++ b/test/runner/reporter.js
@@ -3,11 +3,14 @@ import { getBrowserString } from "./lib/getBrowserString.js";
 import { prettyMs } from "./lib/prettyMs.js";
 import * as Diff from "diff";
 
-const rdquo = /"/g;
+function serializeForDiff( value ) {
 
-function quoteWrap( value ) {
+	// Use naive serialization for everything except types with confusable values
 	if ( typeof value === "string" ) {
-		return `"${ value.replace( rdquo, "\\\"" ) }"`;
+		return JSON.stringify( value );
+	}
+	if ( typeof value === "bigint" ) {
+		return `${ value }n`;
 	}
 	return `${ value }`;
 }
@@ -76,12 +79,12 @@ export function reportTest( test, reportId, { browser, headless } ) {
 
 					// Diff the characters of strings
 					diff = Diff.diffChars( error.expected, error.actual );
-				} else if ( error.expected != null && error.actual != null ) {
+				} else {
 
 					// Diff everything else as words
 					diff = Diff.diffWords(
-						quoteWrap( error.expected ),
-						quoteWrap( error.actual )
+						serializeForDiff( error.expected ),
+						serializeForDiff( error.actual )
 					);
 				}
 
--- a/test/runner/reporter.js
+++ b/test/runner/reporter.js
@@ -3,9 +3,11 @@ import { getBrowserString } from "./lib/getBrowserString.js";
 import { prettyMs } from "./lib/prettyMs.js";
 import * as Diff from "diff";
 
+const rdquo = /"/g;
+
 function quoteWrap( value ) {
 	if ( typeof value === "string" ) {
-		return `"${ value }"`;
+		return `"${ value.replace( rdquo, "\\\"" ) }"`;
 	}
 	return `${ value }`;
 }
@@ -57,7 +59,7 @@ export function reportTest( test, reportId, { browser, headless } ) {
 					diff = Diff.diffJson( error.expected, error.actual );
 				} else if (
 					typeof error.expected === "number" &&
-					typeof error.expected === "number"
+					typeof error.actual === "number"
 				) {
 
 					// Diff numbers directly
--- a/test/runner/reporter.js
+++ b/test/runner/reporter.js
@@ -3,6 +3,13 @@ import { getBrowserString } from "./lib/getBrowserString.js";
 import { prettyMs } from "./lib/prettyMs.js";
 import * as Diff from "diff";
 
+function quoteWrap( value ) {
+	if ( typeof value === "string" ) {
+		return `"${ value }"`;
+	}
+	return `${ value }`;
+}
+
 export function reportTest( test, reportId, { browser, headless } ) {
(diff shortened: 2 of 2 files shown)


test/runner/reporter.js
```diff
@@ -3,6 +3,13 @@ import { getBrowserString } from "./lib/getBrowserString.js";
 import { prettyMs } from "./lib/prettyMs.js";
 import * as Diff from "diff";
 
+function quoteWrap( value ) {
+	if ( typeof value === "string" ) {
+		return `"${ value }"`;
```
mgol, line 8: We should also escape inner quotes in such a case.

mgol (commented):
One comment

test/runner/reporter.js
timmywil: I've tested if the value has quotes and they are already escaped by the template literal.

mgol: What do you mean? `value` is just the value from the test, I assume. If it contains double quotes and you print the string and the printing utility indicates the string by double quotes then yes, it will escape them. But the string doesn't include any escaping.

In other words, if `value === 'foo "bar" baz'`, then this function will return `'"foo "bar" baz"'` which is not what we want.

timmywil: I'm not sure I understand. I *do* want quotes to be left alone and have the whole string still be wrapped by quotes if the value is a string.

timmywil: `'"foo "bar" baz"'` <- this is exactly what I want, but I think it would actually print as `'"foo \"bar\" baz"'`, which is fine with me. I want it clear the value is a string and the other value is not (we already know they are not both strings) and that the string in fact contains double quotes.

test/runner/reporter.js
```diff
@@ -57,30 +68,35 @@ export function reportTest( test, reportId, { browser, headless } ) {
 						diff = [ { removed: true, value: `${ value }` } ];
 					}
 				} else if (
-					typeof error.expected === "boolean" &&
-					typeof error.actual === "boolean"
+					typeof error.expected === "string" &&
```
mgol, line 71: BTW, there's a typo in the number part - it checks for a `number` typeof but it checks `error.expected` twice.

test/runner/reporter.js
mgol: I disagree `"foo "bar" baz"` is what we want. This looks nonsensical and is a JS syntax error. If we indicate a string by outer quotes, inner quotes should be escaped. Right now, for:
```js
assert.equal( 42, "foo \"bar\" baz", "message" );
```
we get:
```
expected: "foo \"bar\" baz"
actual: 42
"foo "bar" baz"42
```

The diff looks extremely confusing to me because it looks like we have two strings in there with a weird `bar` between them. I'd expect to see:
```
expected: "foo \"bar\" baz"
actual: 42
"foo \"bar\" baz"42
```
Or just skip the diff in this case. We get the biggest benefits from the diff from matching actual & expected types. Then we can skip the manual quoting & using `diffWords` completely.

A slightly unrelated remark - we are not printing any data for mismatched symbols. The object we get in the reporter is missing `actual` & `expected` properties there, perhaps because symbols by default are skipped during `JSON.stringify`. But perhaps it's fine here, it's not like we'll have many symbol-related tests...

timmywil: ok, I thought I had seen it print with escape backslashes already, but that was wrong. I've finally come around and will escape the double quotes.

test/runner/reporter.js
```diff
@@ -3,6 +3,15 @@
 import { prettyMs } from "./lib/prettyMs.js";
 import * as Diff from "diff";
 
+const rdquo = /"/g;
+
+function quoteWrap( value ) {
+	if ( typeof value === "string" ) {
+		return `"${ value.replace( rdquo, "\\\"" ) }"`;
```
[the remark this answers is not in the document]
mgol: @timmywil I think this alert was applicable and we need to properly escape at least some backlash characters as well here...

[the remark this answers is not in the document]
timmywil: Do you have a case in mind? I can't think of any.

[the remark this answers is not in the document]
mgol: `quoteWrap('\\"')` would return `'"\\\\""'`. If you try to evaluate the value, i.e. `"\\\\""`, you will get a syntax error since `\\\\` is actually two backslashes and the last but one `"` is effectively unescaped.

Basically, if we escape a quote by prepending it with a backslash, an existing backslash in the input string will lead to nonsensical output.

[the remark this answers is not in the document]
timmywil: My goal is to show that one value is a string while the other is not, (i.e. to differentiate between `"undefined"` and `undefined`. But, I wasn't concerned about syntax errors from copying/pasting console output when including the extra quotes, as long as there was no syntax error when printing it, which is not the issue here (and therefore is not a security issue).

I interpret the outer quotes as discrete characters and never cared if there were other double quotes in the string. In other words, I would never include the other quotes in my copying. It's similar to printing strings as a different color, but you can't see that in github actions.

That said, I don't want to keep going around in circles, so I'd just as soon give up my quest to differentiate `"undefined"` and `undefined`.

test/runner/reporter.js
```diff
@@ -57,30 +70,35 @@ export function reportTest( test, reportId, { browser, headless } ) {
...
+					diff = Diff.diffChars( error.expected, error.actual );
+				} else if ( error.expected != null && error.actual != null ) {
 
-					// Diff everything else as characters
-					diff = Diff.diffChars( `${ error.expected }`, `${ error.actual }` );
+					// Diff everything else as words
+					diff = Diff.diffWords(
+						quoteWrap( error.expected ),
+						quoteWrap( error.actual )
+					);
```
gibson042, line 85: I think the necessary insight here is that we want to diff type-mismatched values _after serialization_, which `quoteWrap` a) fails to convey, and b) currently does so inconsistently (e.g., it escapes `"` in strings by `\`-prefixing, but does not similarly escape `\`).
```suggestion
					// Diff everything else as words (post-serialization)
					diff = Diff.diffWords(
						serializeForDiff( error.expected ),
						serializeForDiff( error.actual )
					);
```
with
```js
function serializeForDiff( value ) {
	// Use naive serialization for everything except types with confusable values
	if ( typeof value === "string" ) {
		return JSON.stringify(value);
	} else if ( typeof value === "bigint" ) {
		return `${ value }n`;
	} else if ( typeof value === "symbol" ) {
		const ctor = Symbol.keyFor(value) !== undefined ? "Symbol.for" : "Symbol";
		return `${ctor}(${JSON.stringify(value.description)})`;
	}
	return `${ value }`;
}
```

timmywil: I like that solution.

mgol: I like that, too! Using `JSON.stringify` on strings is a great idea. ❤️

timmywil: I'll try it and post some examples.

timmywil: Something I realized while testing is that Symbols do not get sent over the wire (they get removed when passed through `JSON.stringify`. So, I moved that part to listeners.js.

image
image

timmywil: Also, I switched to just show the diff all the time when at least one is non-null. I found I preferred to see it as confirmation.

test/runner/listeners.js
```diff
@@ -53,6 +53,14 @@
 					return nu;
 				}
 			}
+
+			// Serialize Symbols as string representations so they are
+			// sent over the wire after being stringified.
+			if ( typeof value === "symbol" ) {
+				var ctor = Symbol.keyFor( value ) !== undefined ? "Symbol.for" : "Symbol";
+				return ctor + "(" + JSON.stringify( value.description ) + ")";
+			}
```
gibson042, line 62: Note that `JSON.stringify` also fails to round-trip bigints, NaN, ±Infinity, and undefined. If we start caring about those cases, we should switch to a higher-order encoding similar to [Endo "Smallcaps"](https://github.com/endojs/endo/blob/1813f35432e266a4998c090f32b184b166d5d47f/packages/marshal/src/encodeToSmallcaps.js#L47-L67) (_in which e.g. the array `[0, 42n, NaN, -Infinity, undefined, "#"]` would be encoded as `[0, "+42", "#NaN", "#-Infinity", "#undefined", "!#"]`_).

test/runner/listeners.js
```diff
@@ -53,6 +53,14 @@
 					return nu;
 				}
 			}
+
+			// Serialize Symbols as string representations so they are
+			// sent over the wire after being stringified.
+			if ( typeof value === "symbol" ) {
+				var ctor = Symbol.keyFor( value ) !== undefined ? "Symbol.for" : "Symbol";
```
gibson042, line 60:
```suggestion
				// We can *describe* unique symbols, but note that their identity
				// (e.g., `Symbol() !== Symbol()`) is lost
				var ctor = Symbol.keyFor( value ) !== undefined ? "Symbol.for" : "Symbol";
```

gibson042 (approved):

mgol (approved):
Thanks for following with the review! LGTM, Richard’s suggestions also look good to apply before merging.

test/runner/listeners.js
timmywil: I don't think we need to go that far.

mgol (approved):

PR #5455, opened 2024-03-22
Tests: improve diffing for values of different types

timmywil:
### Summary ###

3.x version of #5454

Differences are:
- use `diffChars` when both expected and actual are strings
- use `diffWords` for everything after that

Objects, arrays, and numbers are diffed as before.

### Checklist ###


mgol (approved):
LGTM with the same suggestions as in the `main` PR.

PR #5460, opened 2024-03-28
Tests: share queue/browser handling for all worker types

timmywil:
### Summary ###

- one queue to rule them all: browserstack, selenium, and jsdom
- retries and hard retries are now supported in selenium
- selenium tests now re-use browsers in the same way as browserstack

This simplifies things in a way that will be easier to manage across projects. In fact, we can consider adding browserstack tests to UI and it wouldn't change much. Granted, we still may not want to do that because it adds the burden of support for each browser tested.

Sample test run to make sure nothing broke: https://github.com/timmywil/jquery/actions/runs/8468961534/job/23203167696

Will make a 3.x PR when this one gets through review.

### Checklist ###


test/runner/reporter.js
```diff
@@ -115,10 +115,11 @@ export function reportTest( test, reportId, { browser, headless } ) {
 }
 
 export function reportEnd( result, reportId, { browser, headless, modules } ) {
+	const fullBrowser = getBrowserString( browser, headless );
 	console.log(
-		`\n\nTests for ${ chalk.yellow( modules.join( ", " ) ) } on ${ chalk.yellow(
-			getBrowserString( browser, headless )
-		) } finished in ${ prettyMs( result.runtime ) } (${ chalk.bold( reportId ) }).`
```
timmywil, line 121: This just aligns with migrate and UI in terms of the order.

test/runner/browsers.js
```diff
@@ -1,25 +1,248 @@
...
-	"Mobile Safari",
-	"jsdom"
-];
-
-// A function that can be used to update the above list.
-export async function getAvailableBrowsers() {
-	const browsers = await getBrowsers( { flat: true } );
-	const available = [ ...new Set( browsers.map( ( { browser } ) => browser ) ) ];
-	return available.concat( "jsdom" );
+import chalk from "chalk";
```
mgol, line 1: This file is seen as new but most of it is the same as the previous `test/runner/browserstack/browsers.js` with minor modifications. It's unfortunate that we cannot see a better diff here.

One way around it would be to keep this file in `test/runner/browserstack/browsers.js` and only move it in a separate PR later; that should preserve the diff, even across the file rename later.

Do you think it's worth trying that?

test/runner/run.js
```diff
@@ -112,8 +110,6 @@ export async function run( {
 				);
 				report.total = total;
 
-				cleanupJSDOM( reportId, { verbose } );
```
mgol, line 115: Why is this not needed anymore? Do we handle cleanup differently?

test/runner/run.js
```diff
@@ -288,11 +281,7 @@ export async function run( {
 
 	try {
 		console.log( `Starting Run ${ runId }...` );
-		if ( browserstack ) {
-			await runAllBrowserStack( { verbose } );
-		} else {
-			await runAllSelenium( { concurrency, verbose } );
```
mgol, line 294: Do we abandon the concurrency option?

mgol (commented):
Nice work! I have a few questions/remarks.

test/runner/browsers.js
```diff
@@ -1,25 +1,248 @@
...
+			headless,
+			url,
+			verbose
+		} );
+		worker = {
+			quit: () => driver.quit()
+		};
+	}
+
+	worker.debug = !!debug;
```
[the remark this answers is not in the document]
timmywil: Yes, this is intentional. I meant to put it on the worker before.

test/runner/command.js
```diff
@@ -59,15 +59,23 @@ const argv = yargs( process.argv.slice( 2 ) )
 			"Leave the browser open for debugging. Cannot be used with --headless.",
 		conflicts: [ "headless" ]
 	} )
+	.option( "retries", {
+		alias: "r",
+		type: "number",
+		description: "Number of times to retry failed tests by refreshing the URL."
+	} )
```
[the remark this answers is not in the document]
timmywil: We can, but the change of order seemed logical to me. `retries` and `hard-retries` no longer only apply to `browserstack`, and I moved `run-id` below `browserstack` because it *does* only apply to browserstack.

test/runner/browsers.js
timmywil: It is unfortunate, but I'm not sure it's worth all that. Can still diff locally.

test/runner/run.js
timmywil: Yes, I've integrated jsdom into the new queue in a way that it gets cleaned up in the exit hook along with everything else.

test/runner/run.js
timmywil: No, but this argument was being ignored even before this. Instead, `concurrency` is read from the test objects when running them.

test/runner/browsers.js
mgol: OK; this code is relatively new so it's not such big of a deal. If this affected the actual jQuery source code, it'd be worth to apply such a workaround IMO.

mgol (approved):
LGTM

Issue #5459, opened 2024-03-28
Triggering after an `alert()` in an event handler results in a JS error


AllanJard:
### Description ###

It appears that if I trigger a `blur` event, immediately after an `alert()` inside a click event handler, it results in an error:

> Uncaught TypeError: saved.slice is not a function

This is happening in jQuery 3.7.1 and 4.0.0-beta. But not in 3.5.1. I wonder [if this issue might be related](https://github.com/jquery/jquery/issues/4859).

### Link to test case ###

The way to reproduce this is quite simple:

```js
    $('button').on('click', function (e) {
      alert('Hello');
      
      $(this).trigger('blur');
    })
```

https://jsbin.com/vuquzepidi/edit?html%2Cjs%2Coutput=

I've found that if I use `console.log` instead of `alert()` it works just fine. Also if I put the trigger into a `setTimeout` it works just fine. Likewise if I change it to trigger a `focus` event it works. I did try a `focus` before the `blur`, wondering if the alert might cause the button to loose focus, but no change, and even if it did, I would have expected an error.

Apologies, I haven't yet dug into the code to see what might be happening - I can see in the debugger that `saved` is `true`, but I've not looked to see why that is yet.

dmethvin:
I can reproduce this on Firefox, but not Chrome or Edge. If a blur handler is added the error doesn't occur, but you can see that there is an extra focus/blur for Firefox compared to Chrome.

https://jsbin.com/pigorigazo/edit?html,console,output

AllanJard:
Thank you very much :-).

As always, immense respect for every thing you do for jQuery and Javascript in general!

PR #5466, opened 2024-04-02
Event: Increase robustness of an inner native event in leverageNative

mgol:
### Summary ###

In Firefox, alert displayed just before blurring an element dispatches
the native blur event twice which tripped the jQuery logic if a jQuery blur
handler was not attached before the trigger call.

This was because the `leverageNative` logic part for triggering first checked if
setup was done before (which, for example, is done if a jQuery handler was
registered before for this element+event pair) and - if it was not - added
a dummy handler that just returned `true`. The `leverageNative` logic made that
`true` then saved into private data, replacing the previous `saved` array. Since
`true` passed the truthy check, the second native inner handler treated `true`
as an array, crashing on the `slice` call.

The same issue could happen if a handler returning `true` is attached before
triggering. A bare `length` check would not be enough as the user handler may
return an array-like as well. To remove this potential data shape clash, capture
the inner result in an object with a `value` property instead of saving it
directly.

Since it's impossible to call `alert()` in unit tests, simulate the issue by
replacing the `addEventListener` method on a test button with a version that
calls attached blur handlers twice.

Fixes gh-5459
Ref gh-5236

```
main @399a78ee9fc5802509df462a2851aef1b60b7fbc
   raw     gz Filename
   +31    +12 dist/jquery.min.js
   +31    +14 dist/jquery.slim.min.js
   +31    +13 dist-module/jquery.module.min.js
   +31    +13 dist-module/jquery.slim.module.min.js
```

### Checklist ###

### DIFF
--- a/src/event.js
+++ b/src/event.js
@@ -522,11 +522,12 @@ function leverageNative( el, type, isSetup ) {
 			if ( ( event.isTrigger & 1 ) && this[ type ] ) {
 
 				// Interrupt processing of the outer synthetic .trigger()ed event
-				if ( !saved ) {
+				// Detect `saved` of shape `{ value }` and `false`.
+				if ( !saved.length ) {
 
 					// Store arguments for use when handling the inner native event
-					// There will always be at least one argument (an event object), so this array
-					// will not be confused with a leftover capture object.
+					// There will always be at least one argument (an event object),
+					// so this array will not be confused with a leftover capture object.
 					saved = slice.call( arguments );
 					dataPriv.set( this, type, saved );
 
@@ -541,29 +542,36 @@ function leverageNative( el, type, isSetup ) {
 						event.stopImmediatePropagation();
 						event.preventDefault();
 
-						return result;
+						// Support: Chrome 86+
+						// In Chrome, if an element having a focusout handler is
+						// blurred by clicking outside of it, it invokes the handler
+						// synchronously. If that handler calls `.remove()` on
+						// the element, the data is cleared, leaving `result`
+						// undefined. We need to guard against this.
+						return result && result.value;
 					}
 
-				// If this is an inner synthetic event for an event with a bubbling surrogate
-				// (focus or blur), assume that the surrogate already propagated from triggering
-				// the native event and prevent that from happening again here.
-				// This technically gets the ordering wrong w.r.t. to `.trigger()` (in which the
-				// bubbling surrogate propagates *after* the non-bubbling base), but that seems
-				// less bad than duplication.
+				// If this is an inner synthetic event for an event with a bubbling
+				// surrogate (focus or blur), assume that the surrogate already
+				// propagated from triggering the native event and prevent that
+				// from happening again here.
 				} else if ( ( jQuery.event.special[ type ] || {} ).delegateType ) {
 					event.stopPropagation();
 				}
 
 			// If this is a native event triggered above, everything is now in order
 			// Fire an inner synthetic event with the original arguments
-			} else if ( saved ) {
+			// Exclude `saved` of shape `{ value }` and `false`.
+			} else if ( saved.length ) {
 
 				// ...and capture the result
-				dataPriv.set( this, type, jQuery.event.trigger(
-					saved[ 0 ],
-					saved.slice( 1 ),
-					this
-				) );
+				dataPriv.set( this, type, {
+					value: jQuery.event.trigger(
+						saved[ 0 ],
+						saved.slice( 1 ),
+						this
+					)
+				} );
 
 				// Abort handling of the native event by all jQuery handlers while allowing
 				// native handlers on the same element to run. On target, this is achieved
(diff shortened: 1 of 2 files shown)


src/event.js
```diff
@@ -547,16 +547,13 @@ function leverageNative( el, type, isSetup ) {
 				// If this is an inner synthetic event for an event with a bubbling surrogate
 				// (focus or blur), assume that the surrogate already propagated from triggering
 				// the native event and prevent that from happening again here.
-				// This technically gets the ordering wrong w.r.t. to `.trigger()` (in which the
-				// bubbling surrogate propagates *after* the non-bubbling base), but that seems
-				// less bad than duplication.
```
mgol, line 552: This part of the comment is no longer valid since the spec now mandates what browsers ship, i.e. that the bubbling events are fired after the non-bubbling ones.

mgol:
I'm referring to #5236 here as this PR essentially reverts that one (with added tests). The rationale for simplifying the check in #5236 was valid, it's just that we now have a second case where it's needed.

src/event.js
```diff
@@ -547,16 +547,13 @@ function leverageNative( el, type, isSetup ) {
...
 				} else if ( ( jQuery.event.special[ type ] || {} ).delegateType ) {
 					event.stopPropagation();
 				}
 
 			// If this is a native event triggered above, everything is now in order
 			// Fire an inner synthetic event with the original arguments
-			} else if ( saved ) {
+			} else if ( saved.length ) {
```
mgol, line 556: I wonder if I should also replace the upper `if ( !saved ) {` with a `!saved.length` check. I haven't yet found a way to confuse that other `if` since it's already inside an `isTrigger` block but before 3.7.0 both were checking for `saved.length` instead of just `saved`.

mgol, line 556: In some very edge-case'y code, it's possible to have a `blur` handler returning an array and falling into this check, destructuring the incorrect array. But guarding against that would be way more complex and I doubt many people will hit such an issue.

mgol:
@timmywil @gibson042 PR updated & ready for another review.

src/event.js
```diff
@@ -522,11 +522,12 @@ function leverageNative( el, type, isSetup ) {
...
 				// Interrupt processing of the outer synthetic .trigger()ed event
-				if ( !saved ) {
+				// Detect `saved` of shape `{ value }` and `false`.
+				if ( !saved.length ) {
 
 					// Store arguments for use when handling the inner native event
-					// There will always be at least one argument (an event object), so this array
-					// will not be confused with a leftover capture object.
+					// There will always be at least one argument (an event object),
+					// so this array will not be confused with a leftover capture object.
```
gibson042, line 530: We should experiment with checking `isArray(saved)` rather than `saved.length` (as both point replacements in the `if` conditions and as a dedicated variable like `savedIsArgs = Array.isArray(saved)`). If there's no size increase, then this commentary would be unnecessary.

src/event.js
```diff
@@ -541,29 +542,36 @@ function leverageNative( el, type, isSetup ) {
...
+				// propagated from triggering the native event and prevent that
+				// from happening again here.
 				} else if ( ( jQuery.event.special[ type ] || {} ).delegateType ) {
 					event.stopPropagation();
 				}
 
 			// If this is a native event triggered above, everything is now in order
 			// Fire an inner synthetic event with the original arguments
-			} else if ( saved ) {
+			// Exclude `saved` of shape `{ value }` and `false`.
```
gibson042, line 564:
```suggestion
```

gibson042 (approved):

src/event.js
```diff
@@ -522,11 +522,12 @@ function leverageNative( el, type, isSetup ) {
 			if ( ( event.isTrigger & 1 ) && this[ type ] ) {
```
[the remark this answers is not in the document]
timmywil: Thank you!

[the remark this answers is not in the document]
timmywil: Actually, can we be more specific than "non-array saved" in the first case? What other types might it be?

mgol:
```
main @399a78ee9fc5802509df462a2851aef1b60b7fbc
   raw     gz Filename
   +31    +12 dist/jquery.min.js
   +31    +14 dist/jquery.slim.min.js
   +31    +13 dist-module/jquery.module.min.js
   +31    +13 dist-module/jquery.slim.module.min.js
```

src/event.js
mgol: Changing `saved.length` checks to `Array.isArray( saved )` results in a size increase:
```
   raw     gz Filename
   +16     +7 dist/jquery.min.js
   +16     +1 dist/jquery.slim.min.js
   +16     +4 dist-module/jquery.module.min.js
   +16      0 dist-module/jquery.slim.module.min.js
```

src/event.js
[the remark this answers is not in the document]
mgol: I added Richard's comment plus a short listing of possible `saved` data types, PTAL.

gibson042 (approved):

timmywil (approved):

PR #5478, opened 2024-04-24
Tests: Make the beforeunload event tests work regardless of extensions

mgol:
### Summary ###

Some browser extensions, like React DevTools, send messages to the content area. Since our beforeunload event test listens for all messages, it used to catch those as well, failing the test.

Add a `source` field to the payload JSON and check for it before treating the message as coming from our own test to make sure the test passes even with such browser extensions installed.

### Checklist ###

### DIFF
--- a/test/data/event/onbeforeunload.html
+++ b/test/data/event/onbeforeunload.html
@@ -4,17 +4,18 @@
 	<script>
 		function report( event ) {
 			var payload = {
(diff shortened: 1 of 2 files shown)


mgol:
Landed on `main` in 399a78ee9fc5802509df462a2851aef1b60b7fbc and on `3.x-stable` in 5d83fadb6be83bcda54a87c1c0d73534e5aad6fb.

PR #5480, opened 2024-04-25
Fix potential github action smells

ceddy4395:
### Summary ###

Hey! 🙂
I want to contribute the following changes to your workflow:
- Steps should only perform a single command
- Run CI on multiple language versions
- Use cache parameter instead of cache option
- Run tests on multiple OS's

(These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html))
### Checklist ###

### DIFF
--- a/.github/workflows/browserstack-dispatch.yml
+++ b/.github/workflows/browserstack-dispatch.yml
@@ -37,17 +37,21 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
-    environment: browserstack
+    runs-on: ${{matrix.os}}
+    environment: browserstack        
     env:
       BROWSERSTACK_USERNAME: ${{ secrets.BROWSERSTACK_USERNAME }}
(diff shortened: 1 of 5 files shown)


timmywil:
Thanks for your contribution. However, I think we're happy with the way it is.

> Steps should only perform a single command

I normally agree with this, but it's okay to group commands when they are grouped logically to accomplish a single task (for instance, to set up SSH for uploading to Filestash). Feedback in GA logs is not necessary for each of those commands, and those commands are frequently run together. We only care if it failed to set up SSH, so they are grouped as a single action step.

> Use cache parameter instead of cache option

I've used this before and prefer to have more control over the cache. The end result is similar, but we control the version used and the cache fallbacks.

> Run CI on multiple language versions
> Run tests on multiple OS's

There's no need for this when testing jQuery. jQuery is a DOM library that is mainly used in the browser. We do run some smoke tests in multiple Node versions, but jQuery is not a Node or OS library. The browserstack workflow covers all of the browsers jQuery supports.

Thanks.

PR #5490, opened 2024-05-22
CSS: `showOnlyWhen()`

faissaloux:
Helloooo 👋🏻 

### Summary ###
Since we usually on the case where we show/hide elements depending on some condition. In this PR I have added a new method `showOnlyWhen()` that will simplify this more.

#### Before ####
```javascript
const checkbox = $("#the-checkbox");
const otherElem = $("#the-other-element");

if (checkbox.attr("checked")) {
    otherElem.show();
} else {
    otherElem.hide();
}
```
#### After ####
```javascript
const checkbox = $("#the-checkbox");
const otherElem = $("#the-other-element");

otherElem.showOnlyWhen(checkbox.attr("checked"));
```

### Checklist ###

If approved I'll be happy to add documentation for it.

### DIFF
--- a/src/css/showHide.js
+++ b/src/css/showHide.js
@@ -82,6 +82,9 @@ jQuery.fn.extend( {
 	hide: function() {
 		return showHide( this );
 	},
+	showOnlyWhen: function( condition ) {
+		return showHide( this, condition );
+	},
 	toggle: function( state ) {
 		if ( typeof state === "boolean" ) {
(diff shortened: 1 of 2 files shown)


mgol:
Thanks for the contribution but this can easily be done as a userland plugin. The "Before" version is not that complex either.

See:
* https://github.com/jquery/jquery/wiki/Adding-new-features
* https://github.com/jquery/jquery/wiki/API-design-guidelines

PR #5492, opened 2024-05-24
Tests: revert concurrency group change

timmywil:
### Summary ###

It's common for us to merge to main and cherry pick to 3.x-stable, so it's best if concurrency is shared between branches, which is effectively what we had already as it matches on workflow name and browser. Ideally, it could also match on the corresponding commit, but it seems the commit message is not available in the github context.

### Checklist ###


mgol (approved):

timmywil:
The Firefox failure will be addressed in a different PR.

PR #5494, opened 2024-05-27
Update buildFragment.js DOM text reinterpreted as HTML

Shivam7-1:
### Summary ###

By using textContent, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.

### Checklist ###


mgol:
We do need HTML there. Try to run tests and you'll see how many of them fail with such a change.

Issue #5489, opened 2024-05-20
Fix CSS test issues in Firefox coming from CSS zoom support


mgol:
### Description ###

Firefox implemented CSS zoom in version 126. That version fails our CSS tests.

Snapshot from `main`:
<img width="948" alt="Screenshot 2024-05-20 at 18 56 59" src="https://github.com/jquery/jquery/assets/1758366/0d510c94-1e49-4a36-aefd-d26146de4e16">

Snapshot from `3.x-stable`:
<img width="946" alt="Screenshot 2024-05-20 at 18 58 42" src="https://github.com/jquery/jquery/assets/1758366/f852f8fe-74f9-4493-a8c5-96cd88db6d08">

The `main` version may just need a test tweak but the `3.x-stable` one seems like a genuine failure. We'll need to investigate before the next release.

### Link to test case ###

N/A, our own tests are failing.

mgol:
It looks like in Firefox 126 stable zoom affects `offsetWidth` values. In 127 Developer Edition & 128 Nightly it does not, just like it does not in Chrome.

Fortunately, [Firefox 126 will not be ESR](https://whattrainisitnow.com/calendar/).

timmywil:
I'd like to fix this before beta.2, even if just a test fix.

mgol:
PRs:
* `main`: #5495
* `3.x-stable`: #5496

PR #5495, opened 2024-05-29
CSS:Tests: Fix tests & support tests under CSS Zoom

mgol:
### Summary ###

Firefox 126+ implements CSS zoom in a way it affects width computed style
very slightly (`100.008px` instead of `100px`); accept that difference.

Add a test for support tests resolving the same under CSS zoom & without one.
That test uncovered Chrome failing the `reliableTrDimensions` support test
under zoom; the test has been fixed.

Fixes gh-5489
Closes gh-5495
Ref gh-5496

`3.x` version: gh-5496

### Checklist ###

* [x] New tests have been added to show the fix or feature works
* ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~

PR #5496, opened 2024-05-29
Tests: Account for small differences in width under zoom in Firefox

mgol:
### Summary ###

Firefox 126+ implements CSS zoom in a way it affects width computed style very slightly (`100.008px` instead of `100px`); accept that difference.

Also, skip the width setter under zoom test in Firefox 126 completely - that version has CSS zoom affecting `offsetWidth` values. This has been fixed in Firefox 127 so it's not worth working around it.

Add a test for support tests resolving the same under CSS zoom & without one.

Fixes gh-5489
Ref gh-5495

### Checklist ###


mgol:
Size diffs:
```
3.x-stable @37bded35229d4753e2f05290e9304011ea2cbade
   raw     gz Filename
   +33     +1 dist/jquery.min.js
   +33     -1 dist/jquery.slim.min.js
```

timmywil (approved):

PR #5501, opened 2024-06-03
Build: improve specificity of eslint config; add ecma versions

timmywil:
### Summary ###

- Audited the eslint flat config and filled in some gaps
- Added ecma versions where appropriate

### Checklist ###

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -15,26 +15,6 @@ export default [
 		]
 	},
 
-	{
-		files: [
-			"eslint.config.js",
-			"Gruntfile.cjs",
-			"test/bundler_smoke_tests/**/*",
-			"test/node_smoke_tests/**",
-			"test/promises_aplus_adapters/**",
-			"test/middleware-mockserver.cjs"
-		],
-		languageOptions: {
-			globals: {
-				...globals.node
-			}
-		},
-		rules: {
-			...jqueryConfig.rules,
-			strict: [ "error", "global" ]
-		}
-	},
-
 	// Source
 	{
 		files: [ "src/**" ],
@@ -42,6 +22,7 @@ export default [
 			import: importPlugin
 		},
 		languageOptions: {
+			ecmaVersion: 2015,
 
 			// The browser env is not enabled on purpose so that code takes
 			// all browser-only globals from window instead of assuming
@@ -159,7 +140,10 @@ export default [
 	// Tests
 	{
 		files: [
-			"test/**"
+			"test/*",
+			"test/data/**",
+			"test/integration/**",
+			"test/unit/**"
 		],
 		ignores: [
 			"test/data/jquery-3.7.1.js",
@@ -170,11 +154,11 @@ export default [
 			"test/data/core/jquery-iterability-transpiled.js"
 		],
 		languageOptions: {
+			ecmaVersion: 2015,
+			sourceType: "script",
 			globals: {
 				...globals.browser,
 				require: false,
-				Promise: false,
-				Symbol: false,
 				trustedTypes: false,
 				QUnit: false,
 				ajaxTest: false,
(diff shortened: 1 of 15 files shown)


eslint.config.js
```diff
@@ -186,32 +170,43 @@ export default [
...
+				supportjQuery: false,
+				originaljQuery: false,
+				original$: false,
+				baseURL: false,
+				externalHost: false
 			}
 		},
 		rules: {
 			...jqueryConfig.rules,
-			strict: [ "error", "function" ],
```
mgol, line 206: How did it work before?

eslint.config.js
```diff
@@ -317,8 +312,8 @@ export default [
 		],
 
 		languageOptions: {
+			ecmaVersion: 2015,
```
mgol, line 315: This needs to be `5` for `dist` and `2015` for `dist-module`.

eslint.config.js
```diff
@@ -355,5 +349,16 @@ export default [
...
+	},
+
+	{
+		files: [
+			"build/**/*.js",
+			"dist/jquery.bundler-require-wrapper.js",
+			"dist/jquery.bundler-require-wrapper.slim.js"
+		],
+		languageOptions: {
+			sourceType: "commonjs"
```
mgol, line 361: Does this mean ESLint is not following the `type` field from `package.json`?

mgol (changes requested):

eslint.config.js
timmywil: It was ignored because we hadn't set the source type. But, most of these files do not meet this requirement, so I left it out.

eslint.config.js
timmywil: Yea, eslint doesn't seem to respect the extra package.json files

eslint.config.js
```diff
@@ -313,34 +308,18 @@ export default [
...
+			"dist-module/jquery.factory.slim.module.js",
 			"dist/jquery.bundler-require-wrapper.js",
-			"dist/jquery.bundler-require-wrapper.slim.js",
-			"dist-module/jquery.node-module-wrapper.js",
-			"dist-module/jquery.node-module-wrapper.slim.js"
+			"dist/jquery.bundler-require-wrapper.slim.js"
 		],
 
 		languageOptions: {
+			ecmaVersion: 2015,
```
mgol, line 317: Do we set `ecmaVersion` to `2015` here for all `dist` & `dist-module` files and then change it to `5` for `dist` files? Does entries below overwrite what's above for matching files?

It still feels a bit confusing to me. Could we avoid setting `2015` for `dist` at all?

eslint.config.js
```diff
@@ -355,5 +334,30 @@ export default [
...
+	{
+		files: [
+			"dist/jquery.js",
+			"dist/jquery.slim.js",
+			"dist/jquery.factory.js",
+			"dist/jquery.factory.slim.js"
+		],
+
+		languageOptions: {
+			ecmaVersion: 5,
```
mgol, line 348: We should also verify minified files are ES5 - and only that: of course, we don't want to actually run any ESLint rules on minified code. I think we've been doing that before migrating to the flat config.

eslint.config.js
```diff
@@ -186,32 +170,43 @@ export default [
...
+	},
+
+	{
+		files: [
+			"test/unit/core.js"
+		],
+		rules: {
+
+			// Core has several cases where unused vars are expected
+			"no-unused-vars": "off"
```
mgol, line 209: Can't we just prefix them with underscores?

mgol (approved):
LGTM with one question

eslint.config.js
timmywil: No, it's more complicated than that. The vars are used in scope using eval, which eslint doesn't recognize.

timmywil:
I was making the PR for 3.x and noticed an implicit global that I think was unintentional. To avoid that in the future, I added the [no-implicit-globals](https://eslint.org/docs/latest/rules/no-implicit-globals#rule-details) rule to both main and 3.x.

mgol (approved):

PR #5504, opened 2024-06-05
Build: improve specificity of eslint config; add ecma versions

timmywil:
### Summary ###

- Audited the eslint flat config and filled in some gaps
- Added ecma versions where appropriate
- Added no-implicit-globals rule outside of tests, but including dist

I'll make a companion PR for 3.x once this one is approved

### Checklist ###


test/unit/effects.js
```diff
@@ -1596,25 +1590,24 @@ QUnit.test( "User supplied callback called after show when fx off (trac-8892)",
...
+		toggle: [ 1 ],
+		slideToggle: []
+	},
+	$divEmpty = jQuery( "<div></div>" ),
+	$divTest = jQuery( "<div>test</div>" ),
+	$divNone = jQuery( "<div style='display: none;'></div>" ),
+	$divInline = jQuery( "<div style='display: inline;'></div>" ),
+	nullParentDisplay = $divEmpty.css( "display" ),
+	underFragmentDisplay = $divTest.css( "display" ),
+	clock = this.clock;
```
mgol, line 1610: I don't like the indentation changes here; can we restore the previous version?

mgol (changes requested):

test/unit/effects.js
timmywil: Not really. The env variable was unused.

mgol: I mean that all variables declared using the group var here are at the same indentation level as the var itself. That seems wrong.

timmywil: It's like that because the first variable is an object. It's normal. However, we can change that by making the first variable a non-object. I'm ambivalent about making that change, though.

mgol: I'm often indifferent to specific code style choices - e.g. I don't care if a project uses single or double quotes as long as it's consistent - but there are good styling principles I care about and one of them is that the code style doesn't make diffs unnecessarily large. For this reason I dislike aligning values in objects which some projects use - changing one key may modify the whole object. This case seems similar; having the first variable declared in the group affect the whole groups seems bad.

Can we perhaps follow what UI did long time ago and tweak the `one-var` rule to allow multiple vars, as long as they're grouped at the top of the function? Then, if each variable here was declared using a separate `var`, we wouldn't risk further changes.

Also - and that may be a bigger concern for me - when a list of variables is indented like that, especially with me knowing how it's usually indented, I keep having a wrong mental model on which level those variables are declared and it seems to me that we're just setting values here and the declaration happened somewhere earlier.

mgol: Also, AFAIK, we don't just such `var` indendation used anywhere in the project - it's always all indented.

timmywil: > Can we perhaps follow what UI did long time ago and tweak the one-var rule to allow multiple vars, as long as they're grouped at the top of the function?

While that would result in fewer line changes here, that's a bigger change. I've moved another var to top instead.

mgol: Yeah, I agree.

FWIW, issues with a single `var` for multiline declarations is one of the reasons why jQuery projects are the only ones where I follow such a code style...

mgol (approved):
LGTM! Thanks for bearing with me. 😉

PR #5506, opened 2024-06-10
Tests: fix supportsScroll feature test in offset

timmywil:
### Summary ###

- iOS 12 and below do not support scrollTop/Left on the
  window or document. Before 82169df, window.supportsScroll
  was always undefined. There were 2 issues: (1) iOS <=12 do
  support scroll props on the body, so the support test
  wasn't helpful. (2) one test checked the wrong window and
  the value was always undefined, which meant those tests
  never ran in any browser.
### Checklist ###

Sample BrowserStack run: https://github.com/timmywil/jquery/actions/runs/9449886713

### DIFF
--- a/test/unit/offset.js
+++ b/test/unit/offset.js
@@ -4,7 +4,7 @@ if ( !includesModule( "offset" ) ) {
 	return;
 }
 
-var alwaysScrollable,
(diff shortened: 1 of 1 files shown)


mgol (approved):
False positives are the worst!

PR #5508, opened 2024-06-12
Build: upgrade dependencies, including eslint 9.4.0 and uglify 3.7.7

timmywil:
### Summary ###

Upgraded all possible dependencies.

- Sinon is already at the latest version that supports IE9.
- Upgraded uglify to 3.7.7, which is the latest that worked with IE9 (I ran an IE9 browserstack run locally to be sure). I tried 3.9.4, which we are using in jquery-migrate, and 3.8.1, but there were failures in traversing ("Permission denied" errors).
- Upgraded eslint to version 9, which required some changes. Mainly, unused arguments in catch expressions error by default in 9+. The config now makes use of `caughtErrorsIgnorePattern`, which is set to `"^_"`, the same as unused paramaters.

### Checklist ###


build/tasks/promises_aplus_tests.js
```diff
@@ -24,7 +24,7 @@ async function runTests() {
 		spawn(
 			command,
 			[ test ].concat( args ),
-			{ stdio: "inherit" }
```
timmywil, line 27: This change is required to spawn on Windows.

build/tasks/promises_aplus_tests.js
```diff
@@ -24,7 +24,7 @@ async function runTests() {
...
+			{ shell: true, stdio: "inherit" }
```
timmywil, line 27: This change is required to spawn on Windows.

eslint.config.js
```diff
@@ -47,6 +48,10 @@ module.exports = [
 				}
 			],
 			"no-implicit-globals": "error",
+			"no-unused-vars": [
+				"error",
+				{ caughtErrorsIgnorePattern: "^e$" }
```
mgol, line 53: I'd prefer to just use the same style as in other places, i.e. a prefix with `_`. But since it's just for `3.x-stable`, I'm fine with this version - as long as we don't plan to propagate it to `main`.

eslint.config.js
```diff
@@ -105,7 +110,6 @@ module.exports = [
 			"test/unit/**"
 		],
 		ignores: [
-			"test/data/jquery-3.7.1.js",
```
mgol, line 108: I'd keep it, perhaps even generalized to `test/data/jquery-*`. It's essentially a vendor file and we should not lint it, even if the lint passes - especially that future changes to ESLint or our config may make it not pass anymore.

mgol (changes requested):

eslint.config.js
mgol: If you do generalize it, please backport it to `main` as well.

timmywil: It's not linted fyi. eslint 9 was shooting warnings about this file, so I moved it to global ignores at the top, which avoids parsing it altogether. That said, I forgot you had requested this be generic in the past so now is a good time for that.

eslint.config.js
timmywil: I considered that, but I think our instinct is to use `e`, as was evident from about 90% of them. Because it's a separate rule, I think we can consider it separate from function parameters.

mgol: What’s the reason this wasn’t needed on `main`?

BTW, if we’re special casing unused catch parameters in a different way than the rest of the code then I see no point; let’s just not check `catch` at all. What’s the point? It will always have one parameter, whether used or not. Checking `catch` gl for unused parameters will make sense in jQuery 5 where we’ll be able to use a parameterless `catch`.

If we remove `catch` parameter checking, we’ll be able to remove some modifications from this PR.

timmywil: I do see a point, although minor. It's simply to ensure that all of these are the same. Having `err` some places and `e` in others, which not a big deal, is something we can clean up with this rule.

eslint.config.js
timmywil: It's generalized now

eslint.config.js
mgol: @timmywil but this is not a rule ensuring all catch parameters are named the same, only unused ones. I’d rather avoid such special one-off rules.

timmywil: Ensuring the unused ones are all the same is still useful imo.

timmywil: oh, and this wasn't needed on main because we can't upgrade eslint to v9 there yet.

timmywil: FWIW, I'd be more inclined to change this to use `_` than to remove the rule altogether.

mgol: Both options work for me!

timmywil: Ok, done. Also, I've been switching branches quite often and found it annoying that 3.x always failed lint until I removed any built files from a main branch build. I've fixed that by adding them to ignore in eslint and .gitignore.

mgol (approved):

eslint.config.js
mgol: > I've fixed that by adding them to ignore in eslint and .gitignore.

Sounds good to me. I’m not hitting this issue as I keep separate clones for the `main` & `3.x-stable` branches.

PR #5509, opened 2024-06-12
Build: upgrade dependencies; fix bundler tests on windows

timmywil:
### Summary ###

Upgraded all possible dependencies

- Sinon is already at the latest version that supports IE11
- We can't upgrade eslint to version 9 until eslint-plugin-import supports it. See https://github.com/import-js/eslint-plugin-import/issues/2948
- account for newly unused parameters in the slim builds after rollup upgrade. See [comment](https://github.com/jquery/jquery/pull/5509/files#diff-a32a0887ed9d1d707bbb3b845b7df7fd40e673c47e7b60a3ebd896b68d3b8839R361).

### Checklist ###


test/bundler_smoke_tests/lib/run-webpack.js
```diff
@@ -1,16 +1,11 @@
...
-const dirname = path.dirname( fileURLToPath( import.meta.url ) );
-const configPath = path.resolve( dirname, "..", "webpack.config.cjs" );
 
 // See https://webpack.js.org/api/node/#webpack
 export async function runWebpack() {
 	return new Promise( async( resolve, reject ) => {
 		console.log( "Running Webpack" );
 
-		const { default: config } = await import( configPath );
+		const { default: config } = await import( "../webpack.config.cjs" );
```
timmywil, line 8: A full path with `c:` cannot be passed to import. Using the relative path instead was the least complicated solution.

mgol (approved):

test/bundler_smoke_tests/lib/run-webpack.js
mgol: TIL!

timmywil:
test jquery ignore rule generalized

PR #5517, opened 2024-07-11
Release: fix running pre/post release scripts in windows

timmywil:
- also fix tagging the release in the dist repo

### Summary ###

I was attempting the release and ran into some minor issues. I did some more manual dry runs of the pre-release.sh and post-release.sh scripts and this *should* work now.

### Checklist ###

### DIFF
--- a/build/release/post-release.sh
+++ b/build/release/post-release.sh
@@ -33,7 +33,9 @@ cd -
 npm run release:dist $1 $2
 cd $dist
 git add -A
-npm version $1
+git commit -m "Release: $1"
+# -s to sign and annotate tag (recommended for releases)
+git tag -s $1 -m "Release: $1"
 
 # Wait for confirmation from user to push changes to dist repo
 read -p "Press enter to push changes to dist repo"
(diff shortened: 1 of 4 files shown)


build/release/post-release.sh
```diff
@@ -33,7 +33,9 @@ cd -
 npm run release:dist $1 $2
 cd $dist
 git add -A
-npm version $1
+git commit -m "Release: $1"
+# -s to sign and annotate tag (recommended for releases)
+git tag -s $1 -m "Release: $1"
```
timmywil, line 38: `npm version` wasn't quite right. Aside from the need to commit first anyway, the correct version is already set at this point in the package.json. But more importantly, there's no `.npmrc` in the jquery-dist repo, so the tag name would be wrong.

mgol: The correct version is needed not just in `package.json`, but also in the lockfile - and there, depending on the version, even twice! For this reason, it’s be safer to rely on npm version; the format can be tweaked, I believe.

timmywil: This is for jquery-dist. There's no lockfile.

mgol: Ah, right. There are no dependencies there so it wouldn’t be useful to commit one.

mgol (approved):

PR #5519, opened 2024-07-15
Release: ensure builds have the proper version

timmywil:
### Summary ###

I attempted the release this morning and noticed a couple things after the initial steps.

1. The version was not set in the built files. `VERSION` needed to be set in the environment. The `version` is not available in script contexts at the `init` stage. This adds a full build to `after:bump` with the new version, which happens just before publishing to npm. I added `cross-env` to do this so it works on Windows as well.
2. The last build to run in `npm test` was `selector-native`, so the main jquery.js file was not correct.

Also:

- order hooks in execution order
- upgrade actions in the new verify workflow
- verify.js did not need to export and could be run from the file

### Checklist ###

### DIFF
--- a/build/release/verify.js
+++ b/build/release/verify.js
@@ -24,12 +24,13 @@ const REGISTRY_URL = "https://registry.npmjs.org/jquery";
 
 const rstable = /^(\d+\.\d+\.\d+)$/;
 
-export async function verifyRelease( { version } = {} ) {
+async function verifyRelease( { version } = {} ) {
 	if ( !version ) {
 		version = process.env.VERSION || ( await getLatestVersion() );
 	}
-	console.log( `Checking jQuery ${ version }...` );
 	const release = await buildRelease( { version } );
+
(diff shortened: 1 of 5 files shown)


build/release/verify.js
```diff
@@ -139,8 +140,8 @@ async function buildRelease( { version } ) {
 			.filter( ( dirent ) => dirent.isFile() )
 			.map( async( dirent ) => ( {
 				name: dirent.name,
-				path: path.basename( dirent.path ),
-				contents: await readFile( path.join( dirent.path, dirent.name ), "utf8" )
+				path: path.basename( dirent.parentPath ),
+				contents: await readFile( path.join( dirent.parentPath, dirent.name ), "utf8" )
```
timmywil, line 144: This was just because `path` was recently marked deprecated.

mgol (approved):
It'd be a bit cleaner to have scripts (i.e. runnable files) separate from libraries (e.g. files exporting utilities). Not a huge concern, though.

PR #5521, opened 2024-07-17
Release: use correct build date in release verification

timmywil:
### Summary ###

- the date is actually the date of the commit *prior* to the tag commit, as the files are built and then committed.
- also, the CDN should still be checked for non-stable releases, and should use different filenames (including in the map files).
- certain files should be skipped when checking the CDN.
- removed file diffing because it ended up being far too noisy, making it difficult to find the info I needed.
- because the build script required an addition, release verification will not work until the next release.

### Checklist ###


.gitignore
```diff
@@ -31,8 +31,8 @@ npm-debug.log*
 /test/data/qunit-fixture.js
 
 # Release artifacts
-changelog.*
-contributors.*
+changelog.html
+contributors.html
```
timmywil, line 35: `changelog.md` is now committed

.release-it.cjs
```diff
@@ -20,10 +20,12 @@ module.exports = {
 		changelog: "npm run release:changelog -- ${from} ${to}",
 		commitMessage: "Release: ${version}",
 		getLatestTagFromAllRefs: true,
+		pushRepo: "git@github.com:jquery/jquery.git",
```
timmywil, line 23: Pushed to my fork at first. It was easy to fix, but now will be avoided.

build/release/verify.js
```diff
@@ -22,7 +19,15 @@ const SRC_REPO = "https://github.com/jquery/jquery.git";
 const CDN_URL = "https://code.jquery.com";
 const REGISTRY_URL = "https://registry.npmjs.org/jquery";
 
-const rstable = /^(\d+\.\d+\.\d+)$/;
+const excludeFromCDN = [
+	"jquery.factory.js",
+	"jquery.factory.slim.js",
+	"jquery.factory.module.js",
+	"jquery.factory.slim.module.js",
```
mgol, line 26: Since the logic is to basically exclude `package.json` & factory files, it would be safer to exclude the pattern `jquery.factory.*` or regex `/^jquery\.factory\./` and match file names against it.

mgol (approved):
One suggestion. LGTM overall, though.

build/release/verify.js
```diff
@@ -80,16 +93,26 @@ async function verifyRelease( { version } = {} ) {
...
-		console.log( `jQuery ${ version } is reproducible!` );
+		console.log( `jQuery ${ version } is reproducible! All files match!` );
 	} else {
+		console.log();
+		for ( const file of matchingFiles ) {
+			console.log( `✅ ${ file }` );
+		}
+		console.log();
+		for ( const file of mismatchingFiles ) {
+			console.log( `❌ ${ file }` );
```
timmywil, line 113: These characters do seem to print fine in GH actions.

timmywil:
I was debugging the workflow a bit on my fork and made a couple tweaks:

- fix an issue with the build in GH actions where setting the `env` option in the `exec()` was overriding the `PATH` variable
- if the build is not reproducible, print *all* files and whether each matched

.release-it.cjs
```diff
@@ -14,16 +14,19 @@ module.exports = {
 			"sed -i 's/main\\/AUTHORS.txt/${version}\\/AUTHORS.txt/' package.json",
 		"after:bump": "cross-env VERSION=${version} npm run build:all",
 		"before:git:release": "git add -f dist/ dist-module/ changelog.md",
-		"after:release": `bash ./build/release/post-release.sh \${version} ${ blogURL }`
+		"after:release": "echo 'Run the following to complete the release:' && " +
+			`echo './build/release/post-release.sh $\{version} ${ blogURL }'`
 	},
 	git: {
-		changelog: "npm run release:changelog -- ${from} ${to}",
+		changelog: "node build/release/changelog.js ${from} ${to}",
```
mgol, line 21: What's the reason for this change?

build/release/verify.js
```diff
@@ -22,7 +19,12 @@ const SRC_REPO = "https://github.com/jquery/jquery.git";
 const CDN_URL = "https://code.jquery.com";
 const REGISTRY_URL = "https://registry.npmjs.org/jquery";
 
-const rstable = /^(\d+\.\d+\.\d+)$/;
+const excludeFromCDN = [
+	/^package.json$/,
```
mgol, line 23:
```suggestion
	/^package\.json$/,
```

mgol (approved):
Minor remarks to address, otherwise LGTM

.release-it.cjs
timmywil: I tried to describe this in my comment, but I worried that wasn't clear enough. Essentially, when you run an npm script, it first logs the command. The problem with that in this case is that *everything* in stdout is used in the GH release notes. Calling the node script directly fixes that.

mgol: Ah, I see. A code comment in one of these places with this note could be useful.

.husky/commit-msg
```diff
@@ -1,4 +1,2 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
```
mgol, line 2: Why was it needed before and why is it not needed anymore? Doesn't husky rely on it?

The same comment applies to the `pre-commit` hook.

build/release/verify.js
```diff
@@ -33,24 +35,34 @@ async function verifyRelease( { version } = {} ) {
...
-				const cdnContents = await fetch( new URL( file.name, CDN_URL ) ).then(
-					( res ) => res.text()
-				);
-				if ( cdnContents !== file.contents ) {
-					console.log( `${ file.name } is different from the CDN:` );
-					diffFiles( file.contents, cdnContents );
+	// Check all files against the CDN
+	await Promise.all(
+		release.files
+			.filter( ( file ) => !excludeFromCDN.some( ( re ) => re.test( file.name ) ) )
```
mgol, line 44: `!exclude...` is a double negative. I feel this would be a bit easier to read:
```suggestion
			.filter( ( file ) => excludeFromCDN.every( ( re ) => !re.test( file.name ) ) )
```

mgol (approved):
Looks mostly good, I have some minor comments.

mgol:
Looks mostly good, I have some minor comments. Nice work!

.husky/commit-msg
timmywil: It was removed automatically when upgrading husky. Apparently no longer needed.

PR #5522, opened 2024-07-18
Release: migrate release process to release-it (3.x)

timmywil:
### Summary ###

3.x version of release migration, beginning with #5512, and including all changes from subsequent release PRs, including #5521. Some things were left in that were only applicable to the main branch and have no affect on 3.x-stable. I left those things in so that the code can be as close as possible to main, such as leaving in the factory regex in the verify script or leaving in the removal of `dist-module` in the `build:clean` script.

*Authors*
- Checking and updating authors has been migrated to a custom script in the repo

*Changelog*
- changelogplease is no longer maintained
- generate changelog in markdown for GitHub releases
- generate changelog in HTML for blog posts
- generate contributors list in HTML for blog posts

*dist*
- clone dist repo, copy files, and commit/push
- commit tag with dist files on main branch; remove dist files from main branch after release

*cdn*
- clone cdn repo, copy files, and commit/push
- create versioned and unversioned copies in cdn/
- generate md5 sums and archives for Google and MSFT

*build*
- implement reproducible builds and verify release builds
  * uses the last modified date for the latest commit
  * See https://reproducible-builds.org/
- the verify workflow also ensures all files were properly published to the CDN and npm

*docs*
- the new release workflow is documented at build/release/README.md

*verify*
- use the last modified date of the commit before the tag
- use versioned filenames when checking map files on the CDN
- skip factory and package.json files when verifying CDN

*misc*
- now that we don't need the jquery-release script and now that we no longer need to build on Node 10, we can use ESM in all files in the build folder
- limit certain workflows to the main repo (not forks)
- version has been set to the previously released version 3.7.1, as release-it expects
- release-it added the `preReleaseBase` option and we now always set it to `1` in the npm script. This is a noop for stable releases.

Ref jquery/jquery-release#114

### Checklist ###


build/tasks/lib/compareSize.js
```diff
@@ -1,9 +1,9 @@
-import chalk from "chalk";
 import fs from "node:fs/promises";
 import { promisify } from "node:util";
 import zlib from "node:zlib";
 import { exec as nodeExec } from "node:child_process";
-import isCleanWorkingDir from "./lib/isCleanWorkingDir.js";
+import chalk from "chalk";
+import isCleanWorkingDir from "./isCleanWorkingDir.js";
```
mgol, line 6: There's one difference from `main` in this file - the `catch` parameter in line 39 is named `_` instead of `e` - do we want to align?

build/tasks/promises_aplus_tests.js
```diff
@@ -1,17 +1,9 @@
-"use strict";
```
mgol, line 1: Maybe let's rename JS files in `test/promises_aplus_adapters/` to have the `cjs` extension to align with `main`? They will work in Node as well.

mgol (approved):
I compared the output with `main` and I have a few minor comments, but nothing blocking the merge. Nice to see 3.x migrated as well!

build/tasks/lib/compareSize.js
timmywil: `e` should be an error. I wonder why there's no lint error on main.

mgol: The `3.x` ESLint config has this override for `build/**`:
https://github.com/jquery/jquery/blob/3b2330240c9941f8749b4162167b18de8f13ea38/eslint.config.js#L287
which doesn't exist on `main`; that's why:
https://github.com/jquery/jquery/blob/a7d3383e669ff3eee2d2f5ce5a4bb0b9b79bafb1/eslint.config.js#L298-L302

timmywil:
Things to change on main that came up in this PR:

1. Adjust eslint settings for build files to also include rules for unused variables
2. Add package.json comment to dist

build/tasks/lib/compareSize.js
timmywil: Nvm, I forgot. The real reason is we've upgraded eslint in the 3.x branch to v9, which has `caughtErrorsIgnorePatter`, but the main branch is on v8, which is required for the import plugin.

Issue #5214, opened 2023-02-17
jQuery.cleanData traversal issue


PSPvZ:
### Description ###

We use a jQuery.event.special remove handler to write components that clean themselves up when they are removed from the DOM. We ran into an issue where certain components did not execute their cleanup. I've managed to trace it back to the fact that one of the components removed its own element from the DOM. This causes jQuery.cleanData to skip the next element in its list.

From what I can tell jQuery.cleanData iterates over a live list of elements. If one of the already traversed elements is removed from the DOM, all elements in the list behind it will move one index down. This causes the iteration to skip one element. For example, if it's cleaning up [0:A, 1:B, 2:C] and the cleanup of A removes itself, the list becomes [0:B, 1:C]; iteration continues at index 1, so B is skipped.

### Link to test case ###

either https://jsfiddle.net/qy2tx8za/
or https://jsbin.com/yadekowafa/edit?html,css,js,console,output

There are three elements that have the remove handler. When removed, they should log the contents of their data-ext attribute. Elements with the removeself class will remove themselves from the DOM. The first element does this, causing cleanData to skip the second one, but still cleanup the third.

timmywil:
We're going to look into switch the selector list to one that isn't live, i.e. with `querySelectorAll`

AnuragMishraa:
One way to solve this problem is to create a copy of the live list of elements before iterating over it, so that changes to the DOM during cleanup don't affect the original list. 
This way, even if an element removes itself from the DOM during cleanup, it won't affect the iteration of the original list.

Another solution is to reverse the order of iteration, so that you start with the last element in the list and work your way backwards. This way, if an element removes itself from the DOM, it won't affect the cleanup of any earlier elements.

Can you please assign me this issue, I am Interested to work on it.

PSPvZ:
If you choose the "reverse order" solution, be mindful that in the [0:A, 1:B, 2:C] case, C might for some reason remove element A or B (I can't think of why anyone would do this, but it technically could happen). This would lead to C being cleaned up twice, once at position 2 and once at position 1. Even worse would be if C removed both A and B, because position 1 would end up out-of-bounds.

AnuragMishraa:
Yes, that's a good point to keep in mind if you choose the "reverse order" solution. In general, it's always a good idea to be mindful of the potential side effects of DOM manipulation when iterating over a live list of elements, especially if you're modifying the list while iterating over it.

[...]

PR #5523, opened 2024-07-21
Manipulation: Make jQuery.cleanData not skip elements during cleanup

ac-mmi:
This pull request addresses [#5214](https://github.com/jquery/jquery/issues/5214) issue in jQuery.cleanData where certain elements were being skipped during the cleanup process. The problem occurred when an element removed itself from the DOM during its own cleanup, causing subsequent elements in the list to be missed.

### DIFF
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -133,51 +133,46 @@ jQuery.extend( {
 		return clone;
 	},
 
-	cleanData: function( elems ) {
-	    var data, elem, type,
-	        special = jQuery.event.special,
-	        i = 0;
+	cleanData: function(elems) {
+		var data, elem, type,
+		    special = jQuery.event.special,
+		    i = 0;
 
-	    // Convert elems to a native array if it is not already
-	    elems = jQuery.makeArray(elems);
+		// Convert elems to a native array if it is not already
+		elems = jQuery.makeArray(elems);
 
-	    while (i < elems.length) {
-	        elem = elems[i];
+		while (i < elems.length) {
+		  elem = elems[i];
 
-	        if (acceptData(elem)) {
-	            if ((data = elem[dataPriv.expando])) {
-	                if (data.events) {
-	                    for (type in data.events) {
-	                        if (special[type]) {
-	                            jQuery.event.remove(elem, type);
+		  if (acceptData(elem)) {
+		    if ((data = elem[dataPriv.expando])) {
+		      if (data.events) {
+		        for (type in data.events) {
+		          if (special[type]) {
+		            jQuery.event.remove(elem, type);
+		          // This is a shortcut to avoid jQuery.event.remove's overhead
+		          } else {
+		            jQuery.removeEvent(elem, type, data.handle);
+		          }
+		        }
+		      }
 
-	                        // This is a shortcut to avoid jQuery.event.remove's overhead
-	                        } else {
-	                            jQuery.removeEvent(elem, type, data.handle);
-	                        }
-	                    }
-	                }
+		      // Support: Chrome <=35 - 45+
+		      // Assign undefined instead of using delete, see Data#remove
+		      elem[dataPriv.expando] = undefined;
+		    }
+		    if (elem[dataUser.expando]) {
+		      // Support: Chrome <=35 - 45+
+		      // Assign undefined instead of using delete, see Data#remove
+		      elem[dataUser.expando] = undefined;
+		    }
 
-	                // Support: Chrome <=35 - 45+
-	                // Assign undefined instead of using delete, see Data#remove
-	                elem[dataPriv.expando] = undefined;
-	            }
-	            if (elem[dataUser.expando]) {
-
-	                // Support: Chrome <=35 - 45+
-	                // Assign undefined instead of using delete, see Data#remove
-	                elem[dataUser.expando] = undefined;
-	            }
-
-	            // Remove the element from the array while preserving the index positions
-	            elems.splice(i, 1);
-	        } else {
-	            i++;
-	        }
-	    }
-	}
-
-
+		    // Remove the element from the array while preserving the index positions
+		    elems.splice(i, 1);
+		  } else {
+		    i++;
+		  }
+		}
 } );
 
 jQuery.fn.extend( {
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -133,46 +133,49 @@ jQuery.extend( {
 		return clone;
 	},
 
-	cleanData: function(elems) {
+	cleanData: function( elems ) {
 		var data, elem, type,
-		    special = jQuery.event.special,
-		    i = 0;
+			special = jQuery.event.special,
+			i = 0;
 
 		// Convert elems to a native array if it is not already
-		elems = jQuery.makeArray(elems);
+		elems = jQuery.makeArray( elems );
 
-		while (i < elems.length) {
-		  elem = elems[i];
+		while ( i < elems.length ) {
+			elem = elems[ i ];
 
-		  if (acceptData(elem)) {
-		    if ((data = elem[dataPriv.expando])) {
-		      if (data.events) {
-		        for (type in data.events) {
-		          if (special[type]) {
-		            jQuery.event.remove(elem, type);
-		          // This is a shortcut to avoid jQuery.event.remove's overhead
-		          } else {
-		            jQuery.removeEvent(elem, type, data.handle);
-		          }
-		        }
-		      }
+			if ( acceptData( elem ) ) {
+				if ( ( data = elem[ dataPriv.expando ] ) ) {
+					if ( data.events ) {
+						for ( type in data.events ) {
+							if ( special[ type ] ) {
+								jQuery.event.remove( elem, type );
 
-		      // Support: Chrome <=35 - 45+
-		      // Assign undefined instead of using delete, see Data#remove
-		      elem[dataPriv.expando] = undefined;
-		    }
-		    if (elem[dataUser.expando]) {
-		      // Support: Chrome <=35 - 45+
-		      // Assign undefined instead of using delete, see Data#remove
-		      elem[dataUser.expando] = undefined;
-		    }
+								// This is a shortcut to avoid jQuery.event.remove's overhead
+							} else {
+								jQuery.removeEvent( elem, type, data.handle );
+							}
+						}
+					}
 
-		    // Remove the element from the array while preserving the index positions
-		    elems.splice(i, 1);
-		  } else {
-		    i++;
-		  }
+					// Support: Chrome <=35 - 45+
+					// Assign undefined instead of using delete, see Data#remove
+					elem[ dataPriv.expando ] = undefined;
+				}
+				if ( elem[ dataUser.expando ] ) {
+
+					// Support: Chrome <=35 - 45+
+					// Assign undefined instead of using delete, see Data#remove
+					elem[ dataUser.expando ] = undefined;
+				}
+
+				// Remove the element from the array while preserving the index positions
+				elems.splice( i, 1 );
+			} else {
+				i++;
+			}
 		}
+	}
 } );
 
 jQuery.fn.extend( {
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -141,32 +141,32 @@ jQuery.extend( {
 	    // Convert elems to a native array if it is not already
 	    elems = jQuery.makeArray(elems);
 
-	    while ( i < elems.length ) {
-	        elem = elems[ i ];
+	    while (i < elems.length) {
+	        elem = elems[i];
 
-	        if ( acceptData( elem ) ) {
-	            if ( ( data = elem[ dataPriv.expando ] ) ) {
-	                if ( data.events ) {
-	                    for ( type in data.events ) {
-	                        if ( special[ type ] ) {
-	                            jQuery.event.remove( elem, type );
+	        if (acceptData(elem)) {
+	            if ((data = elem[dataPriv.expando])) {
+	                if (data.events) {
+	                    for (type in data.events) {
+	                        if (special[type]) {
+	                            jQuery.event.remove(elem, type);
 
 	                        // This is a shortcut to avoid jQuery.event.remove's overhead
 	                        } else {
-	                            jQuery.removeEvent( elem, type, data.handle );
+	                            jQuery.removeEvent(elem, type, data.handle);
 	                        }
 	                    }
 	                }
 
 	                // Support: Chrome <=35 - 45+
 	                // Assign undefined instead of using delete, see Data#remove
-	                elem[ dataPriv.expando ] = undefined;
+	                elem[dataPriv.expando] = undefined;
 	            }
-	            if ( elem[ dataUser.expando ] ) {
+	            if (elem[dataUser.expando]) {
 
 	                // Support: Chrome <=35 - 45+
 	                // Assign undefined instead of using delete, see Data#remove
-	                elem[ dataUser.expando ] = undefined;
+	                elem[dataUser.expando] = undefined;
 	            }
 
 	            // Remove the element from the array while preserving the index positions
@@ -176,6 +176,7 @@ jQuery.extend( {
 	        }
 	    }
 	}
+
 
 } );
 
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -138,12 +138,7 @@ jQuery.extend( {
 			special = jQuery.event.special,
 			i = 0;
 
-		// Convert elems to a native array if it is not already
-		elems = jQuery.makeArray( elems );
-
-		while ( i < elems.length ) {
-			elem = elems[ i ];
-
+		for ( ; ( elem = elems[ i ] ) !== undefined; i++ ) {
 			if ( acceptData( elem ) ) {
 				if ( ( data = elem[ dataPriv.expando ] ) ) {
 					if ( data.events ) {
@@ -151,7 +146,7 @@ jQuery.extend( {
 							if ( special[ type ] ) {
 								jQuery.event.remove( elem, type );
 
-								// This is a shortcut to avoid jQuery.event.remove's overhead
+							// This is a shortcut to avoid jQuery.event.remove's overhead
 							} else {
 								jQuery.removeEvent( elem, type, data.handle );
 							}
@@ -168,11 +163,6 @@ jQuery.extend( {
 					// Assign undefined instead of using delete, see Data#remove
 					elem[ dataUser.expando ] = undefined;
 				}
-
-				// Remove the element from the array while preserving the index positions
-				elems.splice( i, 1 );
-			} else {
-				i++;
 			}
 		}
 	}
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -134,38 +134,46 @@ jQuery.extend( {
 	},
 
 	cleanData: function( elems ) {
-		var data, elem, type,
-			special = jQuery.event.special,
-			i = 0;
+    var data, elem, type,
+        special = jQuery.event.special,
+        i = 0;
+    
+     elems = jQuery.makeArray(elems);
+    while ( i < elems.length ) {
+        elem = elems[ i ];
 
-		for ( ; ( elem = elems[ i ] ) !== undefined; i++ ) {
-			if ( acceptData( elem ) ) {
-				if ( ( data = elem[ dataPriv.expando ] ) ) {
-					if ( data.events ) {
-						for ( type in data.events ) {
-							if ( special[ type ] ) {
-								jQuery.event.remove( elem, type );
+        if ( acceptData( elem ) ) {
+            if ( ( data = elem[ dataPriv.expando ] ) ) {
+                if ( data.events ) {
+                    for ( type in data.events ) {
+                        if ( special[ type ] ) {
+                            jQuery.event.remove( elem, type );
 
-							// This is a shortcut to avoid jQuery.event.remove's overhead
-							} else {
-								jQuery.removeEvent( elem, type, data.handle );
-							}
-						}
-					}
+                        // This is a shortcut to avoid jQuery.event.remove's overhead
+                        } else {
+                            jQuery.removeEvent( elem, type, data.handle );
+                        }
+                    }
+                }
 
-					// Support: Chrome <=35 - 45+
-					// Assign undefined instead of using delete, see Data#remove
-					elem[ dataPriv.expando ] = undefined;
-				}
-				if ( elem[ dataUser.expando ] ) {
+                // Support: Chrome <=35 - 45+
+                // Assign undefined instead of using delete, see Data#remove
+                elem[ dataPriv.expando ] = undefined;
+            }
+            if ( elem[ dataUser.expando ] ) {
 
-					// Support: Chrome <=35 - 45+
-					// Assign undefined instead of using delete, see Data#remove
-					elem[ dataUser.expando ] = undefined;
-				}
-			}
-		}
-	}
+                // Support: Chrome <=35 - 45+
+                // Assign undefined instead of using delete, see Data#remove
+                elem[ dataUser.expando ] = undefined;
+            }
+
+            // Remove the element from the array while preserving the index positions
+            elems.splice(i, 1);
+        } else {
+            i++;
+        }
+    }
+}
 } );
 
 jQuery.fn.extend( {
--- a/src/manipulation.js
+++ b/src/manipulation.js
@@ -134,46 +134,49 @@ jQuery.extend( {
 	},
 
 	cleanData: function( elems ) {
-    var data, elem, type,
-        special = jQuery.event.special,
-        i = 0;
-    
-     elems = jQuery.makeArray(elems);
-    while ( i < elems.length ) {
-        elem = elems[ i ];
+	    var data, elem, type,
+	        special = jQuery.event.special,
+	        i = 0;
 
-        if ( acceptData( elem ) ) {
-            if ( ( data = elem[ dataPriv.expando ] ) ) {
-                if ( data.events ) {
-                    for ( type in data.events ) {
-                        if ( special[ type ] ) {
-                            jQuery.event.remove( elem, type );
+	    // Convert elems to a native array if it is not already
+	    elems = jQuery.makeArray(elems);
 
-                        // This is a shortcut to avoid jQuery.event.remove's overhead
-                        } else {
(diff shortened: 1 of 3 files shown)


mgol (changes requested):
Thanks for the PR. Please don't change code formatting, though. Run `npm test` locally & you'll see the build fails; you need to fix it.

Also, all such changes need tests.

ac-mmi:
> Thanks for the PR. Please don't change code formatting, though. Run `npm test` locally & you'll see the build fails; you need to fix it.
> 
> Also, all such changes need tests.

Done

ac-mmi:
@mgol  Done

src/manipulation.js
```diff
@@ -134,38 +134,50 @@ jQuery.extend( {
...
+	            }
+	            if (elem[dataUser.expando]) {
+
+	                // Support: Chrome <=35 - 45+
+	                // Assign undefined instead of using delete, see Data#remove
+	                elem[dataUser.expando] = undefined;
+	            }
+
+	            // Remove the element from the array while preserving the index positions
+	            elems.splice(i, 1);
```
mgol, line 173: For large `elems` arrays, this could be very expensive. I also don't see why we'd need to do that.

mgol (changes requested):
`jQuery.cleanData` is not an API you'd call yourself; jQuery calls it. One of the suggested solutions on the issue was to update the internal `getAll` function to avoid `getElementsByTagName` which returns a live collection and just falling back to always use `querySelectorAll`. That also wouldn't require any changes to `jQuery.cleanData` and it'd reduce size instead of increasing it.

We'd need some perf tests done for this change, though, to make sure we don't regress significantly.

Also, as I mentioned before, all such changes need tests; there are no added in this PR. Make also sure existing tests & static checks are passing; `npm test` definitely does not pass on your PR.

The majority of the work here is with the perf tests & writing good unit tests, not in actual code changes.

ac-mmi:
@mgol All test pass done.
Screenshot (1809)

src/manipulation.js
ac-mmi: @mgol  reason was to preserve the index position while deletion

mgol:
Most of [my last review](https://github.com/jquery/jquery/pull/5523#pullrequestreview-2202714588) still applies.

ac-mmi:
@mgol Done changed getAll.js

mgol (changes requested):
We still need tests for what this PR is supposed to fix. Also, we need to have some performance tests comparing before & after done with test cases & results posted in a comment on this PR. This is to make sure we don’t regress too much on performance.

ac-mmi:
@mgol I conducted a performance test on the updated `getAll()` functionality using 1000 span items. Here are the results:

Original implementation time: 1 ms
New implementation time: 8.6 ms
These results show that the new implementation is significantly slower than the original.

mgol:
@ac-mmi can you share the test code, please?

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,66 @@ testIframe(
...
+
+	return ret;
+}
+
+QUnit.test( "should retrieve all elements with a specific tag", function( assert ) {
+    assert.expect( 1 );
+    var context = document.createElement( "div" );
+    context.innerHTML = "<span></span><span></span>";
+
+    var result = getAll( context, "span" );
```
mgol, line 3135: This test is calling a locally defined `getAll`; this is not actually testing your changes.

What we need instead is a test confirming the issue that led to this PR is fixed - i.e. #5214.

mgol (changes requested):

test/unit/manipulation.js
ac-mmi: added the test for the issue addressing [#5214](https://github.com/jquery/jquery/issues/5214)

ac-mmi:
@mgol added the test for the issue addressing https://github.com/jquery/jquery/issues/5214

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,71 @@ testIframe(
...
+      <div class="guarded" data-elt="three">
+        Guarded 3
+      </div>
+    </div>
+  ` );
+
+	// JavaScript code to be tested
+
+  jQuery.event.special[ "_se-custom-destroy" ] = {
+    remove: function( _handleObj ) {
```
mgol, line 3130: Here & in many other places there are still indentation issues - code in a block not indented, spaces used instead of tabs, etc. Please configure your editor to have visible whitespace, it will make it easier for you to properly follow the style guide.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,71 @@ testIframe(
...
+		var $t = jQuery( this );
+		console.log( $t.data( "elt" ) );
```
mgol, line 3132: Instead of patching `console.log`, use `assert.step` with `assert.verifySteps`:
https://qunitjs.com/api/assert/step/

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,71 @@ testIframe(
...
+      </div>
+      <div class="guarded" data-elt="three">
+        Guarded 3
+      </div>
+    </div>
+  ` );
+
+	// JavaScript code to be tested
+
+  jQuery.event.special[ "_se-custom-destroy" ] = {
```
mgol, line 3129: Where is this name coming from? What does `se` stand for?

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,71 @@ testIframe(
...
+
+QUnit.test( "should handle custom '_se-custom-destroy' event correctly", function( assert ) {
+
+	// Set up the HTML structure using innerHTML within #qunit-fixture
+
+	assert.expect( 2 );
+
+	var fixture = jQuery( "#qunit-fixture" );
+
+	fixture.html( `
```
mgol, line 3113: Don't replace the whole fixture, just append to the existing one. See some nearby tests to see how this is commonly done.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,71 @@ testIframe(
...
+    remove: function( _handleObj ) {
+		var $t = jQuery( this );
+		console.log( $t.data( "elt" ) );
+		if ( $t.is( ".removeself" ) ) {
+			$t.remove();
+			}
+		}
+	};
+
+	// Attach the custom event handler
```
mgol, line 3139:
```suggestion
	// Attach an empty handler to trigger the `remove`
	// logic for the custom event when the element is removed.
```

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,71 @@ testIframe(
...
+
```
mgol, line 3140: This empty line is not needed - there should be no space between a comment and the line it's about.

This applies to many places here.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,71 @@ testIframe(
...
+	logMessages.push( message );
+	};
+
+	// Trigger the event by emptying the container
+
+	jQuery( "#container" ).empty( );
+
+	// Verify that the elements with class 'removeself' were removed
+
+	assert.equal( jQuery( ".removeself" ).length, 0, "Elements with class 'removeself' should be removed" );
```
mgol, line 3159: This doesn't tell us much as we're removing almost everything.

Instead, I'd add one `assert.step` to the contents of the `remove` function and then verify those were called.

mgol (changes requested):
We're getting closer, but I still have some remarks.

Please also share the code for measuring performance so that I can run it by myself.

ac-mmi:
@mgol  Code for measuring performance
`function measurePerformance(fn, context, tag, iterations = 1000) {
    var startTime = performance.now();
    for (let i = 0; i < iterations; i++) {
        fn(context, tag);
    }
    var endTime = performance.now();
    return endTime - startTime;
}

var context = document.createElement("div");
context.innerHTML = "<span></span>".repeat(1000);

// Original implementation using getElementsByTagName
function originalGetAll(context, tag) {
    var ret;
    if (typeof context.getElementsByTagName !== "undefined") {
        ret = context.getElementsByTagName(tag || "*");
    } else if (typeof context.querySelectorAll !== "undefined") {
        ret = context.querySelectorAll(tag || "*");
    } else {
        ret = [];
    }

    if (tag === undefined || (tag && nodeName(context, tag))) {
        return jQuery.merge([context], ret);
    }

    return ret;
}

// New implementation using querySelectorAll
function newGetAll(context, tag) {
    var ret;
    if (typeof context.querySelectorAll !== "undefined") {
        ret = context.querySelectorAll(tag || "*");
    } else {
        ret = [];
    }

    if (tag === undefined || (tag && nodeName(context, tag))) {
        return jQuery.merge([context], ret);
    }

    return ret;
}

function nodeName( elem, name ) {
	return elem.nodeName && elem.nodeName.toLowerCase() === name.toLowerCase();
}

console.log("New implementation time: " + measurePerformance(newGetAll, context, "span") + "ms");

console.log("Original implementation time: " + measurePerformance(originalGetAll, context, "span") + "ms");

`

ac-mmi:
`
function measurePerformance(fn, context, tag, iterations = 1000) {
    var startTime = performance.now();
    for (let i = 0; i < iterations; i++) {
        fn(context, tag);
    }
    var endTime = performance.now();
    return endTime - startTime;
}

var context = document.createElement("div");
context.innerHTML = "<span></span>".repeat(1000);

// Original implementation using getElementsByTagName
function originalGetAll(context, tag) {
    var ret;
    if (typeof context.getElementsByTagName !== "undefined") {
        ret = context.getElementsByTagName(tag || "*");
    } else if (typeof context.querySelectorAll !== "undefined") {
        ret = context.querySelectorAll(tag || "*");
    } else {
        ret = [];
    }

    if (tag === undefined || (tag && nodeName(context, tag))) {
        return jQuery.merge([context], ret);
    }

    return ret;
}

// New implementation using querySelectorAll
function newGetAll(context, tag) {
    var ret;
    if (typeof context.querySelectorAll !== "undefined") {
        ret = context.querySelectorAll(tag || "*");
    } else {
        ret = [];
    }

    if (tag === undefined || (tag && nodeName(context, tag))) {
        return jQuery.merge([context], ret);
    }

    return ret;
}

function nodeName( elem, name ) {
	return elem.nodeName && elem.nodeName.toLowerCase() === name.toLowerCase();
}

console.log("New implementation time: " + measurePerformance(newGetAll, context, "span") + "ms");

console.log("Original implementation time: " + measurePerformance(originalGetAll, context, "span") + "ms");

`

ac-mmi:
@mgol  Code for measuring performance
```
function measurePerformance(fn, context, tag, iterations = 1000) {
    var startTime = performance.now();
    for (let i = 0; i < iterations; i++) {
        fn(context, tag);
    }
    var endTime = performance.now();
    return endTime - startTime;
}

var context = document.createElement("div");
context.innerHTML = "<span></span>".repeat(1000);

// Original implementation using getElementsByTagName
function originalGetAll(context, tag) {
    var ret;
    if (typeof context.getElementsByTagName !== "undefined") {
        ret = context.getElementsByTagName(tag || "*");
    } else if (typeof context.querySelectorAll !== "undefined") {
        ret = context.querySelectorAll(tag || "*");
    } else {
        ret = [];
    }

    if (tag === undefined || (tag && nodeName(context, tag))) {
        return jQuery.merge([context], ret);
    }

    return ret;
}

// New implementation using querySelectorAll
function newGetAll(context, tag) {
    var ret;
    if (typeof context.querySelectorAll !== "undefined") {
        ret = context.querySelectorAll(tag || "*");
    } else {
        ret = [];
    }

    if (tag === undefined || (tag && nodeName(context, tag))) {
        return jQuery.merge([context], ret);
    }

    return ret;
}

function nodeName( elem, name ) {
	return elem.nodeName && elem.nodeName.toLowerCase() === name.toLowerCase();
}

console.log("New implementation time: " + measurePerformance(newGetAll, context, "span") + "ms");

console.log("Original implementation time: " + measurePerformance(originalGetAll, context, "span") + "ms");


```

mgol:
## Construction of a perf test

@ac-mmi the perf tests you provided test just raw performance of various versions of `getAll`. However, what's more important is its effect on real jQuery APIs. Most of `getAll` usage is meant to just find scripts which are usually limited in numbers. However, two methods: `remove` and `empty` - use it to find all the nodes within the current one.

I created a test case using three versions of jQuery:
1. A real jQuery Git one.
2. With a patched `getAll` to remove the `getElementsByTagName` path.
3. With a patched `getAll` to wrap the `getElementsByTagName` path with an additional `slice` call:
    ```js
    if ( typeof context.getElementsByTagName !== "undefined" ) {
    	ret = arr.slice.call(context.getElementsByTagName( tag || "*" ));
    ```
    (`arr` is https://github.com/jquery/jquery/blob/dbc9dac7aecb106b66050342ff8daf1ecdd4239f/src/var/arr.js)

The first version is the fastest but the third one is faster than the second one. The differences between (2) & (3) in Chrome & Safari are small enough that I'd opt to (2), but in Firefox they are massive.

Test case: https://jsperf.app/gifera/1/preview

## Environment

Tests done on macOS Sonoma 14.6.1 (23G93) on a MacBook Pro with the M3 Max chip.

Browsers:
1. Chrome 127.0.6533.100
2. Firefox 129.0
3. Safari 17.6 (19618.3.11.11.5)

## Results

|               | Chrome                    | Firefox                  | Safari                    |
|---------------|---------------------------|--------------------------|---------------------------|
| jQuery Git    | 3,343,353 ±2.70% (🏆)     | 6,055,305 ±2.15% (🏆)     | 7,645,512 ±3.95% (🏆)     |
| no gEBTN      | 2,640,264 ±2.28% (21% ⬇️) | 3,533,072 ±2.11% (42% ⬇️) | 6,172,783 ±4.52% (19% ⬇️) |
| gEBTN + slice | 2,854,539 ±2.40% (15% ⬇️) | 4,848,646 ±1.58% (20% ⬇️) | 6,012,539 ±6.10% (21% ⬇️) |

## Conclusions

The perf difference in Firefox is staggering so I'd go with the `slice` version. That will get us a 15-21% perf cut on `empty` on large-ish structures, hopefully that's acceptable to fix the issue at hand.

ac-mmi:
@mgol Done

mgol:
@ac-mmi please have a look at my other comments, I had a couple of them to the added tests.

mgol (changes requested):
Comments about tests still need to get resolved.

test/unit/manipulation.js
ac-mmi: thats the name given to the event in the code mentioned in  issue [#5214](https://github.com/jquery/jquery/issues/5214)

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
...
+
+	// Set up the HTML structure by appending to the existing fixture
+
+	var fixture = jQuery( "#qunit-fixture" );
+
+	fixture.append( `
+	<div id="container">
+	  <div class="guarded removeself" data-elt="one">
+	    Guarded 1
+	  </div>
```
mgol, line 3119: Indentation is still incorrect. For example, in these few lines, there are:
* tabs mixed with spaces - we only use tabs
* the string inside of the call is indented as the call above - it should be indented more

Also, these tests need to work in IE 11, so no ES 6 is allowed. You need to use regular concatenated strings, as other tests do.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
...
 	}
 );
+
+// Custom event handler test case
+
+QUnit.test( "should handle custom '_se-custom-destroy' event correctly", function( assert ) {
+
+	// Define the expected number of assertions
+
+	assert.expect( 4 );
```
mgol, line 3109: There should be no empty line between a comment & code this comment describes.

This applies to many places in this PR; please analyze all the comments & fix as appropriate.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
...
+
+	// Set up the HTML structure by appending to the existing fixture
+
+	var fixture = jQuery( "#qunit-fixture" );
```
mgol, line 3113: Is there any value in this variable? If anything, it makes more sense to save the element created from the template below to a variable and then call `.appendTo( "#qunit-fixture" )` on it.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
...
+	    Guarded 2
+	  </div>
+	  <div class="guarded" data-elt="three">
+	    Guarded 3
+	  </div>
+	</div>
+	` );
+
+	// Define the custom event handler
+	jQuery.event.special[ "_se-custom-destroy" ] = {
```
mgol, line 3130: We don't need to use the same event name as in the bug report. I don't think it's a good idea as tests don't carry the same context. Let's use something generic, like `removeondestroy`.

test/unit/manipulation.js
mgol: This comment has not been addressed; did you see it? If you're missing comments, I recommend navigating to the Files tab & double checking that all comments are addressed before re-requesting review.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
 		} );
 	}
 );
+
+// Custom event handler test case
+
+QUnit.test( "should handle custom '_se-custom-destroy' event correctly", function( assert ) {
```
mgol, line 3105: This is not about any specific event; also, this wording doesn't follow the style from other tests. How about:
```suggestion
QUnit.test( "Cleans data for events removing elements on `remove` (gh-5214)", function( assert ) {
```
Maybe others have a better suggestion for a more succinct name.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
 		} );
 	}
 );
+
+// Custom event handler test case
```
mgol, line 3103: This comment is unnecessary.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
...
+
+QUnit.test( "should handle custom '_se-custom-destroy' event correctly", function( assert ) {
+
+	// Define the expected number of assertions
```
mgol, line 3107: Avoid comments that don't bring any additional value. If it's obvious from the code what it does, a comment shouldn't be added.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
...
+
+	assert.expect( 4 );
+
+	// Set up the HTML structure by appending to the existing fixture
```
mgol, line 3111: Please remove this comment.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
...
+		var $t = jQuery( this );
+		assert.step( $t.data( "elt" ) );
+		if ( $t.is( ".removeself" ) ) {
+			$t.remove();
+		}
+	}
+	};
+
+	// Attach an empty handler to trigger the `remove`
+	// logic for the custom event when the element is removed.
```
mgol, line 3141: For comparison - this is an example of a non-obvious comment that definitely adds useful context. Such comments should stay, others - not so much.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
...
+
+	jQuery( ".guarded" ).on( "_se-custom-destroy", function( ) { } );
+
+	// Trigger the event by emptying the container
```
mgol, line 3145: This is not triggering the event; it's triggering its removal logic. Let's put that in words here:
```suggestion
	// Trigger the event's removal logic by emptying the container
```

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
...
+
+	jQuery( "#container" ).empty();
+
+	// Verify the steps to ensure the expected elements were processed
```
mgol, line 3149: This comment is unnecessary.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,55 @@ testIframe(
...
+
+	assert.verifySteps( [ "one", "two", "three" ], "Steps should match the data-elt attributes" );
```
mgol, line 3151:
```suggestion
	assert.verifySteps( [ "one", "two", "three" ], "All elements were processed in order" );
```

mgol (changes requested):
Thanks for the updates; see my latest comments. Please make sure to address them all before re-requesting review - I bumped one of the older comments that was not addressed.

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,43 @@ testIframe(
...
+		</div>
+		<div class="guarded" data-elt="three">
+			Guarded 3
+		</div>
+	</div>
+	` ).appendTo( "#qunit-fixture" );
+
+	// Define the custom event handler
+	jQuery.event.special.removeondestroy = {
+	remove: function( _handleObj ) {
```
mgol, line 3123: `_handleObj` is not used so let's just remove it:
```suggestion
	remove: function() {
```

test/unit/manipulation.js
mgol: This has not been done; in particular, the last section:
> Also, these tests need to work in IE 11, so no ES 6 is allowed. You need to use regular concatenated strings, as other tests do.

ac-mmi: done

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,43 @@ testIframe(
...
+
+	// Define the custom event handler
+	jQuery.event.special.removeondestroy = {
+	remove: function( ) {
+		var $t = jQuery( this );
+		assert.step( $t.data( "elt" ) );
+		if ( $t.is( ".removeself" ) ) {
+			$t.remove();
+		}
+	}
```
mgol, line 3129: These lines should be indented one level deeper.

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,43 @@ testIframe(
 		} );
 	}
 );
+
+QUnit.test( "should handle custom 'removeondestroy' event correctly", function( assert ) {
```
mgol, line 3103: This event is just an example, let's describe the issue here, with a reference to the issue number as well. How about:
```suggestion
QUnit.test( "should handle node removal in event's remove hook (gh-5214)", function( assert ) {
```

mgol (changes requested):
Just a few nits, other than that, mostly looks good!

test/unit/manipulation.js
ac-mmi: Done

test/unit/manipulation.js
ac-mmi: Done

mgol (approved):
Thanks, this LGTM, but I want to discuss it with the team before merging.

timmywil (approved):
LGTM

src/manipulation/getAll.js
```diff
@@ -8,7 +9,7 @@ export function getAll( context, tag ) {
 	var ret;
 
 	if ( typeof context.getElementsByTagName !== "undefined" ) {
-		ret = context.getElementsByTagName( tag || "*" );
+		ret = arr.slice.call( context.getElementsByTagName( tag || "*" ) );
```
gibson042, line 12:
```suggestion
		// Use slice to snapshot the live collection from gEBTN
		ret = arr.slice.call( context.getElementsByTagName( tag || "*" ) );
```

gibson042 (approved):

test/unit/manipulation.js
```diff
@@ -3099,3 +3099,43 @@ testIframe(
...
+			"<div class='guarded removeself' data-elt='one'>" +
+				"Guarded 1" +
+			"</div>" +
+			"<div class='guarded' data-elt='two'>" +
+				"Guarded 2" +
+			"</div>" +
+			"<div class='guarded' data-elt='three'>" +
+				"Guarded 3" +
+			"</div>" +
+		"</div>"
```
mgol, line 3118: I've just noticed this uses different indendation based on the intended HTML indentation. This poses two issues:
1. The final HTML string does not have this indentation at all.
2. If we enable indent ESLint rules for test, this code will be auto-fixed to start in one line.

When doing indentation in HTML strings, indentation should be a part of the string, not done outside. See https://github.com/jquery/jquery/blob/6d78c0768d9aa6ba213678724c89af69a1958df6/test/unit/css.js#L646-L670 for an example of such in-string indentation.

mgol (changes requested):
I noticed one more issue.

When doing the change, don't remove the comment we added to `src/manipulation/getAll.js` via a suggestion comment.

test/unit/manipulation.js
ac-mmi: Done

mgol (approved):
Thanks!

mgol:
Landed on `main` in 3cad5c435aa2333c39baa55a8bceb2b6bf1e2721 & on `3.x-stable` in 752e9119abea4524d72f48fef03ffda0e255eb33. Thanks!

PR #5527, opened 2024-07-30
Tests: replace express with basic Node server

timmywil:
### Summary ###

- Removed express, body-parser, and express-body-parser-error-handler
- The new server has support for basic connect-style middleware. No changes were required in existing middleware.
- The server logs responses when run via `npm run test:server`. Output for CLI test commands (`npm run test:unit`) is left unchanged.

### Checklist ###


test/runner/createTestServer.js
```diff
@@ -1,70 +1,186 @@
...
 	// Hook up mock server
-	app.use( mockServer() );
+	use( mockServer() );
 
 	// Serve static files
-	app.post( "/test/data/name.html", ( _req, res ) => {
-		res.send( nameHTML );
-	} );
+	const validMimeTypes = {
+		".js": "application/javascript",
```
mgol, line 129: Do we want to omit `.mjs` & `.cjs` on purpose? It might make sense since many HTTP servers don't serve a JS mime type for those, but I was wondering if this was deliberate. If this was, a code comment could help.

test/runner/createTestServer.js
```diff
@@ -1,70 +1,186 @@
...
+			req.url.startsWith( "/external/" )
+		) {
+			const file = req.parsedUrl.pathname.slice( 1 );
+			const ext = file.slice( file.lastIndexOf( "." ) );
 
-	app.use( "/dist", express.static( "dist" ) );
-	app.use( "/src", express.static( "src" ) );
-	app.use( "/test", express.static( "test" ) );
-	app.use( "/external", express.static( "external" ) );
+			// Allow POST to .html files in tests
```
mgol, line 152: Where do we rely on it and how, considering that the HTML response will not parse a POST body?

test/runner/createTestServer.js
```diff
@@ -1,70 +1,186 @@
...
+					res.writeHead( 500 );
+					res.end();
+				} );
+			} else {
+				console.error( `Invalid file extension: ${ ext }` );
+				res.writeHead( 404 );
+				res.end();
+			}
+		} else {
+			next();
```
mgol, line 181: This `if` above is pretty big; can we short-circuit the `next()` call here by reversing the initial `if`? I think you've already done it for some other bigger route handlers.

mgol (approved):
I left a few comments but nothing major. Good to see this size decrease!

test/runner/createTestServer.js
timmywil: There's one place in ajax tests that expected a 200 for a POST to a static HTML file.

test/runner/createTestServer.js
timmywil: I didn't consider serving `.mjs` and `.cjs`. I'll add a comment.

Krinkle (approved):

PR #5547, opened 2024-09-04
Tests: Run tests on both real Firefox ESRs

mgol:
### Summary ###

1. At the same time, there may be two supported versions of Firefox ESR. Run tests on both, installed locally.
2. Don't run tests on Firefox 115 on BrowserStack - it was added as there's an ESR version of Firefox 115, but ESR versions may be different, e.g. for some time ServiceWorker was disabled on ESR versions: https://bugzilla.mozilla.org/show_bug.cgi?id=1547023

Example test runs:
* https://github.com/mgol/jquery/actions/runs/10710118685/job/29696215652
* https://github.com/mgol/jquery/actions/runs/10710118685/job/29696215823

### Checklist ###

### DIFF
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -31,7 +31,6 @@ jobs:
           - 'Edge_latest-1'
           - 'Firefox_latest'
           - 'Firefox_latest-1'
-          - 'Firefox_115'
           - '__iOS_17'
           - '__iOS_16'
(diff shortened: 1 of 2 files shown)


timmywil (approved):
This looks good, and has the benefit of not having to explicitly set the ESR version.

mgol:
`3.x` version: #5548

Issue #5542, opened 2024-08-26
ESLint should enforce ES 5 in tests


mgol:
### Description ###

#5523 used a template literal in the test, with GitHub Actions all green. ESLint should catch it and it should not allow ES2015+ syntax in test files.

neogy-akash:
Hey @timmywil  @mgol  is it an inherent problem with eslint or can I give the below  override in the rules defined in **eslint-config-jquery**  https://github.com/jquery/eslint-config-jquery

```
overrides: [
		{
			files: ["test/*", "test/data/**", "test/integration/**", "test/unit/**"],
			excludedFiles: [
				"test/data/badcall.js",
				"test/data/badjson.js",
				"test/data/support/csp.js",
				"test/data/support/getComputedSupport.js",
				"test/data/core/jquery-iterability-transpiled.js",
			],
			rules: {
				quotes: ["error", "single", { "allowTemplateLiterals": false }],
			},
		},
	],
```

mgol:
We just need to configure ESLint to require ES5 for test files, as we already do for source. Mentioning individual ES2015+ language features is not necessary.

neogy-akash:
> We just need to configure ESLint to require ES5 for test files, as we already do for source. Mentioning individual ES2015+ language features is not necessary.

Hey, thanks for the reply....should I change the **ecmaversion** property for test files as '5' from '2015 in the eslint config file ?

```
// Tests
	{
		files: [
			"test/*",
			"test/data/**",
			"test/integration/**",
			"test/unit/**"
		],
		ignores: [
			"test/data/badcall.js",
			"test/data/badjson.js",
			"test/data/support/csp.js",
			"test/data/support/getComputedSupport.js",
			"test/data/core/jquery-iterability-transpiled.js"
		],
		languageOptions: {
			_**ecmaVersion: 2015,**_
			sourceType: "script",
			globals: {
				...globals.browser,
				require: false,
				trustedTypes: false,
				QUnit: false,
				ajaxTest: false,
				testIframe: false,
				createDashboardXML: false,
				createWithFriesXML: false,
				createXMLFragment: false,
				includesModule: false,
				moduleTeardown: false,
				url: false,
				q: false,
```

mgol:
Yes, I think that should be enough, but you'll need to check.

neogy-akash:
> Yes, I think that should be enough, but you'll need to check.

[...]

PR #5563, opened 2024-10-07
Build: Enforce ECMAScript 5 in tests via ESLint

neogy-akash:
### Summary ###

#5542 Changed the `ecmaVersion`  for unit tests to ES5 from ES2015. Added **Promise** and **Symbol** keyword in `globals ` and given them the value `false` to   ignore them while linting tests. Polyfill added for this as well as per @mgol.

### Checklist ###

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -159,7 +159,7 @@ export default [
 			"test/data/core/jquery-iterability-transpiled.js"
 		],
 		languageOptions: {
-			ecmaVersion: 2015,
+			ecmaVersion: 5,
 			sourceType: "script",
 			globals: {
 				...globals.browser,
@@ -187,7 +187,9 @@ export default [
 				originaljQuery: false,
 				original$: false,
 				baseURL: false,
-				externalHost: false
+				externalHost: false,
+				Promise: false,
+				Symbol: false
 			}
 		},
 		rules: {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -164,6 +164,8 @@ export default [
 			globals: {
 				...globals.browser,
 				require: false,
+				Promise: false,
+				Symbol: false,
 				trustedTypes: false,
 				QUnit: false,
 				ajaxTest: false,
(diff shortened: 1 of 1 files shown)


eslint.config.js
```diff
@@ -187,7 +187,9 @@ export default [
 				originaljQuery: false,
 				original$: false,
 				baseURL: false,
-				externalHost: false
+				externalHost: false,
+				Promise: false,
+				Symbol: false
```
mgol, line 192: Please add these two directly above `trustedTypes` so that it's close to other built-ins. Also, the diff will be nicer.

mgol (changes requested):
One comment, otherwise looks good

eslint.config.js
neogy-akash: Hi...resolved this.

timmywil (approved):

mgol (approved):

mgol:
Landed on `main` in d74fc265de2bca3060da2e6f5ec00371b16e43ca and on `3.x-stable` in ce7db4adf0524afe1cdac2937374f2aeba073845. Thanks!

neogy-akash:
> Landed on `main` in [d74fc26](https://github.com/jquery/jquery/commit/d74fc265de2bca3060da2e6f5ec00371b16e43ca) and on `3.x-stable` in [ce7db4a](https://github.com/jquery/jquery/commit/ce7db4adf0524afe1cdac2937374f2aeba073845). Thanks!

This was my first open source contribution ever. Thanks for the guidance! @mgol

miyasudokoro:
@neogy-akash I've just made a [pull request for updating eslint-config-jquery](https://github.com/jquery/eslint-config-jquery/pull/24) because it is full of deprecated tags. To update the tags, it would now have to pull in @stylistic/js as a plugin. Your code only pulls in the "rules" object but not the "plugins" object that would make it work, so I think another change in jquery itself will be necessary.

Of course, I don't even know if the maintainer of that project will accept my pull request, but I've done what I could.

PR #5566, opened 2024-10-13
Add jQuery Network Error Test and Hover Test HTML Files

AbdellahBoutmad:
### Summary ###

This PR introduces two HTML test pages: one for testing jQuery network error handling in Firefox, which logs AJAX errors when the server is stopped, and another for demonstrating hover event tracking. Each hover test can be activated with a button, helping developers troubleshoot AJAX and hover functionalities.

### Checklist ###

### DIFF
--- a/test/hovertest.html
+++ b/test/hovertest.html
@@ -1,158 +1,153 @@
-<html>
+<!DOCTYPE html>
+<html lang="en">
 <head>
-<title>Hover tests</title>
-<script src="jquery.js"></script>
-<style>
-/* Remove body dimensions so we can test enter/leave to surrounding browser chrome */
-body, html {
-	border: 0;
-	margin: 0;
-	padding: 0;
-}
-p {
-	margin: 2px 0;
-}
-.hover-box {
-	background: #f33;
-	padding: 3px;
-	margin: 10px 40px 20px 0;
-}
-.hover-status {
-	background: #f66;
-	padding: 1px;
-}
-.hover-inside {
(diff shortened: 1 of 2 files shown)


mgol (changes requested):
Thanks for the PR. A few remarks:
1. In the PR description, you claim the PR introduces two new files. It doesn't actually do that; it modifies existing ones.
2. The templates in the PR use spaces for indentation. Please keep tabs.
3. Many line breaks in JS code are modified, making the code harder to read. Don't merge chains into longer, more condensed lines.
4. I'm open to improving the files, but if we want to do it, I'd like to do it right:
    * Replace removed or deprecated APIs; that includes replacing `.live()`, `.delegate()` & `.click()` with proper delegated `.on()`. Note that `.live()` examples should be replaced, not removed, unless there's an equivalent example already using `delegate` or - preferably - `.on()`.
    * Fix the JS code to follow the jQuery code style. The simplest way to do it may be to copy it to an existing JS file, make sure ESLint passes and then copy back to HTML.

There's also another similar file for manual testing - `delegatetest.html` - that could use similar treatment, but it can be done separately.

It might also be good to discuss with the team whether these files actually serve any purpose. We're not using them at the moment as can be seen by the `.live()` errors being ignored for so long, they are not a part of any CI or a release process, etc... Marking for team discussion.

mgol:
Since the PR was closed, I extracted my last remark to an issue: #5571.

PR #5569, opened 2024-10-16
Selector: Remove jQuery.expr.filters

mgol:
### Summary ###

`jQuery.expr.filters` have been deprecated since jQuery `3.0.0` but this removal was missed along removing other deprecated APIs as it never made its way into the `deprecated` module. Let's remove it now. Migrate already patches the API.

It's unfortunate that we didn't also rename `jQuery.expr.setFilters` for
consistency, though... Perhaps we could still do it?

### Checklist ###

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -846,7 +846,7 @@ for ( i in { submit: true, reset: true } ) {
 
 // Easy API for creating new setFilters
 function setFilters() {}
-setFilters.prototype = jQuery.expr.filters = jQuery.expr.pseudos;
+setFilters.prototype = jQuery.expr.pseudos;
 jQuery.expr.setFilters = new setFilters();
 
 function addCombinator( matcher, combinator, base ) {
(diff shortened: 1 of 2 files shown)


mgol:
If we decide to rename `jQuery.expr.setFilters`, we should also do it in `3.x`, perhaps releasing `3.8.0` instead of `3.7.0`. That will also give us an opportunity to move deprecated selector stuff into the deprecated module there:
https://github.com/jquery/jquery/blob/d41f0cd9821126835bfe72dabb21767224cd7ea7/src/selector.js#L2095-L2097

mgol:
PR for the above: #5570

timmywil:
Won't this break a lot and not give us much space savings?

mgol:
We've decided against this - `setFilters` is likely going away in `5.0.0` so it's not worth renaking it at this point. Instead, I'll submit a version of #5570 to the `main` branch.

PR #5570, opened 2024-10-16
Selector: Properly deprecate `jQuery.expr[ ":" ]`/`jQuery.expr.filters`

mgol:
### Summary ###

Those APIs have formally been deprecated since `3.0.0`, but they never made its way into the deprecated module.

Ref gh-5580

PR for `main`: #5580

~~Note: this PR only makes sense in its form if we decide to release it as `3.8.0`; otherwise, we should avoid moving stuff between modules.~~

### Checklist ###


mgol:
PR for `main`: #5580

timmywil (approved):

PR #5579, opened 2024-11-04
Build: Make middleware-mockserver not crash on reading nonexistent files

mgol:
### Summary ###

`fs.readFileSync` crashes when a non-existing file is passed to it. Some APIs of `middleware-mockserver` read a file the path of which depends on query parameters, making it possible to crash it by providing such a parameter. The old PHP server doesn't have these issues.

To fix this, wrap all `fs.readFileSync` occurrences with a function that falls back to the string `"ERROR"`.

### Checklist ###

### DIFF
--- a/test/middleware-mockserver.cjs
+++ b/test/middleware-mockserver.cjs
@@ -8,6 +8,19 @@ const multiparty = require( "multiparty" );
 let cspLog = "";
 
 /**
+ * Like `readFileSync`, but on error returns "ERROR"
+ * without crashing.
+ * @param path
+ */
+function readFileSync( path ) {
+	try {
+		return fs.readFileSync( path );
+	} catch ( e ) {
+		return "ERROR";
+	}
+}
+
+/**
  * Keep in sync with /test/mock.php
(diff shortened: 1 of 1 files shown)


timmywil:
I'm curious what prompted this.

mgol:
> I'm curious what prompted this.

@timmywil I was testing whether you can abuse the param-based `readFile` from https://github.com/jquery/jquery/blob/329661fd538a07993a2fcfa2a75fdd7f5667f86c/test/middleware-mockserver.cjs#L263-L264 and, while testing, I was surprised to see everything stopped working. Then I realized if I pass anything other than `external` or `globaleval`, it will crash the server.

I felt it's a good property of even a dev server to not crash on a simple GET request like that.

timmywil (approved):

mgol:
Landed on `main` in d5ebb464debab6ac39fe065e93c8a7ae1de8547e and on `3.x-stable` in be4f9eccde6e911798b732dc9d19d13812e4339b.

Issue #5582, opened 2024-11-17
Also cache resolved CSS property names that are unprefixed


vlakoff:
Coming from discussion https://github.com/jquery/jquery/discussions/5516, in particular this message: https://github.com/jquery/jquery/discussions/5516#discussioncomment-11270207.

Excerpt:
> However, I noticed that when querying a CSS property name that exists unprefixed, it doesn't get cached into the `vendorProps` map (contrarily to the prefixed names, that get cached).
> 
> Therefore, when querying unprefixed names, each time there is one lookup in the `vendorProps` map then one lookup in the `emptyStyle` object, whereas we could save the latter (after the cache is populated on first access).

Current code: [finalPropName.js](https://github.com/jquery/jquery/blob/d5ebb464debab6ac39fe065e93c8a7ae1de8547e/src/css/finalPropName.js)

Also, I just noticed that previously the unprefixed names were cached at well, but it is no longer the case since commit https://github.com/jquery/jquery/commit/2b5f5d5e90b37f4a735738a6d0b6f22affbea340.

timmywil:
Thanks for opening an issue. I don't know if it would actually make a noticeable difference in performance, but it does seem likely that the change in cache behavior was unintentional.

timmywil:
We discussed this in the meeting. We're leaning towards removing the cache altogether on the main branch as it doesn't seem to add much value.

mgol:
We went in a different direction and in #5583 we removed the cache altogether. See that PR for a more detailed discussion.

PR #5583, opened 2024-11-18
CSS: Drop the cache in finalPropName

mgol:
### Summary ###

The `finalPropName` util caches properties detected to require a vendor prefix. This used to cache unprefixed properties as well, but it was reported that this logic broke accidentally during a refactor. Since fewer & fewer properties require a vendor prefix and caching a few basic checks likely has negligible perf benefits, opt to saving a few bytes and remove the cache.

Ref gh-5582

Size comparison:
```
main @d5ebb464debab6ac39fe065e93c8a7ae1de8547e
   raw     gz Filename
    -9     -3 dist/jquery.min.js
    -9     -7 dist/jquery.slim.min.js
    -9     -4 dist-module/jquery.module.min.js
    -9     -7 dist-module/jquery.slim.module.min.js
```

### Checklist ###

### DIFF
--- a/src/css/finalPropName.js
+++ b/src/css/finalPropName.js
@@ -1,8 +1,7 @@
 import { document } from "../var/document.js";
 
 var cssPrefixes = [ "Webkit", "Moz", "ms" ],
-	emptyStyle = document.createElement( "div" ).style,
-	vendorProps = {};
+	emptyStyle = document.createElement( "div" ).style;
 
 // Return a vendor-prefixed property or undefined
 function vendorPropName( name ) {
@@ -21,13 +20,8 @@ function vendorPropName( name ) {
 
 // Return a potentially-mapped vendor prefixed property
 export function finalPropName( name ) {
-	var final = vendorProps[ name ];
-
-	if ( final ) {
-		return final;
-	}
 	if ( name in emptyStyle ) {
 		return name;
 	}
-	return vendorProps[ name ] = vendorPropName( name ) || name;
+	return vendorPropName( name ) || name;
 }
--- a/test/unit/css.js
+++ b/test/unit/css.js
@@ -1705,7 +1705,7 @@ QUnit.test( "Do not throw on frame elements from css method (trac-15098)", funct
 ( function() {
 	var vendorPrefixes = [ "Webkit", "Moz", "ms" ];
 
-	QUnit.test( "Don't default to a cached previously used wrong prefixed name (gh-2015)", function( assert ) {
+	QUnit.test( "Don't default to a previously used wrong prefixed name (gh-2015)", function( assert ) {
 
 		// Note: this test needs a property we know is only supported in a prefixed version
 		// by at least one of our main supported browsers. This may get out of date so let's
@@ -1761,7 +1761,7 @@ QUnit.test( "Do not throw on frame elements from css method (trac-15098)", funct
 
 } )();
 
-QUnit.test( "Don't detect fake set properties on a node when caching the prefixed version", function( assert ) {
+QUnit.test( "Don't update existing unsupported prefixed properties", function( assert ) {
 	assert.expect( 1 );
 
 	var elem = jQuery( "<div></div>" ),
@@ -1769,7 +1769,7 @@ QUnit.test( "Don't detect fake set properties on a node when caching the prefixe
 	style.MozFakeProperty = "old value";
 	elem.css( "fakeProperty", "new value" );
 
-	assert.equal( style.MozFakeProperty, "old value", "Fake prefixed property is not cached" );
+	assert.equal( style.MozFakeProperty, "old value", "Fake prefixed property is not set" );
 } );
 
 QUnit.test( "Don't set fake prefixed properties when a regular one is missing", function( assert ) {
--- a/test/unit/css.js
+++ b/test/unit/css.js
@@ -1759,18 +1759,32 @@ QUnit.test( "Do not throw on frame elements from css method (trac-15098)", funct
 		assert.equal( elemStyle.undefined, undefined, "Nothing writes to node.style.undefined" );
 	} );
 
-	QUnit.test( "Don't detect fake set properties on a node when caching the prefixed version", function( assert ) {
-		assert.expect( 1 );
-
-		var elem = jQuery( "<div></div>" ),
-			style = elem[ 0 ].style;
-		style.MozFakeProperty = "old value";
-		elem.css( "fakeProperty", "new value" );
-
-		assert.equal( style.MozFakeProperty, "old value", "Fake prefixed property is not cached" );
-	} );
-
 } )();
+
+QUnit.test( "Don't detect fake set properties on a node when caching the prefixed version", function( assert ) {
+	assert.expect( 1 );
+
+	var elem = jQuery( "<div></div>" ),
+		style = elem[ 0 ].style;
+	style.MozFakeProperty = "old value";
+	elem.css( "fakeProperty", "new value" );
+
+	assert.equal( style.MozFakeProperty, "old value", "Fake prefixed property is not cached" );
+} );
+
+QUnit.test( "Don't set fake prefixed properties when a regular one is missing", function( assert ) {
+	assert.expect( 5 );
+
+	var elem = jQuery( "<div></div>" ),
+		style = elem[ 0 ].style;
+	elem.css( "fakeProperty", "fake value" );
+
+	assert.strictEqual( style.fakeProperty, "fake value", "Fake unprefixed property is set" );
+	assert.strictEqual( style.webkitFakeProperty, undefined, "Fake prefixed property is not set (webkit)" );
+	assert.strictEqual( style.WebkitFakeProperty, undefined, "Fake prefixed property is not set (Webkit)" );
+	assert.strictEqual( style.MozFakeProperty, undefined, "Fake prefixed property is not set (Moz)" );
+	assert.strictEqual( style.msFakeProperty, undefined, "Fake prefixed property is not set (ms)" );
+} );
 
 // IE doesn't support CSS variables.
 QUnit.testUnlessIE( "css(--customProperty)", function( assert ) {


timmywil:
The cache no longer provides much value (I'm not 100% sure it ever did). Ironically, the suggestion you got from AI could very well be from training on previous versions of jQuery. But, I like your suggestion to inline `vendorPropName`.

timmywil:
There's no way to be certain of that, even if the code formatting is different. The AI was likely trained on at least some jQuery code given its ubiquity, but that's just an educated guess. Still, it's irrelevant to question of the usefulness of the cache.

mgol:
Inlining the `vendorPropName` function causes an increase in size:
```
drop-finalPropName-cache @d426bda80dd8d460c070b8d70969a81602609477
   raw     gz Filename
    +2     +3 dist/jquery.min.js
    +2     +5 dist/jquery.slim.min.js
    +2     +4 dist-module/jquery.module.min.js
    +2     +6 dist-module/jquery.slim.module.min.js
```

mgol:
Ah, no, this was because the dirty state has a different version string with `.dirty` at the end and that inflates the size... I thought we were handling this somehow, @timmywil? 🤔 

The final size difference is in favor:
```
last run
   raw     gz Filename
   -10     -3 dist/jquery.min.js
   -10     -1 dist/jquery.slim.min.js
   -10     -2 dist-module/jquery.module.min.js
   -10      0 dist-module/jquery.slim.module.min.js
```

I updated the PR.

mgol:
@vlakoff Caching is not always beneficial. We have two possible outcomes:
1. The final property is unprefixed. With the new implementation, we'll just invoke the `name in emptyStyle` check and return `name`. With the old one, we'd first check in the cache and return if present. Otherwise, we'd do the  `name in emptyStyle` check and return `name`. So at the first time we have two checks and then only one - no better than when cache is missing!
2. The final property is prefixed. With the new implementation, we'll do up to a few checks for each call. I doubt this will be on any way expensive - those are very simple checks. Also, the number of properties requiring a prefix gets smaller & smaller over time as new vendor-prefixed properties are no longer being introduced.

For these reasons, we felt `4.0.0` is a good version to drop this cache. We'll keep it in 3.x just in case it matters more for older browsers.

vlakoff:
**Caution:** The PR is currently erroneous.

As I wrote above:

> Just be aware the `name` parameter is overwritten in `vendorPropName()`, so a new variable has to be introduced to preserve the original `name`.

mgol:
Why do we need to preserve the original `name`?

vlakoff:
> 1. The final property is unprefixed. With the new implementation, we'll just invoke the `name in emptyStyle` check and return `name`. With the old one, we'd first check in the cache and return if present. Otherwise, we'd do the  `name in emptyStyle` check and return `name`. So at the first time we have two checks and then only one - no better than when cache is missing!

Lookup is faster on a plain object than on the [style](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style) property (instance of [CSSStyleDeclaration](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration)). Both are fast, but the former is significantly faster (sorry, already deleted my local benchmarks). I suppose the function may be called a lot, and gains add up.

> Why do we need to preserve the original `name`?

Fallback if the supplied value is not a recognized property name. I don't know where this happens in real situations, but I certaintly would not risk removing this fallback.

mgol:
It’d be good to see some benchmarks to see what numbers we’re talking about.

You’re right about the fallback; I need to check if there are any measurable consequences of this and if the answer’s yes, add a test for this.

timmywil:
> .dirty at the end and that inflates the size... I thought we were handling this somehow

I thought so too.

timmywil:
> Lookup is faster on a plain object than on the style property

I suspect you're right about this, and probably by a lot. But, the question isn't, which is faster? The question is, is it fast enough without the cache, which might just take up space? Considering we've been without the cache for most properties (i.e. the ones that don't have a vendor prefix) since 3.4.0 and haven't seen evidence that it affects real world use cases, I doubt adding back the cache is necessary.

timmywil:
Off topic, but the compare size version regex wasn't accounting for the `.2` in the `beta.2` version string.

vlakoff:
Good news, I found back my benchmark:
```js
( function () {
    const nb = 100000;

    const emptyStyle = document.createElement( "div" ).style;
    const finalProps = {};
    finalProps.color = "color";

    const T1 = new Date();
    for ( let i = nb; i--; ) {
        if ( "color" in finalProps ) {}
    }
    const T2 = new Date();
    for ( let i = nb; i--; ) {
        if ( "color" in emptyStyle ) {}
    }
    const T3 = new Date();

    console.log( T2 - T1 );
    console.log( T3 - T2 );
} )();
```

Results for 10,000,000 iterations (in ms, lower is better) on Chrome:
```
6
949
```

The number of iterations is quite high. Although the former is significantly faster, both are very fast.
For perspective, 10,000 accesses take about 1 ms.

Same benchmark (10,000,000 iterations) on Firefox:
```
215
362
```

… Agree the cache can be removed.

mgol:
Regarding the benchmark, it's even better - we don't expose these APIs publicly so their raw execution doesn't matter - it only matters how it influences public APIs. The real difference would be even lower.

mgol:
I started wondering - do we need to auto-add vendor prefixes in 4.x at all? In the past, this was a godsend - instead of calling the `border-radius` getter or setter 5 times, only one was required. Since that time, browser makes agreed to not introduce new vendor prefixes and the old ones gradually disappeared.

I run a small script in supported browsers to detect prefixed CSS props that don't have the identically named unprefixed version and I got the following:

| browser              | webkit* | Moz* | ms* |
|----------------------|---------|------|-----|
| Chrome 131.0.6778.86 |      70 |    0 |   0 |
| Firefox 132.0.2.     |      14 |   25 |   0 |
| Safari 18.1.1        |      82 |    0 |   0 |
| IE 11                |       0 |    0 |  55 |

If you look closer, e.g. at Chrome, you'll notice lots of these properties fall into one of the following buckets:
* like [`-webkit-border-before`](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-border-before): the unprefixed version is [`border-block-start`](https://developer.mozilla.org/en-US/docs/Web/CSS/border-block-start) - calling the setter with `border-block-start` won't pick up prefixed values and calling it with `border-before` would look invalid as there's no such property and there are no plans for it. Also, only the `-webkit`-prefixed version is supported anywhere and other versions are not likely to show up.
* like [`-webkit-box-align`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-align): legacy API; unprefixed versions use different names & syntax
* like [`-webkit-line-clamp`](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp): uses the `-webkit` prefix everywhere, including in Firefox - it should just be used directly as prefixed as there's no unprefixed version - and even if one arrives, it will work differently

There's also `user-select` which requires the `-webkit-` prefix in Safari & `-ms-` in IE, but those APIs seem quite uncommon on those lists.

mgol:
More details in a separate comment, so that it's easier to skip it:

<details><summary>The script I used for the detection</summary>

```js
(() => {
'use strict';

const cssPrefixes = ['webkit', 'Moz', 'ms'];

const prefixedProps = new Map();

const emptyStyle = document.createElement('div').style;

function unprefix(prop, prefix) {
    const withoutPrefix = prop.slice(prefix.length);
    return withoutPrefix[ 0 ].toLowerCase() + withoutPrefix.slice( 1 );
}

for (const prefix of cssPrefixes) {
    prefixedProps.set(prefix, new Set());
}

for (const prop in emptyStyle) {
    for (const prefix of cssPrefixes) {
        if (prop.startsWith(prefix)
                && !(unprefix(prop, prefix) in emptyStyle)) {
            prefixedProps.get(prefix).add(prop);
        }
    }
}

for (const [prefix, propsForPrefix] of prefixedProps) {
    console.log(`props for prefix ${prefix}`, propsForPrefix);
}

})();
```

</details>

<details><summary>The ES5 version of the script for IE</summary>

```js
(function() {
'use strict';

const cssPrefixes = ['webkit', 'Moz', 'ms'];

const prefixedProps = new Map();

const emptyStyle = document.createElement('div').style;

function unprefix(prop, prefix) {
    const withoutPrefix = prop.slice(prefix.length);
    return withoutPrefix[ 0 ].toLowerCase() + withoutPrefix.slice( 1 );
}

for (const prefix of cssPrefixes) {
    prefixedProps.set(prefix, new Set());
}

for (const prop in emptyStyle) {
    for (const prefix of cssPrefixes) {
        if (prop.startsWith(prefix)
                && !(unprefix(prop, prefix) in emptyStyle)) {
            prefixedProps.get(prefix).add(prop);
        }
    }
}

for (const [prefix, propsForPrefix] of prefixedProps) {
    console.log(`props for prefix ${prefix}`, propsForPrefix);
}

})();
```

</details>


<details><summary>The list of detected `-webkit-` properties in Chrome</summary>

* -webkit-border-after
* -webkit-border-after-color
* -webkit-border-after-style
* -webkit-border-after-width
* -webkit-border-before
* -webkit-border-before-color
* -webkit-border-before-style
* -webkit-border-before-width
* -webkit-border-end
* -webkit-border-end-color
* -webkit-border-end-style
* -webkit-border-end-width
* -webkit-border-horizontal-spacing
* -webkit-border-start
* -webkit-border-start-color
* -webkit-border-start-style
* -webkit-border-start-width
* -webkit-border-vertical-spacing
* -webkit-box-align
* -webkit-box-direction
* -webkit-box-flex
* -webkit-box-ordinal-group
* -webkit-box-orient
* -webkit-box-pack
* -webkit-box-reflect
* -webkit-column-break-after
* -webkit-column-break-before
* -webkit-column-break-inside
* -webkit-font-smoothing
* -webkit-line-clamp
* -webkit-locale
* -webkit-logical-height
* -webkit-logical-width
* -webkit-margin-after
* -webkit-margin-before
* -webkit-margin-end
* -webkit-margin-start
* -webkit-mask-box-image
* -webkit-mask-box-image-outset
* -webkit-mask-box-image-repeat
* -webkit-mask-box-image-slice
* -webkit-mask-box-image-source
* -webkit-mask-box-image-width
* -webkit-mask-position-x
* -webkit-mask-position-y
* -webkit-max-logical-height
* -webkit-max-logical-width
* -webkit-min-logical-height
* -webkit-min-logical-width
* -webkit-padding-after
* -webkit-padding-before
* -webkit-padding-end
* -webkit-padding-start
* -webkit-perspective-origin-x
* -webkit-perspective-origin-y
* -webkit-print-color-adjust
* -webkit-rtl-ordering
* -webkit-tap-highlight-color
* -webkit-text-combine
* -webkit-text-decorations-in-effect
* -webkit-text-fill-color
* -webkit-text-security
* -webkit-text-stroke
* -webkit-text-stroke-color
* -webkit-text-stroke-width
* -webkit-transform-origin-x
* -webkit-transform-origin-y
* -webkit-transform-origin-z
* -webkit-user-drag
* -webkit-user-modify

</details>


<details><summary>The list of detected `-moz-` properties in Firefox</summary>

* -moz-box-align
* -moz-box-direction
* -moz-box-orient
* -moz-box-pack
* -moz-float-edge
* -moz-orient
* -moz-osx-font-smoothing
* -moz-text-size-adjust
* -moz-user-input
* -moz-window-dragging
* -moz-force-broken-image-icon
* -moz-box-ordinal-group
* -moz-box-flex
* -moz-border-end-style
* -moz-border-start-style
* -moz-margin-end
* -moz-margin-start
* -moz-border-end-width
* -moz-border-start-width
* -moz-padding-end
* -moz-padding-start
* -moz-border-end-color
* -moz-border-start-color
* -moz-border-start
* -moz-border-end

</details>


<details><summary>The list of detected `-webkit-` properties in Firefox</summary>

* -webkit-box-align
* -webkit-box-direction
* -webkit-box-orient
* -webkit-box-pack
* -webkit-font-smoothing
* -webkit-text-size-adjust
* -webkit-line-clamp
* -webkit-text-security
* -webkit-box-ordinal-group
* -webkit-box-flex
* -webkit-text-stroke-width
* -webkit-text-fill-color
* -webkit-text-stroke-color
* -webkit-text-stroke

</details>


<details><summary>The list of detected `-webkit-` properties in Safari</summary>

* -webkit-border-after
* -webkit-border-after-color
* -webkit-border-after-style
* -webkit-border-after-width
* -webkit-border-before
* -webkit-border-before-color
* -webkit-border-before-style
* -webkit-border-before-width
* -webkit-border-end
* -webkit-border-end-color
* -webkit-border-end-style
* -webkit-border-end-width
* -webkit-border-horizontal-spacing
* -webkit-border-start
* -webkit-border-start-color
* -webkit-border-start-style
* -webkit-border-start-width
* -webkit-border-vertical-spacing
* -webkit-box-align
* -webkit-box-decoration-break
* -webkit-box-direction
* -webkit-box-flex
* -webkit-box-flex-group
* -webkit-box-lines
* -webkit-box-ordinal-group
* -webkit-box-orient
* -webkit-box-pack
* -webkit-box-reflect
* -webkit-column-axis
* -webkit-column-break-after
* -webkit-column-break-before
* -webkit-column-break-inside
* -webkit-column-progression
* -webkit-cursor-visibility
* -webkit-font-smoothing
* -webkit-hyphenate-limit-after
* -webkit-hyphenate-limit-before
* -webkit-hyphenate-limit-lines
* -webkit-initial-letter
* -webkit-line-align
* -webkit-line-box-contain
* -webkit-line-clamp
* -webkit-line-grid
* -webkit-line-snap
* -webkit-locale
* -webkit-logical-height
* -webkit-logical-width
* -webkit-margin-after
* -webkit-margin-before
* -webkit-margin-end
* -webkit-margin-start
* -webkit-mask-box-image
* -webkit-mask-box-image-outset
* -webkit-mask-box-image-repeat
* -webkit-mask-box-image-slice
* -webkit-mask-box-image-source
* -webkit-mask-box-image-width
* -webkit-mask-position-x
* -webkit-mask-position-y
* -webkit-mask-source-type
* -webkit-max-logical-height
* -webkit-max-logical-width
* -webkit-min-logical-height
* -webkit-min-logical-width
* -webkit-nbsp-mode
* -webkit-padding-after
* -webkit-padding-before
* -webkit-padding-end
* -webkit-padding-start
* -webkit-rtl-ordering
* -webkit-ruby-position
* -webkit-text-combine
* -webkit-text-decorations-in-effect
* -webkit-text-fill-color
* -webkit-text-security
* -webkit-text-stroke
* -webkit-text-stroke-color
* -webkit-text-stroke-width
* -webkit-text-zoom
* -webkit-user-drag
* -webkit-user-modify
* -webkit-user-select

</details>


<details><summary>The list of detected `-ms-` properties in IE 11</summary>

* -ms-block-progression
* -ms-interpolation-mode
* -ms-content-zoom-chaining
* -ms-content-zoom-limit
* -ms-content-zoom-limit-max
* -ms-content-zoom-limit-min
* -ms-content-zoom-snap
* -ms-content-zoom-snap-points
* -ms-content-zoom-snap-type
* -ms-content-zooming
* -ms-flex-align
* -ms-flex-item-align
* -ms-flex-line-pack
* -ms-flex-negative
* -ms-flex-order
* -ms-flex-pack
* -ms-flex-positive
* -ms-flex-preferred-size
* -ms-flow-from
* -ms-flow-into
* -ms-grid-column
* -ms-grid-column-align
* -ms-grid-column-span
* -ms-grid-columns
* -ms-grid-row
* -ms-grid-row-align
* -ms-grid-row-span
* -ms-grid-rows
* -ms-high-contrast-adjust
* -ms-hyphenate-limit-chars
* -ms-hyphenate-limit-lines
* -ms-hyphenate-limit-zone
* -ms-hyphens
* -ms-ime-align
* -ms-overflow-style
* -ms-scroll-chaining
* -ms-scroll-limit
* -ms-scroll-limit-x-max
* -ms-scroll-limit-x-min
* -ms-scroll-limit-y-max
* -ms-scroll-limit-y-min
* -ms-scroll-rails
* -ms-scroll-snap-points-x
* -ms-scroll-snap-points-y
* -ms-scroll-snap-type
* -ms-scroll-snap-x
* -ms-scroll-snap-y
* -ms-scroll-translation
* -ms-text-combine-horizontal
* -ms-text-size-adjust
* -ms-touch-select
* -ms-user-select
* -ms-wrap-flow
* -ms-wrap-margin
* -ms-wrap-through

</details>

vlakoff:
I have an infamous example: `user-select` on Safari.

* Chrome:
    * supported `-webkit-` prefixed since 2008-12-11
    * supported unprefixed since 2016-10-12
* Firefox:
    * supported `-moz-` prefixed since 2004-11-09
    * supported `-webkit-` prefixed since 2016-09-20
    * supported unprefixed since 2019-09-03
* Safari:
    * supported `-webkit-` prefixed since 2007-10-26
    * still not supported unprefixed 🤡

Refs:
* https://developer.mozilla.org/en-US/docs/Web/CSS/user-select
* https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#browser_compatibility
* https://caniuse.com/mdn-css_properties_user-select

vlakoff:
I just noticed you already mentioned that one specifically:

> There's also `user-select` which requires the `-webkit-` prefix in Safari & `-ms-` in IE, but those APIs seem quite uncommon on those lists.

Personally, I use this property quite often, and without jQuery, I have to handle both the unprefixed and `-webkit-` prefixed properties, because of Safari. I have tried reporting this issue to the developers a few times.

mgol:
Yeah, `user-select` is a bummer. It's possible we'll have to wait with this for `5.0.0`. I still created an issue and marked it for the team discussion: #5585.

src/css/finalPropName.js
```diff
@@ -17,17 +19,6 @@ function vendorPropName( name ) {
...
-	var final = vendorProps[ name ];
-
-	if ( final ) {
-		return final;
-	}
-	if ( name in emptyStyle ) {
-		return name;
-	}
-	return vendorProps[ name ] = vendorPropName( name ) || name;
+	return name;
```
vlakoff, line 23: Just a reminder: as stated in https://github.com/jquery/jquery/pull/5583#issuecomment-2487013528, we have to avoid overwriting the `name` parameter, line 17 above.

mgol:
I reverted locally to the `main` implementation with cache removed, i.e. with the `vendorPropName` function still existed and then to one with a single function. The latter showed the following size comparison (after applying PR #5586):

```
drop-finalPropName-cache
   raw     gz     br Filename
    -5     +1    +11 dist/jquery.min.js
    -5     +1    +13 dist/jquery.slim.min.js
    -5     +2    -60 dist-module/jquery.module.min.js
    -5     +1     -3 dist-module/jquery.slim.module.min.js
```

Fluctuations in Brotli sizes are pretty interesting. But the more popular no-module builds are larger, so using two functions makes sense, I think.

test/unit/css.js
```diff
@@ -1759,18 +1759,32 @@ QUnit.test( "Do not throw on frame elements from css method (trac-15098)", funct
...
 
-	QUnit.test( "Don't detect fake set properties on a node when caching the prefixed version", function( assert ) {
-		assert.expect( 1 );
+} )();
 
-		var elem = jQuery( "<div></div>" ),
-			style = elem[ 0 ].style;
-		style.MozFakeProperty = "old value";
-		elem.css( "fakeProperty", "new value" );
+QUnit.test( "Don't detect fake set properties on a node when caching the prefixed version", function( assert ) {
```
timmywil, line 1764:
```suggestion
QUnit.test( "Don't set prefixed versions of fake properties", function( assert ) {
```

Since the cache is gone.

test/unit/css.js
```diff
@@ -1759,18 +1759,32 @@ QUnit.test( "Do not throw on frame elements from css method (trac-15098)", funct
...
 
-		assert.equal( style.MozFakeProperty, "old value", "Fake prefixed property is not cached" );
-	} );
+	var elem = jQuery( "<div></div>" ),
+		style = elem[ 0 ].style;
+	style.MozFakeProperty = "old value";
+	elem.css( "fakeProperty", "new value" );
 
-} )();
+	assert.equal( style.MozFakeProperty, "old value", "Fake prefixed property is not cached" );
```
timmywil, line 1772:
```suggestion
	assert.equal( style.MozFakeProperty, "old value", "Fake prefixed property is not set" );
```

test/unit/css.js
mgol: This is not a new test, just moved outside of a block where it shouldn't be. The new one is below. There are a few more tests that mention caching.

This test specifically checks that setting a fake prefixed property doesn't affect subsequent setters. The title you proposed doesn't contain this bit. I'll think about another way to rephrase this.

mgol: Maybe "Don't update existing unsupported prefixed properties"?

timmywil (approved):

Issue #4284, opened 2019-01-23
jQuery.param returns `&undefined=` as part of the string when looping over a empty array element


jkesseler:
### Description ###
`jQuery.param` returns `&undefined=` as part of the string when looping over a empty array element

Given the following code: 
```javascript
    var condition = false;
    var myArray = [
        { name: 'name1', value: 'value1' },
        condition && { name: 'name2', value: 'value2' },
        { name: 'name3', value: 'value3' },
    ];

    var serialized  = $.param(myArray);
   // output= 'name1=value1&undefined=&name3=value3'
```
* What do you expect to happen?
The returned string does not include '`&undefined=`'

*  What actually happens?
The returned string includes '`&undefined=`'

### Link to test case ###
http://jsfiddle.net/vedghafn/

Is this desired behavior? The [documentation page on `jQuery.param()`](http://api.jquery.com/jquery.param/) does not mention this. A simple solution would be to check if `this.name` actually exists before calling `add()` in [serialize.js](https://github.com/jquery/jquery/blob/5bdc85b82b84e5459462ddad9002f22d1ce74f21/src/serialize.js#L81)

mgol:
This is mentioned in docs at https://api.jquery.com/jQuery.param/:
> If the object passed is in an Array, it must be an array of objects in the format returned by `.serializeArray()`
>
> ```js
> [
>   { name: "first", value: "Rick" },
>   { name: "last", value: "Astley" },
>   { name: "job", value: "Rock Star" }
> ]
> ```

Therefore, an `undefined` entry is not supported.

If you can use ES6, this can be worked around via:
```js
var myArray = [
    { name: 'name1', value: 'value1' },
    ...condition ? [{ name: 'name2', value: 'value2' }] : [],
    { name: 'name3', value: 'value3' },
];
```
Otherwise, try using `concat` or `push` conditionally.

PR #5592, opened 2024-12-10
Fix: Skip undefined values during serialization

DeloYong:
### Summary ###
This PR improves the `param` function by adding a check to skip `undefined` values during serialization. This ensures that `undefined` values are not included in the final query string.

### Checklist ###


timmywil:
Thanks for your contribution. However, this is a breaking change and `undefined` is not supported.

See https://github.com/jquery/jquery/issues/4284

PR #5604, opened 2024-12-31
Tests: migrate test runner to jquery-test-runner

timmywil:
### Summary ###

I've moved all the code from `test/runner` into a package called [jquery-test-runner](https://github.com/timmywil/jquery-test-runner). The changes to this repo are fairly simple. It uses the binary included in that package to run tests in browsers using selenium and browserstack.

Some things to note:

- I've tested jquery-test-runner with jquery, jquery-migrate, and jquery-ui and all their workflows have full green runs in my forks after migration.
- The `jtr` (alias for `jquery-test-runner`) binary includes several commands. The default command is for test runs (`jtr` or `jtr run`). jtr also includes `jtr serve`, which has a `-p` (`--port`) option (rather than an env var). See `npx jtr --help` for all commands and options.
- I've migrated yargs to commander in `jquery-test-runner`.
- I based the beginning infra of jquery-test-runner on the jquery repo's `test/runner` folder, which means the recent changes to allow Edge runs in IE-mode have been kept, `express` is not used for the test server, etc. jquery-migrate and jquery-ui will now be able to take advantage of those changes.
- `jquery-test-runner` can take a YAML config file, with support for all of the same options as the command line. I considered JSON and JS, but I liked how terse the YAML was. See [jtr-isolate.yml](https://github.com/jquery/jquery/blob/4618e612af11091df8879e6a980a93bc3edff15f/jtr-isolate.yml) for an example, which is used to create a separate run for each module in BrowserStack runs while still sharing browser instances.
- Upgraded the Node version used for testing to 22.x from 20.x in all workflows.

Once we merge this to `jquery`, I'll make PRs for `jquery-migrate` and `jquery-ui`.

~~At some point, we can move jquery-test-runner over to the jquery org.~~ moved

### Checklist ###


jtr-isolate.yml
```diff
@@ -0,0 +1,30 @@
...
+    - effects
+    - event
+    - manipulation
+    - offset
+    - queue
+    - selector
+    - serialize
+    - support
+    - traversing
+    - tween
```
mgol, line 27: Is this making each module run individual?

mgol (commented):
Code-wise looks good but where is the runner code? https://www.npmjs.com/package/jquery-test-runner points to https://github.com/jquery/jquery-test-runner which is a 404 to me. I should be seeing all the repos in the org with my permissions.

jtr-isolate.yml
timmywil: Yes, same as now. Each module runs on its own page, but the browser instance is shared (so it doesn't have to be restarted for each test).

timmywil:
> where is the runner code?

Currently at https://github.com/timmywil/jquery-test-runner. I plan to move to the jquery org.

mgol:
@timmywil Can you move the repo and add me (`mgol`) and the OpenJSF user (`openjsfoundation`) as co-owners of the npm package before merging this PR?

mgol (approved):
Approving to unblock you but please remember about adding the owners to the runner package!

PR #5606, opened 2025-01-02
Build: Stop testing on iOS 10

mgol:
### Summary ###

As of January 2025, iOS 10 is a tier 4 device on BrowserStack: https://www.browserstack.com/device-tiers
That leads to devices with this iOS version often not being available and failing our tests. Remove it from the test matrix. Also, add comments explaining the status of tests on various iOS versions, including iOS 7 that we stopped testing on a long time ago.

### Checklist ###

* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com: not API, but see https://github.com/jquery/jquery.com/pull/248

### DIFF
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -48,9 +48,19 @@ jobs:
           - '__iOS_13'
           - '__iOS_12'
           - '__iOS_11'
-          - '__iOS_10'
+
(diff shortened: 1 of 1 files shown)


timmywil (approved):
Merging this as iOS 10 is not working anyway.

PR #5607, opened 2025-01-05
Attributes: Support the `until-found` value for the `hidden` attribute

mgol:
### Summary ###

The `hidden` attribute used to be a boolean one but it gained a new `until-found` eventually. This led us to change the way we handle boolean attributes in jQuery 4.0 in gh-5452 to avoid these issues in the future.

That said, currently from the attributes we treat as boolean only `hidden` has gained an extra value, let's support it.

Ref gh-5388
Ref gh-5452

The cost is currently quite high:
```
3.x-stable @fc874a0e125b01c2d91b4a8a06213bf1f1e5bfb6
   raw     gz     br Filename
  +157    +66    +58 dist/jquery.min.js
  +157    +52    +36 dist/jquery.slim.min.js
```

It's possible it could be lowered if we removed the infra for adding future attributes to the list and just mentioned `hidden` & `until-found` explicitly in both the getter & setter.

Another alternative is to avoid changing the `3.x` line and instead add a note to the docs, showing how to define an attrHook for `hidden` that will achieve the same result.

### Checklist ###

### DIFF
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -106,17 +106,19 @@ jQuery.extend( {
 // Hooks for boolean attributes
 boolHook = {
 	set: function( elem, value, name ) {
+		var strValue = String( value );
+
 		if ( value === false ) {
 
 			// Remove boolean attributes when set to false
 			jQuery.removeAttr( elem, name );
-		} else if (
-			name.toLowerCase() !== "hidden" ||
-				String( value ).toLowerCase() !== "until-found"
-		) {
-			elem.setAttribute( name, name );
 		} else {
-			elem.setAttribute( name, value );
+			elem.setAttribute( name,
+				name.toLowerCase() === "hidden" &&
+						strValue.toLowerCase() === "until-found" ?
+					strValue :
+					name
+			);
 		}
 		return name;
 	}
@@ -136,13 +138,10 @@ jQuery.each( jQuery.expr.match.bool.source.match( /\w+/g ), function( _i, name )
 			attrHandle[ lowercaseName ] = ret;
 
 			ret = getter( elem, name, isXML );
-			if ( ret != null ) {
-				if ( lowercaseName !== "hidden" ||
-					ret.toLowerCase() !== "until-found"
-				) {
-					ret = lowercaseName;
-				}
-			}
+			ret = ret == null ||
+					( lowercaseName === "hidden" && ret.toLowerCase() === "until-found" ) ?
+				ret :
+				lowercaseName;
 
 			attrHandle[ lowercaseName ] = handle;
 		}
--- a/src/attributes/attr.js
+++ b/src/attributes/attr.js
@@ -10,7 +10,13 @@ define( [
 "use strict";
 
 var boolHook,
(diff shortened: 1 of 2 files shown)


timmywil (approved):
LGTM!

mgol:
@gibson042 I pushed the smaller version in the second commit, you can compare the two & suggest improvements.

src/attributes/attr.js
```diff
@@ -110,8 +110,13 @@ boolHook = {
...
 			// Remove boolean attributes when set to false
 			jQuery.removeAttr( elem, name );
-		} else {
+		} else if (
+			name.toLowerCase() !== "hidden" ||
+				String( value ).toLowerCase() !== "until-found"
+		) {
 			elem.setAttribute( name, name );
+		} else {
+			elem.setAttribute( name, value );
```
gibson042, line 119:
```suggestion
        } else {
            elem.setAttribute( name, name.toLowerCase() === "hidden" &&
                strValue.toLowerCase() === "until-found" ? strValue : name );
```
(with `var strValue = String( value );`)

src/attributes/attr.js
```diff
@@ -129,9 +134,16 @@ jQuery.each( jQuery.expr.match.bool.source.match( /\w+/g ), function( _i, name )
...
-				null;
+
+			ret = getter( elem, name, isXML );
+			if ( ret != null ) {
+				if ( lowercaseName !== "hidden" ||
+					ret.toLowerCase() !== "until-found"
+				) {
+					ret = lowercaseName;
+				}
+			}
```
gibson042, line 145:
```suggestion
            ret = ret == null || lowercaseName === "hidden" &&
                ret.toLowerCase() === "until-found" ? ret : lowercaseName;
```

gibson042 (approved):
Compression pass:
```
   -52     -6 dist/jquery.js                                                   
   -13    -17 dist/jquery.min.js
```

mgol:
@gibson042 I applied your suggestions but I got a different diff compared to my second commit:
```
3.x-hidden-until-found @1f668bf0a2f354a35bd0e9b125fb398901b63c90
   raw     gz     br Filename
   -13     -9    +67 dist/jquery.min.js
   -13     -6    -21 dist/jquery.slim.min.js
```
I pushed it, can you have a look?

The diff from `3.x-stable` right now:
```
3.x-stable @fc874a0e125b01c2d91b4a8a06213bf1f1e5bfb6
   raw     gz     br Filename
  +126    +42    +72 dist/jquery.min.js
  +126    +37    +13 dist/jquery.slim.min.js
```

mgol:
@gibson042 I am getting a different diff for this version of `boolHook` vs. what's in the PR:
```
3.x-hidden-until-found @9c8e3e3c41927744995729d9270b051cdc5cde5e
   raw     gz     br Filename
    +4     +2    -43 dist/jquery.min.js
    +4     +1    +10 dist/jquery.slim.min.js
```

mgol:
I'm going to merge it as-is; feel free to submit a PR with various size improvements if you find them.

PR #5622, opened 2025-02-02
Error Handling

marsel365:
### Summary ###
1) If newContext.querySelectorAll throws an exception (e.g., due to an invalid selector), the find method returns an empty array without error information.
2) If tokenize returns an empty array (e.g., due to an invalid selector), the code continues execution, which may lead to errors.
3) In the block if ( nodeName( elem, "template" ) ), accessing elem.content might cause an error if the browser does not support HTMLTemplateElement.
4) Wrap = jQuery(html).eq(0).clone(true) only takes the first element from the set. If the user passes HTML with multiple root elements, they will be ignored. This might contradict expectations.

### DIFF
--- a/src/traversing.js
+++ b/src/traversing.js
@@ -156,7 +156,8 @@ jQuery.each( {
 		// Treat the template element as a regular one in browsers that
 		// don't support it.
 		if ( nodeName( elem, "template" ) ) {
-			elem = elem.content || elem;
+			// Safe access to content
+			elem = ("content" in elem) ? elem.content : elem;
 		}
 
 		return jQuery.merge( [], elem.childNodes );
--- a/src/wrap.js
+++ b/src/wrap.js
@@ -14,7 +14,11 @@ jQuery.fn.extend( {
 			}
 
 			// The elements to wrap the target around
-			wrap = jQuery( html, this[ 0 ].ownerDocument ).eq( 0 ).clone( true );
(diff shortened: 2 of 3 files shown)


src/traversing.js
```diff
@@ -156,7 +156,8 @@ jQuery.each( {
 		// Treat the template element as a regular one in browsers that
 		// don't support it.
 		if ( nodeName( elem, "template" ) ) {
-			elem = elem.content || elem;
```
mgol, line 159: What’s the issue with the previous code?

src/wrap.js
```diff
@@ -14,7 +14,11 @@ jQuery.fn.extend( {
 			}
 
 			// The elements to wrap the target around
-			wrap = jQuery( html, this[ 0 ].ownerDocument ).eq( 0 ).clone( true );
+			wrapper = jQuery( html, this[ 0 ].ownerDocument );
```
mgol, line 17: This doesn’t work at all as there’s no such variable, it will just crash. Did you run this code locally?

mgol (changes requested):
Please provide concrete code examples with their current behavior and what you’d expect instead. From the description, I don’t understand what you’re really trying to achieve here - and your changes don’t work, failing many existing tests, crashing on non-existing variables, etc.

All changes also need to have new tests added, confirming newly introduced behavior.

PR #5623, opened 2025-02-07
Fix: focus event fires on disabled inputs

### LINKED ISSUE #5620: `focus` event fires on disabled inputs
## Problem Description
Triggering focus on an input element causes a `focus` event to be emitted even when the input element is _disabled_.
This is a regression introduced with jQuery 3.7 - the bug is reproducible on v3.7.0 and v3.7.1.

## Steps to reproduce the problem
1. Create a disabled input element
2. Attach a `focus` event handler on the input using jQuery
3. Trigger a `focus` event on the input using jQuery

This can be observed with the following code snippet, where clicking the button triggers the focus event on the disabled input, and hence the assertion fails.
```html
<input id="theDisabledInput" disabled />
<button onclick="focusTheDisabledInput()">Focus the disabled input</button>

<script src="https://code.jquery.com/jquery-3.7.1.js"></script>
<script>
    $("#theDisabledInput").on("focus", (event) => {
        console.assert(!event.target.disabled, "'focus' event fired on a disabled input element!");
    });

    function focusTheDisabledInput() {
        $("#theDisabledInput").focus();
    }
</script>
```

*Note*: it is important to both attach the event listener and trigger the event using jQuery, in order to be able to reproduce the issue

ac-mmi:
This PR addresses the issue where calling .focus() on a disabled input still triggers a focus event in jQuery 3.7+, which is inconsistent with native behavior.

**Changes Made:**

1. Added a flag in trigger.js to prevent focus events from firing on disabled inputs.
2. Ensures .focus() on a disabled element does nothing, matching native JavaScript behavior.

**Impact:**

1. Fixes an inconsistency without affecting other focusable elements.
2. No breaking changes.

Fixes [#5620](https://github.com/jquery/jquery/issues/5620)


mgol:
Please make sure submitted PRs pass existing tests.

mgol:
I’m sorry, I mixed up PRs, let me reopen this one.

I don’t think the changes will be that small, though; we also need to handle at least `blur`, get new tests added and clearly state which cases we want to support and which we give up on for now - e.g., there are many quirks with `focus` & `click` propagation in elements inside disabled `fieldset`s.

ac-mmi:
@mgol I've updated this PR based on the earlier discussion.

The latest changes add regression tests for disabled controls (including controls inside a disabled <fieldset>) and extend the existing guard in trigger.js to cover synthetic blur in addition to focus, using :disabled so fieldset-disabled controls are handled as well.

I re-ran the event tests and the full test suite after the changes.

PR #5627, opened 2025-02-11
Fix: Animation promise gets stuck in infinite loop, eventually crashing the browser

### LINKED ISSUE #5534: Animation promise gets stuck in infinite loop, eventually crashing the browser
### Description ###
If a `then()` handler is added to the `animation` promise (the first argument of the `start` callback), jQuery gets stuck in an endless loop. The website becomes unresponsive, the CPU usage jumps to 100% and the memory consumption increases significantly and steadily. Eventually, the process crashes with an "Out of Memory" error.

```
$elem.animate({left: 350}, {
  start: animation => {
    animation.then(() => console.log('end'));
  }
});
```

We discovered this problem when we executed multiple animations simultaneously with `queue: false`. We stored the `animation` promises from the individual `start` callbacks and used them to wait for all animations to complete with `$.when`. When we added another `then()` handler to the combined promise, the bug appeared.

Further observations:
* The problem only occurs when a `then()` handler is added to the `animation` promise from the `start()` callback. It does **not** occur when handler is added to the promise returned by `promise()`. It does **not** occur when a `done()` handler is used.
* To pause the infinite loop, a breakpoint can be set in the `mightThrow` function.
* Otherwise, the loop can only be stopped by closing the browser tab or killing the corresponding process with the browser's task manager.

We have created a small demo page to reproduce the problem, see below. Tick the first two checkboxes to start the infinite loop.

jQuery: 3.7.1
Browsers: Chrome 127, Edge 127, Firefox 129
OS: Windows 10

### Link to test case ###

https://output.jsbin.com/febijutiva

[jquery-animate-promise.html.txt](https://github.com/user-attachments/files/16659237/jquery-animate-promise.html.txt)

ac-mmi:
This pull request addresses an issue in the jQuery Deferred implementation where an infinite loop was being triggered when the `.then()` statement was executed after an animation completion. This issue occurred because the state was being continuously set to "resolved," leading to unintended behavior and performance degradation.

**Changes Made:**

**1. Introduced Flag to Track State Resolution:**

- A flag `isResolvedOrRejected` was introduced to track whether the deferred state has already been set to either "resolved" or "rejected." This ensures that once the state changes, further state transitions do not occur, preventing continuous updates.

**2. State Change Condition:**

- Modified the code to check if the state has already been set to "resolved" or "rejected" before triggering another state change. This prevents redundant calls to `resolveWith `and `fireWith`.

**3. Controlled Execution of `.then()` Handlers:**

- Ensured that the `.then()` handlers (which are responsible for resolving the deferred promise) are only executed once when the animation completes, thus preventing the infinite loop of state changes.

**4.Smoother Animation Execution:**

- By eliminating the redundant state changes, the animation now executes smoothly without interruptions caused by the repeated triggering of state transitions.

**Testing:**
- Verified that after an animation completes, the resolved state is only set once, and the animation continues without issues.
- Ensured that the animation behavior is smooth without any unexpected state changes.

Fixes #5534


mgol:
See https://github.com/jquery/jquery/pull/5623#issuecomment-2651095215

mgol:
This PR heavily modifies Callbacks code in ways that break our test suite. The description also looks quite similar to what you can get out of various AI tools nowadays; did you work on it yourself?

When submitting PRs, please make sure the proposed changes really solve the issue they claim they solve; the PR should pass tests, add new ones confirming the fix and it needs to pass existing tests. Also, changes should be as small as possible. Seeing a big PR failing tests I don’t really see a way to proceed.

Issue #5628, opened 2025-02-11
Reconsider the width/height css hooks CSS swap behavior


mgol:
### Summary

We have `width` & `height` CSS hooks. The `get` one:
https://github.com/jquery/jquery/blob/667321eb2d1c4328b993c25fbe2342a01ec4f87f/src/css.js#L316-L335
checks for `display` to be `none` or to start with `table-` with some exceptions and then it applies some temporary styles:
https://github.com/jquery/jquery/blob/667321eb2d1c4328b993c25fbe2342a01ec4f87f/src/css.js#L27
before measuring the size.

### The `display: none` case

Now, removing this swap doesn't make any test fail in Firefox or Chrome and only one, related to table columns, in Safari. We don't test the `display: none` case. I've been wondering if that one makes any sense. Even just applying `position: absolute` may affect the element styles beyond what we'd see if we just gave it a temporary `display: block`. There's no way to always properly measure hidden elements. Can we get rid of this one in 4.0?

### Table columns

As for table columns, the support comment states:
```js
// Support: Safari <=8 - 12+, Chrome <=73+
// Table columns in WebKit/Blink have non-zero offsetWidth & zero
// getBoundingClientRect().width unless display is changed.
// Support: IE <=11+
// Running getBoundingClientRect on a disconnected node
// in IE throws an error.
```
Modern versions of Safari or Chrome do not match what the comment says. I noticed different cross-browser issues, though. For a column with `span="2"` (denoted as `c`) that is sized implicitly by the size of its cells, I got the following values in an example table:

|                                   | Chrome      | Safari    | Firefox            | IE 11  |
|-----------------------------------|-------------|-----------|--------------------|--------|
| `c.offsetWidth`                   | 172         | 172       | 172                | 0      |
| `c.getBoundingClientRect().width` | 171.59375   | 0         | 171.61666870117188 | 0      |
| `getComputedStyle(c).width`       | "171.594px" | "0px"     | "83.65px"          | "auto" |

Some conclusions:
* Chrome reports all the values and they are roughly equal - barring any transforms that would affect `getBoundingClientRect`.
* Safari only has a non-zero `offsetWidth` - the comment stated that one was 0 as well.
* Firefox reports all the values but `getComputedStyle(c).width` only reports the size of the first column over which the `col` element spans.
* In IE, none of these values makes sense. If the `col` element has an explicit `width` set via CSS then IE starts reporting it instead of `"auto"`. However, our `cssShow` styles have no effect on reported values.

[...]

PR #5634, opened 2025-02-24
CSS: Fix dimensions of table `<col>` elements

mgol:
### Summary ###

3.x version of PR #5630. ~~In this one, we are keeping the style swap even for table col elements for compat reasons.~~

Changes:
1. Fix measurements of `<col span="2">` elements in Firefox.
2. Fix measurements of all implicitly sized `<col>` elements in Safari.

Firefox always reports computed width as if `span` was 1. In Safari, computed width for columns is always 0. Work around both issues by using `offsetWidth`.

In IE/Edge, `<col>` computed width is `"auto"` unless `width` is set explicitly via CSS so measurements there remain incorrect. Because of the lack of a proper workaround, we accept this limitation.

Fixes gh-5628

### Checklist ###


gibson042 (approved):

timmywil (approved):

mgol:
Size diff:
```
3.x-stable @6b7d74a1ea9a0004baf720ffb84f53f261e78c58
   raw     gz     br Filename
  +185    +42    -17 dist/jquery.min.js
  +185    +42    +27 dist/jquery.slim.min.js
```

PR #5640, opened 2025-03-17
Tests: Add tests for `jQuery.get( String, null-ish, null-ish, String )`

mgol:
### Summary ###

Also, fix `mock.php` formatting to not fail the
`jQuery.get( String, null, String )` test in PHP mode.

Ref gh-4989
Ref jquery/api.jquery.com#1208

### Checklist ###

* [x] New tests have been added to show the fix or feature works
* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com - there's already a PR at jquery/api.jquery.com#1208

PR #5646, opened 2025-03-31
Tests: Fix tests for `jQuery.get( String, null-ish, null-ish, String )`

mgol:
### Summary ###

The original test's `text` variant just repeated the previous test with 3 parameters; the goal was to use 4 ones. This fixes it.

Ref gh-5640
Ref gh-5645

### Checklist ###

* [x] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com - PR already at https://github.com/jquery/api.jquery.com/pull/1208

### DIFF
--- a/test/unit/ajax.js
+++ b/test/unit/ajax.js
@@ -2823,7 +2823,7 @@ if ( typeof window.ArrayBuffer === "undefined" || typeof new XMLHttpRequest().re
 					done();
 				} );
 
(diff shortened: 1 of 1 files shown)


mgol:
I only noticed it post merge when trying to backport the test to `3.x-stable` and see it fail as the `3.x` line does not support the `jQuery.get( String, null, String )` signature. 🤦🏻

timmywil (approved):

PR #5652, opened 2025-04-07
WIP Build: Make the JavaScript middleware work on Nginx as well

mgol:
### Summary ###

Extract handlers independent from `middleware-mockserver.cjs` into a separate `middleware-handlers.js` file; add a separate wrapper for working with the nginx JavaScript engine.

I wanted to get some opinions on whether this makes sense. nginx can be used with njs or QuickJS; I used the latter as njs is very limited (it doesn't even support named imports). JS feature support is pretty good but most Node.js modules are missing. On the other hand, handlers returning the status code, headers & body don't need a lot beyond the ability to read files (which works) & read the full body (available out of the box). I'm also reading `import.meta.dirname` which is not supported either so I'm just passing the dirname from nginx.

Right now the nginx version only fails the multipart form data test, I haven't migrated that part. The rest passes.

A caveat: the nginx version available on macOS via Homebrew doesn't include the js module and it's impossible to add one. There's a separate tap for full nginx but [it's broken at the moment](https://github.com/denji/homebrew-nginx/issues/419).

I tried using the Docker `nginx:alpine` image and it worked without major issues, though. I could even use the config I currently use on macOS with only minor changes to enable the JS engine. The command to set this up:
```
docker run -d \
  --name nginx-njs \
  -p 80:80 \
  -v /Users/mgol/projects/public/jquery/jquery-core:/Users/mgol/projects/public/jquery/jquery-core \
  -v /Users/mgol/www/vhosts:/Users/mgol/www/vhosts \
  -v /opt/homebrew/etc/nginx:/opt/homebrew/etc/nginx:ro \
  -v /opt/homebrew/var/log/nginx:/opt/homebrew/var/log/nginx \
  nginx:alpine
```
Then I just had to drop the `/etc/nginx` folder and create a symlink to `/opt/homebrew/etc/nginx` in its place and add some config to the js module. It shouldn't be hard to create a Dockerfile for this if we wanted to make it easier for others as well.

If we were to land this, I'd prefer to do this in multiple PRs to not lose history on `middleware-mockserver.cjs`.

### Checklist ###


timmywil:
We discussed this and I believe we're going to look into more options using docker.

PR #5653, opened 2025-04-08
Core: Formally deprecate push/sort/splice

mgol:
### Summary ###

Since we inlined Sizzle into jQuery, we haven't been actually using these three
methods internally. Therefore, we could add deprecation warnings in Migrate 3.x
for them with jQuery 3.7.0 or newer - both 3.7.0 & 3.7.1 pass all the tests
with them removed (while 3.6.4 does not).

Some basic tests were added as well.

### Checklist ###

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -121,13 +121,7 @@ jQuery.fn = jQuery.prototype = {
 
 	end: function() {
 		return this.prevObject || this.constructor();
-	},
-
-	// For internal use only.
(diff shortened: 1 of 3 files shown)


mgol:
If we decided to land this, the next version would probably have to be 3.8.0, not 3.7.2. But it's been 1.5 years since 3.7.1 and `3.x-stable` had lots of changes landed during that time so this might actually be good.

timmywil (approved):

PR #5663, opened 2025-05-26
DOMEval make use of script nonce value for CSP nonce- validation

Agossi:
### Summary ###
Enabling the use of script nonce tags in jquery which allowed to make a CSP Policy without unsafe-eval for jquery.
the nonce will be picked from the script tag if set.
Or if it is already set nothing will be done.

### Checklist ###

### DIFF
--- a/src/core/DOMEval.js
+++ b/src/core/DOMEval.js
@@ -20,6 +20,13 @@ export function DOMEval( code, node, doc ) {
 		}
 	}
 
+	if (!script.nonce) {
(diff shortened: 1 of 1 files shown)


mgol:
I’m going to close this PR for now as I don’t think it fixes a real issue. If you provide more details we can consider reopening.

It’s a good approach to open an issue before opening a PR in most cases, with a test case - or add a test case in the PR from the start.

PR #5687, opened 2025-08-06
Build: Test on Safari 18 & 17 instead of "latest-1"

mgol:
### Summary ###

JTR doesn't take into account the jump from Safari 18 to 26, so we need to specify versions explicitly. Also, while BrowserStack already added macOS Tahoe with Safari 26, it's not a stable release yet, so we need to test on Safari 17 as well.

Ref jquery/jquery-test-runner#17

### Checklist ###

### DIFF
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -23,7 +23,13 @@ jobs:
         BROWSER:
           - 'IE_11'
           - 'Safari_latest'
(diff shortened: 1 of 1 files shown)


mgol:
Landed on `main` in 958369f08815618b02e485bc29fac81cf4f2e1f6 and on `3.x-stable` in 497414b22bafa6960eb19c51738183e874363663.

PR #5688, opened 2025-08-06
Tests: Use releases.jquery.com as external host for AJAX testing

mgol:
### Summary ###

Previously, we used example.com. However, that one is rate-limited, causing us frequent unit test failures.

### Checklist ###

### DIFF
--- a/test/data/testinit.js
+++ b/test/data/testinit.js
@@ -19,7 +19,7 @@ var parentUrl = window.location.protocol + "//" + window.location.host,
 	];
 
 // see RFC 2606
-this.externalHost = "example.com";
+this.externalHost = "releases.jquery.com";
 this.hasPHP = true;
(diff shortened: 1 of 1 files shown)


test/data/testinit.js
```diff
@@ -19,7 +19,7 @@ var parentUrl = window.location.protocol + "//" + window.location.host,
 	];
 
 // see RFC 2606
-this.externalHost = "example.com";
+this.externalHost = "releases.jquery.com";
```
mgol, line 22: Note that this will trigger a request to https://releases.jquery.com on every AJAX test run. But it's served by Fastly and the traffic will be waaay lower than the normal CDN one, so I hope we're fine. Thoughts, @Krinkle?

Krinkle: That's fine, yeah.

PR #5689, opened 2025-08-06
Release: Only run browserless tests during the release

mgol:
### Summary ###

`npm test` is susceptible to any flakiness in tests, making releases less predictable. Our process requires us to check CI before releases or run `npm test` manually if there are issues with that, we don't need to repeat it all during the release process.

In addition to that, currently the whole pre-release part is silent and the full `npm test` takes a long time to finish while no feedback is presented to the person running the release. It will be more bearable with just browserless tests.

### Checklist ###

### DIFF
--- a/build/release/pre-release.sh
+++ b/build/release/pre-release.sh
@@ -12,9 +12,10 @@ npm run release:clean
 # Check authors
 npm run authors:check
 
-# Run tests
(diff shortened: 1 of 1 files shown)


mgol:
Tests are failing because of the issue I'm fixing in #5688.

PR #5709, opened 2025-09-28
Fix: do postpone init unqouted in preFilter.js

vvepretskiy:
This PR addresses the issue where we do init unqouted before we need it.

Changes Made:

Moved init unqouted in check

Impact:

No breaking changes.

### Checklist ###


timmywil:
Thank you for your contribution. However, I think this change is unnecessary. The assignment here is instant, so there's no performance loss, and it's more readable.

PR #5711, opened 2025-09-28
Fix: move init data in block where they are required ajax.js

vvepretskiy:
This PR addresses the issue where we do init variables where they are not needed.

Changes Made:

Moved variables in block where they are required 

Impact:

No breaking changes.

### Checklist ###


mgol:
Thanks, but var declarations are hoisted to the top of the nearest function scope, which is why the jQuery code style forbids declarations in blocks.

PR #5716, opened 2025-10-22
Tests: Fix selector tests in Chrome 141

mgol:
### Summary ###

Chrome 141 incorrectly matches `:lang()` selectors with values with a trailing `-`. This is fixed in Chrome 142, so just skip these tests in v141. See https://issues.chromium.org/issues/451355198

`main` version of #5717

### Checklist ###


timmywil (approved):

mgol:
Chrome 142 stable is out (more precisely, on macOS it's 142.0.7444.60) and it contains the fix, so excluding v141 will be enough. Merging!

PR #5720, opened 2025-11-01
Build: Verify release using dist repo tarball

### LINKED ISSUE #5693: Fix release verification
The release verification for `4.0.0-rc.1` failed: https://github.com/jquery/jquery/actions/runs/16886583472/job/47850858969

It's probably caused by us now publishing a different folder (`tmp/release/dist`). We should update the script to work again before the final release.

timmywil:
### Summary ###

3.x version of #5719 

- Builds the tarball using the dist repo after running `npm run release:dist` from the build.
- Pulls the blog URL from the dist repo README
- Update the default Node in workflows to 24.x, which is now in LTS
- Ensure all commits and tags are signed. This change was prompted when I noticed the tag on jquery/jquery for 4.0.0-rc.1 was not signed. We were already signing the tag for jquery-dist, and the team signs commits with global git config, but release-it does not sign tags by default.
- If the tag does not yet exist in the dist repo (which can happen as a race condition during the release), it falls back to cloning the main branch. That should still work.

Ref gh-5693

**Note**: this doesn't work for any previous 3.x release, but it should work for any future ones.

### Checklist ###

### DIFF
--- a/.github/workflows/browser-tests.yml
+++ b/.github/workflows/browser-tests.yml
@@ -9,7 +9,7 @@ permissions:
   contents: read # to fetch code (actions/checkout)
 
 env:
-  NODE_VERSION: 22.x
+  NODE_VERSION: 24.x
 
 jobs:
   build-and-test:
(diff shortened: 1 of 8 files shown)


mgol (dismissed):
LGTM assuming you tested it locally.

timmywil:
@mgol I realized there could be a race condition. Because the dist tag is pushed after the main repo tag, it could run the workflow before the dist tag exists. I made it so it falls back to cloning the main branch. This should make it so verify works during a release whether the tag is there or not, but still support verifying previous releases (>= 4.0.0-rc.1).

mgol (approved):

PR #5727, opened 2025-11-17
Build: upgrade all dev dependencies (except sinon & uglify-js)

timmywil:
### Summary ###

3.x version of #5726

### Checklist ###


mgol (approved):
LGTM, but we used to have problems with IE 9, so please do a test JTR run with that browser on this branch before merging.

timmywil:
I just ran IE9 locally and it looks good. I think updating uglify is what caused failures in IE9 before, which has not been changed here.

PR #5728, opened 2025-11-20
Tests: Fix the "outside view position" test in Headless Chrome

mgol:
### Summary ###

In Headless Chrome, when run locally on macOS together with a few other modules, the scroll handler for the appended element in the "outside view position" dimensions test doesn't fire, timing out the test & failing it as a result.

While the scroll handler may not fire, the new position data is available immediately, so just do the checks directly, without relying on scroll handlers.

`3.x` version: gh-5729

Ref gh-5729

### Checklist ###


timmywil (approved):
LGTM. Strange that it only happens when run with other modules, but we may as well make the tests simpler.

PR #5729, opened 2025-11-20
Tests: Fix the "outside view position" test in Headless Chrome (3.x)

mgol:
### Summary ###

In Headless Chrome, when run locally on macOS together with a few other modules, the scroll handler for the appended element in the "outside view position" dimensions test doesn't fire, timing out the test & failing it as a result.

While the scroll handler may not fire, the new position data is available immediately, so just do the checks directly, without relying on scroll handlers.

`main` version: gh-5728

Ref gh-5728

### Checklist ###


timmywil (approved):

mgol:
I tested it on my fork and BrowserStack was green.

PR #5763, opened 2026-01-28
Release: fetch contributors with one fetch; other changelog fixes

timmywil:
- Fetch all contributors with a single GitHub API request. The 4.0 changelog run hit the rate limit because it was doing a fetch for each commit. I had just kept the list from the beta release until I could fix it, but the blog post has now been updated.
- Automatically omit Build, Release, and Upgrade commits (we had merged a couple dependabot commits in the 4.0 release that started with "Upgrade:")
- Fix sorting for scopes that may not have the right casing (the case for one commit that started with `"docs:"`)
- Exclude headers from the changelog HTML for easy copy/pasting into the blog post
- Make contributors.html exactly what I need to copy/paste into the blog post

### DIFF
--- a/build/release/changelog.js
+++ b/build/release/changelog.js
@@ -172,6 +172,8 @@ function addHeaders( commits ) {
 }
 
 async function getContributors() {
+
+	// https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#compare-two-commits
 	const response = await fetch(
 		`https://api.github.com/repos/jquery/jquery/compare/${ prevVersion }...${ nextVersion }`,
 		{
--- a/build/release/changelog.js
+++ b/build/release/changelog.js
@@ -17,6 +17,9 @@ const prevVersion = argv[ 2 ];
 const nextVersion = argv[ 3 ];
 const blogUrl = process.env.BLOG_URL;
 
+// Dependabot used "Upgrade:" for some commits
+const omitScopes = [ "build", "release", "upgrade" ];
+
 if ( !prevVersion || !nextVersion ) {
 	throw new Error( "Usage: `node changelog.js PREV_VERSION NEXT_VERSION`" );
 }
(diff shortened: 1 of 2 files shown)


build/release/changelog.js
```diff
@@ -168,33 +186,27 @@ async function getGitHubContributor( sha ) {
...
 		throw new Error( JSON.stringify( data ) );
 	}
-	return { name: data.commit.author.name, url: data.author.html_url };
-}
 
-function uniqueContributors( contributors ) {
-	const seen = {};
+	const contributors = data.commits.map( ( commit ) => ( {
+
+		// Normalize to avoid duplicates due to different Unicode forms
```
mgol: Nice!

mgol (approved):
Mostly LGTM, just a few minor remarks.

build/release/changelog.js
```diff
@@ -168,33 +186,27 @@ async function getGitHubContributor( sha ) {
...
 
 		// The data may contain multiple helpful fields
 		throw new Error( JSON.stringify( data ) );
 	}
-	return { name: data.commit.author.name, url: data.author.html_url };
-}
 
-function uniqueContributors( contributors ) {
-	const seen = {};
+	const contributors = data.commits.map( ( commit ) => ( {
```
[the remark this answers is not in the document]
timmywil: It's not visible when viewing changes in this PR, but it's line 218.

build/release/changelog.js
mgol: It will also be faster & more reliable to run Git locally and avoid a remote API request.

timmywil: We use the GitHub API because the blog links to GitHub handles.

mgol: OK, never mind; I was hoping for an easy & quick solution, but it looks like there's no easy way to achieve what I wanted with the current API, so what we have right now is good enough.

PR #5764, opened 2026-01-28
Release: fetch contributors with one fetch; other changelog fixes

timmywil:
3.x copy of #5763 

- Fetch all contributors with a single GitHub API request.
- Automatically omit Build, Release, and Upgrade commits (we had merged a couple dependabot commits that started with "Upgrade:")
- Fix sorting for scopes that may not have the right casing (the case for one commit that started with `"docs:"`)
- Exclude headers from the changelog HTML for easy copy/pasting into the blog post
- Make contributors.html exactly what I need to copy/paste into the blog post

### DIFF
--- a/build/release/changelog.js
+++ b/build/release/changelog.js
@@ -172,6 +172,8 @@ function addHeaders( commits ) {
 }
 
 async function getContributors() {
+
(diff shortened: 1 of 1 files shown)


mgol (approved):
Is this identical to gh-5763? If so, then LGTM, but please address my remarks from gh-5763 and backport any changes here.

PR #5775, opened 2026-02-18
CSS: Use createElementNS to support XSLT documents

### LINKED ISSUE #4730: finalPropName.js: emptyStyle undefined for document generated by in-browser XSLT
### Description ###
src/css/finalPropName.js defines a variable: 
```
var cssPrefixes = [ "Webkit", "Moz", "ms" ],
	emptyStyle = document.createElement( "div" ).style,
	vendorProps = {};
```

that is used to do CSS property name translation. Under Chromium (tested v81) and Firefox (tested v76) if the document was loaded via an in-browser XSLT transformation then `document.createElement( "div" ).style` is undefined and attempts to use the `emptyStyle` variable in the remainder of the finalPropName code fail. 

Recommended solution: Change emptyStyle line above to `emptyStyle = document.createElementNS("http://www.w3.org/1999/xhtml", "div" ).style`. This seems to work regardless of whether the document is XSLT or XML or not and in both Chromium and Firefox. 

### Link to test case ###
(Simple XSLT transformed XHTML. The transformation used is a no-op.) 
http://thermal.cnde.iastate.edu/jqbug.xhtml (You will see an error in the browser console). Compare to http://thermal.cnde.iastate.edu/jqbug_fixed.xhtml (Difference is a patched jquery with above fix in place plus another patch based on current master branch) 

The issue with .style being undefined is easy enough to demonstrate. Load either of the above test case URLs. In the Chromium/Chrome console enter: `document.createElement( "div" ).style`. It will show as undefined. Then try entering: 
`document.createElementNS("http://www.w3.org/1999/xhtml","div").style` and it will give the correct object.

apoorvdarshan:
## Summary
- In XSLT-transformed documents, `document.createElement("div").style` is `undefined` because elements are created in the XML namespace rather than XHTML
- Changed `document.createElement` to `document.createElementNS("http://www.w3.org/1999/xhtml", ...)` in `src/css/finalPropName.js` and `src/css/support.js` so that `.style` is always available
- This is the fix recommended in the original issue discussion

## Files changed
- **`src/css/finalPropName.js`** — `emptyStyle` element creation now uses XHTML namespace
- **`src/css/support.js`** — `table`, `col`, `tr`, `td` element creation now uses XHTML namespace (the `table.style` guard on line 19 already existed, but the root cause is the same)

## How to reproduce the original bug
1. Load a page via in-browser XSLT transformation
2. In the console: `document.createElement("div").style` → `undefined`
3. Compare: `document.createElementNS("http://www.w3.org/1999/xhtml", "div").style` → `CSSStyleDeclaration`

Fixes gh-4730

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -1,5 +1,5 @@
 import { jQuery } from "./core.js";
-import { document } from "./var/document.js";
+import { createElement } from "./var/createElement.js";
 import { rnothtmlwhite } from "./var/rnothtmlwhite.js";
 import { location } from "./ajax/var/location.js";
 import { nonce } from "./ajax/var/nonce.js";
@@ -44,7 +44,7 @@ var
 	allTypes = "*/".concat( "*" ),
 
 	// Anchor tag for parsing the document origin
-	originAnchor = document.createElement( "a" );
+	originAnchor = createElement( "a" );
 
 originAnchor.href = location.href;
 
@@ -541,7 +541,7 @@ jQuery.extend( {
 
 		// A cross-domain request is in order when the origin doesn't match the current origin.
 		if ( s.crossDomain == null ) {
-			urlAnchor = document.createElement( "a" );
+			urlAnchor = createElement( "a" );
 
 			// Support: IE <=8 - 11+
 			// IE throws exception on accessing the href property if url is malformed,
--- a/src/css/finalPropName.js
+++ b/src/css/finalPropName.js
@@ -1,7 +1,7 @@
-import { document } from "../var/document.js";
+import { createElement } from "../var/createElement.js";
 
 var cssPrefixes = [ "Webkit", "Moz", "ms" ],
-	emptyStyle = document.createElementNS( "http://www.w3.org/1999/xhtml", "div" ).style;
+	emptyStyle = createElement( "div" ).style;
 
 // Return a vendor-prefixed property or undefined
 function vendorPropName( name ) {
--- a/src/css/finalPropName.js
+++ b/src/css/finalPropName.js
@@ -1,7 +1,7 @@
 import { document } from "../var/document.js";
 
 var cssPrefixes = [ "Webkit", "Moz", "ms" ],
-	emptyStyle = document.createElement( "div" ).style;
+	emptyStyle = document.createElementNS( "http://www.w3.org/1999/xhtml", "div" ).style;
 
 // Return a vendor-prefixed property or undefined
 function vendorPropName( name ) {
--- a/src/css/support.js
+++ b/src/css/support.js
@@ -1,11 +1,11 @@
 import { jQuery } from "../core.js";
-import { document } from "../var/document.js";
+import { createElement } from "../var/createElement.js";
 import { documentElement } from "../var/documentElement.js";
 import { support } from "../var/support.js";
 import { isIE } from "../var/isIE.js";
 
 var reliableTrDimensionsVal, reliableColDimensionsVal,
-	table = document.createElementNS( "http://www.w3.org/1999/xhtml", "table" );
+	table = createElement( "table" );
 
 // Executing table tests requires only one layout, so they're executed
 // at the same time to save the second computation.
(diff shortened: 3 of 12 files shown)


mgol (changes requested):
Thanks for the PR. However, source changes are just a small part of what would be required to support XML documents, even on a basic level.

See my comment at https://github.com/jquery/jquery/issues/4730#issuecomment-685868151. Some infrastructure parts have changed, but you'd still need to modify `mock.php` & `mockserver.js` to support serving true XML documents and then define a few tests using them via `testIframe`. You can look at some tests already using `testIframe` to see how to use it; focus on the ones that pass `mock.php?action=...` as the URL.

Also, we have many more `document.createElement` usage, it seems we should replace them all? Or maybe create file `./src/var/createElement.js` that would wrap the namespace logic? That'd also require measuring the size impact in both cases (running the build locally will tell you the sizes & the diff from a previous build).

Would you like to try that?

apoorvdarshan:
Thanks for the review @mgol! I've pushed a follow-up commit addressing your feedback:

### Changes

1. **Created `src/var/createElement.js` helper** — Wraps `document.createElementNS("http://www.w3.org/1999/xhtml", tag)` so the namespace logic is centralized in one place.

2. **Replaced all `document.createElement` calls in `src/`** with the new helper:
   - `src/ajax.js` — 2 calls (anchor elements for URL parsing)
   - `src/css/finalPropName.js` — 1 call (was already using inline `createElementNS`)
   - `src/css/support.js` — 4 calls (were already using inline `createElementNS`)

3. **Added XML document test via `testIframe`** — Serves an XHTML page as `application/xml` through a new `xmlCss` mock action (added to both `mock.php` and `middleware-mockserver.cjs`). The test verifies:
   - `document.createElement("div").style` is indeed `undefined` in the XML context (confirming the bug condition)
   - `jQuery.css("width")` still works correctly (confirming our fix)

### Size impact

| | Before | After | Diff |
|---|---|---|---|
| jquery.js | 256,155 | 256,282 | +127 |
| jquery.min.js | 78,931 | 78,761 | **-170** |
| jquery.min.js (gzip) | 27,639 | 27,633 | **-6** |

The minified size actually **decreased** because the repeated `"http://www.w3.org/1999/xhtml"` namespace string is now only in one place.

All CSS, ajax, and support tests pass on both Chrome and Firefox.

mgol:
The size impact table is comparing with the previous iteration, right? Against `main` it's a size increase:
```
main @7aa0179514a2aa9e1d90fbc6a4b332b8f1ae1292
   raw     gz     br Filename
   -17    +33    +20 dist/jquery.min.js
    +9    +40    +57 dist/jquery.slim.min.js
   -17    +34    +66 dist-module/jquery.module.min.js
    +9    +40    -27 dist-module/jquery.slim.module.min.js
```

apoorvdarshan:
You're right, the size table was comparing against the previous commit which already had inline `createElementNS` calls, not against `main`. Against `main` it's a net size increase as you showed. Thanks for clarifying!

mgol (changes requested):

apoorvdarshan:
Pushed a follow-up commit addressing the review feedback:

1. **Removed redundant comment** from `xmlDocument.xhtml`
2. **Single `var` declaration** in test scripts instead of multiple `var` statements
3. **Wrapped test code in try-catch**, passing a `threw` parameter to `startIframeTest` so the test doesn't hang if jQuery crashes without the fix
4. **Added ajax cross-domain detection test** (`mock.php?action=xmlAjax`) -- verifies that in an XML document, same-origin URLs are detected as same-origin and cross-origin URLs are detected as cross-domain

mgol (changes requested):

apoorvdarshan:
Pushed another commit addressing the latest feedback:

1. **Var ordering** - uninitialized vars first, initialized (`threw = false`) last, applied to both test pages
2. **Dropped empty line** between assignments in CSS test page
3. **Renamed files** - `xmlDocCss.xhtml` and `xmlDocCrossDomainDetection.xhtml` for unique names
4. **Updated createElement.js comment** to reference XML documents generally, not just XSLT

mgol (approved):
Thanks! This LGTM. I would not backport it to the `3.x-stable` line as it has enough additional support tests that may cause more issues.

I want to discuss it with the team before merging, though, so please wait a bit more.

BTW, did you use AI to help with making the changes? I don't mind either way as the changes are high-quality and all my feedback was addressed; in the past, we've had issues with AI-submitted contributions that didn't make any sense, so that'd be the first for me here.

apoorvdarshan:
Thanks for the review! Happy to wait for the team discussion.

And yes, I used Claude Code (Opus 4.6) to assist with the changes.

gibson042 (approved):

timmywil (approved):

mgol:
Landed, thanks!

Issue #3000, opened 2016-03-14
`deferred.resolve` doesn't unwrap promises


mgol:
This:
```
jQuery.Deferred().resolve($.when(1)).then(d => console.log(d.then));
jQuery.Deferred().resolve(Promise.resolve(1)).then(d => console.log(d.then));
```
should log undefined but instead logs promises' `then` methods, diverging from native `Promise.resolve` behavior.

mgol:
cc @gibson042 @jaubourg: is this intended?

jaubourg:
AFAIK, @gibson042 preferred the current approach. The rewrite I proposed back then handled the unwrapping so it's not difficult to handle if needs be.

mgol:
Native promises unwrap when they can, I'm not sure it's even possible to get the native promise as a parameter to the `then` callback. The problem with not unwrapping is that attaching `.then(d => d)` may change the final result which is IMO really unexpected. Compare:
```js
jQuery.Deferred().resolve($.when(1)).then(d => console.log(d)); // logs a promise
jQuery.Deferred().resolve($.when(1)).then(d => d).then(d => console.log(d)); // logs 1
```

Unfortunately, I've noticed it late in the 3.0.0 cycle so it may be too late for 3.0.0 but then it may only make it to 4.0.0 as it's a breaking change.

Thoughts, @jquery/core?

@jaubourg Would it require only trivial changes?

jaubourg:
I'm not that familiar with the new Deferred code but I don't think it would be very hard to detect if the input a promise.

gibson042:
It is intended for [2.x compatibility](https://github.com/jquery/jquery/blob/5c01cb1cc4a41b29d6739d061de8217c33037639/test/unit/deferred.js#L660-L663), which loomed much more heavily when the code was written. Now, I've had all the fight beaten out of me on this point, much as I still despise the resolve/reject asymmetry.

I will note however, that idiomatic casting is best achieved with `jQuery.when( maybeThenable )` (cf. https://github.com/jquery/jquery/issues/2018#issuecomment-70531678 ) and fulfilled value value unwrapping with `deferred.then().then(…)`.

dmethvin:
What is the benefit of *not* unwrapping?

gibson042:
[...]

PR #5784, opened 2026-02-25
Deferred: Preserve synchronous resolution in jQuery.when for Deferreds

### LINKED ISSUE #4798: jQuery.when: unnecessary asynchronous behavior when called with only one argument
### Description ###
The function "jQuery.when" shows apparently unneccessary asynchronous behavior when called with only one argument.

For example executing the code below
```
const deferred = jQuery.Deferred();
const promise = jQuery.when(deferred);
promise.done(() => console.log("promise done handler called"));
promise.always(() => console.log("promise always handler called"));
console.log("before deferred resolving the promise state is", promise.state());
deferred.resolve();
console.log("after deferred resolving the promise state is", promise.state());
```
in a javascript console produces the following console output
```
"before deferred resolving the promise state is" "pending"
"after deferred resolving the promise state is" "pending"
undefined
"promise done handler called"
"promise always handler called"
```

There really seems to be no reason for the handlers added to the promise returned by the function "jQuery.when" to not be called synchronously while the deferred object (provided as first and only argument to that call of the function "jQuery.when") is being resolved.

When I add any single additional argument (like `undefined`) to that call of the function "jQuery.when"
```
const promise = jQuery.when(deferred, undefined);
```
the unnecessary asynchronous behavior disappears and the handlers are called as synchronously as expected.

### Link to test case ###
I was able to reproduce the behavior using the console at https://codepen.io/mgol/pen/wNWJbZ without any problems, thereby I hope no link to a test case will be necessary here.

Nicknamess96:
When `jQuery.when()` is called with a single pending jQuery Deferred, the returned
promise now resolves synchronously (via `.done()`/`.fail()`) instead of being
scheduled asynchronously via `setTimeout` through `.then()`.

## The Problem

As reported in #4798, `jQuery.when(deferred)` unnecessarily introduces asynchronous
behavior when the single argument is a jQuery Deferred:

```js
const deferred = jQuery.Deferred();
const promise = jQuery.when(deferred);
promise.done(() => console.log("done"));
console.log("before");
deferred.resolve();
console.log("after");
// Before fix: "before", "after", "done" (async via setTimeout)
// After fix:  "before", "done", "after" (synchronous)
```

This happens because `jQuery.when` returns `primary.then()` for pending single-value
cases, and `.then()` schedules handlers via `window.setTimeout` per the Promises/A+
spec. However, jQuery Deferreds use `.done()`/`.fail()` for synchronous notification,
so this async scheduling is unnecessary when the argument is a Deferred.

## The Fix

When the single argument has a `.promise()` method (i.e., is a jQuery Deferred),
pipe resolution through a subordinate Deferred using `.done()`/`.fail()` instead
of `.then()`. The resolved value is re-adopted via `adoptValue` to preserve
secondary thenable unwrapping (cf. gh-3000).

Non-jQuery thenables (native Promises, custom thenables) still use the `.then()`
path, preserving their expected async behavior.

## Tests

Added a new test case `"jQuery.when(pendingDeferred) - synchronous resolution
(gh-4798)"` that verifies:
- `.done()` handler fires synchronously on resolve
- `.fail()` handler fires synchronously on reject
- Resolved values are correctly propagated
- Multi-value resolution preserves all values
- Promise state updates synchronously

All existing tests pass, including the Promises/A+ compliance suite (872/872).

Fixes gh-4798

### DIFF
--- a/src/deferred.js
+++ b/src/deferred.js
@@ -356,6 +356,9 @@ jQuery.extend( {
 			// the primary Deferred
 			primary = jQuery.Deferred(),
 
+			// Deferred used to synchronously pipe single-Deferred values (gh-4798)
+			subordinate,
+
 			// subordinate callback factory
 			updateFunc = function( i ) {
 				return function( value ) {
@@ -375,6 +378,19 @@ jQuery.extend( {
 			// Use .then() to unwrap secondary thenables (cf. gh-3000)
 			if ( primary.state() === "pending" ||
 				typeof( resolveValues[ i ] && resolveValues[ i ].then ) === "function" ) {
+
+				// Privilege jQuery Deferreds to preserve synchronous resolution
+				// when possible (cf. gh-4798). Pipe through a subordinate Deferred
+				// using .done()/.fail() (which are synchronous) and re-adopt the
+				// resolved value to handle secondary thenables.
+				if ( singleValue && typeof singleValue.promise === "function" ) {
+					subordinate = jQuery.Deferred();
+					primary.done( function() {
+						adoptValue( resolveValues[ i ], subordinate.resolve,
+							subordinate.reject );
+					} ).fail( subordinate.reject );
+					return subordinate.promise();
+				}
 
 				return primary.then();
 			}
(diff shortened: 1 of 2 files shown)


Nicknamess96:
Friendly ping — this fixes a regression where jQuery.when doesn't preserve synchronous resolution for Deferreds. Tests pass. Happy to address any feedback.

PR #5793, opened 2026-03-06
Tests: Add jQuery.ajaxPrefilter & jQuery.ajaxTransport tests

mgol:
### Summary ###

This is a prerequiste to gh-4634 which is a prerequisite to gh-4405.

Ref gh-4405
Ref gh-4634

### Checklist ###

* [x] New tests have been added to show the fix or feature works
* ~~If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com~~

PR #5795, opened 2026-03-09
Tests: Re-enable the AJAX binary data tests in IE

mgol:
### Summary ###

The previous check was only filtering out IE <10 which we don't test against on this branch anyway. In gh-5793, this test was erroneously disabled in IE 11 as well. This change re-enables it globally.

Ref gh-5793

### Checklist ###


mgol:
I've already fixed that on `3.x-stable` before landing: https://github.com/jquery/jquery/commit/ef30ce1f390d4c55b1cb06601d96f2351a48f761#diff-14e061231fbbb8627b9f909372ec44c3317d79a19afe4e9c8383e268695102c4R2178-R2188

timmywil (approved):

PR #5822, opened 2026-04-13
Selector: Be more strict in quick selection paths

mgol:
### Summary ###

Since we now pass tag/class selectors via querySelectorAll in the quick path,
fix the regex to be more strict in parsing tags/classes and only allow letters
at the first character. The spec allows for more, but:
1. Not all our supported browsers follow that (probably only IE 11 does not).
2. A complete regex would be more complex.

The new version should work everywhere and match the vast majority of simple
use cases; more complex cases can fall back to regular selection.

Also, since IE doesn't support identifiers starting with a dash, and identifiers
starting with a double dash are not escaped via `jQuery.escapeSelector`,
add fallbacks to TAG & CLASS selection in case qSA throws.

Size diffs:
```
main @53a2bcf984ec5d1955d35ea9565e847cbeea0200
   raw     gz     br Filename
   +61    +29    -34 dist/jquery.min.js
   +61    +27    +25 dist/jquery.slim.min.js
   +61    +31    -43 dist-module/jquery.module.min.js
   +61    +30     -5 dist-module/jquery.slim.module.min.js
```

### Checklist ###

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -5,6 +5,7 @@ import { indexOf } from "./var/indexOf.js";
 import { pop } from "./var/pop.js";
 import { push } from "./var/push.js";
 import { whitespace } from "./var/whitespace.js";
+import { rdoubleDash } from "./var/rdoubleDash.js";
 import { rbuggyQSA } from "./selector/rbuggyQSA.js";
 import { rtrimCSS } from "./var/rtrimCSS.js";
 import { isIE } from "./var/isIE.js";
@@ -62,7 +63,7 @@ var i,
 	rheader = /^h\d$/i,
 
 	// Easily-parseable/retrievable ID or TAG or CLASS selectors
-	rquickExpr = /^(?:#([\w-]+)|(\.?[\w-]+))$/,
+	rquickExpr = /^(?:#([\w-]+)|(\.?[a-z][\w-]*))$/i,
 
 	// Used for iframes; see `setDocument`.
 	// Support: IE 9 - 11+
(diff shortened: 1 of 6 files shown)


test/unit/selector.js
```diff
@@ -267,6 +267,34 @@ QUnit.test( "broken selectors throw", function( assert ) {
...
+QUnit[ QUnit.jQuerySelectors ? "test" : "skip" ](
+	"lenient identifier parsing", function( assert ) {
+
+	// Test that jQuery identifier parsing is more lenient than
+	// querySelectorAll's. Note: this may change in a jQuery major
+	// version bump.
+
+	assert.expect( 9 );
+
+	try {
```
Krinkle: Do we need the try-catch? QUnit catches uncaught errors by default and reports them as test failures. Subsequent tests still run either way, to give a comoplete test report.

mgol: Oh, yeah, good point. Originally, I had some error messages for each failed assertion in a separate try-catch, but later I unified it, so yeah, probably not a lot of gain at wrapping it.

mgol: Removed.

src/selector.js
```diff
@@ -132,9 +133,9 @@ function find( selector, context, results, seed ) {
 					// `querySelectorAll` is, depending on the browser, either on par
 					// perf-wise with `getElementsByTagName` & `getElementsByClassName`
 					// or even faster, so we don't use `gEBTN` & `gEBCN` anymore.
-					// Note: we can only get tags & classes matching `[\w-]+` here thanks
-					// to `rquickExpr`, so there's no need to wrap them with
-					// `jQuery.escapeSelector`.
+					// Note: we can only get tags & classes matching `/[a-z][\w-]*/i`
+					// here thanks to `rquickExpr`, so there's no need to wrap them
+					// with `jQuery.escapeSelector`.
```
gibson042:
```suggestion
					// Note: thanks to the restrictions of `rquickExpr`, there's no
					// need to wrap them with `jQuery.escapeSelector`.
```

gibson042 (approved):

timmywil (approved):

PR #5825, opened 2026-04-20
Build: Upgrade dependencies; move pre-commit to pre-push

timmywil:
- upgraded most node dependencies, excluding sinon, but including eslint v9, which required some updates to the no-unused-args rule. The new `caughtErrorsIgnorePattern` allows for ignoring with a `_` prefix and setting a more readable name or simply using `e`, which we are prone to do.
- removed stale rules for old test runner scripts in the eslint config.
- moved the pre-commit hook to pre-push. I prefer this as you really only need to do the check when making a PR.


eslint.config.js
```diff
@@ -191,7 +191,7 @@ export default [
 
 			"no-unused-vars": [
 				"error",
-				{ args: "after-used", argsIgnorePattern: "^_" }
+				{ args: "after-used", argsIgnorePattern: "^_", caughtErrorsIgnorePattern: "^_|^e$" }
```
mgol: Could you break that into multiple lines? It got a bit long.

.husky/pre-push
mgol, line 1: How will this work for multi-commit PRs? Does it verify every commit individually? Did you check?

mgol (approved):
One question, one suggestion, but LGTM otherwise. Thanks! It makes sense to do it this way rather than merge 3 separate deps update PRs.

.husky/pre-push
timmywil: It does not, but that is intentional. We squash all PRs now to retain signatures, so commit atomicity is not as important as separating concerns between PRs.

mgol: It's not about commit atomicity, but more about verifying the format of commit messages. I get that we squash anyway so this doesn't directly translate, but I wonder what's the value-add then. Is it that most PRs are single-commit and we accept the other ones may not be checked in such detail?

PR #4634, opened 2020-03-01
Ajax: Support an alternative completeCallback API for transports

mgol:
### Summary ###

Apart from the existing API:

```js
function( status, statusText, responses, headers ) {}
```
a new API is now available:
```js
function( { status, statusText, responses, headers } ) {}
```

This makes it possible to add new parameters in the future without relying on
their order among parameters and being able to provide them selectively.

Ref gh-4405

+37 bytes. Not changing existing transports would make that smaller but we'll need to change the XHR one anyway to land #4405 and then the build gets smaller if we update it in all the places I modified in the PR.

*Note:* This still needs tests. That said, we don't have any direct tests for [`jQuery.ajaxTransport`](https://api.jquery.com/jQuery.ajaxTransport/), its only implicitly tested by the virtue of two core transports using this API. Therefore, before landing this I'd like to write a few tests for the API first. I'm opening this draft PR now to gain feedback whether this is a direction in which we'd like to go.

### Checklist ###

* [x] All authors have signed the CLA at https://cla.js.foundation/jquery/jquery
* [ ] New tests have been added to show the fix or feature works
* [x] Grunt build and unit tests pass locally with these changes
* [ ] If needed, a docs issue/PR was created at https://github.com/jquery/api.jquery.com

PR #5826, opened 2026-04-21
Ajax: Add the `responseURL` field to jqXHR

### LINKED ISSUE #4339: Adding responseURL field to jqXHR

mgol:
### Summary ###

Expose `xhr.responseURL` (the final URL after redirects, including any redirects followed by the browser) as `jqXHR.responseURL`. The XHR transport passes it through the object-form `completeCallback` introduced in gh-4634, so the positional signature stays untouched.

Following the existing `responseText`/`responseXML`/`responseJSON` pattern, `jqXHR.responseURL` is only set when a transport reports it. As a result:

* modern-browser XHR: the final URL string
* IE 11 XHR: `undefined` (no native support)
* script / JSONP / aborted requests: `undefined` (transport has no way to know the URL)

A reusable `redirect` mock action has been added to both the Node middleware and `mock.php` so the redirect path can be exercised.

Fixes gh-4339
Ref gh-4634
Ref gh-5793

### Checklist ###

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -713,7 +713,8 @@ jQuery.extend( {
 
 		// Callback for when everything is done
 		function done( status, nativeStatusText, responses, headers ) {
-			var isSuccess, success, error, response, modified, statusText;
+			var isSuccess, success, error, response, modified, statusText,
+				responseURL;
 
 			if ( typeof status === "object" ) {
 
@@ -721,6 +722,7 @@ jQuery.extend( {
 				nativeStatusText = status.statusText;
 				responses = status.responses;
 				headers = status.headers;
+				responseURL = status.responseURL;
 
 				status = status.status;
 			}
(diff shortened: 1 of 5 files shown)


timmywil (approved):
LGTM

gibson042 (approved):

PR #5829, opened 2026-05-01
Preserve CSP nonce in script evaluation paths

metsw24-max:
### Summary ###

Fixes loss of CSP nonce during internal script evaluation in jQuery.

When scripts are evaluated via `DOMEval` or `domManip`, the `nonce` attribute is copied using property access (`node[i]`). However, `nonce` is not consistently exposed as a property across browsers and is often only available via `getAttribute("nonce")`.

This can result in the nonce being dropped when scripts are recreated or executed causing them to fail under strict Content Security Policy (CSP).

This patch:
- Reads nonce via `getAttribute("nonce")` when available
- Falls back to property access for compatibility
- Applies the fix in both `DOMEval` and `domManip` paths

Security impact:
Preserving nonce ensures compatibility with strict CSP and prevents applications from weakening their policies (e.g., enabling `unsafe-inline`), which would reduce protection against script injection (XSS).

### Checklist ###

### DIFF
--- a/src/core/DOMEval.js
+++ b/src/core/DOMEval.js
@@ -11,12 +11,20 @@ export function DOMEval( code, node, doc ) {
 	doc = doc || document;
 
 	var i,
+		val,
 		script = doc.createElement( "script" );
 
 	script.text = code;
 	for ( i in preservedScriptAttributes ) {
-		if ( node && node[ i ] ) {
-			script[ i ] = node[ i ];
+		if ( node ) {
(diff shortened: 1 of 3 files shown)


mgol:
> However, nonce is not consistently exposed as a property across browsers and is often only available via getAttribute("nonce").

This is not true.

If you see an issue with the current code, open an issue first with a test case & clear steps to reproduce.

PR #5830, opened 2026-05-01
Selector: Optimize jQuery.uniqueSort duplicate removal

metsw24-max:
### Summary

This PR refactors the duplicate removal logic in jQuery.uniqueSort to improve performance and reduce memory overhead.

The previous implementation relied on repeated splice operations inside a loop and used an auxiliary duplicates array, resulting in O(N²) time complexity and O(N) additional space in the worst case.

This patch replaces that logic with a two-pointer array compaction approach that performs deduplication in a single pass and trims the array with a single splice call.

### Key improvements:
- Reduces time complexity of duplicate removal from O(N²) to O(N)
- Eliminates auxiliary array usage (O(N) → O(1) space)
- Preserves existing behavior and output ordering
- Keeps in-place mutation semantics unchanged

### DIFF
--- a/src/selector/uniqueSort.js
+++ b/src/selector/uniqueSort.js
@@ -70,9 +70,7 @@ function sortOrder( a, b ) {
  * @param {ArrayLike} results
  */
 jQuery.uniqueSort = function( results ) {
-	var elem,
-		duplicates = [],
-		j = 0,
+	var j = 0,
 		i = 0;
 
 	hasDuplicate = false;
@@ -80,14 +78,13 @@ jQuery.uniqueSort = function( results ) {
 	sort.call( results, sortOrder );
 
 	if ( hasDuplicate ) {
-		while ( ( elem = results[ i++ ] ) ) {
-			if ( elem === results[ i ] ) {
-				j = duplicates.push( i );
+		j = 1;
+		for ( i = 1; i < results.length; i++ ) {
+			if ( results[ i ] !== results[ i - 1 ] ) {
+				results[ j++ ] = results[ i ];
 			}
 		}
-		while ( j-- ) {
-			splice.call( results, duplicates[ j ], 1 );
-		}
+		splice.call( results, j, results.length - j );
 	}
 
 	return results;
--- a/src/selector/uniqueSort.js
+++ b/src/selector/uniqueSort.js
@@ -70,21 +70,29 @@ function sortOrder( a, b ) {
  * @param {ArrayLike} results
  */
 jQuery.uniqueSort = function( results ) {
-	var j = 0,
-		i = 0;
+	var j = 1,
+		i = 1;
 
 	hasDuplicate = false;
 
 	sort.call( results, sortOrder );
 
 	if ( hasDuplicate ) {
-		j = 1;
-		for ( i = 1; i < results.length; i++ ) {
+
+		// Pack the first instance of each unique element into the start of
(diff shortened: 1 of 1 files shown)


src/selector/uniqueSort.js
```diff
@@ -70,24 +70,21 @@ function sortOrder( a, b ) {
  * @param {ArrayLike} results
  */
 jQuery.uniqueSort = function( results ) {
-	var elem,
-		duplicates = [],
-		j = 0,
+	var j = 0,
 		i = 0;
```
gibson042: These can both just start at one, eliminating the initial reassignment.
```suggestion
	var j = 1,
		i = 1;
```

src/selector/uniqueSort.js
```diff
@@ -70,24 +70,21 @@ function sortOrder( a, b ) {
...
 	hasDuplicate = false;
 
 	sort.call( results, sortOrder );
 
 	if ( hasDuplicate ) {
-		while ( ( elem = results[ i++ ] ) ) {
-			if ( elem === results[ i ] ) {
-				j = duplicates.push( i );
+		j = 1;
+		for ( i = 1; i < results.length; i++ ) {
```
gibson042:
```suggestion
		for ( ; i < results.length; i++ ) {
```
This deviates from most of our loops, which have the form they do because of a historical need to work with NodeLists in which the numeric `length` might be shadowed by a constituent element with id "length". But performance-wise, it's much preferable to avoid walking off the end of a collection, especially when it is an array (#3769), and I'd like to migrate to this pattern (and I think we're safe here even without broad migration).

And all that aside, this should be preceded by an explanatory comment, e.g.
```suggestion
		// Pack the first instance of each unique element into the start of
		// results (starting at index 1 because index 0 is always kept), then
		// splice away the tail of duplicates.
		for ( ; i < results.length; i++ ) {
```

src/selector/uniqueSort.js
```diff
@@ -70,24 +70,21 @@ function sortOrder( a, b ) {
...
+			if ( results[ i ] !== results[ i - 1 ] ) {
+				results[ j++ ] = results[ i ];
 			}
 		}
-		while ( j-- ) {
-			splice.call( results, duplicates[ j ], 1 );
-		}
+		splice.call( results, j, results.length - j );
```
gibson042: deleteCount can be omitted to delete all elements from the specified index.
```suggestion
		splice.call( results, j );
```
And we can also get a substantial speedup of this final step when dealing with an actual array rather than a jQuery instance (about 3x in V8/Chrome and 1.5x in JavaScriptCore/Safari):
```suggestion
		if (Array.isArray(results)) {
			results.length = j;
		} else {
			splice.call( results, j );
		}
```

gibson042 (approved):
This makes sense intuitively, and I also confirmed that this is significantly faster in practice after refactoring the new implementation for an apples-to-apples comparison regarding how to loop: https://jsfiddle.net/fnr2jLgz/

I'd like to accept this, but we should iterate on optimizing its post-gzip size.

metsw24-max:
@gibson042 i updated as per u suggested

timmywil (approved):

mgol (approved):
Thanks!

mgol:
Landed on `main` in 35789e7687d31512072098ce0d826079904938d2. Thanks!

PR #5831, opened 2026-05-02
Propagate csp nonce

jmestwa-coder:
### Summary

Automatically propagates the page’s CSP nonce to dynamically executed scripts.

Under strict Content Security Policy (`script-src 'nonce-...'`), dynamically
inserted scripts (e.g., via `jQuery.globalEval` or `$.getScript`) may fail to
execute unless a nonce is explicitly provided. This patch ensures the nonce is
applied automatically when available.

Key points:
- Adds a small helper to read the nonce from the current execution context
- Applies nonce in `DOMEval` for inline script execution
- Applies nonce in `ajaxPrefilter("script")` only when script-tag transport is already used
- Preserves existing behavior (no transport changes, no API changes)

---

### Checklist

### DIFF
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -1,4 +1,5 @@
 import { jQuery } from "../core.js";
+import { getScriptNonce } from "../core/getScriptNonce.js";
 import { document } from "../var/document.js";
 
 import "../ajax.js";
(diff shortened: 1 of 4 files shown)


timmywil:
Thank you for your contribution. However, the first step is to open an issue with a reduced test case for verification. If this addresses an existing issue, please reference it here and we can revisit.

jmestwa-coder:
Thanks for the guidance. I’ve opened an issue with a reduced test case here: #5835 .

Happy to continue discussion there.

PR #5834, opened 2026-05-03
Prevent prototype pollution via constructor in jQuery.extend

metsw24-max:
This patch hardens the `jQuery.extend()` method against prototype pollution by explicitly blocking the `constructor` property during deep recursive merges. 

**How it works:**
In `src/core.js`, the iteration loop inside `jQuery.extend` was already filtering out the `__proto__` key to fix the well-known CVE-2019-11358. However, it left the `constructor` key unchecked. By simply adding `name === "constructor"` to the ignore list, we prevent maliciously crafted JSON payloads (e.g., `{"constructor": {"prototype": {"polluted": true}}}`) from traversing into `constructor.prototype`. 

**Security Impact:**
While the patch is a "one-liner special" and easy to write, the security impact is highly significant and proactive. Deep-merge prototype pollution is a notoriously error-prone design pattern that allows attackers to inject malicious properties into base objects. By explicitly cutting off the `constructor` traversal path, this patch provides demonstrable, secure-by-design mitigation against object injection vulnerabilities without breaking expected library functionality.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -150,7 +150,7 @@ jQuery.extend = jQuery.fn.extend = function() {
 
 				// Prevent Object.prototype pollution
 				// Prevent never-ending loop
-				if ( name === "__proto__" || target === copy ) {
+				if ( name === "__proto__" || name === "constructor" || target === copy ) {
 					continue;
 				}
 


mgol:
Thank you for your contribution. However, the first step is to open an issue with a reduced test case for verification. If this addresses an existing issue, please reference it here and we can revisit.

PR #5837, opened 2026-05-08
fix: apply CSP nonce to dynamically executed scripts via globalEval (gh-5835)

### LINKED ISSUE #5835: CSP nonce not applied to dynamically executed scripts (globalEval, getScript)
### Description

When using a strict Content Security Policy (CSP) with a nonce (e.g. `script-src 'nonce-...'`),
dynamically executed scripts via jQuery fail to execute unless a nonce is explicitly provided.

This affects common APIs such as:

* `jQuery.globalEval`
* `$.getScript`

---

### Reproducible Test Case

```html
<!DOCTYPE html>
<html>
<head>
  <meta http-equiv="Content-Security-Policy" content="script-src 'nonce-abc123'">
  <script nonce="abc123" src="https://code.jquery.com/jquery-3.7.1.js"></script>
</head>
<body>

<script nonce="abc123">
  // Attempt to load and execute script dynamically
  $.getScript("data:text/javascript,console.log('executed')");
</script>

</body>
</html>
```

---

### Current Behavior

The dynamically inserted script is blocked by the browser due to missing nonce:

> Refused to execute script because it violates the Content Security Policy

---

### Expected Behavior

Scripts dynamically inserted by jQuery should be able to execute under CSP when a valid nonce
is already present on the page, without requiring manual `scriptAttrs` configuration.

---

### Notes

* Currently, developers must manually pass `scriptAttrs: { nonce: ... }`, which is error-prone
  and not consistently applied across codebases.
* This makes it difficult to use jQuery in environments with strict CSP policies.

---

### Environment

* jQuery: 3.x
* Browser: (any modern browser enforcing CSP)

Liuwei1125:
### Problem

When using a strict CSP with a nonce (e.g. `script-src 'nonce-abc123'`), dynamically executed scripts via `$.getScript()` and `jQuery.globalEval()` fail because jQuery does not apply the nonce to the injected script element.

The issue has two parts:

1. `$.ajax({ dataType: "script" })` and `$.getScript()` use `jQuery.globalEval(code)` without passing any nonce, so `DOMEval` creates a script element without a nonce attribute.

2. `jQuery.globalEval(code, options)` itself has no fallback when `options` is `undefined` or lacks a nonce — it just calls `DOMEval(code, { nonce: undefined })`, which does not apply any nonce to the script element.

### Fix

**src/ajax/script.js** — Pass `scriptAttrs.nonce` to `globalEval` in the text script converter:

```js
jQuery.globalEval( text, { nonce: s.scriptAttrs && s.scriptAttrs.nonce } );
```

This ensures `$.getScript()` and `$.ajax({ dataType: "script" })` work under strict CSP without requiring manual `scriptAttrs` configuration.

**src/core.js** — Add a fallback in `globalEval` that reads the nonce from the first `<script nonce>` element on the page when no explicit `options.nonce` is provided:

```js
nonce: ( options && options.nonce ) ||
    document.querySelector( "script[nonce]" )?.nonce
```

This covers all callers that invoke `globalEval` without an explicit nonce (e.g. `domManip` when inserting scripts, `_evalUrl`, etc.), so they automatically pick up the page's nonce.

### Testing

Added two unit tests in `test/unit/core.js`:
- `globalEval - nonce fallback reads from page script element (gh-5835)` — verifies the fallback works
- `globalEval with explicit nonce option` — verifies explicit nonce takes precedence

### Related

- Fixes: https://github.com/jquery/jquery/issues/5835

### DIFF
--- a/src/ajax/script.js
+++ b/src/ajax/script.js
@@ -36,7 +36,7 @@ jQuery.ajaxSetup( {
 	},
 	converters: {
 		"text script": function( text ) {
-			jQuery.globalEval( text );
+			jQuery.globalEval( text, { nonce: s.scriptAttrs && s.scriptAttrs.nonce } );
 			return text;
 		}
 	}
--- a/src/core.js
+++ b/src/core.js
@@ -231,7 +231,12 @@ jQuery.extend( {
 	// Evaluates a script in a provided context; falls back to the global one
 	// if not specified.
 	globalEval: function( code, options, doc ) {
-		DOMEval( code, { nonce: options && options.nonce }, doc );
+		DOMEval( code, {
+			nonce: ( options && options.nonce ) ||
+				// Fallback: read the nonce from the first script element on the page with a nonce attribute.
+				// This is needed because the nonce may not be available through ajaxSettings (gh-5835).
+				document.querySelector( "script[nonce]" )?.nonce
+		}, doc );
 	},
 
 	each: function( obj, callback ) {
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -176,6 +176,30 @@ QUnit.test( "globalEval", function( assert ) {
 	assert.equal( window.globalEvalTest, 3, "Test context (this) is the window object" );
 } );
 
+QUnit.test( "globalEval - nonce fallback reads from page script element (gh-5835)", function( assert ) {
+	assert.expect( 1 );
+	Globals.register( "cspNonceFallbackTest" );
+
+	var script = document.createElement( "script" );
+	script.setAttribute( "nonce", "test-nonce-1234" );
+	document.head.appendChild( script );
+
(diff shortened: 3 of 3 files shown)


mgol (changes requested):
Thanks for the PR. However, I said it in https://github.com/jquery/jquery/issues/5835#issuecomment-4425636371 already, but I don't like the approach of automatically prefilling nonces from possibly unrelated scripts. I think some friction is warranted, and I'd be wary of adding such a feature when the web platform doesn't have it; this has a potential of introducing security issues.

Also, tests like that need to be iframe tests as in the original PR https://github.com/jquery/jquery/pull/5831. This PR's tests pass even without any source changes. When adding tests showing a fix works, please always verify that without source changes the tests fails!

PR #5839, opened 2026-05-11
Core: Refactor loops to avoid going behind array bounds

mgol:
### Summary ###

Reaching behind array bounds - which is what effectively happens if you check for `null`/`undefined` value instead of doing a `length` check - can trigger deopts in various JS engines. Switch to regular `length`-based loops.

In the past, this was impractical as IE <9 used to clobber the `length` property if an element with the `name="length"` attribute was present, but this is no longer a concern.

### Checklist ###

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -1171,7 +1171,6 @@ function matcherFromGroupMatchers( elementMatchers, setMatchers ) {
 			// Add elements passing elementMatchers directly to results
 			for ( ; ( elem = elems[ i ] ) != null; i++ ) {
 				if ( byElement && elem ) {
-					j = 0;
 
 					// Support: IE 11+
 					// IE sometimes throws a "Permission denied" error when strict-comparing
@@ -1181,8 +1180,10 @@ function matcherFromGroupMatchers( elementMatchers, setMatchers ) {
 						setDocument( elem );
 						xml = !documentIsHTML;
 					}
-					while ( ( matcher = elementMatchers[ j++ ] ) ) {
-						if ( matcher( elem, context || document, xml ) ) {
+					matcher = undefined;
+					for ( j = 0; j < elementMatchers.length; j++ ) {
+						if ( elementMatchers[ j ]( elem, context || document, xml ) ) {
+							matcher = elementMatchers[ j ];
 							push.call( results, elem );
 							break;
 						}
@@ -1219,8 +1220,8 @@ function matcherFromGroupMatchers( elementMatchers, setMatchers ) {
 			// case, which will result in a "00" `matchedCount` that differs from `i` but is also
 			// numerically zero.
 			if ( bySet && i !== matchedCount ) {
-				j = 0;
-				while ( ( matcher = setMatchers[ j++ ] ) ) {
+				for ( j = 0; j < setMatchers.length; j++ ) {
+					matcher = setMatchers[ j ];
 					matcher( unmatched, setMatched, context, xml );
 				}
 
--- a/src/selector/uniqueSort.js
+++ b/src/selector/uniqueSort.js
@@ -70,8 +70,8 @@ function sortOrder( a, b ) {
  * @param {ArrayLike} results
  */
 jQuery.uniqueSort = function( results ) {
-	var j = 1,
-		i = 1;
+	var i,
+		j = 1;
 
 	hasDuplicate = false;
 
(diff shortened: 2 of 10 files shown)


mgol:
Current size diffs:
```
main @35789e7687d31512072098ce0d826079904938d2
   raw     gz     br Filename
  +146    +38    +35 dist/jquery.min.js
  +133    +35    +47 dist/jquery.slim.min.js
  +146    +41    +54 dist-module/jquery.module.min.js
  +133    +33    +27 dist-module/jquery.slim.module.min.js
```

src/selector/uniqueSort.js
```diff
@@ -82,7 +82,7 @@ jQuery.uniqueSort = function( results ) {
 		// Pack the first instance of each unique element into the start of
 		// results (starting at index 1 because index 0 is always kept), then
 		// splice away the tail of duplicates.
-		for ( ; i < results.length; i++ ) {
+		for ( i = 1; i < results.length; i++ ) {
```
mgol, line 85: @gibson042 changes in this file resulted in the following size diffs compared to if I didn't touch the file:
```
last run
   raw     gz     br Filename
    +1     -7    +17 dist/jquery.min.js
    +1     -6    -18 dist/jquery.slim.min.js
    +1     -6    +35 dist-module/jquery.module.min.js
    +1     -6    +31 dist-module/jquery.slim.module.min.js
```

mgol: It's possible it'd go the other way round if we avoided initializers in loops and moved them to declarations, but it's not always possible, as sometimes the same variable is used in multiple loops.

timmywil (approved):

src/selector.js
```diff
@@ -1171,7 +1171,6 @@ function matcherFromGroupMatchers( elementMatchers, setMatchers ) {
 			// Add elements passing elementMatchers directly to results
 			for ( ; ( elem = elems[ i ] ) != null; i++ ) {
```
gibson042, line 1172: Can this loop not be refactored in the same way?

gibson042 (approved):
Just one comment, and potential size reductions at https://github.com/mgol/jquery/pull/18 .

src/selector.js
mgol: No tests fail with such a refactor. It changes behavior for collections including `null` or `undefined`, throwing in more cases, but we're already changing behavior in other places, so perhaps it's fine? These are unsupported inputs anyway.

For example. given `$elems = $( [ div1, div2, null, div3 ] )`, assuming each `divN` has a class `clazzN`:

| Expression                              | `main`                    | PR 5839                 |
| --------------------------------------- | ------------------------- |------------------------ |
| `$.find.matches( "div", $elems.get() )` | `[ div1, div2 ]`          | `[ div1, div2, div3 ]`  |
| `$elems.hasClass( "clazz1" )`           | `true`                    | `true`                  |
| `$elems.hasClass( "clazz2" )`           | `true`                    | `true`                  |
| `$elems.hasClass( "clazz3" )`           | `false`                   | Throws a `TypeError`    |
| `$elems.text()`                         | Text of `div1` and `div2` | Throws a `TypeError`    |

`main` doesn't universally stop on `null` either, e.g. e.g. `$elems.addClass( "foo" )` or `$elems.parent()` throw.

mgol:
I applied most changes from @gibson042's PR https://github.com/mgol/jquery/pull/18 with my tweaks that pushed it even further; that changes it from the original increases:
```
main @35789e7687d31512072098ce0d826079904938d2
   raw     gz     br Filename
  +146    +38    +35 dist/jquery.min.js
  +133    +35    +47 dist/jquery.slim.min.js
  +146    +41    +54 dist-module/jquery.module.min.js
  +133    +33    +27 dist-module/jquery.slim.module.min.js
```
to decreases:
```
main @32bf84bcbb82c6d24d18107b85a534fd8bb0d6f1
   raw     gz     br Filename
  +149    -15    -58 dist/jquery.min.js
  +126    -19    -21 dist/jquery.slim.min.js
  +149    -14     +9 dist-module/jquery.module.min.js
  +126    -21    +12 dist-module/jquery.slim.module.min.js
```
That changed a lot! 🎉

src/selector.js
gibson042: Yeah, I'd like to fully abandon the `( elem = elems[ i ] ) != null` pattern, and none of those behaviors seem important to preserve.

gibson042 (approved):

PR #5843, opened 2026-05-28
Selector: Unescape class name in filter.CLASS

rootvector2:
Noticed `filter.CLASS` interpolates the raw token text into the cached
regex, while `filter.ID` and `filter.TAG` both run the captured identifier
through `unescapeSelector` first. So when `.foo\41 bar` reaches the
seed-based filter path (e.g. `.filter()` on a multi-element set), it
compares against the literal text `foo\41 bar`, plus `\41` becomes a
JS-regex octal/backreference. Unescape the captured class first and
switch to indexOf, matching the pattern already used by ATTR `~=`.

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -39,7 +39,6 @@ var i,
 	// Instance-specific data
 	dirruns = 0,
 	done = 0,
-	classCache = createCache(),
 	compilerCache = createCache(),
 	nonnativeSelectorCache = createCache(),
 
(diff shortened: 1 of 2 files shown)


src/selector.js
```diff
@@ -453,19 +453,20 @@ jQuery.expr = {
...
+				return pattern;
+			}
+			return classCache( unescapedClass, function( elem ) {
+				var classAttr = typeof elem.className === "string" && elem.className ||
+					typeof elem.getAttribute !== "undefined" &&
+						elem.getAttribute( "class" ) ||
+					"";
+				return ( " " + classAttr.replace( rwhitespace, " " ) + " " )
+					.indexOf( " " + unescapedClass + " " ) > -1;
+			} );
```
gibson042: Since we're no longer constructing a RegExp instance, I don't think `classCache` carries its own weight. Let's drop it.
```suggestion
			var unescapedClass = " " + unescapeSelector( className ) + " ";
			return function( elem ) {
			    var classes = getClass( elem );
			    return ( " " + classes.replace( rwhitespace, " " ) + " " )
			        .indexOf( unescapedClass ) > -1;
			};
```

gibson042 (approved):

src/selector.js
rootvector2: done. dropped `classCache` and its now-unused declaration, and the matcher returns directly. left the class read inline rather than calling `getClass` since that lives in `attributes/classes.js` and importing it here would pull the class-manipulation methods into the selector module. can switch to it if you'd prefer.

rootvector2:
good to merge?

timmywil (approved):

PR #5844, opened 2026-05-29
Core: Ignore inherited properties in jQuery( html, props )

rootvector2:
The `$( html, props )` branch in src/core/init.js walks `context` with a bare for-in, so enumerable keys inherited from a polluted Object.prototype get applied as element attributes or method calls. Guard with hasOwn so only own properties are used, matching the prototype check already in jQuery.extend.

### DIFF
--- a/src/core/init.js
+++ b/src/core/init.js
@@ -1,6 +1,7 @@
 // Initialize a jQuery object
 import { jQuery } from "../core.js";
 import { document } from "../var/document.js";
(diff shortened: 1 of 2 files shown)


mgol (changes requested):
Thanks for the PR, but we explicitly don't support environments with a modified `Object.prototype`: https://github.com/jquery/jquery/wiki/Won't-Fix#modifying-objectprototype

PR #5845, opened 2026-05-29
Selector: Decode invalid escape code points to U+FFFD

rootvector2:
`unescapeSelector` decodes CSS hex escapes that the spec maps to U+FFFD wrong:

    \0       -> U+0000   (literal NUL)
    \D800    -> U+D800   (lone surrogate)
    \110000  -> garbage  (above the Unicode max)

Return U+FFFD for null, surrogate, and out-of-range escapes. Reachable via the seeded `.filter()` matcher; test added alongside the others.

### DIFF
--- a/src/selector/unescapeSelector.js
+++ b/src/selector/unescapeSelector.js
@@ -5,7 +5,7 @@ import { whitespace } from "../var/whitespace.js";
 var runescape = new RegExp( "\\\\[\\da-fA-F]{1,6}" + whitespace +
 	"?|\\\\([^\\r\\n\\f])", "g" ),
 	funescape = function( escape, nonHex ) {
-		var high = "0x" + escape.slice( 1 ) - 0x10000;
+		var codePoint = "0x" + escape.slice( 1 ) - 0;
 
 		if ( nonHex ) {
 
@@ -13,13 +13,24 @@ var runescape = new RegExp( "\\\\[\\da-fA-F]{1,6}" + whitespace +
 			return nonHex;
 		}
 
+		// Per the CSS spec, a NULL, surrogate, or out-of-range code point is
+		// replaced with the REPLACEMENT CHARACTER (U+FFFD).
+		// https://www.w3.org/TR/css-syntax-3/#consume-escaped-code-point
+		if ( !codePoint || codePoint > 0x10FFFF ||
+			( codePoint >= 0xD800 && codePoint < 0xE000 ) ) {
+			return "\uFFFD";
+		}
+
 		// Replace a hexadecimal escape sequence with the encoded Unicode code point
 		// Support: IE <=11+
 		// For values outside the Basic Multilingual Plane (BMP), manually construct a
 		// surrogate pair
-		return high < 0 ?
-			String.fromCharCode( high + 0x10000 ) :
-			String.fromCharCode( high >> 10 | 0xD800, high & 0x3FF | 0xDC00 );
+		return codePoint < 0x10000 ?
+			String.fromCharCode( codePoint ) :
+			String.fromCharCode(
+				( codePoint - 0x10000 ) >> 10 | 0xD800,
+				( codePoint - 0x10000 ) & 0x3FF | 0xDC00
+			);
 	};
 
 export function unescapeSelector( sel ) {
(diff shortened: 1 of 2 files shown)


timmywil:
The workflow failures are unrelated. Chrome has been updated in CI and the failures we saw in beta related to `:enabled` and `:disabled` pseudos  have made it to Chrome stable.

src/selector/unescapeSelector.js
```diff
@@ -5,21 +5,31 @@ import { whitespace } from "../var/whitespace.js";
...
 			// Strip the backslash prefix from a non-hex escape sequence
 			return nonHex;
 		}
 
+		// Per the CSS spec, a NULL, surrogate, or out-of-range code point is
+		// replaced with the REPLACEMENT CHARACTER (U+FFFD).
+		if ( codePoint === 0 || codePoint > 0x10FFFF ||
+			( codePoint >= 0xD800 && codePoint <= 0xDFFF ) ) {
+			return "\uFFFD";
+		}
```
gibson042: gzippability improvements:
```suggestion
		var codePoint = "0x" + escape.slice( 1 ) - 0;

		if ( nonHex ) {

			// Strip the backslash prefix from a non-hex escape sequence
			return nonHex;
		}

		// Per the CSS spec, a NULL, surrogate, or out-of-range code point is
		// replaced with the REPLACEMENT CHARACTER (U+FFFD).
		// https://www.w3.org/TR/css-syntax-3/#consume-escaped-code-point
		if ( !codePoint || codePoint > 0x10FFFF ||
			( codePoint >= 0xD800 && codePoint < 0xE000 ) ) {
			return "\uFFFD";
		}
```

src/selector/unescapeSelector.js
```diff
@@ -5,21 +5,31 @@ import { whitespace } from "../var/whitespace.js";
...
 		// surrogate pair
-		return high < 0 ?
-			String.fromCharCode( high + 0x10000 ) :
-			String.fromCharCode( high >> 10 | 0xD800, high & 0x3FF | 0xDC00 );
+		return codePoint > 0xFFFF ?
+			String.fromCharCode(
+				( codePoint - 0x10000 ) >> 10 | 0xD800,
+				( codePoint - 0x10000 ) & 0x3FF | 0xDC00
+			) :
+			String.fromCharCode( codePoint );
```
gibson042: gzippability improvements:
```suggestion
		return codePoint < 0x10000 ?
			String.fromCharCode( codePoint ) :
			String.fromCharCode(
				( codePoint - 0x10000 ) >> 10 | 0xD800,
				( codePoint - 0x10000 ) & 0x3FF | 0xDC00
			);
```

test/unit/selector.js
```diff
@@ -844,6 +844,26 @@ QUnit.test( "attributes - special characters", function( assert ) {
 		"Long numeric escape (non-BMP)" );
 } );
 
+QUnit.test( "attributes - invalid escaped code points", function( assert ) {
```
gibson042, line 847: Let's also include a test case with complete and off-by-one coverage, e.g. that `[data-attr='\0 \1 \D7FF \D800 \DFFF \E000 \10FFFF \110000']` matches an element with attribute value `"\uFFFD\u0001\uD7FF\uFFFD\uFFFD\uE000\uDBFF\uDFFF\uFFFD"`.

gibson042 (approved):
Thanks! This should be good to go after a few small tweaks.

src/selector/unescapeSelector.js
rootvector2: applied. went with the `"0x" + escape.slice( 1 ) - 0` form, `!codePoint`, and `< 0xE000`, and added the spec link.

src/selector/unescapeSelector.js
rootvector2: done, BMP branch first now.

test/unit/selector.js
rootvector2: added it as a fifth assertion: seeded an element with value `�퟿��􏿿�` and matched it with the full `\0 \1 \D7FF \D800 \DFFF \E000 \10FFFF \110000` list, so both sides of each boundary are covered.

timmywil:
Please rebase when you get a chance.

rootvector2:
rebased onto main, clean. should be good now.

rootvector2:
any update?

PR #5861, opened 2026-07-05
Selector: Decode invalid CSS escapes to the replacement character

madib06ops:
1. `funescape` decoded a zero, a surrogate half (U+D800-U+DFFF), or a value above U+10FFFF into a literal NUL, a lone surrogate, or a corrupt surrogate pair.
2. jQuery's own matcher then decoded such an escape differently from the browser's native CSS parser, so the same selector could match different elements on the engine path (`.filter()`/`.is()` over a multi-element set) than through `querySelectorAll`.

Return U+FFFD for those code points per the CSS Syntax rules; valid BMP and astral escapes are unchanged. Added a regression test next to the existing escape cases.


madib06ops:
any update?

gibson042:
Thank you for the contribution, but this was already handled by #5845.

PR #5863, opened 2026-07-08
Selector: unescape class names in the CLASS filter matcher

madib06ops:
expr.filter.CLASS in src/selector.js splices the raw, still-escaped class-selector text straight into a RegExp and never runs it through unescapeSelector, unlike the sibling ID, TAG, and ATTR matchers which all decode their argument first. A hexadecimal CSS escape such as `.\66oo` (the class `foo`, since 0x66 is `f`) is then read as a regex octal escape, so the jQuery matcher selects a different class than native querySelectorAll. This path is reached via .filter()/.is()/.not() over multi-element sets, where the compiled matcher runs instead of qSA. The fix decodes the name with unescapeSelector and compares whitespace-delimited tokens directly, so escaped class names resolve the same way on both paths.

### DIFF
--- a/src/selector.js
+++ b/src/selector.js
@@ -455,16 +455,20 @@ jQuery.expr = {
 		CLASS: function( className ) {
 			var pattern = classCache[ className + " " ];
 
+			// Decode the class name and match whitespace-delimited tokens directly.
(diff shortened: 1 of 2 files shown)


mgol:
Thanks, but this looks like a duplicate of PR #5843.

PR #5864, opened 2026-07-14
Ajax: Trim trailing whitespace from response header values

madib06ops:
jqXHR.getResponseHeader parses the raw header block with rheaders, which strips leading whitespace after the colon but leaves trailing spaces and tabs on the value, so `Content-Type: text/html ` comes back with the trailing space while native getResponseHeader returns it trimmed per RFC 7230. A custom transport or the object-based completeCallback passing an untrimmed header block surfaces the divergence. Trim trailing tabs and spaces in the value capture so the parse matches native XHR.


madib06ops:
any update?

PR #5867, opened 2026-07-16
CSS: Read custom properties from detached elements

### LINKED ISSUE #5865: CSS: .css() returns undefined for custom properties on detached elements
### Description

Calling `.css()` to read a CSS custom property from a detached element returns `undefined`, even when jQuery just set that inline property.

For a detached element in Chrome, `getComputedStyle( elem ).getPropertyValue( "--foo" )` returns an empty string and `computed[ "--foo" ]` is `undefined`. That leaves `ret` as `undefined`, so the current `ret === ""` detached fallback is skipped. The fallback delegates to `jQuery.style()`, whose getter also reads `style[ name ]`; custom properties require `getPropertyValue()`.

Minimal reproduction against current `main` (`7e1efd77`):

```js
const elem = jQuery( "<div>" ).css( "--foo", "5" );

console.log( elem[ 0 ].style.getPropertyValue( "--foo" ) ); // "5"
console.log( elem.css( "--foo" ) );                         // undefined
```

Expected: `elem.css( "--foo" )` returns `"5"`, matching attached elements and the value stored in the inline style.

Actual: it returns `undefined`.

Tested in Chrome 150 on macOS. The existing custom-property tests only exercise attached elements.

### Link to test case

The reduced code above reproduces after loading the [latest development build](https://releases.jquery.com/git/jquery-git.js). It was also verified against a local build of `main` at `7e1efd77`.

### Screenshot

Detached custom property returns undefined before the fix

Kevinjohn:
Fixes gh-5865

### Summary

* Read inline CSS custom properties through `getPropertyValue()`.
* Use the existing inline-style fallback when a detached custom property's computed value is `undefined`.
* Add regression coverage for reading a custom property from a detached element.

The setter stores custom properties with `setProperty()`, but the getter used bracket access. In Chrome, the computed-style path also yields `undefined` for a detached custom property, so the existing empty-string fallback was skipped.

### Screenshots

Before — the inline value is `"5"`, but `.css()` returns `undefined`:

Detached custom property returns undefined before the fix

After — the same detached element returns `"5"`:

Detached custom property returns the inline value after the fix

### Validation

* `npm run test`
* Focused Chrome CSS regression: 61 passed, 0 skipped

### Checklist

* [x] A docs issue/PR is not needed; this restores consistent existing `.css()` behavior

### DIFF
--- a/src/css.js
+++ b/src/css.js
@@ -259,7 +259,7 @@ jQuery.extend( {
 			}
 
 			// Otherwise just get the value from the style object
-			return style[ name ];
+			return isCustomProp ? style.getPropertyValue( name ) : style[ name ];
 		}
 	},
 
--- a/src/css/curCSS.js
+++ b/src/css/curCSS.js
@@ -48,7 +48,7 @@ export function curCSS( elem, name, computed ) {
 			ret = ret.replace( rtrimCSS, "$1" ) || undefined;
 		}
 
-		if ( ret === "" && !isAttached( elem ) ) {
+		if ( ( ret === "" || ( isCustomProp && ret === undefined ) ) && !isAttached( elem ) ) {
 			ret = jQuery.style( elem, name );
 		}
 	}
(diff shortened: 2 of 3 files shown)


timmywil:

PR #5869, opened 2026-07-17
Event: Restore trigger state after native method failure

### LINKED ISSUE #5868: Event: Restore trigger state when a native method throws
When a same-named native method throws during `.trigger()`, jQuery can silently stop handling later events of that type.

### At a glance — visual summary

**Before — the later native event is suppressed**

Visual summary of the failure: expected two handler calls, but only one occurs

**After — temporary event state is restored**

Visual summary of the intended result: both handler calls occur

### Actual Chrome screenshots

**Failing — upstream `main` at `7e1efd77`**

Only this regression test is added to the upstream source. QUnit reports one failed test and two failed assertions, including **Expected: 2, Result: 1**.

Actual Chrome QUnit screenshot failing on upstream main

**Passing — PR commit `21f17ad9`**

The identical test reports **3 assertions of 3 passed, 0 failed**.

Actual Chrome QUnit screenshot passing after the fix

### Expected

- The native exception is propagated.
- jQuery restores its temporary event state.
- Later native events still reach jQuery handlers.
- The inline `on<type>` handler is restored.

### Actual

- `jQuery.event.triggered` remains set to the event type.
- Later native events of that type are discarded by jQuery's event handle guard.
- The inline `on<type>` handler remains cleared.

### Reduced reproduction

```js
const elem = document.createElement( "div" );
let handlerCalls = 0;

elem.boom = function() {
	throw new Error( "native method failed" );
};

jQuery( elem ).on( "boom", function() {
	handlerCalls++;
} );

try {
	jQuery( elem ).trigger( "boom" );
} catch ( error ) {
	console.log( error.message );
}

elem.dispatchEvent( new Event( "boom" ) );
console.log( handlerCalls ); // 1; expected 2
```

This reproduces in jQuery 4.0.0 and the latest `main` build. The example uses a plain DOM element, so it does not depend on a browser-specific native method.

### Cause

`trigger()` sets its event guard and temporarily clears the inline handler before calling `elem[ type ]()`. The cleanup runs only after that call returns, so a thrown exception skips it.

Kevinjohn:
When a same-named native method throws during `.trigger()`, jQuery can silently stop handling later events of that type.

### Summary ###

This restores jQuery's temporary event state with a `finally` block, while preserving the native error for the caller. 

Basically I wrapped it in a try/finally.

Fixes #5868.

### At a glance — visual summary ###

**Before — the later native event is suppressed**

Visual summary of the failure: expected two handler calls, but only one occurs

**After — temporary event state is restored**

Visual summary of the intended result: both handler calls occur

### Actual Chrome screenshots ###

**Failing — upstream `main` at `7e1efd77`**

Only the regression test is added to the upstream source. QUnit reports one failed test and two failed assertions, including **Expected: 2, Result: 1**.

Actual Chrome QUnit screenshot failing on upstream main

**Passing — this PR at `21f17ad9`**

The identical test reports **3 assertions of 3 passed, 0 failed**.

Actual Chrome QUnit screenshot passing after the fix

### Validation ###

* [x] Added a regression test for state and inline-handler restoration
* [x] `npm test`
* [x] Focused event suite in headless Chrome and Firefox

### Checklist ###

* [x] No documentation changes are needed

### DIFF
--- a/src/event/trigger.js
+++ b/src/event/trigger.js
@@ -137,20 +137,22 @@ jQuery.extend( jQuery.event, {
 					// Prevent re-triggering of the same event, since we already bubbled it above
 					jQuery.event.triggered = type;
 
-					if ( event.isPropagationStopped() ) {
-						lastElement.addEventListener( type, stopPropagationCallback );
-					}
+					try {
+						if ( event.isPropagationStopped() ) {
+							lastElement.addEventListener( type, stopPropagationCallback );
+						}
 
-					elem[ type ]();
+						elem[ type ]();
+					} finally {
+						if ( event.isPropagationStopped() ) {
+							lastElement.removeEventListener( type, stopPropagationCallback );
+						}
 
-					if ( event.isPropagationStopped() ) {
-						lastElement.removeEventListener( type, stopPropagationCallback );
-					}
+						jQuery.event.triggered = undefined;
 
-					jQuery.event.triggered = undefined;
-
-					if ( tmp ) {
-						elem[ ontype ] = tmp;
+						if ( tmp ) {
+							elem[ ontype ] = tmp;
+						}
 					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


timmywil:

PR #5871, opened 2026-07-17
Deferred: Warn on AggregateError exceptions

### LINKED ISSUE #5870: Deferred: AggregateError exceptions do not trigger a warning
AggregateError exceptions thrown from Deferred callbacks are rejected without the development warning emitted for other native error types.

- **Sequence:** Throw an `AggregateError` from a Deferred `.then()` callback.
- **Expected:** `jQuery.Deferred.exceptionHook` calls `console.warn`, matching `TypeError` and the other native `Error` subclasses it handles.
- **Actual:** The Deferred rejects, but no warning is emitted, so the exception is easier to miss.

### Reproduction

Tested against `jquery/jquery@7e1efd77` in Chrome 150.0.7871.125 on macOS.

```js
const previousWarn = console.warn;
let warned = false;

console.warn = function() {
    warned = true;
};

jQuery.Deferred().resolve().then(function() {
    throw new AggregateError([ new Error( "first" ) ], "all failed" );
}).catch(function() {
    console.log( warned ); // false
    console.warn = previousWarn;
});
```

`jQuery.Deferred.exceptionHook` filters by `error.name`. Its native-error allowlist predates `AggregateError` and does not include `"AggregateError"`.

### Proposed fix

Add `AggregateError` to the existing allowlist and cover it with a focused regression test.

### Actual screenshots

Both captures use the same filtered QUnit test and Chrome version.

**Before — `7e1efd77` plus the regression test only**

Image

**After — same revision and test, with `AggregateError` added to the allowlist**

Image

Kevinjohn:
AggregateError exceptions thrown from Deferred callbacks now produce the same development warning as other native `Error` subclasses.

- **Sequence:** Throw an `AggregateError` from a Deferred `.then()` callback.
- **Expected:** `jQuery.Deferred.exceptionHook` calls `console.warn`.
- **Actual before this change:** The Deferred rejects without emitting a warning.

Fixes gh-5870

### Summary

- Add `AggregateError` to the existing native-error name allowlist.
- Add a focused regression test for the warning.

### Actual screenshots

Both captures use Chrome 150.0.7871.125 and the same filtered QUnit test.

**Before — `7e1efd77` plus the regression test only (Expected: `true`, Result: `undefined`)**

QUnit AggregateError test failing before the fix

**After — `c8041ca6` (1 assertion passed, 0 failed)**

QUnit AggregateError test passing after the fix

### Validation

- `npm run test:chrome -- --flag module=deferred` — 29 passed, 0 skipped
- `npm test` — full suite passed

### Checklist

* [x] No API documentation change is needed

### DIFF
--- a/src/deferred/exceptionHook.js
+++ b/src/deferred/exceptionHook.js
@@ -4,7 +4,7 @@ import "../deferred.js";
 
 // These usually indicate a programmer mistake during development,
 // warn about them ASAP rather than swallowing them by default.
-var rerrorNames = /^(Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
+var rerrorNames = /^(Aggregate|Eval|Internal|Range|Reference|Syntax|Type|URI)Error$/;
 
 // If `jQuery.Deferred.getErrorHook` is defined, `asyncError` is an error
 // captured before the async barrier to get the original error cause
--- a/test/unit/deferred.js
+++ b/test/unit/deferred.js
@@ -614,6 +614,26 @@ QUnit.test( "jQuery.Deferred.exceptionHook", function( assert ) {
 	defer.resolve();
 } );
 
+QUnit.test( "jQuery.Deferred.exceptionHook warns on AggregateError", function( assert ) {
+
+	assert.expect( 1 );
+
+	var warned,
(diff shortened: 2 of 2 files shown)


timmywil:

PR #5875, opened 2026-07-17
Callbacks: Restore firing state after exceptions

### LINKED ISSUE #5874: Callbacks: Restore firing state after a callback throws
A `jQuery.Callbacks("memory")` list stops invoking callbacks added after an earlier callback throws.

- Sequence: Create a memory callback list, add a throwing callback, call `.fire("remembered")`, catch the propagated error, then add another callback.
- Expected: The exception propagates and the later callback immediately receives `"remembered"`.
- Actual: The exception propagates, but the later callback is never invoked.

### Reduced reproduction

```js
const callbacks = jQuery.Callbacks( "memory" );
let result = "not called";

callbacks.add( function() {
	throw new Error( "callback error" );
} );

try {
	callbacks.fire( "remembered" );
} catch ( error ) {}

callbacks.add( function( value ) {
	result = value;
} );

console.log( result ); // "not called"; expected "remembered"
```

Tested against `jquery/jquery@7e1efd77` in Chrome 150.0.0.0 on macOS.

`fire()` sets the internal `firing` flag before invoking callbacks. If one throws, the cleanup is skipped and the list remains inert. The exception itself should continue to propagate.

Related context: #1823 reported the same cleanup problem through ready handlers. Ready was later changed to execute handlers asynchronously, but direct `jQuery.Callbacks` usage still leaves the callback list in the firing state.

A focused fix can move the existing state cleanup into a `finally` block without changing normal callback behavior.

### Actual screenshots

**Before — upstream `7e1efd77` with the regression test only**

QUnit failure before callback cleanup fix

**After — focused patch applied**

QUnit pass after callback cleanup fix

Kevinjohn:
A `jQuery.Callbacks("memory")` list now remains usable after a callback throws.

- Sequence: Fire a memory callback list containing a throwing callback, catch the propagated error, then add another callback.
- Expected: The later callback immediately receives the memorized value.
- Actual before this change: The later callback is never invoked because the list remains in the firing state.

Fixes gh-5874

### Summary

- Move the existing callback-list cleanup into a `finally` block.
- Preserve exception propagation and the memorized value.
- Add focused regression coverage for adding a callback after an exception.

### Actual screenshots

**Before — upstream `7e1efd77` with the regression test only**

QUnit failure before callback cleanup fix

**After — commit `92884236`**

QUnit pass after callback cleanup fix

### Validation

- `npm run test:unit -- -b chrome --headless -f module=callbacks` — 52 passed
- `npm test` — full suite passed, including Chrome (1,189 passed) and Firefox (1,190 passed)

### Checklist

* [x] No documentation changes are needed for this internal state repair

### DIFF
--- a/src/callbacks.js
+++ b/src/callbacks.js
@@ -71,38 +71,42 @@ jQuery.Callbacks = function( options ) {
 			// Execute callbacks for all pending executions,
 			// respecting firingIndex overrides and runtime changes
 			fired = firing = true;
-			for ( ; queue.length; firingIndex = -1 ) {
-				memory = queue.shift();
-				while ( ++firingIndex < list.length ) {
+			try {
+				for ( ; queue.length; firingIndex = -1 ) {
+					memory = queue.shift();
+					while ( ++firingIndex < list.length ) {
 
-					// Run callback and check for early termination
-					if ( list[ firingIndex ].apply( memory[ 0 ], memory[ 1 ] ) === false &&
-						options.stopOnFalse ) {
+						// Run callback and check for early termination
+						if ( list[ firingIndex ].apply( memory[ 0 ], memory[ 1 ] ) === false &&
+							options.stopOnFalse ) {
 
-						// Jump to end and forget the data so .add doesn't re-fire
-						firingIndex = list.length;
-						memory = false;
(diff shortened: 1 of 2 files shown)


timmywil:

PR #5877, opened 2026-07-17
CSS: Restore temporary styles after measurement errors

### LINKED ISSUE #5876: CSS: Restore temporary styles when dimension measurement throws
Measuring a hidden element can permanently change its inline styles when a CSS hook throws.

- Sequence: Create an element with `display: none`, `position: relative`, and `visibility: visible`; install a throwing `boxSizing` hook; call `.width()`; then catch the propagated error.
- Expected: The error propagates and the inline styles remain `none / relative / visible`.
- Actual: The error propagates, but the inline styles become `block / absolute / hidden`.

### Reduced reproduction

```js
const elem = jQuery( "<div>" )
	.css( {
		display: "none",
		position: "relative",
		visibility: "visible"
	} )
	.appendTo( document.body );
const oldHook = jQuery.cssHooks.boxSizing;

jQuery.cssHooks.boxSizing = {
	get: function() {
		throw new Error( "measurement error" );
	}
};

try {
	elem.width();
} catch ( error ) {}

if ( oldHook ) {
	jQuery.cssHooks.boxSizing = oldHook;
} else {
	delete jQuery.cssHooks.boxSizing;
}

console.log(
	elem[ 0 ].style.display,
	elem[ 0 ].style.position,
	elem[ 0 ].style.visibility
);
// "block", "absolute", "hidden"
// Expected: "none", "relative", "visible"
```

Tested against `jquery/jquery@7e1efd77` in Chrome 150.0.0.0 on macOS.

The internal CSS `swap` helper applies temporary measurement styles and restores them only after its callback returns. If dimension calculation throws, restoration is skipped. A `try/finally` can preserve exception propagation while guaranteeing cleanup.

Related context: #5628 discusses the hidden-dimension swap itself, but not restoration when measurement throws.

### Actual screenshots

**Before — upstream `7e1efd77` with the regression test only**

QUnit failure showing temporary CSS styles are not restored

**After — focused patch applied**

QUnit pass after temporary CSS style restoration fix

Kevinjohn:
Hidden-element dimension measurement now restores jQuery's temporary inline styles when a CSS hook throws.

- Sequence: Measure a hidden element through `.width()` while a `boxSizing` hook throws.
- Expected: The error propagates and `display / position / visibility` remain `none / relative / visible`.
- Actual before this change: The error propagates, but those styles remain `block / absolute / hidden`.

Fixes gh-5876

### Summary

- Restore temporary measurement styles from a `finally` block.
- Preserve callback return values and exception propagation.
- Add focused regression coverage through the public `.width()` API.

### Actual screenshots

**Before — upstream `7e1efd77` with the regression test only**

QUnit failure showing temporary CSS styles are not restored

**After — commit `8f485fe0`**

QUnit pass after temporary CSS style restoration fix

### Validation

- `npm run test:unit -- -b chrome --headless -f module=dimensions` — 31 passed
- `npm test` — full suite passed, including Chrome (1,189 passed) and Firefox (1,190 passed)

### Checklist

* [x] No documentation changes are needed for this internal cleanup

### DIFF
--- a/src/css/var/swap.js
+++ b/src/css/var/swap.js
@@ -1,6 +1,6 @@
 // A method for quickly swapping in/out CSS properties to get correct calculations.
 export function swap( elem, options, callback ) {
-	var ret, name,
+	var name,
 		old = {};
 
 	// Remember the old values, and insert the new ones
@@ -9,12 +9,13 @@ export function swap( elem, options, callback ) {
 		elem.style[ name ] = options[ name ];
 	}
 
-	ret = callback.call( elem );
+	try {
+		return callback.call( elem );
+	} finally {
 
-	// Revert the old values
-	for ( name in options ) {
-		elem.style[ name ] = old[ name ];
+		// Revert the old values
+		for ( name in options ) {
+			elem.style[ name ] = old[ name ];
+		}
 	}
-
-	return ret;
 }
--- a/test/unit/dimensions.js
+++ b/test/unit/dimensions.js
@@ -385,6 +385,42 @@ QUnit.test( "hidden element with implicit content-based dimensions", function( a
 	assert.strictEqual( div.height(), 40, "height of a hidden element" );
 } );
 
+QUnit.test(
+	"temporary styles are restored when dimension measurement throws (gh-5876)", function( assert ) {
+	assert.expect( 4 );
+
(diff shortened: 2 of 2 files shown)


timmywil:

PR #5882, opened 2026-07-17
Manipulation: Restore script types after insertion errors

### LINKED ISSUE #5881: Manipulation: DOM insertion errors leave script types masked
### Description ###

A failed DOM insertion leaves caller-owned `<script>` elements with jQuery's temporary `type` prefix.

- Sequence: pass a script with `type="application/json"` to `.before()` where the browser rejects the insertion with `HierarchyRequestError`.
- Expected: the native exception propagates and the script keeps `type="application/json"`.
- Actual: the exception propagates, but the script is left as `type="true/application/json"`.
- A script without a `type` is similarly left as `type="false/"`.
- Observed in Chrome 150. The failure depends on a browser-thrown DOM insertion exception; the skipped jQuery cleanup is not browser-specific.

`domManip` masks script types before invoking the insertion callback and restores them only after the callback returns. If the callback throws, restoration is skipped. The affected sequence is in [`domManip.js`](https://github.com/jquery/jquery/blob/7e1efd77598a527c67d2a674a1259787f7e441fd/src/manipulation/domManip.js#L54-L79).

```js
const script = document.createElement( "script" );
script.setAttribute( "type", "application/json" );

try {
	jQuery( document.documentElement ).before( script );
} catch ( error ) {
	console.log( error.name ); // HierarchyRequestError
}

console.log( script.getAttribute( "type" ) );
// Expected: "application/json"
// Result:   "true/application/json"
```

### Actual screenshots

**Before — upstream `main` at `7e1efd77598a527c67d2a674a1259787f7e441fd`, QUnit 2.26.0, Chrome 150, source modules enabled.** The native-error assertion passes; the cleanup assertion fails with `Expected: "application/json"` and `Result: "true/application/json"`.

Failing QUnit test showing the script type left as true/application/json

**After — the same revision and browser with a local proof patch that restores masked script types in a `finally` block.** The focused regression test passes both assertions.

Passing QUnit test after restoring script types in finally

The proof patch also passes the complete manipulation module:

```text
152 passed. 0 skipped.
All tests passed!
```

### Link to test case ###

The focused regression test used for both screenshots:

```js
QUnit.test( "script type is restored when insertion throws", function( assert ) {
	assert.expect( 2 );

	var script = document.createElement( "script" );

	script.setAttribute( "type", "application/json" );

	assert.throws( function() {
		jQuery( document.documentElement ).before( script );
	}, "Invalid insertion throws" );

	assert.strictEqual(
		script.getAttribute( "type" ),
		"application/json",
		"Script type is restored"
	);
} );
```

Kevinjohn:
### Summary ###

A failed DOM insertion no longer leaves caller-owned `<script>` elements with jQuery's temporary `type` prefix.

Fixes #5881.

- Sequence: pass a script with `type="application/json"` to `.before()` where the browser rejects the insertion with `HierarchyRequestError`.
- Expected: the native exception propagates and the script keeps `type="application/json"`.
- Before: the exception propagates, but the script is left as `type="true/application/json"`.
- After: the exception propagates and the original script type is restored.

`domManip` now restores every masked script type in a `finally` block around the clone and insertion loop. This keeps cleanup reliable without swallowing the browser exception. The regression test covers both error propagation and restoration of the caller-owned element.

### Actual screenshots ###

**Before — upstream `main` at `7e1efd77598a527c67d2a674a1259787f7e441fd`, QUnit 2.26.0, Chrome 150, source modules enabled.**

Failing QUnit regression showing expected application/json and actual true/application/json

**After — commit `a4e6e3e9049ffd8a0b8d936f82ca7acd77297fd8`, the same focused test in Chrome 150 with source modules enabled.**

Passing focused QUnit regression with two of two assertions passing

**Complete manipulation module — commit `a4e6e3e9049ffd8a0b8d936f82ca7acd77297fd8`, Chrome 150 with source modules enabled.**

Complete manipulation module with 152 tests and 1035 assertions passing

### Validation ###

- `npm run test` — all builds, lint, browserless, Chrome/Firefox, ESM, slim, no-deprecated, and selector-native checks pass.
- Focused Chrome regression: 1 passed; 2 of 2 assertions passed.
- Complete manipulation module in Chrome: 152 passed; 1035 of 1035 assertions passed.
- Complete manipulation module in Firefox: 152 passed.

### Checklist ###

### DIFF
--- a/src/manipulation/domManip.js
+++ b/src/manipulation/domManip.js
@@ -54,29 +54,35 @@ export function domManip( collection, args, callback, ignored ) {
 			scripts = jQuery.map( getAll( fragment, "script" ), disableScript );
 			hasScripts = scripts.length;
 
-			// Use the original fragment for the last item
-			// instead of the first because it can end up
-			// being emptied incorrectly in certain situations (trac-8070).
-			for ( ; i < l; i++ ) {
-				node = fragment;
+			try {
 
-				if ( i !== iNoClone ) {
-					node = jQuery.clone( node, true, true );
+				// Use the original fragment for the last item
+				// instead of the first because it can end up
+				// being emptied incorrectly in certain situations (trac-8070).
+				for ( ; i < l; i++ ) {
+					node = fragment;
 
-					// Keep references to cloned scripts for later restoration
-					if ( hasScripts ) {
-						jQuery.merge( scripts, getAll( node, "script" ) );
+					if ( i !== iNoClone ) {
+						node = jQuery.clone( node, true, true );
+
+						// Keep references to cloned scripts for later restoration
+						if ( hasScripts ) {
+							jQuery.merge( scripts, getAll( node, "script" ) );
+						}
 					}
-				}
 
-				callback.call( collection[ i ], node, i );
+					callback.call( collection[ i ], node, i );
+				}
+			} finally {
+
+				// Re-enable scripts
+				if ( hasScripts ) {
+					jQuery.map( scripts, restoreScript );
+				}
 			}
 
 			if ( hasScripts ) {
 				doc = scripts[ scripts.length - 1 ].ownerDocument;
-
-				// Re-enable scripts
-				jQuery.map( scripts, restoreScript );
 
 				// Evaluate executable scripts on first document insertion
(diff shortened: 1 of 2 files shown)


timmywil:

PR #5899, opened 2026-08-06
Build: Upgrade dependencies, except for sinon,eslint,jsdom

timmywil:
jsdom 30 dropped Node 20 (engines: ^22.22.2 || ^24.15.0 || >=26) and pulls undici 8, which requires Node >=22.19.0 and calls `webidl.util.markAsUncloneable` — absent in Node 20. That crashed `test:browserless` on the 20.x matrix leg with `'TypeError: webidl.util.markAsUncloneable is not a function' `at jsdom load.


mgol:
Node 20 is EOL: https://github.com/nodejs/Release. Do we need to support it anywhere for jQuery build & test? I know we do for docs, the UI Download Builder & some other things that run on Node 18 with custom Debian patches, but for jQuery maybe not?

mgol (approved):

timmywil:
Yeah, perhaps not. We can drop in a future PR tho.

PR #5902, opened 2026-08-14
Deferred: Don't unwrap a thenable that is this promise

### LINKED ISSUE #5534: Animation promise gets stuck in infinite loop, eventually crashing the browser

arronKler:
### Summary ###

Calling `.then()` on an animation from `start` (or on any deferred that `notify`s with itself) looped inside Deferred/Callbacks until the tab ran out of memory.

`.then()` forwards progress through Identity. Animation's first progress value is the animation promise, so Identity returned that same thenable and `then()` attached another progress handler. Progress has memory, so the handler ran again immediately.

Skip thenable unwrapping when the returned `then` is this promise's own `then`. Foreign thenables still unwrap, including the existing "implicit progress callbacks unwrap thenables" tests.

Fixes #5534

### Checklist ###


arronKler:
recheck

